PR #241, opened 2014-08-11
Make MongoClient serializable

jedestep:
While serializing a MongoClient is hardly a common use case, it seems that major serialization frameworks don't work with MongoClient. These changes make MongoClient, Database and Collection serializable with ```dill```, an extension of ```pickle```.


behackett:
MongoClient is a complicated beast with a lot of internal state that constantly changes. I don't think we can properly support serializing it.

PR #252, opened 2015-02-05
Debug logging

jikamens:
Hi,

I don't intend for this PR to be pulled as-is; it's incomplete (e.g., I only implemented it for replica set client, not plain client) and I'm sure y'all would want to tweak the idea a bit before accepting it. I'm just submitting it as a food-for-thought PR to get your thoughts on the idea behind it.

We were having an issue where many of our ephemeral MongoDB client instances were unable to connect to the primary on one of our replicasets. This issue was proving very difficult to debug, since (a) there's no logging of the internals of mongo_replica_set_client.py's communication with servers or replicaset state refreshes, and (b) we run over SSL which makes debugging via network captures rather challenging.

To address this, what I ended up doing that enabled us to finally track down the problem (current running theory is stale arp cache entries on the MongoDB servers preventing packets from the servers from making it back to the clients) was adding the changes shown in this PR: modifying mongo_replica_set_client.py to log protocol exchanges as well as the process of refreshing replicaset state.

The specific, key discovery this enabled us to make was that [this code](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/mongo_replica_set_client.py#L1219) silently throws away errors, so were never being informed by pymongo that connections to our primary were timing out. So at the very least, even if you don't think any of the logging stuff shown here is a good idea, I'd say something needs to be done to make that information accessible to users rather than throwing it away.

Thanks,

Jonathan Kamens

### DIFF
--- a/pymongo/mongo_replica_set_client.py
+++ b/pymongo/mongo_replica_set_client.py
@@ -64,6 +64,7 @@ from pymongo.thread_util import DummyLock
 
 EMPTY = b("")
 MAX_RETRY = 3
+LOG_PROTOCOL = False
 
 MONITORS = set()
 
@@ -1018,6 +1019,7 @@ class MongoReplicaSetClient(common.BaseObject):
         rqst_id, msg, _ = message.query(0, ns, 0, -1, spec)
         start = time.time()
         try:
+            self.log_protocol('>>>', spec)
             sock_info.sock.sendall(msg)
             response = self.__recv_msg(1, rqst_id, sock_info)
         except:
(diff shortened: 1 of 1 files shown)


behackett:
Thanks for the pull request. It's good timing. We're currently working on a spec for logging and diagnostics in MongoDB drivers. Replica set member discovery and internal timeouts are definitely on the list. Stay tuned.

jikamens:
Thanks! In that case feel free to close the PR, no reason to leave it here cluttering things up, unless it's useful for you to do so.

behackett:
I'll leave it open as a reference for now. Thanks again.

behackett:
We haven't forgotten the request related to this PR. PyMongo 3.1 implemented support for logging command events (basically all application generated operations). We are currently working on a related spec for logging topology related events, which is the topic of this PR. I'll let you know when we have something you can try out.

behackett:
Support for server discovery and monitoring events has been committed to master. This includes published events for topology discovery and topology changes (servers being added or removed from a replica set, servers no longer responding to pings, etc.). It will be released in PyMongo 3.3, planned for this week. See [PYTHON-1067](https://jira.mongodb.org/browse/PYTHON-1067) for details.

PR #253, opened 2015-02-23
allow ssl_cert_reqs options to be passed as string

buckensl:
Fixes https://jira.mongodb.org/browse/PYTHON-842


behackett:
This has been merged to the v2.8 branch and will be released in PyMongo 2.8.1. Similar work has previously been released in PyMongo 3.0.

Thanks again for the patch!

PR #260, opened 2015-04-19
Small fix documentation aggregation

zerc:
Updated the example of Grouping


behackett:
Merged to both master and v2.8. Thanks for the patch.

PR #261, opened 2015-04-19
Update aggregate example

zerc:


behackett:
In PyMongo 3.0 (master on github), aggregate always returns an instance of CommandCursor, so this change is not correct. We already show what your looking for in the v2.x branch of PyMongo:

https://api.mongodb.org/python/2.8/examples/aggregation.html#aggregation-framework

ajdavis:
Hi @zerc , the example is correct for PyMongo 3 and later. `aggregate` now *always* returns a CommandCursor:

http://api.mongodb.org/python/current/api/pymongo/collection.html#pymongo.collection.Collection.aggregate

zerc:
Hi @ajdavis ok, got it. Thank you :)

PR #262, opened 2015-04-29
Add log4mongo-python

oz123:
Log4mongo-python is actively maintained and upaded to the latest pymongo versions (2.8 and 3.X).


behackett:
Merged to both v2.8 and master. Thanks. Looks like a neat project.

PR #263, opened 2015-05-10
Add ability to get BSON from a cursor

wmanley:
To avoid the deserialisation penalty when deserialising to a list of Python dictionaries is not necessary.  For my use-case I want to return BSON to my client so going via Python objects and then JSON is too expensive (~130ms for 6000 documents).

This is a partial solution to [PYTHON-472](https://jira.mongodb.org/browse/PYTHON-472).

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -1278,7 +1278,7 @@ class Collection(common.BaseObject):
                 res = helpers._first_batch(
                     sock_info, namespace, {"ns": self.__full_name},
                     0, slave_ok, codec_options, ReadPreference.PRIMARY)
-                data = res["data"]
+                data = res["data"].decode(codec_options)
                 cursor = {
                     "id": res["cursor_id"],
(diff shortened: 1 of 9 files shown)


wmanley:
@behackett: Would you mind having a look at this?  I've yet to receive any feedback.  Perhaps I should be looking elsewhere?

behackett:
Thanks for sending this patch. We're tracking comprehensive raw BSON support in https://jira.mongodb.org/browse/PYTHON-472, currently scheduled for PyMongo 3.2. We were hoping to have a solution in place for 3.1 but the feature is a bit more tricky than we originally thought. I'm leaving this pull request open as I think there are some good ideas to reference here, but I think the ultimate solution is going to look very different.

behackett:
Support for Raw BSON shipped in PyMongo 3.2, both reading it and writing it. Thanks again for the suggestion.

PR #264, opened 2015-05-18
Conversion fails when the datetime.year is less than 1.

opalfroot:
When converting datetime.datetime objects the millsecond conversion fails if the datetime.year is less than 1.

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -214,6 +214,8 @@ def default(obj):
         # TODO share this code w/ bson.py?
         if obj.utcoffset() is not None:
             obj = obj - obj.utcoffset()
+        if obj.year == 0000:
+            obj = datetime.datetime(0001,01,01)
         millis = int(calendar.timegm(obj.timetuple()) * 1000 +
                      obj.microsecond / 1000)
         return {"$date": millis}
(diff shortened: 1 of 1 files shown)


behackett:
Can you provide an example of the problem you are trying to work around? The python datetime.datetime type does not support year less than 1. How are you getting a datetime where obj.year is 0?

datetime.datetime(0, 1, 1)
ValueError: year is out of range

opalfroot:
Hi and thanks for coming back to me on this. Our production system uses the Java driver running mongodb 2.x We've come across values like:

#     "value" : Date(-62135769600000),
#     "source" : "0001-01-01",
#     "itemType" : "date"

the above translated from epoch == 0000-30-12, I believe this is due to rounding somewhere however the dates are pushed straight into mongo.

If you need any further information please let me know.

Sent from my android device.

-----Original Message-----

behackett:
I see. You're hitting a limitation of python's datetime module. This is documented here:

http://api.mongodb.org/python/current/faq.html#why-do-i-get-overflowerror-decoding-dates-stored-by-another-language-s-driver

There isn't much we can do about it in pymongo. Decoding dates with years < 1 to datetime(1, 1, 1) seems like a bad idea to me. Applications may decide to store the JSON back to the server, corrupting the original document. It would be better to fix the documents with invalid dates.

opalfroot:
After I posted I realised it wasn't the way to go. Could it be that an alternative default function is supplied which can choose to handle convention?

PR #265, opened 2015-05-19
Allow the default method in bson.json_util to be overridden.

opalfroot:
You may wish to handle object conversion differently from the default method. The
new code allows this in the following way.

```
#setup a db connection and grab a user object
db = gettDb()
user = db.user.find({'_id':ObjectId("544a41d40cf29893c1acb223")})[0]

#Alternative default function which will handle datetime objects diffrently
def altdefault(obj):
    if isinstance(obj, datetime):
        if obj.utcoffset() is not None:
            obj = obj - obj.utcoffset()
        if obj.year == 0000:
           obj = datetime.datetime(0001,01,01)
        millis = int(calendar.timegm(obj.timetuple()) * 1000 +
                     obj.microsecond / 1000)
        return {"$date": millis}
    return json_util.default(obj)

#run the dumps function and see the result
print json_util.dumps(user, altdefault)
```

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -100,7 +100,7 @@ _RE_OPT_TABLE = {
 }
 
 
-def dumps(obj, *args, **kwargs):
+def dumps(obj,altdefault=None, *args, **kwargs):
     """Helper function that wraps :class:`json.dumps`.
 
     Recursive function that handles all BSON types including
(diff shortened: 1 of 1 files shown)


behackett:
I don't think this is a change we want to make. As long as your application isn't trying to JSON encode instances of bson.code.Code or bson.binary.Binary you don't need the dumps helper. You can pass your custom encoder to json.dumps directly. If you do need to handle these types (this is not very common) you can monkey patch your decoder over json_utils.default at runtime.

PR #274, opened 2015-06-08
Allow Sequences as parameters to projection, keys, fields.

jaraco:
This PR represents my recommendation for addressing [PYTHON-946](https://jira.mongodb.org/browse/PYTHON-946). I made a couple of minor style changes which I believe makes the code clearer and more concise, but feel free to omit those commits or otherwise re-implement my suggestion.

I also understand that one might want to instead of being more lenient about the input types, simply document the more rigid expectation.

### DIFF
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -280,12 +280,13 @@ def _fields_list_to_dict(fields, option_name):
     """
     if isinstance(fields, collections.Mapping):
         return fields
-    elif isinstance(fields, collections.Sequence):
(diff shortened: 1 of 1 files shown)


behackett:
Thanks. I squashed this to one commit, using the message mentioning PYTHON-946 so our Jira githook picks it up and merged it to master. The related changes are a big improvement. It's nice to no longer have to deal with python 2.4...

jaraco:
Super. Thanks.

PR #277, opened 2015-06-08
PYTHON-889 The aggregate method should always return CommandCursor wh…

aherlihy:
…en passed the cursor option.


behackett:
Merged to v2.9.

PR #283, opened 2015-06-11
Use a standard version variable name (specified by PEP8)

ramnes:
See https://www.python.org/dev/peps/pep-0008/#version-bookkeeping

### DIFF
--- a/pymongo/__init__.py
+++ b/pymongo/__init__.py
@@ -77,7 +77,7 @@ def get_version_string():
         return '.'.join(map(str, version_tuple[:-1])) + version_tuple[-1]
     return '.'.join(map(str, version_tuple))
 
-version = get_version_string()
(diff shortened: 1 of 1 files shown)


behackett:
Since PyMongo has always used git for version control we've never thought to add __version__, but it is a pretty standard convention regardless of version control. I'll merge this for 3.1 and 2.9.

ramnes:
Well, `pymongo.version` was already there actually, but not `pymongo.__version__`, which is much more standard and widely used.

Thanks!

behackett:
Yep, we're on the same page. github's markdown rendering just made my comment less clear.

behackett:
Merged to v2.9.

behackett:
Also merged to master for 3.1. Thanks again for the patch.

PR #288, opened 2015-07-02
PYTHON-307 Replace select with poll on platforms that support it.

aherlihy:
PyMongo sometimes has to deal with >1023 sockets, so replace select with poll. Use select when poll is not available.


behackett:
Merged to master.

PR #289, opened 2015-07-04
PYTHON-957

jsamoocha:
Implemented rich comparison methods for write operations, as indicated in [PYTHON-957](https://jira.mongodb.org/browse/PYTHON-957).

### DIFF
--- a/pymongo/operations.py
+++ b/pymongo/operations.py
@@ -32,6 +32,15 @@ class _WriteOp(object):
         self._doc = doc
         self._upsert = upsert
 
+    def __eq__(self, other):
+        if type(other) == type(self):
+            return (other._filter, other._doc, other._upsert) == \
+                   (self._filter, self._doc, self._upsert)
+        return False
+
+    def __ne__(self, other):
+        return not self == other
+
 
 class InsertOne(_WriteOp):
     """Represents an insert_one operation."""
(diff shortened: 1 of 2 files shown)


pymongo/operations.py
```diff
@@ -32,6 +32,15 @@ def __init__(self, filter=None, doc=None, upsert=None):
         self._doc = doc
         self._upsert = upsert
 
+    def __eq__(self, other):
+        if type(other) == type(self):
+            return (other._filter, other._doc, other._upsert) == \
+                   (self._filter, self._doc, self._upsert)
+        return False
```
behackett: This should return NotImplemented, instead of False.

behackett: Oh, I see what you are doing here. You want to return NotImplemented if the comparison is to a type that doesn't inherit from _WriteOp and False if the type doesn't match.

jsamoocha: You're attributing too much sophistication to my code :-), this was my first attempt at implementing `__eq__()` in Python and I simply wasn't aware of `NotImplemented` as a potential return value, so good catch there.

I changed the return value in a new commit - the behavior of `assertNotEqual()` in tests doesn't seem to be affected by this change.

behackett:
Merged to master. Thanks for the patch!

PR #290, opened 2015-07-10
document gevent monkey patching side effect

ultrabug:
As discussed on http://www.ultrabug.fr/uwsgi-gevent-and-pymongo-3-threads-mayhem/#comments

Cheers

### DIFF
--- a/doc/examples/gevent.rst
+++ b/doc/examples/gevent.rst
@@ -19,12 +19,30 @@ on greenlets instead of threads.
 
 Consideration on gevent's hub termination
 -----------------------------------------
+PyMongo uses threads to monitor your servers' topology and to handle
+graceful reconnections.
+
 When shutting down your application gracefully, your code be may waiting
 for any remaining greenlet to terminate before exiting.
-Because monkey-patched threads are greenlets and PyMongo uses quite a lot
-of them, this may become a **blocking operation**  and may prevent your
-application to exit properly. You therefore must close (or dereference)
-any active MongoClient object before exiting !
+
+This may become a **blocking operation** because monkey-patched threads
+are considered as greenlets and this could prevent your application
+to exit properly. You therefore **must close or dereference** any active
+MongoClient object before exiting !
+
+Code example, this function is called when your application is asked
+to exit or gracefully reload itself by receiving a SIGHUP signal:
+
+.. code-block:: python
+
+    import signal
+
+    def graceful_reload(signum, traceback):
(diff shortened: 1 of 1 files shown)


behackett:
Thanks. We'll sort this out for 3.1.

ajdavis:
Thanks for the patch. I have a couple questions before we proceed.

First, what do you mean that PyMongo uses "quite a lot" of greenlets? Each MongoClient spawns one greenlet to monitor each server in your deployment -- so, if you have one MongoClient and a three-node replica set, the MongoClient spawns only three greenlets to monitor your replica set members. At least, that's the expected behavior -- have you observed anything different from this?

Second, PyMongo's monitor threads are configured with ``thread.daemon = True``. If the thread is actually a greenlet, does "wait for hub" wait for it anyway, even though it's set as "daemon"?

Third, could you update this PR with more practical advice? For example, show how to call MongoClient.close, or how to configure uWSGI with Gevent, or something that our readers can use directly.

Thanks!

ultrabug:
> what do you mean that PyMongo uses "quite a lot" of greenlets? Each MongoClient spawns one greenlet to monitor each server in your deployment

what I mean is that those monitor greenlets are new and are responsible for the possible lock on hub termination, so I meant "more than before" but I'll rephrase

> so, if you have one MongoClient and a three-node replica set, the MongoClient spawns only three greenlets to monitor your replica set members. At least, that's the expected behavior -- have you observed anything different from this?

no that's right

> PyMongo's monitor threads are configured with thread.daemon = True. If the thread is actually a greenlet, does "wait for hub" wait for it anyway, even though it's set as "daemon"?

yes that's what I'd expect since they are monkey patched as greenlets anyway

> Third, could you update this PR with more practical advice?

I'll try to, please bear with me

ajdavis:
Great, thanks for your responses. I look forward to the next version of your patch!

ultrabug:
Ok mates, it took a while but I just updated this PR with a better doc. Hope this feels better now.

ajdavis:
Thanks for this, I plan to merge it after I check with my colleagues. How would you like us to credit you in [contributors.rst](https://github.com/mongodb/mongo-python-driver/blob/master/doc/contributors.rst)? As "ultrabug" or by your full name? If you'd prefer us to use your name, what *is* your name? =)

ultrabug:
Good news, glad to help. ultrabug is fine thanks !

ajdavis:
Merged, thanks!

PR #291, opened 2015-07-20
matched_count should return a non 0 number if self.upserted_id is not none.

macdiesel:
The case that is incorrect here is that when a document is matched, but the data was the same so the upsert didn't happen. 

Also not sure here why n wasn't checked first.

### DIFF
--- a/pymongo/results.py
+++ b/pymongo/results.py
@@ -111,7 +111,7 @@ class UpdateResult(_WriteResult):
         """The number of documents matched for this update."""
         self._raise_if_unacknowledged("matched_count")
         if self.upserted_id is not None:
-            return 0
+            return 1
         return self.__raw_result.get("n", 0)
 
     @property


behackett:
In the case you describe self.upserted_id will be None, and the value for 'n' will be returned:

  >>> c.test.command('update', 'test', updates=[{'q': {'a': 1}, 'u': {'a': 1}, 'upsert': True}])
  {u'nModified': 0, u'ok': 1, u'upserted': [{u'index': 0, u'_id': ObjectId('55ad4c471503dc1dd7110165')}], u'n': 1, u'electionId': ObjectId('55a959380000000000000000'), u'lastOp': Timestamp(1437420615, 1)}
  >>> c.test.command('update', 'test', updates=[{'q': {'a': 1}, 'u': {'a': 1}, 'upsert': True}])
  {u'electionId': ObjectId('55a959380000000000000000'), u'nModified': 1, u'lastOp': Timestamp(1437420631, 1), u'ok': 1, u'n': 1}
  >>>
  >>> c.test.test.delete_many({})
  <pymongo.results.DeleteResult object at 0x7fa737e5afa0>
  >>> result = c.test.test.replace_one({'a': 1}, {'a': 1}, upsert=True)
  >>> result.upserted_id
  ObjectId('55ad4d181503dc1dd7110166')
  >>> result.matched_count
  0
  >>> result = c.test.test.replace_one({'a': 1}, {'a': 1}, upsert=True)
  >>> result.upserted_id
  >>> result.matched_count
  1

behackett:
Can you provide a test case that demonstrates the behavior you are trying to fix?

macdiesel:
yeah I should have sent one over, busy day.

Here is the raw output this returning what I feel is an erroneous result:

{'updatedExisting': False, u'nModified': 0, u'ok': 1, u'upserted': {u'category': u'problem', u'name': u'p1', u'course': u'toy', u'tag': u'i4x', u'org': u'edX', u'revision': u'draft'}, u'n': 1}

behackett:
What version of MongoDB are you using?

macdiesel:
$ mongod --version
db version v2.6.5
2015-07-20T19:47:06.517+0000 git version: e99d4fcb4279c0279796f237aa92fe3b64560bf6

behackett:
OK, I think I understand. The n field does necessarily specify how many documents were matched. In my examples calling the commands directly you can see the a value of 1 is returned for n even though there are no documents in the collection. You can read the (arguably ambiguous) definition of 'n' for update here:

http://docs.mongodb.org/manual/reference/command/update/#update.n

Note also that the mongo shell has the same behavior as the python driver, for the same reason:

```
repl0:PRIMARY> db.test.update({a: 1}, {a: 1}, {upsert: true})
WriteResult({
	"nMatched" : 0,
	"nUpserted" : 1,
	"nModified" : 0,
	"_id" : ObjectId("55ad524d1503dc1dd7110167")
})
repl0:PRIMARY> db.test.update({a: 1}, {a: 1}, {upsert: true})
WriteResult({ "nMatched" : 1, "nUpserted" : 0, "nModified" : 1 })
```

PR #292, opened 2015-07-22
insert_many(): Added support for generators

ameily:
While porting my application over to use the latest pymongo, I noticed that the new `insert_many()` function does not accept generators. This pull request adds a check in `insert_many()` for the Python `GeneratorType`. Also, I added a single unit test for inserting a generator.

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -16,7 +16,6 @@
 
 import collections
 import warnings
-from types import GeneratorType
 
 from bson.code import Code
 from bson.objectid import ObjectId
@@ -469,7 +468,7 @@ class Collection(common.BaseObject):
                                    self.write_concern.acknowledged)
 
     def insert_many(self, documents, ordered=True):
-        """Insert a list or generator of documents.
+        """Insert an iterable of documents.
 
           >>> db.test.count()
           0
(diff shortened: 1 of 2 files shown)


pymongo/collection.py
```diff
@@ -491,7 +492,7 @@ def insert_many(self, documents, ordered=True):
 
         .. versionadded:: 3.0
         """
-        if not isinstance(documents, list) or not documents:
+        if not isinstance(documents, (list, GeneratorType)) or not documents:
```
behackett: I think it might make more sense to get rid of list and instead do isinstance(documents, Iterable). Since we validate the documents we really only need to be sure we can iterate "documents".

pymongo/collection.py
```diff
@@ -468,7 +469,7 @@ def insert_one(self, document):
                                    self.write_concern.acknowledged)
 
     def insert_many(self, documents, ordered=True):
-        """Insert a list of documents.
+        """Insert a list or generator of documents.
```
behackett: Maybe "Insert an iterable of documents."

pymongo/collection.py
```diff
@@ -479,7 +480,7 @@ def insert_many(self, documents, ordered=True):
           2
 
         :Parameters:
-          - `documents`: A list of documents to insert.
+          - `documents`: A list or generator of documents to insert.
```
behackett: Same here.

behackett:
Good idea.

behackett:
Merged to master. Thanks again for the patch.

PR #293, opened 2015-08-14
Unsigned Int64 Encoding Support

ameily:
There was an issue with encoding an unsigned 64-bit int/long. This would occur when encoding a 640bit value larger than `(2^63 - 1)`, which is still a valid int64. Now, the size of Python ints/longs are intelligently detected and the encoded BSON is correct. However, decoding an unsigned int64 will still be decoded as a signed int64. It is up to the end-user to properly know whether the value was originally signed or unsigned and to perform the conversion manually, which is the standard across the other MongoDB drivers.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -134,7 +134,7 @@ Here's a basic example (for more see the *examples* section of the docs):
   >>> db.my_collection.find_one()
   {u'x': 10, u'_id': ObjectId('4aba15ebe23f6b53b0000000')}
   >>> for item in db.my_collection.find():
-  ...     print item["x"]
+  ...     print(item["x"])
   ...
   10
   8
@@ -142,7 +142,7 @@ Here's a basic example (for more see the *examples* section of the docs):
   >>> db.my_collection.create_index("x")
   u'x_1'
   >>> for item in db.my_collection.find().sort("x", pymongo.ASCENDING):
-  ...     print item["x"]
+  ...     print(item["x"])
   ...
   8
   10
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -1912,6 +1912,12 @@ static PyObject* get_value(PyObject* self, const char* buffer,
             }
             Py_XDECREF(utc_type);
             value = PyObject_Call(replace, args, kwargs);
+            if (!value) {
+                Py_DECREF(replace);
+                Py_DECREF(args);
+                Py_DECREF(kwargs);
+                goto invalid;
+            }
 
             /* convert to local time */
             if (options->tzinfo != Py_None) {
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -53,6 +53,37 @@ struct module_state {
     PyObject* Mapping;
 };
 
+
+/* pyint_value is a union for dealing with signed/unsigned 32bit and 64bit
+ * Python numbers for both Python 2.X and Python 3.X. */
+#if LLONG_MAX == LONG_MAX
+  /* On platforms where sizeof(long) == sizeof(long long). */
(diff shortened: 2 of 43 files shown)


behackett:
The problem here is that BSON doesn't provide unsigned integer types (32 or 64 bit) only signed integers. So MongoDB and any other driver will overflow when decoding these values:

http://bsonspec.org/spec.html

ameily:
My PR basically just encodes an unsigned int64 to a signed int64 by casting the value instead of raising an exception. I was encountering a problem where I was loading a JSON file that contained a 64-bit int that was larger than the signed max `(2^63 - 1)` and pymongo was raising an `OverflowError` during BSON encoding. MongoDB will not overflow when decoding the value because it is a valid signed, albeit negative, value when it is decoded. I've verified that this is the behavior of the C# driver.

For example, here is a session I performed using my PR. Normally an `OverflowError` would be raised in the call to `insert_one`:

```python
import pymongo
import ctypes

c = pymongo("mongodb://localhost")
c.testdb.pr293.insert_one({
    "val": (2**64) - 1
}) # ObjectId('.....')

data = c.testdb.pr293.find_one() #{ '_id': '....', 'val': -1 }

# I know I stored an unsigned Int64 value in the "val" field, convert it back
val = ctypes.c_uint64(data['val']).value # 18446744073709551615
val == (2**64) - 1 # True
```

This PR does not break backwards compatibility. It simply catches the OverflowError that was being raised by the `PyLong_AsLongLong()` function, and instead calls `PyLong_AsUnsignedLongLong()`, which doesn't raise an exception if the value is a valid unsigned 64-bit value. I verified that the PR works for Python 2.7, 3.3, and 3.4.

behackett:
But you're silently changing the value that was passed to PyMongo without the user being aware of it. That seems like a bad idea to me. BSON and MongoDB don't support unsigned 64 bit integers. The stored value will be interpreted differently in different drivers and the server. The .net driver's behavior seems like a bug to me.

behackett:
The .net driver's behavior is configurable and casting unsigned 64 bit integers isn't the default. You have to opt in and recognize that only the .net driver will interpret the integer the way you intended. This also has implications for update operations server side.

ameily:
I wouldn't say that the value is being changed. Rather, the Python value is cast to a BSON-compatible signed int64 only if the value is 64bits and greater than the signed max value. In the code it is just a matter of calling `unsigned long long PyLong_AsUnsignedLongLong()` instead of `long long PyLong_AsLongLong()`, neither the Python type nor the BSON value itself are modified, it is just a cast prior to the value hits the BSON encoder. Additionally, I don't see how different drivers will interpret the value different when querying the server. An Int64 is still stored server-side regardless. The BSON spec is preserved, this PR merely safely coerces a specific subset of Python int64 values to fit into MongoDB rather than raising an OverflowError.

I was originally doing this translation in Python, but performing this in the BSON C code seemed like a much better choice:

1. The BSON encoder is already iterating over each document's field so it is a preexisting choke point.
2. I imagine that C has to be considerable faster when inserting thousands of documents at a time, which I am doing in my project.
3. I figured handling these large unsigned int64 values could be useful in the main Python driver rather than a fork or another library.

That being said, I could make this new behavior configurable. I do think that this functionality should exist somewhere within the Python MongoDB driver, even if you have to explicitly enable it.

behackett:
Hi @ameily. I've given this feature more thought and I still don't think we should do it. We can't just consider how the python driver will handle these types. Other drivers will interpret the values as signed integers. So will the server, which at the very least means the sort order for the values will be undefined. 

The right thing to do is add unsigned integer types to the BSON spec and server. This is already tracked in https://jira.mongodb.org/browse/SERVER-4996.

PR #294, opened 2015-09-21
fix bug with dupplicates ObjectId in bulk insert

develf:
without this fix we get errors:
```
>> dbs.coll.insert_many([{'x':1}]*10)

raise BulkWriteError(full_result)
pymongo.errors.BulkWriteError: batch op errors occurred
```
or
```
>> for i in range(1000):
>>     data = {'x':1}
>>     dbs.coll.insert_one(data)

raise DuplicateKeyError(error.get("errmsg"), 11000, error)
pymongo.errors.DuplicateKeyError: insertDocument :: caused by :: 11000 E11000 duplicate key error index: dbs.coll.$_id_  dup key: { : ObjectId('56008f92c341cf5ab45db695') }
```

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -534,6 +534,7 @@ class Collection(common.BaseObject):
         .. versionadded:: 3.0
         """
         common.validate_is_mutable_mapping("document", document)
+        document = document.copy()
         if "_id" not in document:
             document["_id"] = ObjectId()
         with self._socket_for_writes() as sock_info:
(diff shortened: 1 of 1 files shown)


behackett:
This is because PyMongo modifies the original document and your call is passing 10 references to the same object. PyMongo has always worked this way and won't change. You can avoid this by passing 10 distinct documents instead:

c.foo.bar.insert_many({'x': 1} for _ in range(10))

PR #295, opened 2015-10-01
Preserve order of fields on JSON

paulo-raca:
I want to preserve field order on my application.

To do that, mongo has to use with ordered dicts instead of the standard `dict`.
Python's OrderedDict and Mongo's SON are good candidates for that.

As far as I can tell, there are 2 places where I need to make the change:
- On MongoClient's `document_class` (When reading from MongoDB)
- On json_util's `loads` (When parsing a JSON file)

I do both on my application, but it seems like it would be a reasonable default behavior

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -116,9 +116,9 @@ def dumps(obj, *args, **kwargs):
 def loads(s, *args, **kwargs):
     """Helper function that wraps :class:`json.loads`.
 
-    Automatically passes the object_hook for BSON type conversion.
+    Automatically passes the object_pairs_hook for BSON type conversion.
     """
-    kwargs['object_hook'] = lambda dct: object_hook(dct)
+    kwargs['object_pairs_hook'] = object_pairs_hook
     return json.loads(s, *args, **kwargs)
 
 
@@ -136,7 +136,8 @@ def _json_convert(obj):
         return obj
 
 
-def object_hook(dct):
+def object_pairs_hook(pairs):
+    dct = SON(pairs)
     if "$oid" in dct:
         return ObjectId(str(dct["$oid"]))
     if "$ref" in dct:
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -77,7 +77,7 @@ class MongoClient(common.BaseObject):
             self,
             host=None,
             port=None,
-            document_class=dict,
+            document_class=SON,
             tz_aware=False,
             connect=True,
             **kwargs):


behackett:
> I want to preserve field order on my application.

MongoDB *does not* preserve key order itself, so there is no way to enforce this in your application. If a document grows in size beyond the space allocated for it on disk it will reorder the top level fields when moving the document.

> I do both on my application, but it seems like it would be a reasonable default behavior

There are a few problems with this change. The first it that it would break the expectations of *all* existing applications. The second is that SON and OrderedDict are not as efficient or fast to serialize as dict (they both inherit from dict, and are implemented in pure python). Another problem is that OrderedDict, which would be the more likely candidate if we ever made such a change, does not exist in python 2.6, a python version we will have to support for many more years because of Red Hat Enterprise Linux.

Thanks for the suggestion though, and the patch. Unfortunately we can't take it.

paulo-raca:
Except for `_id`, the [docs says it _does_ preserve the field order](http://docs.mongodb.org/manual/core/document/#document-field-order). Is there anything I am missing?

I don't think the change would break expectations of existing applications, since right now there are no expectations about the field order at all.

But I do agree about the performance considerations you mentioned, and I agree `dict` should be the default behavior.

My next suggestion is to add a `document_class` parameter to `bson.json_util import.loads`. My current solution is too hackish:

```
from bson.json_util import object_hook as bson_object_hook
from json import loads as json_loads
from bson import SON

#Hack to preserve order of object fields
def loads(*args, **kwargs):
    kwargs['object_pairs_hook'] = lambda x: bson_object_hook(SON(x))
    return json_loads(*args, **kwargs)
```

behackett:
Sorry for the long silence on this.

> Is there anything I am missing?

It looks like we made some progress there in MongoDB 2.6, though the docs do mention one case where things could still be reordered.

> I don't think the change would break expectations of existing applications, since right now there are no expectations about the field order at all.

What I mean is applications currently expect a dict, not a SON or OrderedDict. That change of type may  not matter to most applications, but it will for sure break some applications somewhere. It would have to be a major version bump change if we did it (i.e. PyMongo 4.0).

> My next suggestion is to add a document_class parameter to bson.json_util import.loads

I agree, there does need to be a solution for this. Perhaps we need to implement our own object_pairs_hook that you can use instead of object_hook. It's a shame the two don't work together.

behackett:
It might make more sense for us to just document how to do this, rather than enforce returning SON instances (since we have to support python 2.6). This seems to work:

```
>>> import json
>>> from bson import json_util
>>> from bson.objectid import ObjectId
>>> doc = {'_id': ObjectId(), 'sub': {'x': 1}}
>>> s = json.dumps(doc, default=json_util.default)
>>> s
'{"_id": {"$oid": "562aa60efba5224ae8efd4bb"}, "sub": {"x": 1}}'
>>> from collections import OrderedDict  # Could also use bson.son.SON
>>> def my_object_pairs_hook(pairs):
...     dct = OrderedDict(pairs)
...     return json_util.object_hook(dct)
... 
>>> json.loads(s, object_pairs_hook=my_object_pairs_hook)
OrderedDict([(u'_id', ObjectId('562aa60efba5224ae8efd4bb')), (u'sub', OrderedDict([(u'x', 1)]))])
```

behackett:
I'm reopening this for the time being. We're not going to take the patch, but we do need to decide how to support returning ordered data structures when decoding mongo extended JSON. As I've said above, that will likely just be a documentation change since it's trivial for users to implement with a copy/paste example and we won't have to enforce use of a particular ordered data structure.

behackett:
[PYTHON-1111](https://jira.mongodb.org/browse/PYTHON-1111) tracks solving this problem, along with a number of other json_util issues.

PR #297, opened 2015-10-19
Utilities to convert datetime to/from milliseconds.

estobbart:
Spawned from [SO](https://stackoverflow.com/questions/28595338/bson-json-util-datetime-encode-and-decode-best-practice) discussion

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -100,6 +100,23 @@ _RE_OPT_TABLE = {
 }
 
 
+def dt_utc_milliseconds(obj):
+    """Helper function to convert a datetime obj to milliseconds in utc.
+    """
+    # TODO share this code w/ bson.py?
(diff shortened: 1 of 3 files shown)


behackett:
Can you tell me what you are using these helpers for?

estobbart:
It was compound problem. When returning from json_utils a python datetime becomes aware even if it by default in python is naive. And a python datetime includes microseconds and the Date obj stores milliseconds. This made it so a dumps/loads can't be compared for this type, seems to be only this type.

I was hoping for utilities to help with this. We use a lot of dumps/loads roundtrips for testing.

behackett:
[PYTHON-1111](https://jira.mongodb.org/browse/PYTHON-1111) tracks solving this problem, along with a number of other json_util issues.

PR #298, opened 2015-11-08
PYTHON-1009 Switch from "Monotime" to "monotonic"

felixonmars:
This implements the same try order as tornado, and updated docs to recommend monotonic only.


behackett:

PR #299, opened 2015-11-12
Compare IndexModel to SON from list_indexes.

estobbart:
When IndexModel objects are created by the user, you can check if they exist by comparing to a collections list_index method.

### DIFF
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -52,7 +52,7 @@ def _index_list(key_or_list, direction=None):
     else:
         if isinstance(key_or_list, string_type):
             return [(key_or_list, ASCENDING)]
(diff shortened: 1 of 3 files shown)


marcus-ahle:
:+1:

behackett:
The problem with this change is that it makes various assumptions about the structure of the index documents returned by MongoDB and the structure of those documents change from server release to server release. I think a better idea would be to open a ticket against the server requesting a stable index document format that can be passed unaltered to the createIndexes command.

estobbart:
@behackett Thanks. My assumption was that consistency problem was wrapped by IndexModel.

estobbart:
https://jira.mongodb.org/browse/SERVER-21882

PR #300, opened 2015-12-04
Use codec options in aggregate command

hedgepigdaniel:
Before this the aggregate command ignored the codec options. this resulted in in aggregate command returning documents with naive datetime objects despite the tz_aware attribute in the collections codec options being set to True.

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -1802,6 +1802,7 @@ class Collection(common.BaseObject):
 
         command_kwargs = {
             'pipeline': pipeline,
+            'codec_options': self.codec_options,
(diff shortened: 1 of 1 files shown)


behackett:
Thanks for the patch. I was confused until I realized this is for PyMongo 2.9.x.

behackett:
I opened https://jira.mongodb.org/browse/PYTHON-1027 to track this. We'll merge your PR for PyMongo 2.9.1.

As a workaround you can pass tz_aware=True when creating an instance of MongoClient.

behackett:
Merged to v2.9. Thanks again for the patch!

PR #302, opened 2016-02-20
PYTHON-1057: allow connect option to be passed as string

ffix:
It fixes [PYTHON-1057](https://jira.mongodb.org/browse/PYTHON-1057) (Connect option validator should use validate_boolean_or_string).


behackett:

PR #303, opened 2016-04-12
Added cursor find command for chaining

primal100:
Enables filter chaining on the cursor similar to how many ORM's are implemented.


behackett:
I think this will be really confusing. Users will expect Cursor.find() to support all the options of Collection.find() (regardless of it not making any sense). It's not difficult to build a dictionary iteratively and pass it to Collection.find() once.

PR #308, opened 2016-07-22
PYTHON-1111 Add JSONOptions for dumps and loads

ShaneHarvey:


behackett:
Merged to 34dev. Thanks.

PR #309, opened 2016-07-29
[PYTHON-1123] Cannot customize some properties of MongoClient.codec_options

oxymor0n:
this PR addresses [PYTHON-1123](https://jira.mongodb.org/browse/PYTHON-1123)

### DIFF
--- a/pymongo/common.py
+++ b/pymongo/common.py
@@ -17,6 +17,7 @@
 
 import collections
 import warnings
+import datetime
 
 from bson.binary import (STANDARD, PYTHON_LEGACY,
                          JAVA_LEGACY, CSHARP_LEGACY)
@@ -419,6 +420,27 @@ def validate_ok_for_update(update):
         raise ValueError('update only works with $ operators')
 
 
+_UNICODE_DECODE_ERROR_HANDLERS = frozenset(['strict', 'replace', 'ignore'])
+
+
+def validate_unicode_decode_error_handler(dummy, value):
+    """Validate the Unicode decode error handler option of CodecOptions.
+    """
+    if value not in _UNICODE_DECODE_ERROR_HANDLERS:
+        raise ValueError("%s is an invalid Unicode decode error handler. "
+                         "Must be one of "
+                         "%s" % (value, tuple(_UNICODE_DECODE_ERROR_HANDLERS)))
+    return value
+
+
+def validate_tzinfo(dummy, value):
+    """Validate the tzinfo option
+    """
+    if value is not None and not isinstance(value, datetime.tzinfo):
+        raise TypeError("%s must be an instance of datetime.tzinfo" % value)
+    return value
+
+
 # journal is an alias for j,
 # wtimeoutms is an alias for wtimeout,
 URI_VALIDATORS = {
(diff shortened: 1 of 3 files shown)


ajdavis:
The enhancement to MongoClient parameters looks quite nice to me. I notice that we can't specify `tzinfo` via the URI, but I think that's ok.

pymongo/mongo_client.py
```diff
@@ -81,7 +81,7 @@ def __init__(
             host=None,
             port=None,
             document_class=dict,
-            tz_aware=False,
+            tz_aware=None,
             connect=True,
```
llvtt: I think `connect` has a similar problem to `tz_aware`, where it can't be specified in the URI. Would you mind making a similar change for connect as you have for tz_aware on L357?

pymongo/common.py
```diff
@@ -17,6 +17,7 @@
 
 import collections
 import warnings
+import datetime
```
llvtt: This is a minor nit: we try to keep the imports alphabetical in pymongo. Could you move `import datetime` above `import warnings`?

llvtt:
@oxymor0n, thanks a lot for your work on this! I made just a couple of small comments. After they're resolved, I think this will be ready to merge. This will go into version 3.3.1 of PyMongo.

oxymor0n:
@ajdavis `tzinfo` requires its value to be a subclass of `datetime.tzinfo`, so it cannot be specified via the URL parameters in a straightforward manner. There is a way to map IANA time zone database identifier (e.g. `US/Eastern`, `Europe/Amsterdam`, etc.) to `tzinfo` objects, but it requires a dependency on the third-party `pytz` library, so I was wary of introducing an additional dependency to the driver without consulting anyone from the mongo team. Let me know if you guys would like that.

pymongo/common.py
```diff
@@ -450,7 +472,8 @@ def validate_ok_for_update(update):
     'tz_aware': validate_boolean_or_string,
     'uuidrepresentation': validate_uuid_representation,
     'connect': validate_boolean_or_string,
-    'minpoolsize': validate_non_negative_integer
+    'minpoolsize': validate_non_negative_integer,
+    'unicode_decode_error_handler': validate_unicode_decode_error_handler
```
oxymor0n: @llvtt I've also just noticed that Mongo URI parameters use camelCase instead of snake_case. Should change this from `unicode_decode_error_handler` to `unicodedecodeerrorhandler` to conform with the convention?

ajdavis: There's a mix of parameter styles. You can see that tz_aware is snake-cased. The principle, more or less, is that standard MongoDB URI options are camel-cased, and PyMongo-specific options are snake-cased. This isn't deliberate, but it's a consequence of how we use the same code to validate MongoClient() parameters and URI options, and we've decided to live with it.

Short answer: let's leave unicode_decode_error_handler snake-cased.

oxymor0n: cool, that makes sense. thanks for the clarification!

llvtt:
This all looks good to me. Thanks again for your hard work, @oxymor0n. One final request: We've decided not to do a 3.3.1 release and instead put this in 3.4, which is on the "34dev" branch in this repo. Would you mind reopening this pull request over on that branch? Thanks again!

oxymor0n:
@llvtt thanks, see #310

PR #313, opened 2016-08-27
write concern for update_one and update_many operation

sumitkumar1209:


behackett:
Thanks for the patch, but I'm not going to take it. These methods, along with insert_one/many, delete_one/many, etc. don't take a write concern by design. If you want to change the write concern use Collection.with_options:

https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.with_options

sumitkumar1209:
@behackett I agree with your point. Can you point me to the reference explaining why from version 3.2 find_one_and_update family methods have writeConcern enabled, as I got the idea from the same. Do they require any special treatment internally? 

> Changed in version 3.2: Respects write concern.
> 
> Warning Starting in PyMongo 3.2, this command uses the WriteConcern of this Collection when connected to MongoDB >= 3.2. Note that using an elevated write concern with this command may be slower compared to using the default write concern.

behackett:
MongoDB 3.2 added support for writeConcern to the findAndModify command, so all the drivers started applying it:

https://jira.mongodb.org/browse/DRIVERS-224

MongoDB 3.4 will go the last mile, accepting writeConcern for all commands that write:

https://jira.mongodb.org/browse/DRIVERS-290

sumitkumar1209:
So, PyMongo won't give support for update commands even if MongoDB supports it or are they removing it from newer versions and we have to use Collection.with_options for the workaround?

behackett:
PyMongo has always supported write concern for inserts, updates, and deletes, all the way back to the 0.x versions. What changed for PyMongo 3.x is that write concern is only set at the client, database, or collection level, not per-operation. So yes, you must use Collection.with_options to change the write concern.

sumitkumar1209:
Sorry to bug you, probably for the last time. Can you direct me to any reference where I can read on why this decision was made to support only for client, database or collection level as before 3.x PyMongo had support for it.

behackett:
No reference. It was a decision I made during the development of PyMongo 3.0.

PR #314, opened 2016-09-09
V2.8: max_pool_size does not work very well in some cases

yunstanford:
Each motor’s MongoReplicaSetClient has a “Monitor”. The Monitor checks the state of each mongo node in our replica set every 30 seconds by default. The Monitor firstly tries to find an available connection from connection pool for the ping check,  if it cannot find any available connection, it “forces" the connection pool to create a new connection for its ping check, even if  # of connections in pool  > max_pool_size.  yeah, it  just ignore the max_pool_size.

Let's think about a edge case. 

1. let's say we have 3 mongo instances (1 primary and 2 secondaries), and assume that reads and writes keep coming
2. set write_concern = 3 and max_pool_size = 10
3. one secondary instance goes down (maybe for maintainance)
4. all connections in connection pool will be "blocked" quickly due to write_concern, because  write_concern = 3  but we only have 2 mongo instances now.
5. However, the monitor checks and refreshes the state of replica set every 30 seconds(default), which forces the connection pool to create a new connection every 30 seconds. After the monitor check, the new created connection will be returned to the pool ,then it will quickly be used by new write operation and will be “blocked” due to write_concern issue.  This scenario keeps occurring again and again until all available connections of mongo instance have been used up. 

This is very bad although it is a edge case. 

In my understanding, all "forced" created connections for "refresh check" should drop after use instead of returning it to connection pool again.


behackett:
I can't tell if you intend this to be a bug report or a pull request, or if this is really about PyMongo or Motor.

The 2.x branch of PyMongo is in maintenance mode. We're only fixing critical bugs. PyMongo 3.x addresses what I think are your concerns by using a thread to monitor each individual replica set member. One member going down has no effect on other members.

yunstanford:
This should be a bug report. It is about Pymongo's MongoReplicaSetClient class. the monitor forces connection pool to create new connections if it cannot find any available connection in the pool, which makes max_pool_size meaningless.

ajdavis:
Thanks for the report. In the future, to report bugs in MongoDB projects, please use Jira:

https://github.com/mongodb/motor#how-to-ask-for-help

That's a good analysis of a case I hadn't considered. But luckily we've already fixed this problem in PyMongo 3 by using separate connections exclusively for monitoring. This conforms to the Server Discovery And Monitoring Spec:

https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#servers-are-monitored-with-dedicated-sockets

Motor will soon upgrade its PyMongo dependency to PyMongo 3.2 and later, so this problem will be fixed in Motor too:

https://jira.mongodb.org/browse/MOTOR-87

yunstanford:
Nice. Thanks for your reply.

PR #318, opened 2016-11-20
[PYTHON-1191] Added KeyEscaper SONManipulators.

todofixthis:
Introduces two new `SONManipulators` to transparently escape/unescape illegal keys when storing/retrieving documents:

- `DeterministicKeyEscaper` uses MD5 to generate "safe" versions of escaped keys. This method is slower and requires more storage space, but the escaped keys are potentially queryable.
- `NonDeterministicKeyEscaper` uses an internal counter to generate "safe" versions of escaped keys. This method is faster and requires less storage space, but the escaped keys should be considered not queryable.

For more information, refer to:
- https://jira.mongodb.org/browse/PYTHON-1191 (JIRA ticket)
- http://stackoverflow.com/q/40639393/ (original context)

**Note:** In order to achieve compatibility with Python 2 and 3 for this feature, I added the `six` library to the project dependencies (in `setup.py`). I'm not sure that this is the right context to introduce a new dependency, but I don't know this project well enough yet to propose an alternate approach. Definitely open to guidance on this!

### DIFF
--- a/pymongo/son_manipulator.py
+++ b/pymongo/son_manipulator.py
@@ -19,8 +19,16 @@ installed on a database by calling
 `pymongo.database.Database.add_son_manipulator`."""
 
 import collections
+from abc import ABCMeta, abstractmethod as abstract_method
+from codecs import decode
+from collections import Mapping as MappingType
+from hashlib import md5
+from sys import getdefaultencoding as get_default_encoding
+
+from six import binary_type, iteritems, text_type, with_metaclass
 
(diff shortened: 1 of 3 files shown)


ajdavis:
Thanks for proposing this idea. For reasons explained here we won't accept it:

https://jira.mongodb.org/browse/PYTHON-119

If you disagree, I invite you to maintain your own library that provides this feature as a wrapper around PyMongo.

PR #319, opened 2017-01-30
PYTHON-1214 - Implement the extended JSON specification.

llvtt:


bson/json_util.py
```diff
@@ -415,13 +379,135 @@ def object_hook(dct, json_options=DEFAULT_JSON_OPTIONS):
...
+    frozenset(['$binary', '$type']): _get_binary,
+    frozenset(['$code']): lambda d, _: Code(d['$code']),
+    frozenset(['$code', '$scope']): lambda d, _: Code(
+        d['$code'], d['$scope']),
+    frozenset(['$timestamp']): lambda d, _: Timestamp(
+        int(d['$timestamp']) >> 32, int(d['$timestamp']) & 0xff)
+}
+
+
+def canonical_object_hook(dct, json_options=DEFAULT_JSON_OPTIONS):
```
behackett: Should json_options default to CANONICAL_JSON_OPTIONS here?

bson/json_util.py
```diff
@@ -499,4 +587,19 @@ def default(obj, json_options=DEFAULT_JSON_OPTIONS):
             return {"$uuid": obj.hex}
     if isinstance(obj, Decimal128):
         return {"$numberDecimal": str(obj)}
+    if isinstance(obj, bool):
```
behackett: I think the json module already does this for you?

test/test_bson_corpus.py
```diff
@@ -170,12 +186,18 @@ def run_test(self):
                         normalized_cE)
 
                 if 'lossy' not in valid_case:
-                    self.assertEqual(encode_bson(decode_extjson(E)), cB)
-
-                    if E != cE:
-                        self.assertEqual(
-                            encode_bson(decode_extjson(cE)),
-                            cB)
+                    # Skip tests for document type in Python 2.6 that have
```
behackett: Maybe this should be tested if simplejson is installed?

bson/json_util.py
llvtt: This is here to prevent booleans from matching the following case `isinstance(obj, integer_types`:

```python
>>> isinstance(True, int)
True
>>> issubclass(bool, int)
True
```

test/test_bson_corpus.py
llvtt: Done.

bson/json_util.py
```diff
@@ -499,4 +587,19 @@ def default(obj, json_options=DEFAULT_JSON_OPTIONS):
...
+        return obj
+    if json_options.canonical_extended_json and isinstance(obj, integer_types):
+        return {'$numberInt': text_type(obj)}
```
ShaneHarvey: This will not produce canonical JSON when the int is outside the bounds of a 32-bit integer. It should produce '$numberInt', '$numberLong', or error when its outside 64 bits.

llvtt: Good catch! Fixed.

bson/json_util.py
```diff
@@ -469,11 +555,15 @@ def default(obj, json_options=DEFAULT_JSON_OPTIONS):
...
+        if json_options.canonical_extended_json:
+            return {'$timestamp': str((obj.time << 32) + obj.inc)}
         return {"$timestamp": SON([("t", obj.time), ("i", obj.inc)])}
     if isinstance(obj, Code):
         if obj.scope is None:
             return SON([('$code', str(obj))])
-        return SON([('$code', str(obj)), ('$scope', obj.scope)])
+        return SON([
+            ('$code', str(obj)),
+            ('$scope', _json_convert(obj.scope or {}, json_options))])
```
ShaneHarvey: We already check for none above so is the `or {}` necessary?

test/test_bson_corpus.py
```diff
@@ -65,18 +58,41 @@ class TestBSONCorpus(unittest.TestCase):
...
 to_bson_uuid_04 = functools.partial(BSON.encode,
                                     codec_options=codec_options_uuid_04)
 to_bson = functools.partial(BSON.encode, codec_options=codec_options)
 decode_bson = lambda bbytes: BSON(bbytes).decode(codec_options=codec_options)
 
 
+class TestBSONCorpus(unittest.TestCase):
+    def test_all_bson_types(self):
+        # Because we can't round-trip all BSON types, make this test a special
+        # case, instead of mangling our create_test function below.
```
ShaneHarvey: Can you update the comment with the BSON types that are not preserved?

test/test_bson_corpus.py
```diff
@@ -170,12 +186,19 @@ def run_test(self):
...
+                            bson_type == '0x03' and
+                            len(decode_extjson(E)) > 1):
+                        self.assertEqual(encode_bson(decode_extjson(E)), cB)
+
+                        if E != cE:
+                            self.assertEqual(
+                                encode_bson(decode_extjson(cE)),
+                                cB)
 
             for decode_error_case in case_spec.get('decodeErrors', []):
```
ajdavis: This is overindented - it runs once per valid case or, if there are no valid cases, it doesn't run at all. That's why top.json seems to pass, even though the "Document with keys that start with $" test isn't right.

llvtt: Fixed

llvtt:
We compare json_util's JSON output to the extended JSON in the bson corpus tests by running the extended JSON from the corpus tests through Python's json parser (*not* bson.json_util's parser). Thus, forms like `{"$oid": <whatever>}` aren't converted to BSON types. This comparison still tests the extended JSON output, but without having to worry about whether there are spaces before/after `:`, `{`, `,`, etc. Key ordering should also be disregarded in Python 2.6 et al, since `object_pairs_hook` doesn't exist, but it's still worthwhile test JSON output minus the key ordering in these situations.

I've updated the test runner to preserve key ordering when possible in all the bson corpus JSON tests.

bson/json_util.py
```diff
@@ -486,13 +492,13 @@ def _get_dbpointer(doc, dummy):
 
 
 def canonical_object_hook(dct, json_options=CANONICAL_JSON_OPTIONS):
-    keyset = frozenset(dct)
+    keyset = frozenset(key for key in dct if key.startswith('$'))
```
ShaneHarvey: The spec says what happens when extra $ keys are present (https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#id43), but what if an extra non-$ keys exist?

Does `'{"$numberInt": "123", "foo": "bar"}'` get parsed into a bson Int32 or a document `{"$numberInt": "123", "foo": "bar"}`?

llvtt: It is parsed into an int32. Because the extra keys do not start with `$`, they don't "count against" the type of the document. Here's a test that demonstrates the expected behavior with DBPointer: https://github.com/mongodb/specifications/blob/6cee58a86008ed3ee4586825ac0acfeab8b5263b/source/bson-corpus/tests/dbpointer.json#L18-L23. Note that this behavior is consistent json_util's non-canonical behavior with `object_hook`, e.g., if we see `$oid` in a dictionary, interpret the `$oid` field as an ObjectId hex string and forget about any other keys.

ShaneHarvey: Thanks for the clarification!

bson/json_util.py
```diff
@@ -415,13 +379,141 @@ def object_hook(dct, json_options=DEFAULT_JSON_OPTIONS):
...
+    frozenset(['$ref', '$id', '$db']): lambda d, _: DBRef(
+        d.pop('$ref'), d.pop('$id'), d.pop('$db'), **d),
+    frozenset(['$regex', '$options']): lambda d, _: Regex(
+        d['$regex'], d['$options']),
+    frozenset(['$binary', '$type']): _get_binary,
+    frozenset(['$code']): lambda d, _: Code(d['$code']),
+    frozenset(['$code', '$scope']): lambda d, _: Code(
+        d['$code'], d['$scope']),
+    frozenset(['$timestamp']): lambda d, _: Timestamp(
+        int(d['$timestamp']) >> 32, int(d['$timestamp']) & 0xff)
```
ShaneHarvey: The increment is a 4-byte value not 1-byte, so it should be `& 0xffffffff`.

bson/json_util.py
```diff
@@ -201,11 +209,13 @@ class JSONOptions(CodecOptions):
        <http://www.mongodb.org/display/DOCS/Mongo+Extended+JSON>`_.
 
     .. versionadded:: 3.4
+
```
behackett: Add a ".. versionchanged:: 3.5" here to document when the canonical_extended_json option was added.

bson/json_util.py
```diff
@@ -415,13 +379,141 @@ def object_hook(dct, json_options=DEFAULT_JSON_OPTIONS):
...
+        millis = int(dtm["$numberLong"])
+    # mongoexport before 2.6
+    else:
+        millis = int(dtm)
+    return bson._millis_to_datetime(millis, json_options)
+
+
+def _get_dbpointer(doc, json_options):
+    dbref = doc['$dbPointer']
+    if isinstance(dbref, DBRef):
```
behackett: Why would the value of $dbPointer be a DBRef instance? Shouldn't it just be a dict like any other JSON object?

bson/json_util.py
```diff
@@ -80,6 +80,7 @@
 import base64
```
ShaneHarvey: Should we update the module docstring at the top to call out the canonical extended JSON format? I think a link to the spec and an example input/output would be in order.

bson/json_util.py
llvtt: No- the json parser works bottom-up. `object_hook` is called on the leaf elements of a document first, then the results are used in turn to construct their parent objects.

bson/json_util.py
llvtt: Added.

bson/json_util.py
llvtt: Added.

bson/json_util.py
behackett: Ah, right. Thanks.

bson/json_util.py
behackett: I see. Thanks.

behackett:
LGTM. Make sure the doctests pass in both python 2 and python3, squash this whole thing to a single commit, then push.

bson/json_util.py
```diff
@@ -269,7 +286,7 @@ def _arguments_repr(self):
 """
 
 CANONICAL_JSON_OPTIONS = JSONOptions(canonical_extended_json=True)
-""":class:`JSONOptions` for canonical extended JSON.
+""":class:`JSONOptions` for `canonical extended JSON`_.
```
ShaneHarvey: The `canonical extended JSON` reference can be reused here?

llvtt: It can.

llvtt:
Added doctests pass. Merged in https://github.com/mongodb/mongo-python-driver/commit/74c0d55314d978193e7fa39502833ac1d85a0561.

PR #320, opened 2017-02-10
Sdam tags

tap1r:
Add live demo script tags.py


behackett:
I think this may be better to leave in @aherlihy's fork for now. We're not going to merge this to PyMongo until a full spec process has happened.

PR #322, opened 2017-03-03
Link to createIndexes command from create_indexes documentation.

llvtt:
Fix markup for Collation class in IndexModel documentation.

For https://jira.mongodb.org/browse/PYMODM-55

There was some confusion in PyMODM about how to create an index that used the "weights" property for a text index. The [PyMODM docs](https://pymodm.readthedocs.io/en/latest/api/index.html#pymodm.MongoModel) link to PyMongo's [IndexModel](http://api.mongodb.com/python/current/api/pymongo/operations.html#pymongo.operations.IndexModel). This links to [create_indexes](http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.create_indexes). From here, there's a dead end.

These changes link to the createIndexes command in the MongoDB manual, which provides a table showing all available options for the command. This also fixes some markup for the "collation" parameter in constructing an IndexModel.

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -1367,7 +1367,7 @@ class Collection(common.BaseObject):
           - `indexes`: A list of :class:`~pymongo.operations.IndexModel`
             instances.
 
-        .. note:: `create_indexes` uses the ``createIndexes`` command
+        .. note:: `create_indexes` uses the `createIndexes`_ command
            introduced in MongoDB **2.6** and cannot be used with earlier
            versions.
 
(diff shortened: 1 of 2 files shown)


pymongo/collection.py
```diff
@@ -1379,6 +1379,8 @@ def create_indexes(self, indexes):
            Apply this collection's write concern automatically to this operation
            when connected to MongoDB >= 3.4.
         .. versionadded:: 3.0
+
+        .. _createIndexes: https://docs.mongodb.com/manual/indexes/
```
behackett: Let's use dochub links instead. They are guaranteed to resolve to the right thing seemingly forever.

http://dochub.mongodb.org/core/indexes

llvtt: According to Kay: There is no dochub link to this. Dochub links are created upon request of an engineer, so that they can put a link in their docs to a page that does not yet exist. When pages are moved in the MongoDB manual, there are redirects set up so that we avoid link rot.

PR #323, opened 2017-05-23
Slightly faster Cursor.next()

tzachyrm:
only referencing `self.__collection.database` when actually needed


behackett:
The v2.9 branch of PyMongo is in maintenance mode, bug fixes only. Did you intend to apply this change to that branch?

PR #325, opened 2017-06-07
Don't change readpreference value at uri parsing

wronglink:
For some reason [validate_read_preference_mode](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/common.py#L305) returns value, that is not valid to be validated again. This PR fixes this problem and allows `MongoClient` options could be parsed outside of it and passed as keyword arguments like celery mongo backend [does it](https://github.com/celery/celery/blob/master/celery/backends/mongodb.py#L77).

### DIFF
--- a/pymongo/client_options.py
+++ b/pymongo/client_options.py
@@ -22,7 +22,7 @@ from pymongo.errors import ConfigurationError
 from pymongo.monitoring import _EventListeners
 from pymongo.pool import PoolOptions
 from pymongo.read_concern import ReadConcern
-from pymongo.read_preferences import make_read_preference
+from pymongo.read_preferences import make_read_preference, read_pref_mode_from_name
 from pymongo.ssl_support import get_ssl_context
 from pymongo.write_concern import WriteConcern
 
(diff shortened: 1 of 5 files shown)


behackett:
This is an interesting change. It's technically backward breaking, but it's unlikely anyone would notice. I haven't tested it out yet, but I assume the MongoClient repr still works properly?

wronglink:
Kind of. It doesn't fail, but it represents `readpreference` as a string, instead of integer:

```python
In [1]: from pymongo import MongoClient

In [2]: repr(MongoClient('mongodb://localhost/dbname?readpreference=nearest'))
Out[2]: "MongoClient(host=['localhost:27017'], document_class=dict, tz_aware=False, connect=True, readpreference='nearest')"
```

Master branch:
```python
In [1]: from pymongo import MongoClient

In [2]: repr(MongoClient('mongodb://localhost/dbname?readpreference=nearest'))
Out[2]: "MongoClient(host=['localhost:27017'], document_class=dict, tz_aware=False, connect=True, readpreference=4)"
```

Yes it's backward breaking, but I'd say that it's better readable.

ShaneHarvey:
Merged in https://github.com/mongodb/mongo-python-driver/commit/9f07ff35a18ea456b555ba79c227d99f5d008cc5, thanks @wronglink!

PR #326, opened 2017-06-07
Pass current field name to raw BSON decoder

wronglink:
This PR makes both python and c BSON decoder implementation pass current decoding field name to raw BSON decoder (e.g. RawBSONDocument). It is needed to add some name-dependent decoding behaviour.
In addition to context sharing between decoding levels from #324 it can give custom BSON decoder such feature like decoding map (e.g. `foo` array would be decoded to `FooClass` object, and the inner `foo.bar` array would be wrapped by `BarClass`)

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -134,7 +134,7 @@ def _get_string(data, position, obj_end, opts, dummy):
                          opts.unicode_decode_error_handler, True)[0], end + 1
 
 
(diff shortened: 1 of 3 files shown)


behackett:
Thanks again for your contributions. As mentioned in #324, I think using raw documents for this use case is the wrong thing to do.

PR #330, opened 2017-07-19
Fix bug: SON object and list can not be compared

tvytlx:
The program's current behavior will always call `collection.create_index(index)` regardless of whether this index exists or not, so when you use a mongo user which without `createIndex` privilege, you can't write to the `gridfs`, though those indexes are already created by admin.

### DIFF
--- a/gridfs/grid_file.py
+++ b/gridfs/grid_file.py
@@ -19,6 +19,7 @@ import os
 
 from hashlib import md5
 
+from bson.son import SON
 from bson.binary import Binary
 from bson.objectid import ObjectId
 from bson.py3compat import text_type, StringIO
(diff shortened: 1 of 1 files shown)


gridfs/grid_file.py
```diff
@@ -179,7 +179,7 @@ def __create_index(self, collection, index, unique):
                 indexes = list(collection.list_indexes())
             except OperationFailure:
                 indexes = []
-            if index not in indexes:
+            if dict(index) not in [idx['key'].to_dict() for idx in indexes]:
```
tvytlx: According to SON object's compare strategy ([\_\_eq\_\_](https://github.com/mongodb/mongo-python-driver/blob/master/bson/son.py#L206)), it should ensure items's order.
```
In [59]: SON([(u'filename', 1.0), (u'uploadDate', 1.0)]) == SON([(u'uploadDate', 1.0), (u'filename', 1.0)])
Out[59]: False
```
But here we don't know the order returned from `list_indexes()`, so I use dict to compare.

ShaneHarvey: Actually, this code is more wrong than we thought. `list_indexes()` returns the full index spec but we're trying to compare that to just the index keys. Instead we should look at only the `'key'` field in the index spec:
```python
try:
    index_keys =[index_spec['key'] for index_spec in collection.list_indexes()]
except OperationFailure:
    index_keys = []
if index_key not in index_keys:
...
```

On your point about comment ordering: `list_indexes()` returns the indexes as a SON document so the keys will be well ordered. The order of keys in indexes are important; the index `SON([('filename', 1), ('uploadDate', 1)])` is different from `SON([('uploadDate', 1), ('filename', 1)])`. That's why we should not convert these to dictionaries.

tvytlx: Your explanation makes sense, I changed this commit.

PR #332, opened 2017-08-17
ignore rst files for evergreen builds

tychoish:


behackett:
We have doctests run in evergreen:

https://github.com/mongodb/mongo-python-driver/blob/master/.evergreen/config.yml#L1287-L1292

Is there a different change that could be made to only run those tests when changes are made to restructured text files?

tychoish:
this isn't possible with the current system but we can scope the ignores more tightly so that *most* .rst files are ignored, but ones with doctests aren't. The ignore list syntax is "`.gitignore`-ish" (e.g. [this](https://github.com/sabhiram/go-gitignore)), which may make it easy.

.evergreen/config.yml
```diff
@@ -17,6 +17,12 @@ command_type: system
 # That roughly accounts for variable system performance for various buildvariants
 exec_timeout_secs: 1800 # 6 minutes is the longest we'll ever run
 
+# don't schedule tests if a commit only changes non-code files 
+ignore:
+  - "*.rst" 
+  - "LICENSE"
+  - "THIRD-PARTY-NOTICES"
```
ShaneHarvey: Add:
```
- ".gitignore"
- ".travis.yml"
- "CONTRIBUTING.rst"
- "README.rst"
- "RELEASE.rst"
- "./tools/"
```
and remove "*.rst".

tychoish: I'm fine with this change.

ShaneHarvey: Can you also add the changelog: "doc/changelog.rst". I don't think we'd ever put doctests in there and it often gets updated by itself.

PR #335, opened 2017-09-06
PYTHON-1371 - The tailable cursor cannot get document through __getitem__(index) on MongoDB v3.4

caosiyang:
Fix issue and add test case.

### DIFF
--- a/pymongo/cursor.py
+++ b/pymongo/cursor.py
@@ -590,6 +590,7 @@ class Cursor(object):
             clone = self.clone()
             clone.skip(index + self.__skip)
             clone.limit(-1)  # use a hard limit
+            clone.__query_flags &= ~CursorType.TAILABLE  # PYTHON-1371
             for doc in clone:
                 return doc
             raise IndexError("no such item for Cursor instance")
--- a/pymongo/cursor.py
+++ b/pymongo/cursor.py
@@ -590,7 +590,7 @@ class Cursor(object):
             clone = self.clone()
             clone.skip(index + self.__skip)
             clone.limit(-1)  # use a hard limit
-            clone.__query_flags &= ~CursorType.TAILABLE  # PYTHON-1371
+            clone.__query_flags &= ~CursorType.TAILABLE_AWAIT  # PYTHON-1371
             for doc in clone:
                 return doc
             raise IndexError("no such item for Cursor instance")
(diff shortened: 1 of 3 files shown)


pymongo/cursor.py
```diff
@@ -590,6 +590,7 @@ def __getitem__(self, index):
             clone = self.clone()
             clone.skip(index + self.__skip)
             clone.limit(-1)  # use a hard limit
+            clone.__query_flags &= ~CursorType.TAILABLE  # PYTHON-1371
```
behackett: I haven't tested it, but I assume awaitData has the same problem. This should probably use CursorType.TAILABLE_AWAIT.

test/test_cursor.py
```diff
@@ -1133,6 +1133,12 @@ def test_tailable(self):
 
         self.assertEqual(3, db.test.count())
 
+        # __getitem__(index)
+        cursor2 = db.test.find(cursor_type=CursorType.TAILABLE)
```
behackett: Test both TAILABLE and TAILABLE_AWAIT.

test/test_cursor.py
```diff
@@ -1133,6 +1133,12 @@ def test_tailable(self):
...
+        self.assertEqual(4, cursor2[0]["x"])
+        self.assertEqual(5, cursor2[1]["x"])
+        self.assertEqual(6, cursor2[2]["x"])
+
```
ShaneHarvey: Can test this with `CursorType.TAILABLE_AWAIT` too? And another for `__getitem__(slice)`, eg
```python
cursor = db.test.find(cursor_type=CursorType.TAILABLE_AWAIT)
self.assertEqual([4,5], [doc["x"] for doc in cursor[0:2]])
```

pymongo/cursor.py
caosiyang: @behackett You are right. I miss the 'awaitData' bit.

test/test_cursor.py
caosiyang: @behackett OK.  I will test them later.

test/test_cursor.py
caosiyang: @ShaneHarvey Thanks for your advice. I will expand the tests later.

caosiyang:
I commit some new code, please review again.

ShaneHarvey:
Thanks, @caosiyang. I'm running a job on our internal CI to test against all MongoDB versions since our Travis only tests against the default installed MongoDB. I'll let you know the results.

ShaneHarvey:
Thanks @caosiyang LGTM! Merged: https://github.com/mongodb/mongo-python-driver/commit/db57f671dcf018dc90f4b7fc8e3db75d62ee1766

PR #336, opened 2017-09-20
making Database._list_collections a public method :: issue #PYTHON-1355

Jagrut:
reference to the ticket
https://jira.mongodb.org/browse/PYTHON-1355?jql=project%20%3D%20PYTHON%20AND%20assignee%20in%20(EMPTY)


ShaneHarvey:
Hi @Jagrut, thanks for the pull but PYTHON-1355 involves more than just making `_list_collections` public. We need the need method to comply with the [enumerate collections specification](https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst). 

I think the high level API should be:
```python
def list_collections(self, filter=None, batch_size=None, session=None):
```
The spec also contains a detailed [test plan](https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst#test-plan). Let us know if you want to read through the spec and continue working on this.

Jagrut:
Thanks for quick response. I will read and fix this.

Jagrut:
I read the links. what i understood is we have to make it compatible with all the version of mongo server like in some versions it returns the list of documents and in some version it returns the cursor (like pointer) and adding the functionality of filter with capped and non capped collection, etc.
correct me if i am wrong and if you guys have decided something please share with me.

ShaneHarvey:
Yes, you're correct. The new `list_collections` method will need to work and have the same output on each supported MongoDB version (2.6 and up). `Database.collection_names` and `Database._list_collections` already have compatibility logic in them that, I think, follows the [getting collection names](https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst#getting-collection-names) part of the spec.

You can use those as a guide to implement `list_collections` with support for the `filter` and `batch_size` options.

Jagrut:
@ShaneHarvey . Just want to clarify the meaning of batch size . suppose I have 11 collections in 1 database. so, if i specify batch_size 5 only 5 collections should be returned (mongo shell is returning only 5 it is not providing any cursor) or it should return all the 11 collections with 3 rounds. I have done the first approach which follows the mongo shell by running this command 
`db.runCommand( { listCollections: 1, cursor : { batchSize: 5 } } )`

pymongo/database.py
```diff
@@ -541,36 +541,39 @@ def command(self, command, value=1, check=True,
...
                 "id": res["cursor_id"],
                 "firstBatch": data,
                 "ns": coll.full_name,
             }
-            return CommandCursor(coll, cursor, sock_info.address)
+            return CommandCursor(coll, cursor, sock_info.address, batch_size=batch_size)
 
-    def collection_names(self, include_system_collections=True, session=None):
+    def collection_names(self, include_system_collections=True,
+                         session=None, batch_size=0, filter=None):
```
ShaneHarvey: Let's not change this method signature just yet. The spec says:
"If a driver already has a method to perform one of the listed tasks, there is no need to change it. Do not break backwards compatibility when adding new methods."

However, you can change this method and [`Collection.options`](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/collection.py#L1954) to use `list_collections`.

pymongo/database.py
```diff
@@ -541,36 +541,39 @@ def command(self, command, value=1, check=True,
...
+                         session=None, batch_size=0):
         """Internal listCollections helper."""
-        criteria = criteria or {}
-        cmd = SON([("listCollections", 1), ("cursor", {})])
-        if criteria:
-            cmd["filter"] = criteria
+        filter = filter or {}
+        cmd = SON([("listCollections", 1),
+                   ("cursor", {"batchSize": batch_size}),
+                   ("filter",filter)])
```
ShaneHarvey: `filter` should only be added if it's non-empty and `batch_size` only if it's not `None`:
```python
cmd = SON([("listCollections", 1), ("cursor", {})])
if filter:
    cmd["filter"] = filter
if batch_size is not None:
    cmd["cursor"]["batchSize"] = batch_size
```

pymongo/database.py
```diff
@@ -541,36 +541,39 @@ def command(self, command, value=1, check=True,
...
             with self.__client._tmp_session(session, close=False) as s:
                 cursor = self._command(
                     sock_info, cmd, slave_okay, session=s)["cursor"]
                 return CommandCursor(coll, cursor, sock_info.address, session=s,
-                                     explicit_session=session is not None)
+                                     explicit_session=session is not None, batch_size=batch_size)
         else:
             coll = self["system.namespaces"]
+            if "name" in filter:
+                filter["name"] = coll.database.name + "." + filter["name"]
```
ShaneHarvey: Nice, but you also need to validate that `filter["name"]` is a string. A good error might be:
```python
if not isinstance(filter["name"], string_type):
    raise TypeError("filter['name'] must be a string on MongoDB 2.6")
```
https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst#filters

pymongo/database.py
```diff
@@ -586,8 +589,14 @@ def collection_names(self, include_system_collections=True, session=None):
                 ReadPreference.PRIMARY) as (sock_info, slave_okay):
 
             wire_version = sock_info.max_wire_version
-            results = self._list_collections(
-                sock_info, slave_okay, session=session)
+            if batch_size and include_system_collections and wire_version <= 2:
+                batch_size = 2*batch_size - 1
+            elif batch_size and wire_version <= 2:
+                batch_size = 2*batch_size + 1
```
ShaneHarvey: This change is not needed.

pymongo/database.py
```diff
@@ -541,36 +541,39 @@ def command(self, command, value=1, check=True,
                                  check, allowable_errors, read_preference,
                                  codec_options, session=session, **kwargs)
 
-    def _list_collections(self, sock_info, slave_okay, criteria=None,
-                          session=None):
+    def list_collections(self, sock_info, slave_okay, filter=None,
```
ShaneHarvey: I suggest keeping `_list_collections` and having `list_collections` select a socket like this:
```python
    def list_collections(self, filter=None, batch_size=None, session=None):
        """Get info about the collections in this database."""
        with self.__client._socket_for_reads(
                ReadPreference.PRIMARY) as (sock_info, slave_okay):
             return self._list_collections(
                sock_info, slave_okay, filter=filter, batch_size=batch_size,
                 session=session)
```

Jagrut:
@ShaneHarvey . any reviews on new commit?

pymongo/database.py
```diff
@@ -541,36 +541,66 @@ def command(self, command, value=1, check=True,
...
+        # Iterating the cursor to completion may require a socket for getmore.
+        # Ensure we do that outside the "with" block so we don't require more
+        # than one socket at a time.
+        names = [result["name"] for result in results]
+        if wire_version <= 2:
+            # MongoDB 2.6 and older return index namespaces and collection
+            # namespaces prefixed with the database name.
+            names = [n[len(self.__name) + 1:] for n in names
+                     if n.startswith(self.__name + ".") and "$" not in n]
+        return names
```
ShaneHarvey: This needs to return the full `listCollections` response documents not just the names. Because we need to modify those documents before giving them back to the user, I think it makes sense for this method to be a generator:

```python
# run self._list_collections
for result in results:
    if wire_verison <= 2:
        name = result["name"]
        if "$" in name:
            continue
        result["name"] = name.split(".", 1)[1]
    yield result
```

pymongo/database.py
```diff
@@ -586,8 +616,10 @@ def collection_names(self, include_system_collections=True, session=None):
                 ReadPreference.PRIMARY) as (sock_info, slave_okay):
 
             wire_version = sock_info.max_wire_version
-            results = self._list_collections(
-                sock_info, slave_okay, session=session)
+
+            results = self._list_collections(sock_info,
+                                             slave_okay,
+                                             session=session)
```
ShaneHarvey: The ` if wire_version <= 2:` check on line 628 is unnecessary now that `_list_collections` strips the database name on MongoDB 2.6. Also, this should use the public `list_collections` method now.

Jagrut:
@ShaneHarvey  kindly view second last commit (6309093)

behackett:
I'm prefer that we not add a batch_size option to list_collections. The list_indexes helper doesn't have one and the in-progress list_databases helper (which also returns a CommandCursor) can't have one since there isn't actually a cursor behind it. The spec says it's optional, let's skip it.

PR #337, opened 2017-09-28
fix Regex __eq__ method

gzcf:


test/test_bson.py
```diff
@@ -788,6 +788,11 @@ def test_regex_from_native(self):
     def test_regex_hash(self):
         self.assertRaises(TypeError, hash, Regex('hello'))
 
+    def test_regex_comparison(self):
+        re1 = Regex('a')
+        re2 = Regex('b')
+        self.assertNotEqual(re1, re2)
```
ShaneHarvey: Can you add some more assertions that make sure flags are compared properly too? And an `self.assertEqual` with two equivalent `Regex`'s?

gzcf: Of course

ShaneHarvey:
Whoops, linked the wrong ticket in the commit message. It should be [PYTHON-1386](https://jira.mongodb.org/browse/PYTHON-1386).

PR #340, opened 2017-10-24
PYTHON 1310

Smackle:
Added docs for _create_or_update user. However, the dropUser command is in the remove_user functions which is getting deprecated(?) which has caused me some confusion on which docs to write for removing users.


behackett:
Hi. Thanks for you contribution. In PyMongo 4 the helpers for creating and deleting users are going away. They are no longer needed with more modern versions of MongoDB. The work for PYTHON-1310 is to create example documentation that shows users how to do the same thing using pymongo.database.Database.command. You probably want to leave that to use (MongoDB employees) since we'll want to document quirks across all the versions of MongoDB the driver continues to support.

PR #343, opened 2017-11-13
PYTHON-1411 Reproducibility fix for mongodb-python-driver

ramiscoding:
The copyright message constructed depends on the system current date. So the build changes everytime(Every year technically) it runs. This dependency should not be there as part of reproducible-builds effort and has been fixed by removing the date alone.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -30,7 +30,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'PyMongo'
-copyright = u'MongoDB, Inc. 2008-{0}. MongoDB, Mongo, and the leaf logo are registered trademarks of MongoDB, Inc'.format(datetime.date.today().year)
+copyright = u'MongoDB, Inc. MongoDB, Mongo, and the leaf logo are registered trademarks of MongoDB, Inc'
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -30,7 +30,7 @@
 
 # General information about the project.
 project = u'PyMongo'
-copyright = u'MongoDB, Inc. 2008-{0}. MongoDB, Mongo, and the leaf logo are registered trademarks of MongoDB, Inc'.format(datetime.date.today().year)
+copyright = u'MongoDB, Inc. MongoDB, Mongo, and the leaf logo are registered trademarks of MongoDB, Inc'
```
behackett: I see. The date is still necessary, but it can be static. Change it to "2008-present" instead of removing it.

ramiscoding: ok will do that

doc/conf.py
```diff
@@ -30,7 +30,7 @@
```
ShaneHarvey: Can you remove the now unused "import datetime" above?

ramiscoding: yaa sure

behackett:

PR #345, opened 2017-12-07
Up-to-date tools documentation

defrank:
* Mark some ORM-like layers as no longer maintained: MongoKit, MongoAlchemy, Minimongo, Manga
* Add testing driver: mongomock
* Add ORM-like layer: uMongo


defrank:
:+1:

PR #347, opened 2018-01-30
Allow fields to be a set

rdb:
It was previously possible to use a set to specify fields, but this is no longer being allowed due to new validation constraints.

It seems that the validation is being unnecessarily strict.  [PYTHON-946](https://jira.mongodb.org/browse/PYTHON-946) covered the relaxation of the validation to allow sequences; this trivial change extends that to sets as well.  (I'm presuming there was a good reason for not changing it to the more generic `collections.Collection`; otherwise, that would be a better fit.)

### DIFF
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -235,7 +235,7 @@ def _fields_list_to_dict(fields, option_name):
     if isinstance(fields, collections.Mapping):
         return fields
 
-    if isinstance(fields, collections.Sequence):
+    if isinstance(fields, (collections.Sequence, collections.Set)):
         if not all(isinstance(field, string_type) for field in fields):
             raise TypeError("%s must be a list of key names, each an "
                             "instance of %s" % (option_name,
--- a/test/test_collection.py
+++ b/test/test_collection.py
@@ -1743,6 +1743,16 @@ class TestCollection(IntegrationTest):
 
         self.assertTrue("hello" in db.test.find_one(projection=["hello"]))
         self.assertTrue("hello" not in db.test.find_one(projection=["foo"]))
+
+        self.assertTrue("hello" in db.test.find_one(projection=("hello",)))
+        self.assertTrue("hello" not in db.test.find_one(projection=("foo",)))
+
+        self.assertTrue("hello" in db.test.find_one(projection=set(["hello"])))
+        self.assertTrue("hello" not in db.test.find_one(projection=set(["foo"])))
+
+        self.assertTrue("hello" in db.test.find_one(projection=frozenset(["hello"])))
+        self.assertTrue("hello" not in db.test.find_one(projection=frozenset(["foo"])))
+
         self.assertEqual(["_id"], list(db.test.find_one(projection=[])))
 
         self.assertEqual(None, db.test.find_one({"hello": "foo"}))


ShaneHarvey:
It looks like PyMongo 2.X accepted any iterable, including set, iterator, and even string:
```python
>>> c.test.test.find_one(projection='ab')
{u'a': 1, u'_id': ObjectId('5a70b23894761678aac76528'), u'b': 1}
>>> c.test.test.find_one(projection=['a', 'b'])
{u'a': 1, u'_id': ObjectId('5a70b23894761678aac76528'), u'b': 1}
>>> c.test.test.find_one(projection=set(['a', 'b']))
{u'a': 1, u'_id': ObjectId('5a70b23894761678aac76528'), u'b': 1}
>>> c.test.test.find_one(projection=list(['a', 'b']))
{u'a': 1, u'_id': ObjectId('5a70b23894761678aac76528'), u'b': 1}
>>> c.test.test.find_one(projection=iter(['a', 'b']))
{u'a': 1, u'_id': ObjectId('5a70b23894761678aac76528'), u'b': 1}
>>> c.test.test.find_one(projection=[1, 'b'])
  File "/Users/shane/git/mongo-python-driver/venv-python2.7/lib/python2.7/site-packages/pymongo/collection.py", line 1083, in find_one
    *args, **kwargs).max_time_ms(max_time_ms)
  File "/Users/shane/git/mongo-python-driver/venv-python2.7/lib/python2.7/site-packages/pymongo/collection.py", line 1291, in find
    return Cursor(self, *args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/venv-python2.7/lib/python2.7/site-packages/pymongo/cursor.py", line 180, in __init__
    fields = helpers._fields_list_to_dict(fields)
  File "/Users/shane/git/mongo-python-driver/venv-python2.7/lib/python2.7/site-packages/pymongo/helpers.py", line 256, in _fields_list_to_dict
    "each an instance of %s" % (basestring.__name__,))
TypeError: fields must be a list of key names, each an instance of basestring
```
PyMongo 3 only accepts Mapping and Sequence, meaning sets and iterators are not accepted (but strings still are):
```python
>>> c.test.test.find_one(projection=set(['a', 'b']))
  File "pymongo/collection.py", line 1223, in find_one
    cursor = self.find(filter, *args, **kwargs)
  File "pymongo/collection.py", line 1399, in find
    return Cursor(self, *args, **kwargs)
  File "pymongo/cursor.py", line 168, in __init__
    projection = helpers._fields_list_to_dict(projection, "projection")
  File "pymongo/helpers.py", line 246, in _fields_list_to_dict
    "list of key names" % (option_name,))
TypeError: projection must be a mapping or list of key names
>>> c.test.test.find_one(projection=iter(['a', 'b']))
  File "pymongo/collection.py", line 1223, in find_one
    cursor = self.find(filter, *args, **kwargs)
  File "pymongo/collection.py", line 1399, in find
    return Cursor(self, *args, **kwargs)
  File "pymongo/cursor.py", line 168, in __init__
    projection = helpers._fields_list_to_dict(projection, "projection")
  File "pymongo/helpers.py", line 246, in _fields_list_to_dict
    "list of key names" % (option_name,))
TypeError: projection must be a mapping or list of key names
>>> c.test.test.find_one(projection='ab')
{u'a': 1, u'_id': ObjectId('5a70b23894761678aac76528'), u'b': 1}
```

If we're going to change this again for backwards compatibility can you just use `collections.Iterable` and make sure it works with an iterator too? 
```python
     if isinstance(fields, collections.Iterable):
         as_dict = {}
         for field in fields:
             if not isinstance(field, string_type):
                 raise TypeError("%s must be a list of key names, each an "
                             "instance of %s" % (option_name,
                                                 string_type.__name__))
             as_dict[field] = 1
         return as_dict
```
It would be great to add some test cases for this too.

For the record, I don't like that we accept a single string but at this point there are likely a lot of people relying on that behavior.

rdb:
I am happy to change it to use `collections.Collection`.  I am not certain what the use case would be for accepting any iterable; it seem safer to err on the restrictive side, unless pymongo 2.0 accepted non-container iterables as well?

A string is currently accepted, but it doesn't do what one might think it does; it makes every individual character in the string an individual field specification.  I'd personally recommend exempting string as an allowed type.

pymongo/helpers.py
```diff
@@ -235,7 +235,8 @@ def _fields_list_to_dict(fields, option_name):
     if isinstance(fields, collections.Mapping):
         return fields
 
-    if isinstance(fields, collections.Sequence):
+    if isinstance(fields, collections.Sequence) or \
+       isinstance(fields, collections.Set):
```
ShaneHarvey: Please reduce this to a single isinstance:
```python
    if isinstance(fields, (collections.Sequence, collections.Set)):
```

rdb:
Done.  Also added lines to the test case for tuple and frozenset.

PR #348, opened 2018-02-10
Use formatting operator to format error message

kalvdans:
The following one-character patch fixes the ugly looking message:
<pre>
pymongo.errors.InvalidURI: ("Invalid URI scheme: URI must begin with '%s' or '%s'", ('mongodb://', 'mongodb+srv://'))
</pre>


ShaneHarvey:
Thanks @kalvdans!

PR #350, opened 2018-03-16
Improve the check for legal utf8 in the bson module.

stephan-hof:
As discussed here https://jira.mongodb.org/browse/PYTHON-1504 I make a pull request to bring the 'isLegalUTF8' check in line with the python behaviour.

I also added two tests.
* one which tries all bit combinations within the 244 range
* one which tries just a few samples to trigger the problem.

I think the first one is the better one, since it goes much deeper, with the negative side effect of letting the test suite run for more than 5 minutes. Which is typically not acceptable, so I disabled it via 'return' at the top of it.

I enabled 'allow edits from maintainers' on my branch here. So feel free to adjust the unittests to your needs before merging.

### DIFF
--- a/test/test_bson.py
+++ b/test/test_bson.py
@@ -650,6 +650,89 @@ class TestBSON(unittest.TestCase):
             z = {iso8859_bytes: "hello"}
             self.assertRaises(InvalidStringData, BSON.encode, z)
 
+    # Verify that python and bson have the same understanding of
+    # legal utf-8 if the first byte is 0xf4 (244)
+    @staticmethod
+    def _py_is_legal_utf8(x):
+        try:
+            x.decode('utf-8')
+            return True
+        except UnicodeDecodeError:
+            return False
+
+    @staticmethod
+    def _bson_is_legal_utf8(x):
+        try:
+            BSON.encode({'x': x})
+            return True
+        except InvalidStringData:
+            return False
+
+    @unittest.skipIf(PY3, "python3 has strong separation between bytes/unicode")
+    def test_legal_utf8_full_coverage(self):
+        # this tests takes 400 seconds. Which is too long to run each time.
+        # However it is the only one which covers all possible bit combinations
+        # in the 244 space.
+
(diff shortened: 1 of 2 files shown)


test/test_bson.py
```diff
@@ -650,6 +650,89 @@ def test_utf8(self):
...
+    @staticmethod
+    def _bson_is_legal_utf8(x):
+        try:
+            BSON.encode({'x': x})
+            return True
+        except InvalidStringData:
+            return False
+
+    @unittest.skipIf(PY3, "python3 has strong separation between bytes/unicode")
+    def test_legal_utf8_full_coverage(self):
```
ShaneHarvey: Since this test takes so long we will probably separate it out into its own file and only run it once for each python 2.x version. But don't worry about this for now.

test/test_bson.py
```diff
@@ -650,6 +650,89 @@ def test_utf8(self):
...
+        # However it is the only one which covers all possible bit combinations
+        # in the 244 space.
+
+        b1 = chr(0xf4)
+
+        for b2 in map(chr, range(255)):
+            m2 = b1 + b2
+            self.assertEqual(
+                self._py_is_legal_utf8(m2),
+                self._bson_is_legal_utf8(m2)
```
ShaneHarvey: This assertion appears a lot. I suggest adding an assert helper method:
```python
    def assertUTF8MatchesPython(self, string):
        self.assertEqual(
            self._py_is_legal_utf8(string),
            self._bson_is_legal_utf8(string),
            string)
```

ShaneHarvey:
Thanks @stephan-hof! Feel free to make the change I requested. If not, no worries we'll take care of the rest!

stephan-hof:
Thanks for the feedback. I refactored the assert into a separate method, not sure about the naming schema though.

Regarding the separation of the long running test. I'll leave this to you, since I'm pretty sure you have already something in your mind how to do that.

behackett:
Merged to master. Thanks for the patch, and your patience.

PR #352, opened 2018-05-14
Fix compatibility issues with Python >= 3.6

egao1980:
- \$ and \. trigger SyntaxError in Python 3.6
- % format operator is not compatible with Python 3.6

### DIFF
--- a/tools/benchmark.py
+++ b/tools/benchmark.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """MongoDB benchmarking suite."""
+from __future__ import print_function
 
 import time
 import sys
@@ -87,7 +88,7 @@ def timed(name, function, args=[], setup=None):
         function(*args)
         times.append(time.time() - start)
     best_time = min(times)
-    print "%s%d" % (name + (60 - len(name)) * ".", per_trial / best_time)
+    print("{:s}{:d}".format(name + (60 - len(name)) * ".", per_trial / best_time))
     return best_time
 
 
--- a/pymongo/database.py
+++ b/pymongo/database.py
@@ -34,8 +34,8 @@ from pymongo.son_manipulator import SONManipulator
 from pymongo.write_concern import WriteConcern
 
 
-_INDEX_REGEX = {"name": {"$regex": "^(?!.*\$)"}}
-_SYSTEM_FILTER = {"filter": {"name": {"$regex": "^(?!system\.)"}}}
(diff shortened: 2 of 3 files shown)


ShaneHarvey:
Hi @egao1980. What do you mean "$ and . trigger SyntaxError in Python 3.6"?
```
$ python3.6
Python 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 05:52:31)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> {"$regex": "^(?!.*\$)"}
{'$regex': '^(?!.*\\$)'}
>>> ^D
$ python3.7
Python 3.7.0b4 (v3.7.0b4:eb96c37699, May  2 2018, 04:13:13)
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> {"$regex": "^(?!.*\$)"}
{'$regex': '^(?!.*\\$)'}
>>>
```

egao1980:
Markdown ate my slashes. Python 3.6 do not allow non standard escapes. So we need to use raw strings or escape the backslash

egao1980:
This breaks strict tests with 'python -We ...'

python3.6 -We
Python 3.6.5 (default, Mar 30 2018, 06:42:10)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> '\s'
  File "<stdin>", line 1
SyntaxError: invalid escape sequence \s
>>> '\$'
  File "<stdin>", line 1
SyntaxError: invalid escape sequence \$
>>> '\.'
  File "<stdin>", line 1
SyntaxError: invalid escape sequence \.

tools/benchmark.py
```diff
@@ -87,7 +88,7 @@ def timed(name, function, args=[], setup=None):
         function(*args)
         times.append(time.time() - start)
     best_time = min(times)
-    print "%s%d" % (name + (60 - len(name)) * ".", per_trial / best_time)
+    print("{:s}{:d}".format(name + (60 - len(name)) * ".", per_trial / best_time))
```
ShaneHarvey: This should be `{0}{1}` to support python 2.6. or you can just use `%` formatting as before which is valid on all python versions.

ShaneHarvey:
Ah I see. That is a problem. My only comment is on the change to  `benchmark.py`.

PR #354, opened 2018-05-21
Fall back to sha1 if md5 is not available

daviddavis:
In FIPS mode, md5 is not available. In such cases, fall back to sha1.
See https://docs.python.org/3/library/hashlib.html for more info.

This change is based on [something similar to what celery
did](https://github.com/celery/kombu/pull/711/commits/7e164f7f5e12521812a10ac268d8bec8e51e3a13).

This is needed for https://jira.mongodb.org/browse/PYTHON-1520.

### DIFF
--- a/bson/objectid.py
+++ b/bson/objectid.py
@@ -35,7 +35,12 @@ from bson.tz_util import utc
 def _machine_bytes():
     """Get the machine portion of an ObjectId.
     """
-    machine_hash = hashlib.md5()
+    try:
+        machine_hash = hashlib.md5()
+    except ValueError:
+        # md5 isn't available in FIPS mode
+        machine_hash = hashlib.sha1()
+
     if PY3:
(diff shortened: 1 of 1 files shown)


behackett:
Thanks for the pull request. We're currently planning a change to all drivers to avoid this problem and it likely won't involve sha1, which will almost certainly be banned by FIPS in the future. Whatever we change we make will happen in PyMongo's next release. Follow https://jira.mongodb.org/browse/PYTHON-1521 (or https://jira.mongodb.org/browse/PYTHON-1520) for updates.

behackett:
@daviddavis - I've resolved https://jira.mongodb.org/browse/PYTHON-1521. We've switched to using FNV-1a, a non-cryptographic hash avoiding FIPS issues now and in the future.

daviddavis:
@behackett awesome. thank you for the update!

PR #355, opened 2018-06-11
PYTHON-1565 Add helpers and options for Change Streams Follow-on Work

prashantmital:
WIP PR that makes it easier to track changes for https://jira.mongodb.org/browse/PYTHON-1565.


pymongo/change_stream.py
```diff
@@ -101,3 +123,100 @@ def __enter__(self):
...
+
+    @property
+    def _database(self):
+        return self._target
+
+
+class ChangeStreamClient(ChangeStreamDatabase):
+    """ Class for creating a change stream on all collections on a cluster. """
+
+    def _full_pipeline(self, options={}):
```
ShaneHarvey: We prefer not to use mutable default values for arguments. Consider setting allChangesForCluster another way.

pymongo/change_stream.py
```diff
@@ -16,35 +16,63 @@
...
     .. mongodoc:: changeStreams
     """
-    def __init__(self, collection, pipeline, full_document,
+    def __init__(self, target, pipeline, full_document,
                  resume_after=None, max_await_time_ms=None, batch_size=None,
-                 collation=None, session=None):
-        self._collection = collection
+                 collation=None, start_at_operation_time=None, session=None):
+        # Validate inputs
+        if not isinstance(pipeline, list):
```
ShaneHarvey: `None` should be a valid value for `pipeline` since that is the default for the watch methods.

pymongo/collection.py
```diff
@@ -2219,7 +2219,7 @@ def aggregate_raw_batches(self, pipeline, **kwargs):
         return self._aggregate(pipeline, RawBatchCommandCursor, 0,
                                None, False, **kwargs)
 
-    def watch(self, pipeline=None, full_document='default', resume_after=None,
+    def watch(self, pipeline=[], full_document='default', resume_after=None,
```
ShaneHarvey: I suggest changing all the `pipeline` defaults back to `None`. It's generally bad practice to use mutable default arguments so we avoid them (even though it doesn't really matter in this case since the first thing we do is take a deepcopy of the pipeline).

Also since `None` is no longer a valid value that means we technically broke compatibility.

pymongo/change_stream.py
```diff
@@ -16,35 +16,63 @@
...
 
     .. versionadded: 3.6
     .. mongodoc:: changeStreams
     """
-    def __init__(self, collection, pipeline, full_document,
+    def __init__(self, target, pipeline, full_document,
                  resume_after=None, max_await_time_ms=None, batch_size=None,
-                 collation=None, session=None):
-        self._collection = collection
+                 collation=None, start_at_operation_time=None, session=None):
```
ShaneHarvey: Since we always pass all of these arguments when creating a ChangeStream can you make all of them required?

pymongo/change_stream.py
```diff
@@ -16,35 +16,63 @@
...
+        self._target = target
         self._pipeline = copy.deepcopy(pipeline)
         self._full_document = full_document
         self._resume_token = copy.deepcopy(resume_after)
         self._max_await_time_ms = max_await_time_ms
         self._batch_size = batch_size
         self._collation = collation
+        self._start_at_operation_time = (
+            start_at_operation_time or self._default_start_at_operation_time()
+        )
```
ShaneHarvey: Optimization: If resumeAfter is specified then we don't need to run isMaster initialize `_start_at_operation_time`. At least I think that's what the spec says.

pymongo/change_stream.py
```diff
@@ -16,35 +16,63 @@
...
         self._session = session
         self._cursor = self._create_cursor()
 
-    def _full_pipeline(self):
+    def _default_start_at_operation_time(self):
+        """
+        Return the current operationTime timestamp from the server.
+        """
+        return self._database.command('isMaster')['operationTime']
```
ShaneHarvey: The spec says:
> If neither startAtOperationTime nor resumeAfter are specified, and the server version is >= 4.0 drivers MUST set a default startAtOperationTime. A default timestamp can be obtained from the operationTime field on any server response (e.g. isMaster). This allows change streams to be resumed before the first notification is received.

So I think we have to make sure the isMaster.maxWireVersion field is at least 7.

pymongo/change_stream.py
```diff
@@ -16,35 +16,63 @@
...
+    def _create_cursor(self):
+        """
+        Instantiate the cursor corresponding to this ChangeStream.
+        """
+        raise NotImplementedError
+
+    def _full_pipeline(self, options={}):
         """Return the full aggregation pipeline for this ChangeStream."""
-        options = {}
+        options['startAtOperationTime'] = self._start_at_operation_time
```
ShaneHarvey: I don't think we should add startAtOperationTime if the value is None (which it can be if we're connected to <=4.0 or if resumeAfter is specified).

pymongo/change_stream.py
prashantmital: Actually, the default for watch methods is `[]`.

pymongo/change_stream.py
```diff
@@ -101,3 +179,59 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
+
+
+class CollectionChangeStream(ChangeStream):
```
behackett: I don't think we can change the type returned by Collection.watch. I think that should continue to return ChangeStream. Your abstract class can be an internal type line _ChangeStream.

pymongo/change_stream.py
```diff
@@ -16,48 +16,125 @@
...
+            options.update(inject_options)
         full_pipeline = [{'$changeStream': options}]
         full_pipeline.extend(self._pipeline)
         return full_pipeline
 
+    def _run_aggregation_cmd(self, session, explicit_session):
+        """Run the full aggregation pipeline for this ChangeStream and return
+        the corresponding CommandCursor.
+        """
+        with self._database.client._socket_for_reads(self._target._read_preference_for(session)) as (sock_info, slave_ok):
```
behackett: Let's break this up to keep the lines under 80 chars.

read_preference = self._target._read_preference_for(session)
client = self._database.client
with client._socket_for_read(read_preference) as (sock_info, slave_ok):

pymongo/change_stream.py
```diff
@@ -16,48 +16,125 @@
...
+                read_concern=self._target.read_concern,
+                collation=self._collation,
+                session=session,
+                client=self._database.client)
+
+            cursor = result["cursor"]
+
+            if (self._start_at_operation_time is None and
+                self._resume_token is None and
+                cursor.get("_id", None) is None and
```
behackett: get() returns None if the key doesn't exist. No need to pass a second argument.

pymongo/change_stream.py
```diff
@@ -16,48 +16,125 @@
...
+            pipeline = self._full_pipeline()
+            cmd = SON([("aggregate", self._aggregation_target),
+                       ("pipeline", pipeline),
+                       ("cursor", {})])
+
+            result = sock_info.command(
+                self._database.name,
+                cmd,
+                slave_ok,
+                self._target._read_preference_for(session),
```
behackett: If you refactor as I suggested above, this can just be 'read_preference'.

pymongo/change_stream.py
```diff
@@ -16,48 +16,125 @@
...
+                self._resume_token is None and
+                cursor.get("_id", None) is None and
+                sock_info.max_wire_version >= 7):
+                self._start_at_operation_time = result["operationTime"]
+
+            ns = cursor["ns"]
+            _, collname = ns.split(".", 1)
+            aggregation_collection = self._database.get_collection(
+                collname, codec_options=self._target.codec_options,
+                read_preference=self._target._read_preference_for(session),
```
behackett: Same here about reusing read_preference.

pymongo/change_stream.py
```diff
@@ -16,48 +16,125 @@
...
+        """The argument to pass to the aggregate command."""
+        raise NotImplementedError
+
+    @property
+    def _database(self):
+        """The database against which the aggregation commands for
+        this ChangeStream will be run. """
+        raise NotImplementedError
+
+    def _full_pipeline(self, inject_options=None):
```
behackett: inject_options reads like a verb. Maybe just call it 'options', then in the method do:

options = options or {}

test/test_change_stream.py
```diff
@@ -390,5 +504,188 @@ def test_read_concern(self):
...
+            for change, expected_changes in zip(changes, test["result"]["success"]):
+                assert_dict_is_subset(change, expected_changes)
+        
+        finally:
+            # Check for expected events
+            results = self.listener.results
+            for expectation in test["expectations"]:
+                for idx, (event_type, event_desc) in enumerate(expectation.iteritems()):
+                    results_key = event_type.split("_")[1]
+                    event = results[results_key][idx] if len(results[results_key]) else None
```
ShaneHarvey: if ... len(results[results_key]) > idx?

pymongo/change_stream.py
behackett: Oh, I see why this won't work. Well, I still think you should pick a noun for the argument name. Maybe just 'extra_options'?

test/test_change_stream.py
```diff
@@ -390,5 +504,188 @@ def test_read_concern(self):
...
+
+        except OperationFailure as exc:
+            if test["result"].get("error") is None:
+                self.fail("Encountered unexpected error")
+            expected_code = test["result"]["error"]["code"]
+            self.assertEqual(exc.code, expected_code)
+
+        else:
+            # Check for expected output from change streams
+            for change, expected_changes in zip(changes, test["result"]["success"]):
```
ShaneHarvey: After this assert that `changes` and `test["result"]` have the same length. (after so that you get the more informative error first)

test/test_change_stream.py
```diff
@@ -390,5 +504,188 @@ def test_read_concern(self):
...
+                    # Run specified operations
+                    run_operation(self.client, operation)
+                num_expected_changes = len(test["result"]["success"])
+                changes = [
+                    change_stream.next() for _ in range(num_expected_changes)
+                ]
+
+        except OperationFailure as exc:
+            if test["result"].get("error") is None:
+                self.fail("Encountered unexpected error")
```
ShaneHarvey: Just issue a blank `raise` so we know what error caused the test to fail and where it was raised.

pymongo/change_stream.py
prashantmital: The 'correct' choice probably takes into account how users check the 'type' of what is returned by `watch()`. 

If we use a base class `_ChangeStream` and call the change stream associated with collections `ChangeStream`, `isinstance(coll.watch(), ChangeStream)` will be `True`, but `isinstance(db.watch(), ChangeStream)` and `isinstance(client.watch(), ChangeStream) will be `False` (as they will inherit from `_ChangeStream` base class). 

I think it is much more intuitive to have everything returned by `watch()` methods be of the same (implicit) type (as determined by `isinstance`). I can't think of a non-hacky way to do that while simultaneously ensuring that the explicit type from `coll.watch()` is `ChangeStream`.

With that in mind, how do you want to proceed?

pymongo/change_stream.py
prashantmital: I was channeling the Zen of Python 😄 

["Explicit is better than implicit"](https://www.python.org/dev/peps/pep-0020/)

pymongo/change_stream.py
behackett: That's a good point. isinstance checks will still pass. I take back what I wrote. Just note this in the changelog and I think we'll be OK.

pymongo/change_stream.py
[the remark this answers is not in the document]
ShaneHarvey: To follow: I don't think `cursor` will ever have a "_id" field.

pymongo/database.py
```diff
@@ -403,6 +404,14 @@ def _fix_outgoing(self, son, collection):
             son = manipulator.transform_outgoing(son, collection)
         return son
 
+    def watch(self, pipeline=None, full_document='default', resume_after=None,
```
behackett: Need a docstring here to tell users how to use this method.

pymongo/mongo_client.py
```diff
@@ -689,6 +690,14 @@ def _server_property(self, attr_name):
 
         return getattr(server.description, attr_name)
 
+    def watch(self, pipeline=None, full_document='default', resume_after=None,
```
behackett: Also need a docstring here. You'll also have to add watch to the list of methods in doc/api/pymongo/mongo_client.rst.

behackett:
BTW, you can run "python setup.py doc" to build the documentation so you can see if the formatting is correct. You'll have to install Sphinx first.

pymongo/change_stream.py
prashantmital: Adding the following to the changelog:
```
- The :meth:`pymongo.collection.Collection.watch` method now returns an
  instance of the :class:`~pymongo.change_stream.CollectionChangeStream`
  class which is a subclass of :class:`~pymongo.change_stream.ChangeStream`.
```

pymongo/change_stream.py
```diff
@@ -16,48 +16,125 @@
...
         """Return the full aggregation pipeline for this ChangeStream."""
         options = {}
+
         if self._full_document is not None:
             options['fullDocument'] = self._full_document
         if self._resume_token is not None:
             options['resumeAfter'] = self._resume_token
+        if self._start_at_operation_time is not None:
+            options['startAtOperationTime'] = self._start_at_operation_time
```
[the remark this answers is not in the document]
prashantmital: I have used a different pattern from the one you suggested here to eliminate the need for the `inject_options`/`options` kwargs. Let me know if that is satisfactory. We can discuss offline why I wasn't a fan of what you suggested (adding a `_cluster_changes` attr).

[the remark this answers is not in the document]
ShaneHarvey: Great I like your way better!

pymongo/collection.py
```diff
@@ -2435,16 +2435,10 @@ def watch(self, pipeline=None, full_document='default', resume_after=None,
         .. _change streams specification:
```
ShaneHarvey: For the `:Parameters:` docs above add:
```
          - `start_at_operation_time` (optional): If provided, the resulting
            change stream will only return changes that occurred at or after
            the specified :class:`~bson.timestamp.Timestamp`. Requires
            MongoDB >= 4.0.

#etc...

        .. versionchanged:: 3.7
           Added the ``start_at_operation_time`` parameter.
        .. versionadded:: 3.6
```

prashantmital: I added a different description for this option, but I think yours is slightly clearer to understand, albeit a bit verbose. Going with yours!

test/test_change_stream.py
```diff
@@ -390,5 +490,187 @@ def test_read_concern(self):
...
+    )
+    return cmd(**arguments)
+
+
+def assert_dict_is_subset(superdict, subdict):
+    """Check that subdict is a subset of superdict."""
+    exempt_fields = ["documentKey", "_id"]
+    for key, value in iteritems(subdict):
+        if not superdict.has_key(key):
+            assert False
```
ShaneHarvey: `has_key` doesn't exist in Python 3. Just do `if key not in superdict:`

PR #356, opened 2018-06-25
ENH: add helper method to run tests on specific storage engines

prashantmital:
Closes [PYTHON-1597](https://jira.mongodb.org/browse/PYTHON-1597)

This allows us to specify (with the help of a decorator on the `ClientContext`) the storage engine that the server must be running with for a given test to be executed.

### Motivation

Spec-tests introduced in [PYTHON-1565](https://jira.mongodb.org/browse/PYTHON-1565) can only be run when the server is using wired Tiger. These tests failed when the MMAPv1 engine was in use, resulting in [failures on evergreen](https://evergreen.mongodb.com/task/mongo_python_driver_tests_storage_engines__python_version~2.7_storage_engine~mmapv1_test_4.0_standalone_9b632c776c3ace91dd6fd71dd98164d77881f766_18_06_25_02_59_16).

### Comments

I chose to use the `db.serverStatus()` command to obtain storage engine information because: 

1) other drivers already follow this approach in their test infrastructure (e.g. PHP)
2) inferring storage engine information from the output of `serverStatus` seemed to present fewer ambiguities than the output of `getCmdLineOpts`. The latter tends to be confusing because the `getCmdLineOpts['parsed']` dict seems to always contain the `wiredTiger` key and we need to check whether the `engine` key has been set to `mmapv1` to ascertain the storage engine. This seemed a bit counter-intuitive to me but if you're not a fan of calling `serverStatus`, I can use the `cmd_line` information as well.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -162,6 +162,7 @@ class ClientContext(object):
         self.nodes = set()
         self.replica_set_name = None
         self.cmd_line = None
+        self.server_status = None
         self.version = Version(-1)  # Needs to be comparable with Version
         self.auth_enabled = False
         self.test_commands_enabled = False
@@ -225,6 +226,7 @@ class ClientContext(object):
 
             try:
                 self.cmd_line = self.client.admin.command('getCmdLineOpts')
+                self.server_status = self.client.admin.command('serverStatus')
             except pymongo.errors.OperationFailure as e:
                 msg = e.details.get('errmsg', '')
                 if e.code == 13 or 'unauthorized' in msg or 'login' in msg:
@@ -247,6 +249,7 @@ class ClientContext(object):
 
                 # May not have this if OperationFailure was raised earlier.
                 self.cmd_line = self.client.admin.command('getCmdLineOpts')
+                self.server_status = self.client.admin.command('serverStatus')
 
             self.ismaster = ismaster = self.client.admin.command('isMaster')
             self.sessions_enabled = 'logicalSessionTimeoutMinutes' in ismaster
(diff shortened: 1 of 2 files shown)


test/__init__.py
```diff
@@ -247,6 +249,7 @@ def _init_client(self):
 
                 # May not have this if OperationFailure was raised earlier.
                 self.cmd_line = self.client.admin.command('getCmdLineOpts')
+                self.server_status = self.client.admin.command('serverStatus')
```
ShaneHarvey: You only need to run serverStatus once after creating the authenticated client. Just un-indent this line and remove the serverStatus on line 229.

test/__init__.py
```diff
@@ -424,6 +427,15 @@ def require_connection(self, func):
             "Cannot connect to MongoDB on %s" % (self.pair,),
             func=func)
 
+    def require_storage_engine(self, engine):
+        """Run a test only if the server is running the specified storage
+        engine (as determined by the `db.serverStatus` command)."""
+        #server_status = self.client.admin.command("serverStatus")
+        #current_engine = server_status["storageEngine"]["name"]
```
ShaneHarvey: Remove comments.

test/__init__.py
```diff
@@ -424,6 +427,15 @@ def require_connection(self, func):
...
+        return self._require(
+            lambda: engine == self.server_status["storageEngine"]["name"],
```
ShaneHarvey: This will fail when `self.server_status` is None. What if we just keep track of the storage engine instead of the entire server status response?

test/test_change_stream.py
```diff
@@ -54,6 +54,7 @@ class TestClusterChangeStream(IntegrationTest):
 
     @classmethod
     @client_context.require_version_min(4, 0, 0, -1)
+    @client_context.require_storage_engine("wiredTiger")
```
ShaneHarvey: Change streams work of storage engines other than wiredTiger. Like inMemory for example. Can you make these `@client_context.require_not_mmap` instead?

prashantmital:
Ignoring your comments for now as this will need a rethink since it cannot presently work with a sharded cluster.

ShaneHarvey:
> Ignoring your comments for now as this will need a rethink since it cannot presently work with a sharded cluster.

For now we don't need it to work on a sharded cluster. All we need is to skip the spec tests when connected to a *standalone server* running with MMAP.

ShaneHarvey:
Thanks!

PR #361, opened 2018-07-08
Document find_one_and_update return value on no match.

jaraco:
Today I was implementing find_one_and_update, and I wanted to verify the behavior if no value was matched, but found the docs did not adequately document the behavior. I verified the behavior on MongoDB 3.6 and the latest pymongo and documented it here.

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -3019,9 +3019,14 @@ class Collection(common.BaseObject):
           ...    {'_id': 665}, {'$inc': {'count': 1}, '$set': {'done': True}})
           {u'_id': 665, u'done': False, u'count': 25}}
 
-        By default :meth:`find_one_and_update` returns the original version of
-        the document before the update was applied, or ``None`` if no document
-        matches the filter. To return the updated (or inserted in the case of
+        Returns ``None`` if no document matches the filter.
+
+          >>> db.test.find_one_and_update(
+          ...    {'_exists': False}, {'$inc': {'count': 1}})
+
+        When the filter matches, by default :meth:`find_one_and_update`
+        returns the original version of the document before the update was
+        applied. To return the updated (or inserted in the case of
         *upsert*) version of the document instead, use the *return_document*
         option.
 
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -3022,7 +3022,7 @@ class Collection(common.BaseObject):
         By default :meth:`find_one_and_update` returns the original version of
         the document before the update was applied, or ``None`` if no document
         matches the filter. To return the updated (or inserted in the case of
-        *upsert*), version of the document instead, use the *return_document*
+        *upsert*) version of the document instead, use the *return_document*
         option.
 
           >>> from pymongo import ReturnDocument
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -2431,10 +2431,10 @@ class Collection(common.BaseObject):
         """Watch changes on this collection.
 
         Performs an aggregation with an implicit initial ``$changeStream``
-        stage and returns a 
+        stage and returns a
         :class:`~pymongo.change_stream.CollectionChangeStream` cursor which
         iterates over changes on this collection.
-        
+
         Introduced in MongoDB 3.6.
 
         .. code-block:: python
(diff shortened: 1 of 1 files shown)


XinMa1:
OperationFailure: Performing an update on the path '_id' would modify the immutable field '_id'
I find this.When i used find_one_and_update.
mongo: 4.0 pymong: 3.6.1
Anyone can help me.

pymongo/collection.py
```diff
@@ -3030,6 +3030,9 @@ def find_one_and_update(self, filter, update,
           ...     return_document=ReturnDocument.AFTER)
           {u'_id': u'userid', u'seq': 1}
 
+        Regardless the value of **return_document**, if no document matches
+        the filter, the return value is ``None``.
```
ShaneHarvey: This is not true. For example when `upsert=True` and no document matched the filter a new document will be inserted and the return value will be `None` with `BEFORE` but it will be equal to the inserted document with `AFTER`.

This is why the documentation for `AFTER` down below says:        
"If :attr:`ReturnDocument.AFTER`, returns the updated or **inserted** document."

Do you think it would be an improvement to say:
"If :attr:`ReturnDocument.AFTER`, returns the updated or **upserted** document."

jaraco: Hi Shane. Good call.

The condition I was trying to document was the condition where `upsert=False` but `return_document` is `AFTER`, where the return value is `None`. I'll take another cut and try to address all these conditions together.

jaraco:
I think this latest cut more accurately describes the expectation.

XinMa1:
Thanks, I know the reason.
I use save() then  raise  DuplicateKeyError.
       But, save()  add "_id" to data.
       Why don't first to validation the unique key.  Do find. 
       And then to pretreatment data. Do update.

pymongo/collection.py
```diff
@@ -3080,8 +3082,7 @@ def find_one_and_update(self, filter, update,
             document matches the query. Defaults to ``False``.
           - `return_document`: If
             :attr:`ReturnDocument.BEFORE` (the default),
-            returns the original document before it was updated, or ``None``
-            if no document matches. If
```
ShaneHarvey: Can you keep this part? I think it's useful to keep the `None` return here too because this doc string is so large.

pymongo/collection.py
```diff
@@ -3020,8 +3020,10 @@ def find_one_and_update(self, filter, update,
           {u'_id': 665, u'done': False, u'count': 25}}
 
         By default :meth:`find_one_and_update` returns the original version of
-        the document before the update was applied. To return the updated
-        version of the document instead, use the *return_document* option.
+        the document before the update was applied, or ``None`` if no document
+        matches the filter. To return the updated (or inserted in the case of
+        *upsert*), version of the document instead, use the *return_document*
+        option.
```
ShaneHarvey: This part SGTM.

pymongo/collection.py
jaraco: I can, but then the description is asymmetric, describing the `None` return value if no document matches for `BEFORE` but not describing the return value for `AFTER`, which once again makes it unclear what the return value is for `AFTER`. By omitting this statement here, the user is directed to the statement above which applies to both `BEFORE` and `AFTER`.

In order to be consistent, one would also need to describe something similar for the `AFTER`, which makes the docstring even more verbose and redundant.

I'll follow your advice to illustrate what I mean.

jaraco:
I've pushed a new commit that restores the BEFORE description and adds a corresponding description about AFTER. My preference would be to accept the PR on the parent commit, which attempts to be more precise and concise about the behavior, describing the default behavior and allowing the options (upsert mainly) to override this behavior.

pymongo/collection.py
ShaneHarvey: Thanks for pointing that out. I agree that we can revert your latest commit but I think it still needs one change to fix the asymmetry. The docs in the parent commit say:

> By default :meth:`find_one_and_update` returns the original version of
the document before the update was applied, or ``None`` if no document
matches the filter. To return the updated (or inserted in the case of
*upsert*), version of the document instead, use the *return_document*
option.

The asymmetry is still present in this paragraph: it describes `None` for `BEFORE` but not for `AFTER` just like before this change. Can you amend this section to mention when `None` is returned with `AFTER`?

Also can you remove the comma here: `*upsert*), version`?

jaraco: > Also can you remove the comma here?

Yes, easy.

> The asymmetry is still present in this paragraph.

Hmm. I had written it to mean "By default find_one_and_update returns None if no document matches the filter" (independent of "by default find_one_and_update returns the original version of the document before the update was applied"), even though they are in the same sentence. But I can see how your reading betrays my intention. I'll hit it again.

jaraco:
@ShaneHarvey Would you consider my latest suggestion?

PR #362, opened 2018-07-09
PYTHON-1603 Truncate large datetimes properly

ShaneHarvey:


bson/__init__.py
```diff
@@ -837,7 +837,7 @@ def _datetime_to_millis(dtm):
     if dtm.utcoffset() is not None:
         dtm = dtm - dtm.utcoffset()
     return int(calendar.timegm(dtm.timetuple()) * 1000 +
-               dtm.microsecond / 1000)
+               dtm.microsecond // 1000)
```
ShaneHarvey: Should line 821 also use `//`?

test/test_bson.py
```diff
@@ -474,6 +474,13 @@ def test_datetime_encode_decode(self):
         dt2 = BSON.encode({"date": dt1}).decode()["date"]
         self.assertEqual(dt1, dt2)
 
+    def test_large_datetime_truncation(self):
+        # Ensure that a large datetime is truncated correctly.
+        dt1 = datetime.datetime(9999, 1, 1, 1, 1, 1, 999999)
+        dt2 = BSON.encode({"date": dt1}).decode()["date"]
+        self.assertEqual(dt2.microsecond, 999000)
+        self.assertEqual(dt2.second, dt1.second)
```
ShaneHarvey: I can reproduce this on Windows with 32-bit and 64-bit Python 3.X. So this fix does actually work.

PR #363, opened 2018-07-24
Fix - check if dbname (arg) in cache instead database (module)

didibz:


ShaneHarvey:
Thanks @didibz!

didibz:
@ShaneHarvey My pleasure, thanks for accepting the change.

PR #365, opened 2018-08-07
Removed ambiguity of aggregate command.

ayushijain218:
Removed ambiguity for aggregate command in MongoDB 3.6 or higher by adding a note in documentation.


ShaneHarvey:
Hi @ayushijain218 thanks for the pull! Can you explain what ambiguity this change is removing? The added note seems out of place to me because the "cursor" option is not discussed anywhere else in the page.

PR #366, opened 2018-08-09
Remove Travis workarounds

ShaneHarvey:
Looks like https://github.com/travis-ci/travis-ci/issues/943 was fixed.
https://github.com/travis-ci/travis-ci/issues/5485 isn't needed since we don't support Python 3.2 .


prashantmital:
🎉 for clean code!

PR #367, opened 2018-08-10
Avoid failure to instantiate an OperationFailure if details is not a dict

Colin-b:
Hi,

Even if I assume OperationFailure details parameter was always expected to be a dictionary, the check was never explicit,

Prior to version 3.7.0 there was nothing preventing users to instantiate an OperationFailure (or any sub class) with details being something else than a dictionary.

This lack of validation resulted in mongomock providing a string value in this parameter, thus being incompatible with pymongo==3.7.0 or 3.7.1

This PR aim at fixing this regression.

### DIFF
--- a/pymongo/errors.py
+++ b/pymongo/errors.py
@@ -126,8 +126,13 @@ class OperationFailure(PyMongoError):
     """
 
     def __init__(self, error, code=None, details=None):
+        """
+
+        :param details: The complete error document returned by the server.
+        Error labels are expected to be provided with errorLabels key.
+        """
         error_labels = None
-        if details is not None:
+        if details is not None and isinstance(details, dict):
             error_labels = details.get('errorLabels')
         super(OperationFailure, self).__init__(
             error, error_labels=error_labels)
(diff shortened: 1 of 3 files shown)


ShaneHarvey:
Hi @Colin-b, thanks for contributing! Unfortunately, I'm not sure it makes sense for pymongo to allow `details` to be any type just because it happened to work before. The [documentation for `details`](http://api.mongodb.com/python/3.6.0/api/pymongo/errors.html#pymongo.errors.OperationFailure.details) says:

> The complete error document returned by the server.
>
> Depending on the error that occurred, the error document may include useful information beyond just the error message. When connected to a mongos the error document may contain one or more subdocuments if errors occurred on multiple shards.

I think that makes it pretty clear that `details` is a document (ie a dict).

Also, it looks like this issue was fixed by https://github.com/mongomock/mongomock/pull/398 which should be included in the next mongomock release. Perhaps you can try testing with their develop branch?

Colin-b:
Hi @ShaneHarvey ,

Thanks for your response, indeed this is not a big deal as mongomock is not something that I use in production, it was more about adding-back compatibility with other projects that may exists and have improper implementation (not that I know any).

It's up to you, I would rather throw an explicit error in case the wrong type of parameter is provided. Should I let you close it?

PR #369, opened 2018-08-10
PYTHON-1624 Fix failing $out test on MongoDB latest

ShaneHarvey:
There's no need to project out the _id field in this test. I opened [SERVER-36598](https://jira.mongodb.org/browse/SERVER-36598) for the server bug.

### DIFF
--- a/test/test_collection.py
+++ b/test/test_collection.py
@@ -1616,9 +1616,8 @@ class TestCollection(IntegrationTest):
         self.assertEqual([{'foo': [1, 2]}], list(result))
 
         # Test write concern.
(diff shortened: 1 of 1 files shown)


prashantmital:
Can we project out some other field instead (doesn't have to be `_id`). Projecting out a subset of the entire document seems to test an additional code-path here and we might be losing that with this change?

ShaneHarvey:
There's no difference in codepath because PyMongo doesn't don't do anything special with the `pipeline`. It just passes it along to the server. So adding a $project would just add unnecessary noise.

PR #371, opened 2018-08-14
PYTHON-1577 Allow applications to register a custom server selector

prashantmital:
Closes [PYTHON-1577](https://jira.mongodb.org/browse/PYTHON-1577)

### DIFF
--- a/doc/examples/server_selection.rst
+++ b/doc/examples/server_selection.rst
@@ -78,37 +78,31 @@ Server Selection Process
 ------------------------
 
 This section dives deeper into the server selection process for reads and
-writes. Let's first consider the case of **writes**. Assuming it is connected to a
-reachable database, the driver performs the following operations (in order)
-during the selection process:
-
-* Select all writeable servers from the list of known hosts. For a replica set
-  this is the primary, while for a sharded cluster this can be an arbitrary
-  number of ``mongos``
-
-* Apply the user-defined server selector function. Note that the custom server
-  selector is **not** called if there are no servers left from the previous
-  filtering stage.
-
-* Apply the ``localThresholdMS`` setting to the list of remaining hosts. This
-  whittles the host list down to only contain servers whose latency is at most
-  ``localThresholdMS`` milliseconds higher than the lowest observed latency.
-
-* Select a server at random from the remaining host list. The desired
-  operation is then performed against the selected server.
+writes. In the case of a write, the driver performs the following operations
+(in order) during the selection process:
 
 
-For the case of **reads** the process is identical except for the first step.
+#. Select all writeable servers from the list of known hosts. For a replica set
+   this is the primary, while for a sharded cluster this is all the known mongoses.
+
+#. Apply the user-defined server selector function. Note that the custom server
+   selector is **not** called if there are no servers left from the previous
+   filtering stage.
+
+#. Apply the ``localThresholdMS`` setting to the list of remaining hosts. This
+   whittles the host list down to only contain servers whose latency is at most
+   ``localThresholdMS`` milliseconds higher than the lowest observed latency.
+
+#. Select a server at random from the remaining host list. The desired
+   operation is then performed against the selected server.
+
+
+In the case of **reads** the process is identical except for the first step.
 Here, instead of selecting all writeable servers, we select all servers
 matching the user's :class:`~pymongo.read_preferences.ReadPreference` from the
 list of known hosts. As an example, for a 3-member replica set with a
 :class:`~pymongo.read_preferences.Secondary` read preference, we would select
 all available secondaries.
 
-
-In both cases, the user-defined server selector function is applied
-to host pool *after* accounting for the
-:class:`~pymongo.read_preferences.ReadPreference`, but *before* applying the
-latency window specified by ``localThresholdMS``.
 
 .. _server selection algorithm: https://docs.mongodb.com/manual/core/read-preference-mechanics/
--- a/doc/examples/server_selection.rst
+++ b/doc/examples/server_selection.rst
@@ -2,7 +2,7 @@ Server Selector Example
 =======================
 
 Users can exert fine-grained control over the `server selection algorithm`_
-by setting the `serverSelector` option on the :class:`~pymongo.MongoClient`
+by setting the `server_selector` option on the :class:`~pymongo.MongoClient`
 to an appropriate callable. This example shows how to use this functionality
 to prefer servers running on ``localhost``.
 
@@ -70,7 +70,7 @@ server selector.
 
 .. doctest::
 
-   >>> client = MongoClient(serverSelector=server_selector)
+   >>> client = MongoClient(server_selector=server_selector)
 
 
 
--- a/doc/examples/server_selection.rst
+++ b/doc/examples/server_selection.rst
@@ -13,12 +13,6 @@ to prefer servers running on ``localhost``.
    custom server selectors can have unintended consequences such as degraded
    read/write performance.
 
-.. note::
-
-   Directing queries to servers running on ``localhost`` can be desirable
-   when using a sharded cluster with multiple ``mongos``, as locally run
-   queries are likely to see lower latency and higher throughput.
-
 
 .. testsetup::
 
--- a/doc/examples/server_selection.rst
+++ b/doc/examples/server_selection.rst
@@ -7,6 +7,12 @@ to an appropriate callable. This example shows how to use this functionality
 to prefer servers running on ``localhost``.
 
 
+.. warning::
+
+   Use of custom server selector functions is a power user feature. Misusing
+   custom server selectors can have unintended consequences such as degraded
+   read/write performance.
+
 .. note::
 
    Directing queries to servers running on ``localhost`` can be desirable
@@ -29,6 +35,15 @@ To start, we need to write the server selector function that will be used.
 The server selector function should accept a list of
 :class:`~pymongo.server_description.ServerDescription` objects and return a
 list of server descriptions that are suitable for the read or write operation.
+A server selector must not create or modify
+:class:`~pymongo.server_description.ServerDescription` objects, and must return
+the selected instances unchanged.
+
+In this example, we write a server selector that prioritizes servers running on
+``localhost``. This can be desirable when using a sharded cluster with multiple
+``mongos``, as locally run queries are likely to see lower latency and higher
+throughput. Please note, however, that it is highly dependent on the
+application if preferring ``localhost`` is beneficial or not.
 
 In addition to comparing the hostname with ``localhost``, our server selector
 function accounts for the following edge cases:
@@ -45,7 +60,7 @@ function accounts for the following edge cases:
 
 
 The described server selection logic is implemented in the following server
-selctor function:
+selector function:
 
 
 .. doctest::
@@ -62,9 +77,7 @@ selctor function:
 
 
 Finally, we can create a :class:`~pymongo.MongoClient` instance with this
-server selector. Optionally, we can also specify a custom value for the
-``localThresholdMS`` option which is also used by the
-`server selection algorithm`_.
+server selector.
 
 
 .. doctest::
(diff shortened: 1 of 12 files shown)


behackett:
This needs two things:
- A bunch more tests
- Comprehensive documentation with examples.

test/test_server_selection.py
```diff
@@ -32,5 +34,15 @@
 class TestAllScenarios(create_selection_tests(_TEST_PATH)):
     pass
 
+
+class TestCustomServerSelectorFunction(unittest.TestCase):
+    def test_invalid_server_selector(self):
+        _bad_server_selector = 1
```
ShaneHarvey: _bad_server_selector?

test/test_server_selection.py
```diff
@@ -32,5 +34,15 @@
...
+        with self.assertRaisesRegex(ValueError, "must be a callable"):
+            _ = MongoClient(
+                connect=False, serverSelector=list()
+            )
```
ShaneHarvey: I think this can fit on a single line.

test/test_server_selection.py
```diff
@@ -32,5 +36,79 @@
...
+        client = rs_or_single_client(
+            serverSelector=custom_selector,
+            event_listeners=[listener]
+        )
+        self.addCleanup(client.close)
+        coll = client.get_database(
+            'testdb', read_preference=ReadPreference.NEAREST).coll
+        self.addCleanup(client.drop_database, 'testdb')
+
+        # Hack to wait long enough for server to populate node list.
```
prashantmital: @ShaneHarvey @behackett I notice we are `sleep`ing elsewhere in the test suite as well. Does that mean doing so is the only way to ensure that the node list is populated properly? If so, how can one optimize the amount of time we pause execution (assuming we want as small a number here as possible to minimize test suite running time)?

prashantmital:
> This needs two things:
>
>> A bunch more tests
>> Comprehensive documentation with examples.

I have added tests. If there are other cases that are not covered by these, please let me know.
I shall work on documentation and examples tomorrow.

prashantmital:
Some open questions/TODOs:

- [] Update methods that use `TopologyDescription.apply_selector` to use ``TopologyDescription.apply_selectors` if appropriate. AFAICT, this is only applicable to `TopologyDescription.has_readable_server`. However, propagating the `serverSelector` to the `TopologyDescription` would require somewhat significant changes, so I'd like to be sure before making this change. 

- [] Determine (and implement) the correct behavior of `Topology._error_message` when using the `serverSelector`. Since we can no longer definitively point to a particular server selector function as being responsible for the driver failing to select any servers, we will need to decide what needs to be displayed to the user.

test/test_server_selection.py
ShaneHarvey: You could add a [TopologyListener](http://api.mongodb.com/python/3.7.1/api/pymongo/monitoring.html#pymongo.monitoring.TopologyListener) that waits for the topology description, [TopologyDescriptionChangedEvent.new_description.known_servers](http://api.mongodb.com/python/3.7.1/api/pymongo/monitoring.html#pymongo.monitoring.TopologyDescriptionChangedEvent.new_description),  to include all known servers ([isMaster.hosts](https://docs.mongodb.com/manual/reference/command/isMaster/#isMaster.hosts)). The topology listener could then notify the test via a [condition variable](https://docs.python.org/2/library/threading.html#condition-objects).

pymongo/topology.py
```diff
@@ -219,9 +221,8 @@ def select_server(self,
                       server_selection_timeout=None,
                       address=None):
         """Like select_servers, but choose a random server if several match."""
-        return random.choice(self.select_servers(selector,
-                                                 server_selection_timeout,
-                                                 address))
+        return random.choice(self.select_servers(
+            selector, server_selection_timeout, address))
```
ShaneHarvey: No need to change this line.

pymongo/topology.py
```diff
@@ -187,16 +187,18 @@ def select_servers(self,
...
     def _select_servers_loop(self, selector, timeout, address):
         """select_servers() guts. Hold the lock when calling this."""
+        selectors = [selector, self._settings.server_selector]
+
         now = _time()
         end_time = now + timeout
-        server_descriptions = self._description.apply_selector(
-            selector, address)
+        server_descriptions = self._description.apply_selectors(
+            selectors, address)
```
ShaneHarvey: I think it would be clearer to pass the custom server selector as a another argument instead of a list.

test/test_server_selection.py
```diff
@@ -32,5 +36,74 @@
 class TestAllScenarios(create_selection_tests(_TEST_PATH)):
     pass
 
+
+class TestCustomServerSelectorFunction(IntegrationTest):
+    def test_functional_select_max_port_number_host(self):
+        # Selector that returns server with highest port number.
+        def custom_selector(selection):
```
ShaneHarvey: The spec says this should input and output a list of server descriptions:
> The function takes as a parameter a list of server descriptions representing the suitable servers for the read or write operation, and returns a list of server descriptions that should still be considered suitable.

We need to follow that part of the spec.

pymongo/topology_description.py
```diff
@@ -240,18 +240,21 @@ def apply_local_threshold(selection):
...
             # Ignore the read preference, but apply localThresholdMS.
-            return apply_local_threshold(
+            servers = apply_local_threshold(
                 Selection.from_topology_description(self))
         else:
-            return apply_local_threshold(
+            servers = apply_local_threshold(
                 selector(Selection.from_topology_description(self)))
 
+        custom_selector = custom_selector or (lambda x: x)
```
jyemin: The spec requires that the custom selector is applied prior to applying the latency window check.

> 3: Find suitable servers by topology type and operation type
> 4: Filter the suitable servers by calling the optional, application-provided server selector.
> 5: If there are any suitable servers, choose one at random from those within the latency window and return it; otherwise, continue to the next step

Looks like here it's being applied after the latency window check and before the random selection of a single server.

jyemin:
The code appears to be compliant with the spec now, but is there a way to add a unit test that asserts that 

1. the custom selector is passed all servers that match the read preference/all mongo server, regardless of latency threshold
2. the latency window filter is applied to the server returned from the custom selector

prashantmital:
@ShaneHarvey can you review the documentation?

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,82 @@
...
+   ...     if not servers:
+   ...         return server_descriptions
+   ...     return servers
+
+
+
+Finally, we can create a :class:`~pymongo.MongoClient` instance with this
+server selector. Optionally, we can also specify a custom value for the
+``localThresholdMS`` option which is also used by the
+`server selection algorithm`_.
```
ShaneHarvey: I don't think we need to mention passing localThresholdMS here.

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,82 @@
...
+by setting the `serverSelector` option on the :class:`~pymongo.MongoClient`
+to an appropriate callable. This example shows how to use this functionality
+to prefer servers running on ``localhost``.
+
+
+.. note::
+
+   Directing queries to servers running on ``localhost`` can be desirable
+   when using a sharded cluster with multiple ``mongos``, as locally run
+   queries are likely to see lower latency and higher throughput.
```
ShaneHarvey: Can you replace the note with a ".. warning::" that says this is a power user feature? Misusing a custom server selectors may lead to serious issues such as heart attack or stroke. Please consult your doctor before attempting to use server selectors. etc...

The existing note can be moved to a normal paragraph in the "Example: ..." section below. Also please mention that it's highly dependent on the application if preferring localhost is beneficial or not.

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,82 @@
...
+.. doctest::
+
+   >>> client = MongoClient(serverSelector=server_selector)
+
+
+.. note::
+
+   The user-defined server selector function is applied to host pool *after*
+   accounting for the :class:`~pymongo.read_preferences.ReadPreference`,
+   but *before* applying the latency window specified by ``localThresholdMS``.
```
ShaneHarvey: I think we should be a little more specific since there's not much documentation on server selection. Describe what happens with a write operation, selector gets a list of all known writeable servers (replica set primary, or a number of mongoses for a sharded cluster).

The describe what happens for a read, the selector gets a list of all known servers matching the read preference.

Describe what happens if the server selector returns multiple servers, one server, or an empty list.

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,82 @@
...
+
+* No servers are running on ``localhost``. In this case, we allow the default
+  server selection logic to prevail by passing through all server descriptions
+  received by the server selector. Failure to do so would render the client
+  unable to communicate with MongoDB in the event that no servers were running
+  on ``localhost``.
+
+
+The described server selection logic is implemented in the following server
+selctor function:
```
ShaneHarvey: selctor -> selector

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,82 @@
...
+.. _server selection algorithm: https://docs.mongodb.com/manual/core/read-preference-mechanics/
+
+
+Example: Selecting Servers Running on ``localhost``
+---------------------------------------------------
+
+To start, we need to write the server selector function that will be used.
+The server selector function should accept a list of
+:class:`~pymongo.server_description.ServerDescription` objects and return a
+list of server descriptions that are suitable for the read or write operation.
```
ShaneHarvey: Mention that a server selector must not create or modify the ServerDescription objects. It must return the instances unchanged.

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,82 @@
...
+
+In addition to comparing the hostname with ``localhost``, our server selector
+function accounts for the following edge cases:
+
+* The list of server descriptions passed to the server selector is empty. This
+  can happen if the server selection process starts before the client has
+  received successful heartbeats from any of the ``mongod`` or ``mongos``.
```
ShaneHarvey: @jyemin is this how the Java selector works? Should we bother calling the custom selector with an empty list? I'd like us to work the same way across drivers.

test/utils_selection_tests.py
```diff
@@ -139,6 +139,18 @@ def get_topology_type_name(scenario_def):
...
+def get_topology_settings_dict(
+        monitor_class=MockMonitor, heartbeat_frequency=HEARTBEAT_FREQUENCY,
+        pool_class=MockPool, **kwargs):
+    settings = dict(
+        monitor_class=monitor_class,
+        heartbeat_frequency=heartbeat_frequency,
+        pool_class=pool_class
+    )
+    settings.update(kwargs)
+    return settings
```
ShaneHarvey: This would be more concise like:
```
 def get_topology_settings_dict(**kwargs):
    settings = dict(
        monitor_class=MockMonitor,
        heartbeat_frequency=HEARTBEAT_FREQUENCY,
        pool_class=MockPool)
    ...

File: doc/examples/server_selection.rst (hunk 13 above)
Comment (ShaneHarvey) [reply doc/examples/server_selection.rst]: Discussed with Jeff offline, the Java driver does not call the selector if the list of servers is empty. Can you make python do the same?

**Comment by prashantmital** [comment]
@ShaneHarvey made the necessary changes. Please review.

File: pymongo/mongo_client.py [hunk 15]
```diff
@@ -321,6 +321,14 @@ def __init__(
             Defaults to ``-1``, meaning no maximum. If maxStalenessSeconds
             is set, it must be a positive integer greater than or equal to
             90 seconds.
+          - `serverSelector`: (callable) Optional, user-provided function that
```
ShaneHarvey: `server_selector`, we only use camelCase for kwargs that are also URI options.

pymongo/common.py
```diff
@@ -572,6 +579,7 @@ def validate_tzinfo(dummy, value):
     'tzinfo': validate_tzinfo,
     'username': validate_string_or_none,
     'password': validate_string_or_none,
+    'serverselector': validate_is_callable,
```
ShaneHarvey: `server_selector`.

pymongo/mongo_client.py
prashantmital: Ah, I see... I noticed only some kwargs used camelCase but was unable to identify a theme (or didn't spend enough time trying to do so). Good to know how things are organized.

prashantmital:
@ShaneHarvey renamed the kwarg to `server_selector`. Ready for another look.

prashantmital:
@ShaneHarvey not sure what caused the failure, but it all worked fine when I restarted the job. I also confirmed that I can build the docs fine locally.

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,114 @@
...
+matching the user's :class:`~pymongo.read_preferences.ReadPreference` from the
+list of known hosts. As an example, for a 3-member replica set with a
+:class:`~pymongo.read_preferences.Secondary` read preference, we would select
+all available secondaries.
+
+
+In both cases, the user-defined server selector function is applied
+to host pool *after* accounting for the
+:class:`~pymongo.read_preferences.ReadPreference`, but *before* applying the
+latency window specified by ``localThresholdMS``.
```
ShaneHarvey: I don't think we need to repeat this part since it says the "process is identical" above.

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,114 @@
...
+
+* Apply the ``localThresholdMS`` setting to the list of remaining hosts. This
+  whittles the host list down to only contain servers whose latency is at most
+  ``localThresholdMS`` milliseconds higher than the lowest observed latency.
+
+* Select a server at random from the remaining host list. The desired
+  operation is then performed against the selected server.
+
+
+For the case of **reads** the process is identical except for the first step.
```
ShaneHarvey: "*In* the case of" is much more popular than "for": https://books.google.com/ngrams/graph?content=for+the+case+of%2Cin+the+case+of&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cfor%20the%20case%20of%3B%2Cc0%3B.t1%3B%2Cin%20the%20case%20of%3B%2Cc0

doc/examples/server_selection.rst
```diff
@@ -0,0 +1,114 @@
...
+
+
+
+Server Selection Process
+------------------------
+
+This section dives deeper into the server selection process for reads and
+writes. Let's first consider the case of **writes**. Assuming it is connected to a
+reachable database, the driver performs the following operations (in order)
+during the selection process:
```
ShaneHarvey: Shorten to:

> This section dives deeper into the server selection process for reads and
writes. In the case of a *write*, the driver performs the following operations (in order) during the selection process:

pymongo/mongo_client.py
```diff
@@ -321,6 +321,14 @@ def __init__(
...
             is set, it must be a positive integer greater than or equal to
             90 seconds.
+          - `server_selector`: (callable) Optional, user-provided function that
+            augments server selection rules. The function should accept as an
+            argument a list of
+            :class:`~pymongo.server_description.ServerDescription` objects and
+            return a list of server descriptions that should be considered
+            suitable for the desired operation.
+
+          .. seealso:: :doc:`/examples/server_selection`
```
ShaneHarvey: server_selector is not really a Read Preference setting since it applies to both reads and writes. Should we move this to the "Other optional parameters can be passed as keyword arguments" section?

prashantmital: Good point. Out of the current subsections we have in the `MongoClient` docstring, the 'Other optional parameters...` section seems most relevant for this option so I'll move it there.

pymongo/common.py
```diff
@@ -572,6 +579,7 @@ def validate_tzinfo(dummy, value):
     'tzinfo': validate_tzinfo,
     'username': validate_string_or_none,
     'password': validate_string_or_none,
+    'server_selector': validate_is_callable,
```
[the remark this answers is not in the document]
prashantmital: Great point. Added a test for this as well.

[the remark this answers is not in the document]
prashantmital: Rather, edited an existing test to ensure it covers this case.

PR #372, opened 2018-08-22
Provide a docstring to MongoClient

fbidu:
Hello, first time contributor here!

I was doing a little debugging on an old code of mine and started to read around the documentation available. To my surprise, MongoClient's docstring was set to 

```
A base class that provides attributes and methods common
to multiple pymongo classes.

SHOULD NOT BE USED BY DEVELOPERS EXTERNAL TO MONGODB.
```

Confused, I went on researching and found out that MongoClient inherits from a base class and does not provide a new docstring, cause the parent's docstring to be displayed.

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -75,8 +75,15 @@ from pymongo.write_concern import DEFAULT_WRITE_CONCERN
 
 
 class MongoClient(common.BaseObject):
-    """Class that provides a full featured client for a MongoDB
-    instance, a replica set, or a set of mongoses."""
+    """
+    A client-side representation of a MongoDB cluster.
+
+    Instances can represent either a standalone MongoDB server, a replica
+    set, or a sharded cluster. Instances of this class are responsible for
+    maintaining up-to-date state of the cluster, and possibly cache
+    resources related to this, including background threads for monitoring,
+    and connection pools.
+    """
     HOST = "localhost"
     PORT = 27017
     # Define order to retrieve options from ClientOptions for __repr__.
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -75,6 +75,8 @@ from pymongo.write_concern import DEFAULT_WRITE_CONCERN
 
 
 class MongoClient(common.BaseObject):
+    """Class that provides a full featured client for a MongoDB
+    instance, a replica set, or a set of mongoses."""
     HOST = "localhost"
     PORT = 27017
     # Define order to retrieve options from ClientOptions for __repr__.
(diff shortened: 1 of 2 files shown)


behackett:
Huh, I've never noticed that this happens when calling help(MongoClient). The docstring for __init__ is what Sphinx generates:

https://api.mongodb.com/python/current/api/pymongo/mongo_client.html

fbidu:
@behackett Yes, I based myself on the docstring for `__init__` to write a simple one for the class. My intention is to just suppress that misleading warning, since init's docstring is very detailed

ShaneHarvey:
Thanks for finding this! Looks like this was changed in Python 3.5. Here's Python 3.4:
```
$ python3.4 -c 'import pymongo;import pydoc; print(pydoc.render_doc(pymongo.MongoClient)[:400])'
Python Library Documentation: class MongoClient in module pymongo.mongo_client

class MongoClient(pymongo.common.BaseObject)
 |  Method resolution order:
 |      MongoClient
 |      pymongo.common.BaseObject
 |      builtins.object
 |
 |  Methods defined here:
 |
 |  __enter__(self)
 |
 |  __eq__(self, other)
 |
 |  __exit__
```
And then Python 3.5+:
```
$ python3.5 -c 'import pymongo;import pydoc; print(pydoc.render_doc(pymongo.MongoClient)[:400])'
Python Library Documentation: class MongoClient in module pymongo.mongo_client

class MongoClient(pymongo.common.BaseObject)
 |  A base class that provides attributes and methods common
 |  to multiple pymongo classes.
 |
 |  SHOULD NOT BE USED BY DEVELOPERS EXTERNAL TO MONGODB.
 |
 |  Method resolution order:
 |      MongoClient
 |      pymongo.common.BaseObject
 |
```

pymongo/mongo_client.py
```diff
@@ -75,6 +75,8 @@
 
 
 class MongoClient(common.BaseObject):
+    """Class that provides a full featured client for a MongoDB
+    instance, a replica set, or a set of mongoses."""
```
ShaneHarvey: I was curious how other drivers document their MongoClient's. I like [Java's description](http://mongodb.github.io/mongo-java-driver/3.8/javadoc/com/mongodb/client/MongoClient.html) the best:
```
A client-side representation of a MongoDB cluster.

Instances can represent either a standalone MongoDB server, a replica
set, or a sharded cluster. Instances of this class are responsible for
maintaining up-to-date state of the cluster, and possibly cache
resources related to this, including background threads for monitoring,
and connection pools.
```
 What do you think of using the same description here?

behackett: SGTM

prashantmital: +1 I like this for the docstring!

fbidu: Nice description! More well thought out than the one I provided

@behackett thanks for finding this changelog, I glanced over the docs a few times but couldn't find the change itself. In case anyone else gets curious, here is the full issue [bpo-15582](https://bugs.python.org/issue15582)

ShaneHarvey:
Thanks @fbidu!

fbidu:
@ShaneHarvey you're welcome!

PR #375, opened 2018-08-29
GridFS support delete_many

erezarnon:


ShaneHarvey:
Thanks for the pull! I've opened [PYTHON-1637](https://jira.mongodb.org/browse/PYTHON-1637) for us to discuss whether we want to add this method. We strive for all of our drivers to have the same or similar APIs so before adding a method like this we'll consult the [GridFS spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst) and see if other drivers implement something similar.

agolin95:
We have decided not to implement this because it may not work for very large query criteria, but thank you for your contribution!

PR #377, opened 2018-10-01
Fix issues found by coverity

omron93:
I've run coverity scan of python-pymongo 3.6.1 - see log [scan-results.txt](https://github.com/mongodb/mongo-python-driver/files/2434573/scan-results.txt)

I've guessed what is the right way to fix the issues which IMHO aren't false positive. In case of other preferences, I will change it.

` pymongo/database.py` - without this fix, `elif` is never reached.
`mongo_client.py` - `session` can be None, so it needs to be checked.
`test_cursor.py` - typo
`utils.py` - currently `pref.tag_sets" is never used. Using tags from both lists.

### DIFF
--- a/pymongo/database.py
+++ b/pymongo/database.py
@@ -1031,7 +1031,7 @@ class Database(common.BaseObject):
 
             opts["roles"] = [self._default_role(read_only)]
 
-        elif read_only:
+        if read_only:
             warnings.warn("The read_only option is deprecated in MongoDB "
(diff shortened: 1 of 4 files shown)


behackett:
That's interesting. We run coverity also, but don't get any output related to pure Python, only the C extensions. Thanks for this. We'll take a deeper look soon.

ShaneHarvey:
Merged thanks @omron93!

PR #382, opened 2018-11-13
[WIP] Flexible BSON encoder/decoder

prashantmital:
# WIP

***

## 09/13/2018

This PR introduces the `BSONWriter` class which is intended to be the primary API for the user to specify serialization routines for custom types. Currently, only the functionality to support the serialization of arbitrary types in document fields has been implemented.


bson/__init__.py
```diff
@@ -775,6 +775,16 @@ def _name_value_to_bson(name, value, check_keys, opts):
             _ENCODERS[type(value)] = func
             return func(name, value, check_keys, opts)
 
+    # Finally, try to encode the object with registered codec.
```
behackett: I think this should go before the "all else fails" for loop.

bson/__init__.py
```diff
@@ -775,6 +775,16 @@ def _name_value_to_bson(name, value, check_keys, opts):
...
+    codec = opts.custom_codec_map.get(type(value))
+    if codec is not None:
+        from bson.flexible_bson.writer import BSONFieldWriter
```
behackett: Are you doing this to avoid a circular import?

bson/flexible_bson/writer.py
```diff
@@ -0,0 +1,288 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""BSON (Binary JSON) encoding.
+
+"""
+
+from collections import deque
+from enum import Enum
```
behackett: This doesn't exist in Python 2.7, and I'd rather not add a required dependency. You can use trivial classes with class level variables instead.

bson/flexible_bson/writer.py
```diff
@@ -0,0 +1,288 @@
...
+_ID_FIELD_C_NAME = _make_name("_id")
+
+
+class BSONWriter(object):
+    def __init__(self, check_keys=False, codec_options=None):
+        self._check_keys = check_keys
+        self._codec_options = codec_options or DEFAULT_CODEC_OPTIONS
+        self.initialize()
+
+    def initialize(self):
```
behackett: Do you need to do this in a separate method? It seems like it could all just be in the constructor.

bson/flexible_bson/writer.py
```diff
@@ -0,0 +1,288 @@
...
+        # Writer state tracker.
+        self._states = deque()
+        self._update_state(_BSONWriterState.INITIAL)
+
+        # Bytestream in which to place BSON encoded bytes.
+        # Deque used to track streams for nested entities.
+        self._streams = deque()
```
behackett: Can you talk a bit about the advantages of having each subdocument or array be a separate byte stream, instead of using one for the entire document?

bson/flexible_bson/writer.py
```diff
@@ -0,0 +1,288 @@
...
+        # Initialize basic BSONWriter.
+        super(BSONDocumentWriter, self).initialize()
+
+    def get_document_id(self):
+        return self._id
+
+    def set_document_id(self, id):
+        self._id = id
+
+    def _get_field_name(self):
```
behackett: This method name is a bit confusing, since it doesn't just return the name, it also writes it into the BSON stream.

.travis.yml
```diff
@@ -11,5 +11,5 @@ python:
 services:
   - mongodb
 
-script: PYMONGO_MUST_CONNECT=1 python setup.py test
+script: PYMONGO_MUST_CONNECT=1 python setup.py test --no_ext
```
behackett: This is because the tests don't currently work when the C extensions are installed? Don't you have the same issue in Evergreen?

bson/flexible_bson/writer.py
ShaneHarvey: Yes I'd like to know too. I think it would be more efficient to reserve space for the bson document/array size, save the position, and patch the actual size in when the element is ended.

bson/flexible_bson/codec.py
```diff
@@ -0,0 +1,44 @@
...
+        generated_id = ObjectId()
+        obj._id = generated_id
+        return generated_id
+
+    def has_id(self, obj):
+        # Return boolean indicating whether `_id` is set on the object.
+        # Defaults to returning boolean of `get_id()` return value.
+        # User can override this behavior to exclude other inadmissible
+        # values for `_id`.
+        return hasattr(obj, '_id') and bool(obj._id)
```
ShaneHarvey: These comments should become docstrings. Also "Defaults to returning boolean of `get_id()` return value" is confusing because this method never calls get_id. I suggest rewriting as:

```suggestion
        try:
            return self.get_id(obj) is not None
        expect AttributeError:
            return False
```

This way a user only needs to override get_id() and they get has_id() for free. Another thing to note is that not all _ids are True so I think comparing to None is a better approach.

bson/flexible_bson/writer.py
```diff
@@ -0,0 +1,288 @@
...
+
+        # If not at top level document, render this container
+        # and concatenate it to the higher level container.
+        if self._state != _BSONWriterState.TOP_LEVEL:
+            bstream = self._stream.getvalue()
+            self._sizes.pop()
+            self._streams.pop()
+            self._insert_bytes(bstream)
+
+    def start_document(self, name=None):
```
ShaneHarvey: Can you add docstrings to all the public methods of this class?

.travis.yml
ShaneHarvey: Instead of changing the travis or evergreen config, I suggest skipping the tests when the C extensions are installed. Once you get the BSONWriter working with C we can start testing across the board.

jaraco:
I'm very eager to see this functionality, especially on the decoding side, as we have a current performance bottleneck loading large documents from the database that would benefit from an efficient decoder.

Question - does this implementation allow for deferring to the default BSON representation for supported types? That is, say I have a document with a Python Decimal in it: `{"val": decimal.Decimal("1.0")}` - can I supply a BSONWriter that detects and handles `decimal.Decimal` values, transforms each to a bson.decimal128 value, and then defers to the default writer to encode that to BSON? I see the current work in progress implements a BSONWriter. Is it at a stage I can start experimenting with its usage? If so, can you provide a brief example of how it is used?

Also, importantly, on decoding _any_ object, can a BSONReader detect if the object meets certain criteria and if so, transform it? Consider a document that has: `{"val": {"_type": "decimal", "_val": "1.0"}}`. The nested object is readily decodable by the existing BSON readers, but I'd like for a hook to be able to intercept that value, detect the presence of the `{"_type": "_decimal"}` and use that as a signal to convert the entire object to decimal, such that the resulting document is `{"val": decimal.Decimal("1.0")}`. Another way to think about this, I'd like to create a BSONReader that first invokes the default behavior, but then for each object returned, invoke a (possible null) transformation on it.

Although these two examples share some similarities, they are two distinct use cases I'm hoping this work will support.

behackett:
@jaraco see https://github.com/mongodb/mongo-python-driver/pull/405. The example usage implements a transformer from Decimal to Decimal128 and back again.

prashantmital:
Closing this as it is superseded by #405

PR #383, opened 2018-11-14
Remove legacy Python 2.4 support code

prashantmital:
I found this little snippet of code that supported some Python 2.4 behavior hanging around in the `uri_parser` module. I think we can safely remove it now.


ShaneHarvey:
Can you:
- open a jira ticket and link this review?
- add pypy 2 and jython to your evergreen patch for good measure?

PR #385, opened 2018-11-16
Provide a new ensure_valid_collection_name function

Colin-b:
It will avoid copying/pasting code from this project to make sure that a variable is a valid collection name.

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -68,6 +68,24 @@ class ReturnDocument(object):
     """Return the updated/replaced or inserted document."""
 
 
+def ensure_valid_collection_name(collection_name):
+    """Ensure that provided collection name is valid.
+    :raises InvalidName if not.
+    """
+    if not collection_name or ".." in collection_name:
+        raise InvalidName("collection names cannot be empty")
+    if "$" in collection_name and not (collection_name.startswith("oplog.$main") or
+                                       collection_name.startswith("$cmd")):
+        raise InvalidName("collection names must not "
+                          "contain '$': %r" % collection_name)
+    if collection_name[0] == "." or collection_name[-1] == ".":
+        raise InvalidName("collection names must not start "
+                          "or end with '.': %r" % collection_name)
+    if "\x00" in collection_name:
+        raise InvalidName("collection names must not contain the "
+                          "null character")
+
+
 class Collection(common.BaseObject):
     """A Mongo collection.
     """
@@ -159,18 +177,7 @@ class Collection(common.BaseObject):
             raise TypeError("name must be an instance "
                             "of %s" % (string_type.__name__,))
 
-        if not name or ".." in name:
-            raise InvalidName("collection names cannot be empty")
-        if "$" in name and not (name.startswith("oplog.$main") or
-                                name.startswith("$cmd")):
-            raise InvalidName("collection names must not "
-                              "contain '$': %r" % name)
-        if name[0] == "." or name[-1] == ".":
-            raise InvalidName("collection names must not start "
-                              "or end with '.': %r" % name)
-        if "\x00" in name:
-            raise InvalidName("collection names must not contain the "
-                              "null character")
+        ensure_valid_collection_name(name)
         collation = validate_collation_or_none(kwargs.pop('collation', None))
 
         self.__database = database


ShaneHarvey:
Hi @Colin-b, this sounds reasonable but may I ask why you want to validate a collection name? It may not be the cure-all you might think it is. 

Even when `ensure_valid_collection_name` does not raise an error it is still possible that the server will reject the name due to other [naming-restrictions](https://docs.mongodb.com/manual/reference/limits/#naming-restrictions) that we cannot validate on the driver side. For example, there are stricter requirements when the server is running on Windows vs Unix. Also, the validity of a collection name depends on the database it resides in:
> The maximum length of the collection namespace, which includes the database name, the dot (.) separator, and the collection name (i.e. <database>.<collection>), is 120 bytes.

We generally don't want drivers to duplicate the server's validation rules. The main reason we do any validation at all is to avoid obscure errors from the server such as those described in [SERVER-26431](https://jira.mongodb.org/browse/SERVER-26431):

> A database name that's the empty string, combined with a collection name with a dot in it, like "a.b", will result in the namespace "a.b" This will be encoded as the namespace "a.b" in OP_QUERY, which the server will interpret as database of "a" and collection of "b". Not what the user intended.
>
> Similarly, a database name with a dot in it, like "a.b", combined with a collection name like "c" will result in the namespace "a.b.c". This will be encoded as the namespace "a.b.c" in OP_QUERY, which the server will interpret as database of "a" and collection of "b.c". Also not what the user intended.

That said I'm still open to merging this but I think the helper should at least make a note that the server's validation is stricter. Two examples are:
```python
>>> coll_name = "o"*118
>>> client.t[coll_name].insert_one({})
<pymongo.results.InsertOneResult object at 0x10e29e908>
>>> client.t2[coll_name].insert_one({})
  File "pymongo/collection.py", line 693, in insert_one
    session=session),
  File "pymongo/collection.py", line 607, in _insert
    bypass_doc_val, session)
  File "pymongo/collection.py", line 595, in _insert_one
    acknowledged, _insert_command, session)
  File "pymongo/mongo_client.py", line 1268, in _retryable_write
    return self._retry_with_session(retryable, func, s, None)
  File "pymongo/mongo_client.py", line 1221, in _retry_with_session
    return func(session, sock_info, retryable)
  File "pymongo/collection.py", line 590, in _insert_command
    retryable_write=retryable_write)
  File "pymongo/pool.py", line 579, in command
    unacknowledged=unacknowledged)
  File "pymongo/network.py", line 150, in command
    parse_write_concern_error=parse_write_concern_error)
  File "pymongo/helpers.py", line 155, in _check_command_response
    raise OperationFailure(msg % errmsg, code, response)
pymongo.errors.OperationFailure: fully qualified namespace t2.oooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo is too long (max is 120 bytes)
```
And:
```python
>>> client.t['$cmd'].insert_one({})
  File "pymongo/collection.py", line 693, in insert_one
    session=session),
  File "pymongo/collection.py", line 607, in _insert
    bypass_doc_val, session)
  File "pymongo/collection.py", line 595, in _insert_one
    acknowledged, _insert_command, session)
  File "pymongo/mongo_client.py", line 1268, in _retryable_write
    return self._retry_with_session(retryable, func, s, None)
  File "pymongo/mongo_client.py", line 1221, in _retry_with_session
    return func(session, sock_info, retryable)
  File "pymongo/collection.py", line 590, in _insert_command
    retryable_write=retryable_write)
  File "pymongo/pool.py", line 579, in command
    unacknowledged=unacknowledged)
  File "pymongo/network.py", line 150, in command
    parse_write_concern_error=parse_write_concern_error)
  File "pymongo/helpers.py", line 155, in _check_command_response
    raise OperationFailure(msg % errmsg, code, response)
pymongo.errors.OperationFailure: invalid collection name
```

Colin-b:
Hi,

First of, thanks for taking the time to look into this.

I am aware that there are more restriction and that unfortunately this
check is not enough to cover all cases.

I have an application were collection names are provided by users and
stored (as part of a document). Later on this collection might be created
but I want to reject the user request as soon as possible.

After providing this PR I noticed there is a slightly similar function on
database (to validate a collection name), except that this one is not
supposed to be used by public.

If we follow this then you might want to rename it to match your own logic
in naming?

Thanks again

Colin-b:
Apologies for the delay, I no longer need this as a try/except statement will do the trick (I need to create the underlying collection anyway).

PR #386, opened 2018-11-27
Implement URI options spec

prashantmital:
Implements the spec introduced in: https://github.com/mongodb/specifications/pull/414


test/test_uri_spec.py
```diff
@@ -23,44 +25,74 @@
...
+
+CONN_STRING_TEST_PATH = os.path.join(
     os.path.dirname(os.path.realpath(__file__)),
     os.path.join('connection_string', 'test'))
 
+URI_OPTIONS_TEST_PATH = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)), 'uri_options')
+
+TEST_DESC_SKIP_LIST = [
+    "Valid options specific to single-threaded drivers are parsed correctly",
```
saghm: Depending on how you sync the files from the specifications repo, it might be easier to just not include this file at all in your repo.

saghm: Fair enough! If you're interested, a while back we wrote a script to handle updating spec tests that you might be interested in using in the future: https://github.com/mongodb/mongo-go-driver/blob/master/etc/update-spec-tests.sh

prashantmital:
Not sure why this test is failing on Py 2.7:
```
 [2018/11/27 17:08:52.021] ======================================================================
 [2018/11/27 17:08:52.021] FAIL [0.001s]: test_test_uri_options_connection-options (test_uri_spec.TestAllScenarios)
 [2018/11/27 17:08:52.021] ----------------------------------------------------------------------
 [2018/11/27 17:08:52.021] Traceback (most recent call last):
 [2018/11/27 17:08:52.021]   File "/data/mci/4928f855b5cf08bed089bdc80a32cb03/src/test/test_uri_spec.py", line 55, in modified_test_scenario
 [2018/11/27 17:08:52.021]     func(*args, **kwargs)
 [2018/11/27 17:08:52.021]   File "/data/mci/4928f855b5cf08bed089bdc80a32cb03/src/test/test_uri_spec.py", line 92, in run_scenario
 [2018/11/27 17:08:52.021]     expected_warning, "warning") % dsc)
 [2018/11/27 17:08:52.021] AssertionError: Expected warning for test 'Invalid serverSelectionTimeoutMS causes a warning'
 [2018/11/27 17:08:52.021] Stderr:
 [2018/11/27 17:08:52.021] /data/mci/4928f855b5cf08bed089bdc80a32cb03/src/pymongo/common.py:723: UserWarning: connectTimeoutMS must be an integer or float
 [2018/11/27 17:08:52.021]   warnings.warn(str(exc))
 [2018/11/27 17:08:52.021] /data/mci/4928f855b5cf08bed089bdc80a32cb03/src/pymongo/common.py:723: UserWarning: heartbeatFrequencyMS must be an integer or float
 [2018/11/27 17:08:52.021]   warnings.warn(str(exc))
 [2018/11/27 17:08:52.021] /data/mci/4928f855b5cf08bed089bdc80a32cb03/src/pymongo/common.py:723: UserWarning: localThresholdMS must be an integer or float
 [2018/11/27 17:08:52.021]   warnings.warn(str(exc))
 [2018/11/27 17:08:52.021] ----------------------------------------------------------------------
```

I am unable to reproduce it locally. Will try with the EVG python toolchain tomorrow.

prashantmital:
Turns out, we were running into an obscure `warnings`-related bug detailed here: https://bugs.python.org/issue21724. I added the `clear_warnings_registry` helper method to the `test` module to work around this issue.

pymongo/common.py
```diff
@@ -611,17 +684,41 @@ def validate(option, value):
...
     """
     validated_options = {}
     for opt, value in iteritems(options):
         lower = opt.lower()
         try:
-            validator = URI_VALIDATORS.get(lower, raise_config_error)
+            validator = URI_OPTIONS_VALIDATOR_MAP.get(
+                lower, raise_config_error)
             value = validator(opt, value)
-        except (ValueError, ConfigurationError) as exc:
```
ShaneHarvey: Can you extract this change into a separate review for PYTHON-1686?

test/__init__.py
```diff
@@ -718,3 +717,10 @@ def test_cases(suite):
             # unittest.TestSuite
             for case in test_cases(suite_or_case):
                 yield case
+
+
+def clear_warning_registry():
```
ShaneHarvey: Can you add a comment with a link to the python bug tracker?

pymongo/common.py
```diff
@@ -611,17 +684,41 @@ def validate(option, value):
     return lower, value
 
 
-def get_validated_options(options, warn=True):
-    """Validate each entry in options and raise a warning if it is not valid.
-    Returns a copy of options with invalid entries removed
+def get_validated_options(options):
```
ShaneHarvey: Technically all these public methods and globals are public so we need to consider backwards compatibility. Why remove the "warn" option?

pymongo/common.py
```diff
@@ -524,56 +525,77 @@ def validate_tzinfo(dummy, value):
...
-    'w': validate_non_negative_int_or_basestring,
-    'wtimeout': validate_non_negative_integer,
-    'wtimeoutms': validate_non_negative_integer,
-    'fsync': validate_boolean_or_string,
-    'j': validate_boolean_or_string,
+# Dictionary where keys are the names of public URI options, and values
+# are lists of aliases for that option. Aliases of option names are assumed
+# to have been deprecated.
+URI_OPTIONS_ALIAS_MAP = {
+    'journal': ['j', ],
```
ShaneHarvey: Can you remove the unneeded trailing commas?

pymongo/common.py
```diff
@@ -524,56 +525,77 @@ def validate_tzinfo(dummy, value):
...
+    'tlscafile': ['ssl_ca_certs', ],
+    'tlsclientcertfile': ['ssl_certfile', ],
+    'tlsclientkeypassword': ['ssl_pem_passphrase', ],
+    'tlsclientkeyfile': ['ssl_keyfile', ],
+}
+
+# Dictionary where keys are the names of URI options, and values
+# are functions that validate user-input values for that option. If an option
+# alias uses a different validator than its public counterpart, it should be
+# included here as a key, value pair.
```
ShaneHarvey: > If an option alias uses a different validator than its public counterpart, it should be included here as a key, value pair.

Is there an example of such an option?

pymongo/common.py
```diff
@@ -524,56 +525,77 @@ def validate_tzinfo(dummy, value):
...
-    'sockettimeoutms': validate_timeout_or_none,
+# Dictionary where keys are the names of URI options specific to pymongo,
+# and values are functions that validate user-input values for those options.
+NONSPEC_OPTIONS_VALIDATOR_MAP = {
+    'connect': validate_boolean_or_string,
+    'driver': validate_driver_or_none,
+    'fsync': validate_boolean_or_string,
+    'minpoolsize': validate_non_negative_integer,
+    'socketkeepalive': validate_boolean_or_string,
+    'tlscrlfile': validate_readable,
```
ShaneHarvey: Surprising that `tlscrlfile` is not part of the new spec. Why is that?

pymongo/common.py
```diff
@@ -524,56 +525,77 @@ def validate_tzinfo(dummy, value):
...
+    'uuidrepresentation': validate_uuid_representation,
+    'waitqueuemultiple': validate_non_negative_integer_or_none,
     'waitqueuetimeoutms': validate_timeout_or_none,
-    'serverselectiontimeoutms': validate_timeout_or_zero,
-    'heartbeatfrequencyms': validate_timeout_or_none,
-    'maxidletimems': validate_timeout_or_none,
-    'maxstalenessseconds': validate_max_staleness,
 }
 
+# Dictionary where keys are the names of keyword-only options for pymongo,
```
ShaneHarvey: "keyword-only options for for the MongoClient constructor".

pymongo/common.py
```diff
@@ -584,10 +606,61 @@ def validate_tzinfo(dummy, value):
...
+    'tlscafile': 'ssl_ca_certs',
+    'tlsclientcertfile': 'ssl_certfile',
+    'tlsclientkeypassword': 'ssl_pem_passphrase',
+    'tlsclientkeyfile': 'ssl_keyfile',
+}
+
+# Map from deprecated URI option names to the updated option names.
+# Case is preserved for updated option names as they are part of user warnings.
+URI_OPTIONS_DEPRECATION_MAP = {
+    'j': 'journal',
```
ShaneHarvey: Does this change flip the priority of 'j' and 'journal'? It looks like before 'j' would override 'journal' but now it's reversed. I'm not too concerned but we should at least document all potentially breaking changes like this in the changelog.

pymongo/common.py
prashantmital: I thought there would be merge conflicts if I resolve these issues separately since this PR changes the names of many of the options affected which is why I didn't bother with a separate PR for 1686. Do we intend to backport these changes? That might make a good case for doing things as you suggest. If not, I'd be in favor of the resolving 1686 after we merge this (shouldn't involve much more than some additional testing) since no users had complained about that anyway.

pymongo/uri_parser.py
```diff
@@ -129,50 +130,33 @@ def parse_host(entity, default_port=DEFAULT_PORT):
...
-        warnings.warn("Option wtimeout is deprecated, use 'wtimeoutMS'"
-                      " instead")
-
+    options = _tokenize_option_string(opts, delim)
+    options = handle_option_deprecations(options)
     return options
 
 
-def split_options(opts, validate=True, warn=False):
+def split_options(opts, validate=True, warn=False, normalize=True):
```
ShaneHarvey: Needs documentation for `normalize=True`.

pymongo/uri_parser.py
```diff
@@ -129,50 +130,33 @@ def parse_host(entity, default_port=DEFAULT_PORT):
     return host.lower(), port
 
 
-def validate_options(opts, warn=False):
```
ShaneHarvey: We can't remove this method so suddenly since it's a public api.

pymongo/mongo_client.py
```diff
@@ -536,9 +537,13 @@ def __init__(
...
         keyword_opts['connect'] = connect
-        # Validate all keyword options.
-        keyword_opts = dict(common.validate(k, v)
-                            for k, v in keyword_opts.items())
+        # Handle deprecated options in kwarg list.
+        keyword_opts = common.handle_option_deprecations(keyword_opts)
+        # Validate all kwarg options.
+        keyword_opts = common.get_validated_options(keyword_opts)
+        # Change kwarg option names to those used internally.
+        keyword_opts = common.normalize_options(keyword_opts)
```
ShaneHarvey: It seems like there should be a single method that performs these 3 steps. Don't we do the same 3 step process for URI options?

pymongo/common.py
```diff
@@ -631,6 +728,37 @@ def get_validated_options(options, warn=True):
     return validated_options
 
 
+def handle_option_deprecations(options):
```
ShaneHarvey: Is there a benefit to this being a public method? I think it should be private.

pymongo/common.py
```diff
@@ -631,6 +728,37 @@ def get_validated_options(options, warn=True):
...
+                warnings.warn("Deprecated option '%s' ignored in favor of "
+                              "'%s'." % (str(key), renamed_key))
+                continue
+            warnings.warn("Option '%s' is deprecated, use '%s' instead." % (
+                          str(key), renamed_key))
+        undeprecated_options[str(key)] = value
+    return undeprecated_options
+
+
+def normalize_options(options):
```
ShaneHarvey: Probably should be private.

pymongo/common.py
prashantmital: Only the Python driver exposes a CRL file feature. None of the other drivers have bothered to do it and as per a discussion I had with Sam Rossi, there doesn't seem to be much appetite to start now.

pymongo/common.py
prashantmital: `tlsAllowInvalidCertificates` is the renamed analogue of `ssl_cert_reqs`. These options have different validators as `ssl_cert_reqs` must be either `ssl.CERT_NONE` or `ssl.CERT_REQUIRED` while `tlsAllowInvalidCertificates` needs to be either True or False. In such a situation, the validator method also maps the boolean `tlsAllowInvalidCertificates` value to the appropriate `ss.CERT_*` constant.
Would you like me to add this to the comment?

pymongo/common.py
prashantmital: Done.

pymongo/common.py
prashantmital: Done.

pymongo/common.py
```diff
@@ -524,56 +525,78 @@ def validate_tzinfo(dummy, value):
...
-    'fsync': validate_boolean_or_string,
-    'j': validate_boolean_or_string,
+# Dictionary where keys are the names of public URI options, and values
+# are lists of aliases for that option. Aliases of option names are assumed
+# to have been deprecated.
+URI_OPTIONS_ALIAS_MAP = {
+    'journal': ['j'],
+    'wtimeoutms': ['wtimeout'],
+    'tls': ['ssl'],
+    'tlsallowinvalidcertificates': ['ssl_cert_reqs'],
```
behackett: This seems wrong to me. ssl_cert_reqs is a tri-state. It's value can be ssl.CERT_NONE (0), ssl.CERT_OPTIONAL (1), or ssl.CERT_REQUIRED (2). I'm fine with treating CERT_OPTIONAL as the same as CERT_REQUIRED (since CERT_OPTIONAL has no meaning on the client side), but this mapping would seem to make CERT_REQUIRED and tlsallowinvalidcertificates=true as being equivalent when they are actually opposites. tlsallowinvalidcertificates=true is the same as ssl_cert_reqs=ssl.CERT_NONE.

pymongo/common.py
```diff
@@ -524,56 +525,78 @@ def validate_tzinfo(dummy, value):
...
+    'tlsallowinvalidhostnames': ['ssl_match_hostname'],
```
behackett: The same problem exists here. tlsallowinvalidhostname=true should be equivalent to ssl_match_hostname=False.

pymongo/common.py
```diff
@@ -524,56 +525,78 @@ def validate_tzinfo(dummy, value):
...
+    'tlscrlfile': ['ssl_crlfile'],
+    'tlscafile': ['ssl_ca_certs'],
+    'tlsclientcertfile': ['ssl_certfile'],
```
behackett: This has been changed to tlscertificatekeyfile.

pymongo/common.py
```diff
@@ -524,56 +525,78 @@ def validate_tzinfo(dummy, value):
...
+    'tlsclientkeypassword': ['ssl_pem_passphrase'],
```
behackett: tlscertificatekeyfilepassword

pymongo/common.py
```diff
@@ -524,56 +525,78 @@ def validate_tzinfo(dummy, value):
...
+    'tlsclientkeyfile': ['ssl_keyfile'],
```
behackett: This no longer exists in the spec, because it would be confused with the existing server option "sslPEMKeyFile" which isn't the same thing.

pymongo/common.py
```diff
@@ -584,10 +607,61 @@ def validate_tzinfo(dummy, value):
...
+    'tlsclientkeypassword': 'ssl_pem_passphrase',
+    'tlsclientkeyfile': 'ssl_keyfile',
+}
+
+# Map from deprecated URI option names to the updated option names.
+# Case is preserved for updated option names as they are part of user warnings.
+URI_OPTIONS_DEPRECATION_MAP = {
+    'j': 'journal',
+    'wtimeout': 'wTimeoutMS',
+    'ssl': 'tls',
```
behackett: The Atlas team requested that we not deprecate "ssl" and never remove it. We agreed and made ssl a required option in the spec. It can still just be an alias for "tls".

pymongo/common.py
```diff
@@ -584,10 +607,61 @@ def validate_tzinfo(dummy, value):
...
+    'ssl_cert_reqs': 'tlsAllowInvalidCertificates',
+    'ssl_match_hostname': 'tlsAllowInvalidHostnames',
+    'ssl_crlfile': 'tlsCRLFile',
+    'ssl_ca_certs': 'tlsCAFile',
+    'ssl_certfile': 'tlsClientCertFile',
```
behackett: Because of the naming conflict we had with the server, we're only providing a URI option to support a unified file that provides both the cert chain and the private key. Let's not deprecate these three options. There are definitely users that want to store the private key in a separate file.

test/__init__.py
prashantmital: Done.

pymongo/uri_parser.py
prashantmital: Understood. I didn't think about the fact that this is all public API - I'll add it back. I just don't understand why we have a method that does nothing but wrap `common.get_validated_options`.

pymongo/mongo_client.py
prashantmital: I think the methods need to be separate even though in a majority of cases we perform all 3 steps on the URI options. This  is because we use these 3 methods in different ways depending on whether we are parsing (a) a URI string, or (b) keyword arguments passed to the `MongoClient`constructor. Also, we don't always want to perform the normalization step thereby requiring `normalize_options` to be a separate method that can be called after a boolean flag is checked. In light of all this, we must keep these three methods separate to facilitate re-use of this code.

pymongo/common.py
prashantmital: In this case you are correct - I didn't account for the inversion. I have done so in my latest commits by using a lambda to invert the return value of the validator function.

pymongo/common.py
prashantmital: Did you mean to say two options? AFAICT we only need to not deprecate the `ssl_certfile` and `ssl_keyfile` options. Everything else remains the same, is it not?

behackett: Yes, you're right.

pymongo/common.py
```diff
@@ -613,15 +686,22 @@ def validate(option, value):
 
 def get_validated_options(options, warn=True):
     """Validate each entry in options and raise a warning if it is not valid.
-    Returns a copy of options with invalid entries removed
+    Returns a copy of options with invalid entries removed.
+
+    :Parameters:
+        - `opts`: A dict of MongoDB URI options.
+        - `warn` (optional): If ``True`` then warnigns will be logged and
```
behackett: warnigns -> warnings

pymongo/common.py
```diff
@@ -631,6 +711,37 @@ def get_validated_options(options, warn=True):
...
+def _handle_option_deprecations(options):
+    """Issue appropriate warnings when deprecated options are present in the
+    options dictionary. Removes deprecated option key, value pairs if the
+    options dictionary is found to also have the renamed option."""
+    undeprecated_options = {}
+    for key, value in iteritems(options):
+        optname = str(key).lower()
+        if optname in URI_OPTIONS_DEPRECATION_MAP:
+            renamed_key = URI_OPTIONS_DEPRECATION_MAP[optname]
+            if any(renamed_key.lower() == n.lower() for n in options):
```
behackett: You should make the list of lower cased options once outside of the for loop. Then this can be a simple membership test.

pymongo/mongo_client.py
```diff
@@ -368,45 +369,57 @@ def __init__(
...
-            then the ``ssl_ca_certs`` parameter must point to a file of CA
-            certificates. Implies ``ssl=True``. Defaults to
-            ``ssl.CERT_REQUIRED`` if not provided and ``ssl=True``.
-          - `ssl_ca_certs`: The ca_certs file contains a set of concatenated
+          - `tls`: (boolean) If ``True``, create the connection to the server
+            using TLS. Defaults to ``False``.
+          - `ssl`: (boolean) Alias for ``tls``.
+          - `tlsInsecure`: (boolean) Specify whether TLS constraints should be
+            relaxed as much as possible. Setting ``tlsInsecure=True`` implies
+            ``tlsAllowInvalidCertificates=False`` and
```
behackett: Implies True, not False.

pymongo/mongo_client.py
```diff
@@ -368,45 +369,57 @@ def __init__(
...
+            ``tlsAllowInvalidHostnames=False``. Defaults to ``False``. Think
```
behackett: Implies True, not False.

pymongo/mongo_client.py
```diff
@@ -368,45 +369,57 @@ def __init__(
...
+            very carefully before setting this to ``True`` as it could allow
+            malicious third-parties to exploit TLS vulnerabilities to attack
```
behackett: It's not about vulnerabilities. Say "...setting this to ``True`` dramatically reduces the security of TLS..."

pymongo/mongo_client.py
```diff
@@ -368,45 +369,57 @@ def __init__(
...
+            your application.
+          - `tlsAllowInvalidCertificates`: (boolean) Specifies whether a
+            certificate is required from the other side of the connection, and
```
behackett: I know this text is pre-existing, but it's wrong so let's fix it. Since we're the client side the server _must_ and _will_ send us a certificate. This is just about us accepting invalid certificates. So get rid of the part before the and, and clean up the everything below implying that the server doesn't have to send us a certificate.

pymongo/mongo_client.py
```diff
@@ -368,45 +369,57 @@ def __init__(
...
+            whether it will be validated if provided. Certificates are ignored
+            if ``tlsAllowInvalidCertificates=True``. Certificates are required
+            and validated if ``tlsAllowInvalidCertificates=False``. If the
+            value of this parameter is ``False`` and a value is not provided
+            for ``tlsCAFile``, PyMongo will attempt to load system provided CA
+            certificates. If the python version in use does not support loading
+            system CA certificates then the ``tlsCAFile`` parameter must point
+            to a file of CA certificates. ``tlsAllowInvalidCertificates=False``
+            implies ``tls=True``. Defaults to ``False``.
+          - `tlsAllowInvalidHostnames`: (boolean) If ``False`` and
```
behackett: Since the default is False and hostname matching is enabled by default let's change this to only talk about what happens when you set it to True (and continue to strongly discourage setting this to True). Also, starting with CPython 3.7 we don't use ssl.match_hostname since OpenSSL does this work for us. Either way, the implementation details probably aren't useful to anyone using PyMongo, so let's get rid of them.

pymongo/mongo_client.py
```diff
@@ -368,45 +369,57 @@ def __init__(
...
+            python's :mod:`~ssl` module. ``tlsAllowInvalidHostnames=False``
+            implies ``tls=True``. Defaults to ``False``. Think very carefully
+            before setting this to ``True`` as that could make your application
+            vulnerable to man-in-the-middle attacks.
+          - `tlsCAFile`: The file containing a single or a bundle of
             "certification authority" certificates, which are used to validate
             certificates passed from the other end of the connection.
-            Implies ``ssl=True``. Defaults to ``None``.
+            Implies ``tls=True``. Defaults to ``None``.
+          - `tlsCertificateKeyFile`: The client certificate file or the client
```
behackett: or -> "and possibly". This file should never contain _only_ the private key.

pymongo/mongo_client.py
```diff
@@ -368,45 +369,57 @@ def __init__(
...
+            private key file used to identify the local connection against
+            mongod. In the case that they are both needed, the files can either
```
behackett: Both are always needed. We just provide the option of passing them in two separate files. Read this for a more in depth explanation: https://docs.python.org/2/library/ssl.html#certificates

I think the terse original documentation for ssl_certfile and ssl_keyfile is correct.

pymongo/mongo_client.py
```diff
@@ -368,45 +369,57 @@ def __init__(
...
+          - `ssl_keyfile`: The private keyfile used to identify the local
+            connection against mongod. Can be omitted if the keyfile is
+            included with the ``tlsCertificateKeyFile``. Implies ``tls=True``.
+            Defaults to ``None``.
+          - `tlsCertificateKeyFilePassword`: The password or passph rase for
+            decrypting the private key in ``tlsCertificateKeyFile`` or
+            ``ssl_keyfile``. Only necessary if the private key is encrypted.
+            Only supported by python 2.7.9+ (pypy 2.5.1+) and 3.3+. Defaults
+            to ``None``.
           - `ssl_crlfile`: The path to a PEM or DER formatted certificate
```
behackett: Didn't we add a new tlsCRLFile option?

pymongo/uri_parser.py
```diff
@@ -129,6 +130,26 @@ def parse_host(entity, default_port=DEFAULT_PORT):
...
+        key, value = opt.split("=")
+        optname = str(key).lower()
+        if optname == 'readpreferencetags':
+            options.setdefault(optname, []).append(value)
+        else:
+            if optname == 'tlsinsecure':
+                options['tlsAllowInvalidCertificates'] = value
+                options['tlsAllowInvalidHostnames'] = value
+            if key in options:
+                warnings.warn("Duplicate URI option '%s'." % (str(key),))
```
behackett: If the application specifies tlsInsecure and one or both of the tlsAllow* options you added above you're going to trigger this warning unnecessarily.

pymongo/mongo_client.py
prashantmital: Nope. Only the Python driver implements the CRLFile option for some reason.

prashantmital: Sorry, I did add a `tlsCRLFile` option but forgot to document it. I'll add it in the next commits.

pymongo/uri_parser.py
prashantmital: We definitely want _some_ warning to be triggered when `tlsInsecure` is specified along with one of the other `tlsAllow*` options as evidenced by this [spec test](https://github.com/mongodb/specifications/pull/414/files#diff-6e7064dc748fe47120d108a28f561240R87). I agree that the duplicate option warning is not very appropriate in this situation, but I thought it was simpler than introducing a new warning... What kind of warning would you like to see @behackett ?

prashantmital:
Test failures in EVG are due to something in MO, therefore this can be safely merged.

PR #388, opened 2018-12-07
Fix typo in doc

ddc67cd:


behackett:
I didn't right these docs. The person who did left MongoDB many years ago. But I don't think this is a typo. Note that SON is all caps. It's a acronym.

ddc67cd:
Хуй

cjgibson:
> @ddc67cd 
> [Хуй](https://www.bing.com/translator?text=%D0%A5%D1%83%D0%B9)

Lol. That's a little childish, don't you think?
I can't seem to find any other documentation after a quick Googling that explicitly mentions what SON stands for, although I would assume you can always just ping @mdirolf to see if he remembers what it stands for eight years after authoring the original commit.

PR #389, opened 2018-12-13
PYTHON-1704 Close periodic task thread when client is closed

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/PYTHON-1704
Fixes: https://github.com/mongodb/motor/pull/44

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -1030,8 +1030,9 @@ class MongoClient(common.BaseObject):
         session_ids = self._topology.pop_all_sessions()
         if session_ids:
             self._end_sessions(session_ids)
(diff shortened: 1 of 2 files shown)


pymongo/mongo_client.py
```diff
@@ -1136,10 +1139,6 @@ def _send_message_with_response(self, operation, exhaust=False,
           - `address` (optional): Optional address when sending a message
             to a specific server, used for getMore.
         """
-        with self.__lock:
-            # If needed, restart kill-cursors thread after a fork.
-            self._kill_cursors_executor.open()
```
ShaneHarvey: This is handled in _get_topology now.

ajdavis: Acknowledged.

PR #390, opened 2018-12-18
Collection count_documents 'filter' variable defaulted to empty dict

ggbeppe:
This allows the same behavior as with find(), more intuitive.

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -1613,7 +1613,7 @@ class Collection(common.BaseObject):
         cmd.update(kwargs)
         return self._count(cmd)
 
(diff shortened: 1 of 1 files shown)


behackett:
Thanks for your contribution, but we won't be able to take it. The CRUD spec for MongoDB drivers has this to say about countDocuments:

```
Applications must be required to pass a value for filter, but an empty document is supported
```

https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#countdocuments

On top of that, defaulting a keyword argument to a mutable type is a bad idea in Python.

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

PR #391, opened 2019-01-22
Fix compression URI options test failure

prashantmital:
The failure was likely the result of an incorrect spec-test.

### DIFF
--- a/test/test_uri_spec.py
+++ b/test/test_uri_spec.py
@@ -23,6 +23,7 @@ import warnings
 sys.path[0:0] = [""]
 
 from pymongo.common import INTERNAL_URI_OPTION_NAME_MAP, validate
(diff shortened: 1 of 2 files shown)


prashantmital:
> Can we remove the try/except with test.get('options', {})?

I would still need to do something special to handle the cases when `options` is `None`. I don't like the `try...except` either, but I thought this was the most concise way to get it done. 

```
    {
      "description": "Unsupported option values are ignored",
      "uri": "mongodb://example.com/?fsync=ifPossible",
      "valid": true,
      "warning": true,
      "hosts": [
        {
          "type": "hostname",
          "host": "example.com",
          "port": null
        }
      ],
      "auth": null,
      "options": null
    },
```

ShaneHarvey:
How about `(test.get('options') or {}).get('compressors', [])`?

PR #392, opened 2019-01-25
PYTHON-1528 - Support encoding sets, dict view objects, bytearray, range

salty-horse:
Jira: [PYTHON-1528](https://jira.mongodb.org/browse/PYTHON-1528)

This implements encoding support for:

* set
* frozenset
* dict_keys
* dict_values
* dict_items
* xrange (Python 2)
* range (Python 3)
* bytearray

These are of course all one-way conversions, which is detailed in the mapping table in the documentation.

`bytearray`s are also available in Python 2. I made them serialize like `bytes` in Python 3, which means they decode to `bson.Binary` in Python 2.
I don't see a lot of use for serializing ranges, but I added it because it was simple.

Tests are included. I tested with Python 2.7.15rc1 and 3.6.7, with and without the C bson module.

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -28,6 +28,13 @@ float                                    number (real)  both
 string                                   string         py -> bson
 unicode                                  string         both
 list                                     array          both
+set                                      array          py -> bson
+frozenset                                array          py -> bson
+dict_keys                                array          py -> bson
+dict_values                              array          py -> bson
+dict_items                               array          py -> bson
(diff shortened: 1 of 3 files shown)


behackett:
Given the discussion and decision made in PYTHON-1528 I'm going to close this. Note that we're considering adding support for treating any type that implements the buffer API as BSON binary subtype 0. That would include bytearray. See [PYTHON-1648](https://jira.mongodb.org/browse/PYTHON-1648).

PR #393, opened 2019-01-31
PYTHON-1730 Use w:majority when retrying commitTransaction

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/PYTHON-1730

This change makes Python use 'w':majority and 'wtimeout' according to the updated transaction spec: https://github.com/mongodb/specifications/blob/ba1b071/source/transactions/transactions.rst#majority-write-concern-is-used-when-retrying-committransaction

### DIFF
--- a/pymongo/client_session.py
+++ b/pymongo/client_session.py
@@ -391,8 +391,8 @@ class ClientSession(object):
             raise InvalidOperation(
                 "Cannot call commitTransaction after calling abortTransaction")
         elif state is _TxnState.COMMITTED:
-            # We're rerunning the commit, move the state back to "in progress"
-            # so that _in_transaction returns true.
+            # We're explicitly retrying the commit, move the state back to
+            # "in progress" so that _in_transaction returns true.
             self._transaction.state = _TxnState.IN_PROGRESS
             retry = True
 
--- a/pymongo/client_session.py
+++ b/pymongo/client_session.py
@@ -379,7 +379,7 @@ class ClientSession(object):
         .. versionadded:: 3.7
         """
         self._check_ended()
-
+        retry = False
         state = self._transaction.state
         if state is _TxnState.NONE:
             raise InvalidOperation("No transaction started")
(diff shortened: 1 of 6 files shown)


ShaneHarvey:
All failures are due to https://jira.mongodb.org/browse/PYTHON-1734 and https://jira.mongodb.org/browse/PYTHON-1719.

pymongo/client_session.py
```diff
@@ -379,7 +379,7 @@ def commit_transaction(self):
         .. versionadded:: 3.7
         """
         self._check_ended()
-
+        retry = False
```
prashantmital: Let's add a comment to clarify what this flag is trying to track. AFAICT this is present to track explicit retries (application called `commit_transaction` twice).

pymongo/client_session.py
```diff
@@ -439,31 +440,37 @@ def abort_transaction(self):
...
             return self._client.admin._command(
                 sock_info,
                 command_name,
                 session=self,
-                write_concern=self._transaction.opts.write_concern,
+                write_concern=wc,
                 parse_write_concern_error=True)
 
-    def _finish_transaction_with_retry(self, command_name):
+    def _finish_transaction_with_retry(self, command_name, retrying):
```
prashantmital: Can we rename `retrying` to something like `explicit_retry`?

pymongo/client_session.py
```diff
@@ -439,31 +440,37 @@ def abort_transaction(self):
...
             # The transactions spec says to ignore abortTransaction errors.
             pass
         finally:
             self._transaction.state = _TxnState.ABORTED
 
-    def _finish_transaction(self, command_name):
+    def _finish_transaction(self, command_name, retrying):
+        wc = self._transaction.opts.write_concern
+        if retrying and command_name == "commitTransaction":
+            wc_doc = wc.document
```
prashantmital: Not sure about this, but do we want to be modifying `opts.write_concern.document` in-place? Should we be copying this dictionary before messing with it?

pymongo/client_session.py
```diff
@@ -394,9 +394,10 @@ def commit_transaction(self):
             # We're rerunning the commit, move the state back to "in progress"
             # so that _in_transaction returns true.
             self._transaction.state = _TxnState.IN_PROGRESS
+            retry = True
```
prashantmital: Let's also comment here that landing up here means the application has explicitly retried committing the transaction.

PR #394, opened 2019-02-01
PYTHON-1719/PYTHON-1734 Fix test failures on MongoDB 4.1+

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/PYTHON-1719 and https://jira.mongodb.org/browse/PYTHON-1734

I plan to merge this into master and 42dev.

### DIFF
--- a/test/test_database.py
+++ b/test/test_database.py
@@ -391,23 +391,27 @@ class TestDatabase(IntegrationTest):
         self.assertTrue(isinstance(info[0]["ts"], datetime.datetime))
 
     @client_context.require_no_mongos
+    @ignore_deprecations
     def test_errors(self):
-        with ignore_deprecations():
-            # We must call getlasterror, etc. on same socket as last operation.
-            db = rs_or_single_client(maxPoolSize=1).pymongo_test
-            db.reset_error_history()
-            self.assertEqual(None, db.error())
+        # We must call getlasterror, etc. on same socket as last operation.
+        db = rs_or_single_client(maxPoolSize=1).pymongo_test
+        db.reset_error_history()
+        self.assertEqual(None, db.error())
+        if client_context.supports_getpreverror:
             self.assertEqual(None, db.previous_error())
 
-            db.test.insert_one({"_id": 1})
-            unacked = db.test.with_options(write_concern=WriteConcern(w=0))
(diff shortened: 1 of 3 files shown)


test/test_legacy_api.py
```diff
@@ -1263,6 +1263,7 @@ def test_ensure_index(self):
         # Clean up indexes for later tests
         db.test.drop_indexes()
 
+    @client_context.require_version_max(4, 1)  # PYTHON-1734
```
prashantmital: Can we put in a more descriptive `FIXME` comment block here saying that we should revisit limiting this test to server versions < 4.1 once SERVER-39323 has been resolved? Without that, it looks as if we have fixed the issue whereas this is just a band-aid.

prashantmital: Can we put in a more descriptive FIXME comment block here saying that we should revisit limiting this test to server versions < 4.1 once SERVER-39323 has been resolved? Without that, it looks as if we have fixed the issue whereas this is just a band-aid.

test/test_legacy_api.py
```diff
@@ -1323,6 +1324,7 @@ def run(self):
         finally:
             coll.drop()
 
+    @client_context.require_version_max(4, 1)  # PYTHON-1734
```
prashantmital: Same comment as above.

test/test_database.py
```diff
@@ -391,23 +391,27 @@ def test_profiling_info(self):
         self.assertTrue(isinstance(info[0]["ts"], datetime.datetime))
 
     @client_context.require_no_mongos
+    @ignore_deprecations
```
prashantmital: Nice.

test/test_legacy_api.py
ShaneHarvey: I'm going to keep PYTHON-1734 open until we reenable the test.

ShaneHarvey:
> Also, not sure why the EVG status hasn't updated on GH.

I've seen this on other projects too. It should be fixed soon: https://jira.mongodb.org/browse/EVG-5717

test/test_legacy_api.py
prashantmital: Sounds good. Might want to link this PR on that ticket and mention that a temp fix was introduced for the time being?

PR #396, opened 2019-02-01
PYTHON-1735 Remove redundant type checks after document validation.

vasiliykovalev:
https://jira.mongodb.org/browse/PYTHON-1735

### DIFF
--- a/pymongo/bulk.py
+++ b/pymongo/bulk.py
@@ -169,7 +169,7 @@ class _Bulk(object):
         """
         validate_is_document_type("document", document)
         # Generate ObjectId client side.
-        if not (isinstance(document, RawBSONDocument) or '_id' in document):
+        if '_id' not in document:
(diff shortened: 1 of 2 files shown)


behackett:
None of these checks are redundant. They are all special cases specific to RawBSONDocument that don't effect other types that inherit from MutableMapping.

vasiliykovalev:
Sorry, I'm confused. When document type is not `RawBSONDocument`, function `validate_is_document_type` raise TypeError. Line below if statement, where we check again document type is not `RawBSONDocument`.

vasiliykovalev:
I figured out and realized that I was wrong. Thanks for your feedback and examples. I hope I can help to project later :).

ShaneHarvey:
Thanks!

behackett:
No worries. Thanks for the contribution.

PR #398, opened 2019-02-10
Make sure Travis tests both native and compiled BSON implementations

salty-horse:
Currently, Travis only ever checks the compiled BSON implementation.

This change makes it run the whole test suite twice: once with the native implementation, and again with the compiled implementation. This is especially important when making changes to the BSON module, as I have done in #392.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,5 @@ python:
 services:
   - mongodb
 
-script: PYMONGO_MUST_CONNECT=1 python setup.py test
+script: PYMONGO_MUST_CONNECT=1 sh -c 'python setup.py --no_ext test && python setup.py test'
 


ShaneHarvey:
Thanks, this does add some more CPython test coverage on Travis but we're not going to merge it. Travis already tests both bson implementations (CPython uses C, PyPy uses pure python) because we [never build the C extensions on PyPy](https://github.com/mongodb/mongo-python-driver/blob/master/setup.py#L341-L353). 


Another reason is that all of our changes are tested [much more extensively in Evergreen](https://evergreen.mongodb.com/waterfall/mongo-python-driver) (MongoDB's own CI system). Even though the Travis testing is completely redundant we still keep it around so that contributors like yourself can more easily test their changes. However, we want to keep Travis testing as simple and minimal as possible because we don't want to maintain a redundant testing infrastructure. Does that make sense?

salty-horse:
Sure!

PR #400, opened 2019-02-14
Implement SPEC 1153: Adding readConcern support to aggregation with $out

prashantmital:
Note that the newly introduced tests for SPEC-1153 follow the new [v2 CRUD spec-test format](https://github.com/mongodb/specifications/tree/master/source/crud/tests#test-format). 

Additionally, this PR also:

- re-syncs the CRUD (v1) spec-tests
- refactors CRUD (v1 and v2) and Transaction spec test runners


pymongo/collection.py
```diff
@@ -2275,11 +2275,13 @@ def _aggregate(self, pipeline, cursor_class, first_batch_size, session,
...
             cmd.update(kwargs)
-            # Apply this Collection's read concern if $out is not in the
-            # pipeline.
-            if (sock_info.max_wire_version >= 4
-                    and 'readConcern' not in cmd
-                    and not dollar_out):
+            # Apply this Collection's read concern if
+            # - running MongoDB >= 4.2
+            # - pipeline does not have an $out stage on MongoDB >= 3.2
+            # - read concern has not been specified as a kwarg
```
ShaneHarvey: I suggest reordering the comments to match the order in the if statement.

test/test_crud_v1.py
```diff
@@ -158,12 +159,21 @@ def run_operation(collection, test):
...
         drop_collections(self.db)
-        self.db.test.insert_many(scenario_def['data'])
-
-        result = run_operation(self.db.test, test)
+        data = scenario_def.get('data')
+        if data:
+            self.db.test.insert_many(scenario_def['data'])
+
+        # Run operations and check results or errors.
+        expected_result = test.get('outcome', {}).get('result')
```
ShaneHarvey: I think this is supposed to be `test.get('result')`. I don't see outcome.result in the new format.

test/test_crud_v2.py
```diff
@@ -0,0 +1,331 @@
...
+                    for key, val in expected.items():
+                        if val is None:
+                            if key in actual:
+                                self.fail("Unexpected key [%s] in %r" % (
+                                    key, actual))
+                        elif val == {'$exists': True}:
+                            if key not in actual:
+                                self.fail("Expected key [%s] in %r" % (
+                                    key, actual))
+                        elif val == {'$exists': False}:
```
ShaneHarvey: Jeremy put off adding `$exists` so this should be removed, see https://jira.mongodb.org/browse/SPEC-1215.

test/test_crud_v2.py
```diff
@@ -0,0 +1,331 @@
...
+            opts.pop('readPreference'))
+
+    if 'writeConcern' in opts:
+        opts['write_concern'] = WriteConcern(
+            **dict(opts.pop('writeConcern')))
+
+    if 'readConcern' in opts:
+        opts['read_concern'] = ReadConcern(
+            **dict(opts.pop('readConcern')))
+    return opts
```
ShaneHarvey: It would be great if you could move `camel_to_snake`, `camel_to_upper_camel`, `camel_to_snake_args`, and `parse_collection_options` into `test/utils.py` so the various spec runners could reuse them.

test/test_crud_v2.py
```diff
@@ -0,0 +1,331 @@
...
+                o_collection = getattr(database, collname)
+            else:
+                o_collection = collection
+            self.assertEqual(list(o_collection.find()),
+                             expected_outcome['data'])
+
+    return run_scenario
+
+
+def create_tests():
```
ShaneHarvey: It would be great if you could move this `create_tests` into `test/utils.py` and reuse it in both crud spec runners. For now it can simply accept two arguments: the `create_test` method and the `TestAllScenarios` class.

test/test_crud_v2.py
```diff
@@ -0,0 +1,331 @@
+# Copyright 2015-present MongoDB, Inc.
```
ShaneHarvey: 2019-present

prashantmital:
EVG test failures are on `latest` and unrelated to these changes.

test/test_crud_v1.py
```diff
@@ -172,46 +172,9 @@ def run_scenario(self):
...
-                    test_type,
-                    str(test['description'].replace(" ", "_")))
-
-                new_test.__name__ = test_name
-                setattr(TestAllScenarios, new_test.__name__, new_test)
-
-
-create_tests()
+test_creator = TestCreator(create_test, TestAllScenarios, _TEST_PATH)
+test_creator.add_test_modifier(test_creator.enforce_min_max_server_version)
```
ShaneHarvey: As we discussed in person, I think we can call `enforce_min_max_server_version` on all these spec tests (transactions doesn't have the field but there's no harm in  supporting it there too). Then we can just remove `add_test_modifier` and the test specific skipping logic can happen in `create_test`.

ShaneHarvey:
There are a couple EVG test failures on latest that *are* caused by these changes:
```
test_read_concern.TestReadConcern.test_aggregate_out 	
test_session.TestCausalConsistency.test_writes_do_not_include_read_concern 
test_crud_v2.TestAllScenarios.test_v2_aggregate-out-readConcern_readConcern_majority_with_out_stage 
```
The other two failures:
```
test_transactions.TestTransactions.test_transactions_abort_abort_ignores_TransactionAborted
test_transactions.TestTransactions.test_transactions_error-labels_DuplicateKey_errors_do_not_contain_transient_label
```
Will soon be fixed by SPEC-1221.

test/test_crud_v1.py
prashantmital: One issue I am running into while trying to implement this is the fact that some checks are relying upon the test name, which isn't readily available within `create_test` as things are currently written (we compute the test name after creating the test method and then `setattr` on the created method's `__name__`). 
I am modifying `create_test` to also accept the name as a parameter since that seems like the only sensible thing to do. Let me know if you have a better solution.

ShaneHarvey: You could also try using `self.id()`: https://docs.python.org/3/library/unittest.html#unittest.TestCase.id

prashantmital:
@ShaneHarvey time for a final review. 
Also, do we need to set the base branch for this to `42dev` instead of `master`? My $0.02: since its not a backward breaking change, we should merge it into `master` and then cherry-pick the changes onto `42dev`.

test/test_session.py
```diff
@@ -999,6 +996,13 @@ def test_writes_do_not_include_read_concern(self):
             self._test_no_read_concern(
                 lambda coll, session: coll.reindex(session=session))
 
+    @client_context.require_no_standalone
+    @client_context.require_version_max(4, 1, 0)
+    def test_agggreate_out_doe_not_include_read_concern(self):
```
ShaneHarvey: `out_doe_not` -> `out_does_not`

test/test_read_concern.py
```diff
@@ -118,8 +118,14 @@ def test_aggregate_out(self):
         except OperationFailure:
             # "ns doesn't exist"
             pass
-        self.assertNotIn('readConcern',
-                         self.listener.results['started'][0].command)
+
+        # Aggregate with $out supports readConcern MongoDB 4.2 onwards.
+        if client_context.version >= Version(4, 1):
```
ShaneHarvey: I think you can just do `client_context.version >= (4, 1)` and avoid importing or using Version.

test/test_transactions.py
```diff
@@ -188,63 +138,6 @@ def test_transaction_write_concern_override(self):
...
-            for _ in range(UNPIN_TEST_MAX_ATTEMPTS):
-                cursor = coll.find({}, session=s)
-                self.assertTrue(next(cursor))
-                addresses.add(cursor.address)
-                # Break early if we can.
-                if len(addresses) > 1:
-                    break
-
-            self.assertGreater(len(addresses), 1)
-
```
ShaneHarvey: Why did you remove these tests?

test/utils.py
```diff
@@ -377,78 +514,6 @@ def stop(self):
...
-      - `tag_sets` (optional): List of dicts of tags for data-center-aware reads
-    """
-    members = set(members)
-    used = set()
-    for _ in range(100):
-        used.add(read_from_which_host(client, *args, **kwargs))
-
-    testcase.assertEqual(members, used)
-
-
```
ShaneHarvey: Why was this removed?

prashantmital: these methods appear to be unused.

prashantmital: though i have no clue why I removed these, these methods appear to be unused so we should probably not add them back in

test/test_transactions.py
prashantmital: It was a mistake.

test/test_transactions.py
```diff
@@ -476,15 +452,16 @@ def end_sessions(sessions):
...
         if test.get('skipReason'):
             raise unittest.SkipTest(test.get('skipReason'))
 
         listener = OvertCommandListener()
-        # Create a new client, to avoid interference from pooled sessions.
-        # Convert test['clientOptions'] to dict to avoid a Jython bug using
-        # "**" with ScenarioDict.
+        # New client, to avoid interference from pooled sessions.
+        # Convert test['clientOptions'] to dict to avoid a Jython bug using "**"
+        # with ScenarioDict.
```
ShaneHarvey: Can you revert these changes too?

prashantmital:
EVG is 💚

PR #401, opened 2019-02-20
Update docs to use list_collection_names method

wolfhoundjesse:
I encountered DeprecationWarning: collection_names is deprecated.

This is a simple name change, but we could also explain the
deprecation in detail in an aside or blockquote.

### DIFF
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -142,7 +142,7 @@ of the collections in our database:
 
 .. doctest::
 
-  >>> db.collection_names(include_system_collections=False)
+  >>> db.list_collection_names(include_system_collections=False)
   [u'posts']
 
 Getting a Single Document With :meth:`~pymongo.collection.Collection.find_one`


doc/tutorial.rst
```diff
@@ -142,7 +142,7 @@ of the collections in our database:
 
 .. doctest::
 
-  >>> db.collection_names(include_system_collections=False)
+  >>> db.list_collection_names(include_system_collections=False)
```
ShaneHarvey: Thanks for finding this issue! It also made me realize that we didn't properly support  the [Enumerate Collections Specification](https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst). I fixed that by adding support for `filter` argument in https://jira.mongodb.org/browse/PYTHON-1784.

Can you replace the `include_system_collections` parameter with a `filter` that only matches non-system collections, like this:
```
>>> db.list_collection_names(filter={"name": {"$regex": r"^(?!system\.)"}})
[u'posts']
```

ShaneHarvey: I just realized that for this doctest it's fine to just remove the `include_system_collections=False`. We only run the doctests on 4.0+ so we don't need to worry about system collections like (`system.indexes`) showing up in the results.

ShaneHarvey:
Thanks for the contribution! I merged this to master (https://github.com/mongodb/mongo-python-driver/commit/66eb6da601551cea6b32b709aac44495c80e3d5d) and v3.8 (https://github.com/mongodb/mongo-python-driver/commit/170d217a78decde0cefc18f3b6356b8274c51b9d):

PR #405, opened 2019-02-21
PYTHON 1750: Support callbacks for simple types

prashantmital:
Closes https://jira.mongodb.org/browse/PYTHON-1750 .


bson/__init__.py
```diff
@@ -377,6 +377,31 @@ def _get_decimal128(data, position, dummy0, dummy1, dummy2):
     BSONMAX: lambda v, w, x, y, z: (MaxKey(), w)}
 
 
+_BSON_TYPE_MARKER_TO_PYTHON_TYPE_MAP = {
```
behackett: I don't think you need this. You can just do type(value) on the value returned from the standard decoder. I'm guessing you're doing this to avoid the type() call. Is that more expensive than dict lookup?

bson/codec_options.py
```diff
@@ -35,7 +36,7 @@ def _raw_document_class(document_class):
 _options_base = namedtuple(
     'CodecOptions',
     ('document_class', 'tz_aware', 'uuid_representation',
-     'unicode_decode_error_handler', 'tzinfo'))
+     'unicode_decode_error_handler', 'tzinfo', 'encoder_map', 'decoder_map'))
```
behackett: I think we all discussed have a new class to bundle these (and future related options)?

bson/__init__.py
prashantmital: Just a lapse in judgment; I'll rewrite this to use `type(value)`.

bson/codec_options.py
```diff
@@ -32,10 +33,41 @@ def _raw_document_class(document_class):
...
 
+class TypeRegistry(object):
+    def __init__(self, encoder_map=None, decoder_map=None):
+        self.__encoder_map = encoder_map or {}
+        self.__decoder_map = decoder_map or {}
+
+    def copy(self):
+        return type(self)(self.__encoder_map.copy(), self.__decoder_map.copy())
+
+    def _arguments_repr(self):
```
ShaneHarvey: I don't think this will be used anywhere.

bson/codec_options.py
```diff
@@ -32,10 +33,41 @@ def _raw_document_class(document_class):
     return marker == _RAW_BSON_DOCUMENT_MARKER
 
 
+class TypeRegistry(object):
+    def __init__(self, encoder_map=None, decoder_map=None):
+        self.__encoder_map = encoder_map or {}
+        self.__decoder_map = decoder_map or {}
```
ShaneHarvey: I think the registry should make a copy of each map here. That way if the app makes changes to the maps after creating the TypeRegistry, the changes will not be reflected in the TypeRegistry.

bson/codec_options.py
```diff
@@ -32,10 +33,41 @@ def _raw_document_class(document_class):
...
+
+    def copy(self):
+        return type(self)(self.__encoder_map.copy(), self.__decoder_map.copy())
```
ShaneHarvey: Do we still need this when `__init__` creates the copy itself?

bson/codec_options.py
```diff
@@ -32,10 +33,41 @@ def _raw_document_class(document_class):
...
+
+    def _arguments_repr(self):
+        return ('Encode=%r, Decode=%r' % (list(self.__encoder_map.keys()), list(self.__decoder_map.keys())))
+
+    def __repr__(self):
+        return '%s(%s)' % (
+            self.__class__.__name__,
+            'encodable_types=%r, decodable_types=%r' % (
+                [t.__name__ for t in self.__encoder_map.keys()],
+                [t.__name__ for t in self.__decoder_map.keys()]))
```
ShaneHarvey: I think it would be clearer to follow the "`repr` returns an executable representation pattern" so the this should use `__init__`'s argument names (`encoder_map/decoder_map`).

bson/codec_options.py
```diff
@@ -126,9 +160,14 @@ def __new__(cls, document_class=dict,
                 raise ValueError(
                     "cannot specify tzinfo without also setting tz_aware=True")
 
+        type_registry = type_registry or TypeRegistry()
```
ShaneHarvey: I don't think we can have the `or TypeRegistry()` here. Creating a new TypeRegistry for each CodecOptions breaks equality checks. We should leave it as `None` and check for `None` before calling any methods on `opts.type_registry`.

prashantmital: How about implementing `__eq__` for `TypeRegistry` instead? I imagine equality checks on `CodecOptions` instances will be quite rare so performance isn't a major consideration (since comparing `None` values will obviously be faster than comparing 2 dictionaries). Also, I really don't envy the thought of sanity-checking `type_registry` before using it everywhere in our code. 
```
def __eq__(self, other):
    return (self._encoder_map == other._encoder_map) and (self._decoder_map == other._decoder_map)
```

ShaneHarvey: STGM!

bson/codec_options.py
```diff
@@ -33,29 +33,58 @@ def _raw_document_class(document_class):
...
+    def transform_python(self, value):
+        """Convert the given Python object into something serializable."""
+        raise NotImplementedError
+
+
 class TypeRegistry(object):
-    def __init__(self, encoder_map=None, decoder_map=None):
-        self._encoder_map = (encoder_map or {}).copy()
-        self._decoder_map = (decoder_map or {}).copy()
+    def __init__(self, *args):
```
ShaneHarvey: Why `*args` instead of a single argument which is a list or tuple or `TypeCodecBase`'s ? Like
```
    def __init__(self, type_codecs):
```
And if we keep it as `*args` we should still change the name to `*type_codecs` or something else more informative.

bson/codec_options.py
```diff
@@ -33,29 +33,58 @@ def _raw_document_class(document_class):
...
+        for arg in args:
+            if not isinstance(arg, TypeCodecBase):
+                raise TypeError(
+                    "Expected instance of %s" % (TypeCodecBase.__name__,))
+            try:
+                python_type = arg.python_type
+            except NotImplementedError:
+                pass
+            else:
+                self._encoder_map[python_type] = arg.transform_python
```
ShaneHarvey: Should we do something special if  `arg.transform_python` (or `arg.transform_bson`) raises NotImplementedError?

bson/_cbsonmodule.c
```diff
@@ -510,8 +536,24 @@ static int write_element_to_buffer(PyObject* self, buffer_t buffer,
     int result;
     if(Py_EnterRecursiveCall(" while encoding an object to BSON "))
         return 0;
+
+    /* Transform types that have a registered converter.
+     * A new reference is created upon transformation. */
+    PyObject* value_type = PyObject_Type(value);
```
behackett: These declarations have to go above the if block above or this won't compile in VC++.

bson/_cbsonmodule.c
```diff
@@ -2483,6 +2526,13 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
     }
 
     if (value) {
+        PyObject* value_type = PyObject_Type(value);
+        PyObject* converter = NULL;
+        if ((converter = PyDict_GetItem(options->type_registry.decoder_map, value_type)) != NULL) {
+            PyObject* args = PyTuple_Pack(1, value);
+            value = PyObject_CallObject(converter, args);
+        }
+        Py_DECREF(value_type);
```
behackett: You also need to Py_DECREF the previous value for "value".

bson/_cbsonmodule.c
```diff
@@ -510,8 +536,24 @@ static int write_element_to_buffer(PyObject* self, buffer_t buffer,
...
     if(Py_EnterRecursiveCall(" while encoding an object to BSON "))
         return 0;
+
+    /* Transform types that have a registered converter.
+     * A new reference is created upon transformation. */
+    PyObject* value_type = PyObject_Type(value);
+    PyObject* converter = NULL;
+    int new_reference = 0;
+    if ((converter = PyDict_GetItem(options->type_registry.encoder_map, value_type)) != NULL) {
+        PyObject* args = PyTuple_Pack(1, value);
```
behackett: I think you can avoid this by using PyObject_CallFunctionObjArgs instead of PyObject_CallObject.

https://docs.python.org/3/c-api/object.html#c.PyObject_CallFunctionObjArgs

bson/_cbsonmodule.c
```diff
@@ -2483,6 +2526,13 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
     }
 
     if (value) {
+        PyObject* value_type = PyObject_Type(value);
+        PyObject* converter = NULL;
+        if ((converter = PyDict_GetItem(options->type_registry.decoder_map, value_type)) != NULL) {
+            PyObject* args = PyTuple_Pack(1, value);
```
behackett: Same comment here about PyObject_CallFunctionObjArgs

bson/_cbsonmodule.c
ShaneHarvey: Nice! I was looking for a method like that but I couldn't find it in the docs.

bson/codec_options.py
prashantmital: I liked the idea of being able to initialize this with an arbitrary number of `TypeCodecs` directly rather than having to first create a list/tuple and then pass it to `__init__`. Another reason I went with this was that I don't see what else we might want to pass to this constructor in the future since the type codecs are all self-describing. That said, we can start using a list/tuple if you don't find my reasoning to be satisfactory.
I take your point about the variable name and will change it.

bson/codec_options.py
prashantmital: I was planning to include some validation logic into `TypeCodecBase.__init__` to ensure that if a codec advertises support for a type on the encode or decode side, it also implements the corresponding conversion. However, I have no reliable way of figuring out an argument I can use to call the transformer function to test if it raises that exception. The best I could do was call `testvalue = python_type()` followed by a try/except on `arg.transform_python(testvalue)` but that assumes that the default constructor of `python_type` returns a properly initialized object which might not be true for every custom type (though it is true for all types I tried - decimal, int64, set, frozenset). 
What do you think would be the best way to test something like this?

ShaneHarvey: Thanks for explaining that! Now I understand why we can't do this validation. The user will get an exception the first time it's actually used anyway so this SGTM.

bson/codec_options.py
ShaneHarvey: I suspect we might add an extra argument to the registry for https://jira.mongodb.org/browse/PYTHON-1731 (something like `TypeRegistry(fallback_encoder=lambda x: repr(x))`) but there could be a better place to add it. Either way let's cross that bridge when we come to it. Keeping it as is in this PR SGTM.

bson/_cbsonmodule.c
```diff
@@ -455,25 +475,30 @@ static long _type_marker(PyObject* object) {
...
                           &options->tz_aware,
                           &options->uuid_rep,
                           &options->unicode_decode_error_handler,
-                          &options->tzinfo)) {
+                          &options->tzinfo,
+                          &type_registry_obj))
         return 0;
-    }
+
+    convert_type_registry(type_registry_obj, &options->type_registry);
```
ShaneHarvey: Need to check the return value of `convert_type_registry`.

bson/_cbsonmodule.c
```diff
@@ -455,25 +475,30 @@ static long _type_marker(PyObject* object) {
...
 
     type_marker = _type_marker(options->document_class);
     if (type_marker < 0) return 0;
```
ShaneHarvey: There's a potential leak here because `convert_type_registry` increments the reference count of `type_registry_obj` and we can return failure here without a corresponding decref.

I think the simplest fix is to call `convert_type_registry` *after*  `if (type_marker < 0) return 0;`.

bson/_cbsonmodule.c
```diff
@@ -447,6 +447,26 @@ static long _type_marker(PyObject* object) {
...
+ * Return 0 on failure.
+ */
+int convert_type_registry(PyObject* registry_obj, type_registry_t* registry) {
+    registry->encoder_map = PyObject_GetAttrString(registry_obj, "_encoder_map");
+    if (registry->encoder_map == NULL)
+        return 0;
+
+    registry->decoder_map = PyObject_GetAttrString(registry_obj, "_decoder_map");
+    if (registry->decoder_map == NULL)
+        return 0;
```
ShaneHarvey: Before returning `0` here we need to decref the `registry->encoder_map` object.

bson/_cbsonmodule.c
```diff
@@ -501,17 +527,35 @@ void destroy_codec_options(codec_options_t* options) {
     Py_CLEAR(options->document_class);
     Py_CLEAR(options->tzinfo);
     Py_CLEAR(options->options_obj);
+    Py_CLEAR(options->type_registry.registry_obj);
```
ShaneHarvey: Also need to clear the `registry->encoder_map` and `registry->decoder_map`.

bson/codec_options.py
prashantmital: Ok! Leaving it be for now.

bson/_cbsonmodule.c
```diff
@@ -501,17 +527,35 @@ void destroy_codec_options(codec_options_t* options) {
...
+
     if(Py_EnterRecursiveCall(" while encoding an object to BSON "))
         return 0;
-    result = _write_element_to_buffer(self, buffer, type_byte,
-                                      value, check_keys, options);
+
+    if ((converter = PyDict_GetItem(options->type_registry.encoder_map, value_type)) != NULL) {
+        /* Transform types that have a registered converter.
+         * A new reference is created upon transformation. */
+        PyObject* new_value = PyObject_CallFunctionObjArgs(converter, value, NULL);
```
ShaneHarvey: Need to check the return value for NULL and add error handling to decref `value_type`. I suggest rewriting to decref `value_type` ASAP, like this:
```
converter = PyDict_GetItem(options->type_registry.encoder_map, value_type)
Py_DECREF(value_type);
if (converter != NULL) {....
```

bson/_cbsonmodule.c
ShaneHarvey: I suggest writing this function in the style where we:
- declare all `PyObjects *` as NULL upfront.
- have a single error label that xdecrefs all objects
- on any error: `goto <errorlabel>`

bson/_cbsonmodule.c
```diff
@@ -447,6 +447,30 @@ static long _type_marker(PyObject* object) {
     return type;
 }
 
+/* Fill out a type_registry_t* from a TypeRegistry object.
+ *
+ * Return 1 on success. options->document_class is a new reference.
+ * Return 0 on failure.
+ */
+int convert_type_registry(PyObject* registry_obj, type_registry_t* registry) {
+    registry->encoder_map = PyObject_GetAttrString(registry_obj, "_encoder_map");
```
ShaneHarvey: Need to initialize all the `registry` fields to NULL up top. Technically we only need to initialize `registry->decoder_map` but it's good practice if we do it for all the fields.

bson/_cbsonmodule.c
```diff
@@ -455,25 +479,34 @@ static long _type_marker(PyObject* object) {
...
                           &options->unicode_decode_error_handler,
-                          &options->tzinfo)) {
+                          &options->tzinfo,
+                          &type_registry_obj))
         return 0;
-    }
 
-    type_marker = _type_marker(options->document_class);
-    if (type_marker < 0) return 0;
+    if ((type_marker = _type_marker(options->document_class)) < 0)
```
ShaneHarvey: Can you move this back outside the if statement. I think this style of assignment-if-statement makes the code harder to read and more error prone.

bson/_cbsonmodule.c
```diff
@@ -701,6 +754,7 @@ static int _write_element_to_buffer(PyObject* self, buffer_t buffer,
     struct module_state *state = GETSTATE(self);
     PyObject* mapping_type;
     PyObject* uuid_type;
+
```
ShaneHarvey: Nit: Unneeded newline.

bson/codec_options.py
```diff
@@ -32,10 +33,63 @@ def _raw_document_class(document_class):
...
+                pass
+            else:
+                self._decoder_map[bson_type] = codec.transform_bson
+
+    def __repr__(self):
+        return '%s%r' % (self.__class__.__name__, self.__args)
+
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return False
```
behackett: I think you want to return NotImplemented here. https://docs.python.org/3/library/constants.html#NotImplemented

typemap.py
```diff
@@ -0,0 +1,61 @@
+from decimal import Decimal
```
behackett: Are you going to turn this into a set of unittests in test_bson.py?

prashantmital: Just did. See `test_bson_custom_types.py`.

bson/codec_options.py
prashantmital: Ah, good point.

prashantmital:
A couple of doubts/questions:

- Should I also add tests for `decode_iter` and `decode_file_iter` to `test_bson_custom_types.py`?
- Should there be tests that round trip documents with custom types to/from the server (current tests only test BSON library functionality)?

I am also yet to verify driver functionality and behavior when a user-provided type-codec modifies how a fundamental type is read/written. In the event that a codec transforms a type that is also used to represent some internally-used value, we might start running into errors. Thanks, @ShaneHarvey for reminding me of that case.

test/test_bson_custom_types.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
```
ShaneHarvey: Any specific reason for the utf-8 encoding?

prashantmital: ⌘-C + ⌘-V 😀

prashantmital: Removing it.

bson/codec_options.py
```diff
@@ -32,10 +33,63 @@ def _raw_document_class(document_class):
...
+
+class TypeRegistry(object):
+    def __init__(self, *type_codecs):
+        self.__args = type_codecs
+        self._encoder_map = {}
+        self._decoder_map = {}
+        for codec in type_codecs:
+            if not isinstance(codec, TypeCodecBase):
+                raise TypeError("Expected instance of %s, got %s instead" % (
+                    TypeCodecBase.__name__, type(codec).__name__))
```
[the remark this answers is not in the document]
prashantmital: Yes this error message was pretty awful. I've updated it to print the repr.

bson/__init__.py
```diff
@@ -756,10 +757,11 @@ def _name_value_to_bson(name, value, check_keys, opts):
     # Custom encoder (if any) takes precedence over default encoders.
     # Using 'if' instead of 'try...except' for performance since this will
     # usually not be true.
-    # No support for auto-encoding subtypes of registered custom types.
-    custom_encoder = opts.type_registry._encoder_map.get(type(value))
-    if custom_encoder is not None:
-        value = custom_encoder(value)
+    # No support for auto-encoding subtypes of registered cstom types.
```
behackett: cstom -> custom

prashantmital:
I still don't know whether it makes sense to add tests for `decode_iter` and `decode_file_iter` to `test_bson_custom_types.py`?

bson/codec_options.py
```diff
@@ -32,10 +33,82 @@ def _raw_document_class(document_class):
...
+            if not isinstance(codec, TypeCodecBase):
+                raise TypeError(
+                    "Expected an instance of %s, got %r instead" % (
+                    TypeCodecBase.__name__, codec))
+            try:
+                python_type = codec.python_type
+            except NotImplementedError:
+                pass
+            else:
+                self._encoder_map[python_type] = codec.transform_python
```
prashantmital: @ShaneHarvey @behackett do you think a warning should be generated when multiple codecs target the same type (and therefore, one of them gets overridden)? I have added a note about to the function docstring above, but if you think that is not enough, I can add a warning.

doc/examples/custom_type.rst
```diff
@@ -0,0 +1,146 @@
...
+
+  >>> from pymongo import MongoClient
+  >>> client = MongoClient()
+  >>> client.drop_database('custom_type_example')
+  >>> db = client.custom_type_example
+
+
+Since the purpose of the example is to demonstrate working with custom types,
+we’ll need a custom data type to use. For this example, we will be working with
+the :py:class:`~decimal.Decimal` type from Python's standard library. Since the
```
ShaneHarvey: Does `:py:class:` work for you? It doesn't generate a link for me (running Sphinx v1.6.5).

doc/examples/custom_type.rst
```diff
@@ -0,0 +1,146 @@
...
+
+The Type Codec
+--------------
+
+In order to encode custom types, we must first define a **type codec** for our
+type. A type codec describes how an instance of a custom type can be
+*transformed* to/from one of the types :mod:`~bson` already understands, and
+can encode/decode. Type codecs must inherit from
+:class:`~bson.codec_options.TypeCodecBase`. Codecs that facilitate encoding of
+a custom type, must implement the ``python_type`` property, and the
```
ShaneHarvey: `Codecs that facilitate encoding of
a custom type, must implement ...` -> `In order to encode a custom type, a codec must implement the ``python_type`` property and the ``transform_python`` method`.

And same for the next sentence.

doc/examples/custom_type.rst
prashantmital: I needed to add the intersphinx inventory. It works now, but I am not sure whether to link to Python 2 or 3 docs. Changes introduced in `61f9ede` make it so that we link to the standard library docs corresponding to the python version being used to run `sphinx` on to compile PyMongo docs.

doc/examples/custom_type.rst
```diff
@@ -0,0 +1,146 @@
...
+The type registry can then be used to create a custom-type-aware
+:class:`~pymongo.collection.Collection`. Read and write operations
+issued against the resulting collection object transparently manipulate
+documents as they are saved or retrieved from MongoDB.
+
+
+Setup
+-----
+
+We’ll start by getting a clean database to use for the example:
```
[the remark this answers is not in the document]
prashantmital: Good catch! I have no clue where the `’` comes from.

test/test_bson.py
```diff
@@ -909,21 +909,44 @@ def test_bad_id_keys(self):
...
-                "transform_python": lambda self, val: str(val),
-                "bson_type": str,
-                "transform_bson": lambda self, val: types[1](val)}))
-        cls.types = types
-        cls.codecs = codecs
+        class MyIntType(int):
+            pass
+
+        class MyStrType(str):
+            pass
```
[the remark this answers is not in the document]
prashantmital: Good point. I modified the test classes to ensure that PyMongo cannot encode them out of the box.

bson/codec_options.py
```diff
@@ -34,6 +34,14 @@ def _raw_document_class(document_class):
 
 
 class TypeCodecBase(object):
+    """Base class for defining type codec classes which describe how a
+    custom type can be transformed to/from one of the types BSON already
+    understands, and can encode/decode.
+
+    Codec classes must implement the ``python_type`` property, and the
+    ``transform_python`` method to facilitate encoding, or the ``bson_type``
```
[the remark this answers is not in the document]
prashantmital: Done.

doc/conf.py
```diff
@@ -166,6 +167,9 @@
 # If false, no module index is generated.
 #latex_use_modindex = True
 
+PY_VERSION = 3 if bson.py3compat.PY3 else 2
+
 intersphinx_mapping = {
     'gevent': ('http://www.gevent.org/', None),
+    'py': ('https://docs.python.org/%i/' % (PY_VERSION,), None),
```
prashantmital: @ShaneHarvey @behackett should we link to Py2 or Py3 standard library documentation? I feel like this isn't a great solution since we only ever have a single version of the docs hosted so we'd link to whatever Python version is used to compile the docs. Personally, I think we should just link to Python 3 docs.

bson/codec_options.py
ShaneHarvey: Let's not add a warning for now.

doc/conf.py
ShaneHarvey: I think we should always link to Python 3.

doc/examples/custom_type.rst
ShaneHarvey: I think we should always link to Python 3.

prashantmital:
Something strange is going on with Travis that is causing the failures on `pypy`:
image

It shouldn't matter though as I'll be shortly running a large chunk of our EVG suite :)

doc/conf.py
```diff
@@ -166,6 +166,8 @@
 # If false, no module index is generated.
 #latex_use_modindex = True
 
+
 intersphinx_mapping = {
     'gevent': ('http://www.gevent.org/', None),
+    'py': ('https://docs.python.org/3/', None),
```
behackett: Neat! There are _so many_ places we can use this.

PR #406, opened 2019-02-22
PYTHON-1684 Support sharded transactions recovery token

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/PYTHON-1684

This change implements sharded transaction support and testing for MongoDB 4.2 according to [SPEC-1168](https://jira.mongodb.org/browse/SPEC-1168).

In summary:
- sharded transactions have the same API as replica set transactions.
- On the first operation within a sharded transaction, the session is pinned to the selected mongos and all* subsequent operations are routed to the same server.
- When any command within a sharded transaction fails with a transient error, the session is unpinned so that a subsequent `abortTransaction` can complete immediately without waiting for a potentially unreachable mongos.
- The `recoveryToken` field is updated on each command response and sent along with all `commitTransaction` attempts.

To accomplish these goals, these changes refactor how Cursor/CommandCursor decode responses so that a command failure from a `find` or `getMore` can reset the server and unpin the session.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,8 +1,8 @@
 Changelog
 =========
 
-Changes in Version 3.9.0
-------------------------
+Changes in Version 3.9.0.dev0
+-----------------------------
 
 Version 3.9 adds support for MongoDB 4.2. Highlights include:
 
@@ -40,8 +40,8 @@ in this release.
 
 .. _PyMongo 3.9 release notes in JIRA: https://jira.mongodb.org/secure/ReleaseNote.jspa?projectId=10004&version=21787
 
-Changes in Version 3.8.0
-------------------------
+Changes in Version 3.8.0.dev0
+-----------------------------
 
 .. warning:: PyMongo no longer supports Python 2.6. RHEL 6 users should install
   Python 2.7 or newer from `Red Hat Software Collections
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,23 +1,14 @@
 Changelog
 =========
 
-Changes in Version 3.8.0
+Changes in Version 3.9.0
 ------------------------
 
-.. warning:: PyMongo no longer supports Python 2.6. RHEL 6 users should install
-  Python 2.7 or newer from `Red Hat Software Collections
-  <https://www.softwarecollections.org>`_. CentOS 6 users should install Python
-  2.7 or newer from `SCL
-  <https://wiki.centos.org/AdditionalResources/Repositories/SCL>`_
+Version 3.9 adds support for MongoDB 4.2. Highlights include:
 
-.. warning:: PyMongo no longer supports PyPy3 versions older than 3.5. Users
-  must upgrade to PyPy3.5+.
-
-- :class:`~bson.objectid.ObjectId` now implements the `ObjectID specification
-  version 0.2 <https://github.com/mongodb/specifications/blob/master/source/objectid.rst>`_.
-
-
-- Version 3.8.0 implements the `URI options specification`_ in the
+- Support for MongoDB 4.2 sharded transactions. Sharded transactions have
+  the same API as replica set transactions. See :ref:`transactions-ref`.
+- Implement the `URI options specification`_ in the
   :meth:`~pymongo.mongo_client.MongoClient` constructor. Consequently, there are
   a number of changes in connection options:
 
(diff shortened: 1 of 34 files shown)


pymongo/command_cursor.py
```diff
@@ -149,13 +149,14 @@ def duration(): return datetime.datetime.now() - start
...
 
         try:
-            docs = self._unpack_response(reply,
-                                         self.__id,
-                                         self.__collection.codec_options)
-            if from_command:
-                first = docs[0]
-                client._receive_cluster_time(first, self.__session)
-                helpers._check_command_response(first)
+            with client._reset_on_error(self.__address, self.__session):
```
behackett: _reset_on_error catches and resets on OperationFailure, but this previously avoided resetting on OperationFailure. Is this change correct?

pymongo/client_session.py
```diff
@@ -539,6 +558,14 @@ def advance_operation_time(self, operation_time):
                             "of bson.timestamp.Timestamp")
         self._advance_operation_time(operation_time)
 
+    def _process_response(self, reply):
+        """Process a response to a command that was run with this session."""
+        self._advance_cluster_time(reply.get('$clusterTime'))
+        self._advance_operation_time(reply.get("operationTime"))
+        recovery_token = reply.get('recoveryToken')
+        if recovery_token and self._in_transaction:
```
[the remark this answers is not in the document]
ShaneHarvey: As far as I know, recoveryToken is only returned from commands run inside a transaction. Maybe this would be clearer:
```python
        self._advance_operation_time(reply.get("operationTime"))
        if self._in_transaction:
            recovery_token = reply.get('recoveryToken')
            if recovery_token:
                self._transaction.recovery_token = recovery_token
```

pymongo/command_cursor.py
ShaneHarvey: Yes, we are supposed to reset on certain OperationFailures (like ShutdownInProgress). I think it was an bug that we did not. See https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-master-and-node-is-recovering

I should probably make this change in a new ticket so we can include it in 3.8 as well.

doc/changelog.rst
```diff
@@ -1,23 +1,14 @@
 Changelog
 =========
 
-Changes in Version 3.8.0
+Changes in Version 3.9.0
```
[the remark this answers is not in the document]
prashantmital: I like that... I just want to be able to differentiate between a Changelog belonging to a tagged release and any other changelog (in `master` or some other branch). Right now, I can't do this simply by reading the changelog.

pymongo/client_session.py
```diff
@@ -245,12 +259,19 @@ class _Transaction(object):
     def __init__(self, opts):
         self.opts = opts
         self.state = _TxnState.NONE
-        self.transaction_id = 0
```
ShaneHarvey: `transaction_id` is never used.

pymongo/command_cursor.py
prashantmital: SGTM. No one has complained about this either so I don't think we will need to do backport. 🤞

ShaneHarvey: See https://jira.mongodb.org/browse/PYTHON-1759.

pymongo/mongo_client.py
```diff
@@ -1217,30 +1199,51 @@ def _send_message_with_response(self, operation, exhaust=False,
...
+                set_slave_ok,
+                self.__all_credentials,
+                self._event_listeners,
+                exhaust)
 
-        Returns fn()'s return value on success. On error, clears the server's
-        pool and marks the server Unknown.
+    @contextlib.contextmanager
+    def _reset_on_error(self, server_address, session):
+        """Reset the server according to SDAM errors.
```
prashantmital: Can you add a comment here mentioning which "SDAM errors" this docstring is referring to? 
In the [specification](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#error-handling), errors (and the drivers' handling of them) are categorized and explained in terms of what the driver was trying to do when the error was encountered. It is not immediately clear what part of the spec is being implemented by this helper.

test/utils.py
```diff
@@ -147,6 +147,19 @@ def __getitem__(self, item):
             return ScenarioDict({})
 
 
+class CompareType(object):
```
prashantmital: Nice!

test/test_transactions.py
```diff
@@ -52,15 +52,32 @@
 UNPIN_TEST_MAX_ATTEMPTS = 50
 
 
+class CompareTrue(object):
```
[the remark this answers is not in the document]
prashantmital: This is really nice. I'm sure we can use it elsewhere (if we can remember to do so).

prashantmital:
@ShaneHarvey LGTM other than the less than informative docstring on the `_reset_on_error` method. Also a lot of system failures and failures due to SIGKILL on EVG. Any idea why that is?

pymongo/mongo_client.py
ShaneHarvey: Updated to mention "not master" or "node is recovering" errors from the SDAM spec.

ShaneHarvey:
Merged as two commits, one for PYTHON-1759: https://github.com/mongodb/mongo-python-driver/commit/05565783d9c1a50e1ff76200ca6149db4f45a2d0 and one for PYTHON-1684: https://github.com/mongodb/mongo-python-driver/commit/a886922157e95d41ebdbc4526d6fb0d4a67f1a15

PR #407, opened 2019-02-28
Add changelog for 3.7.2

ShaneHarvey:
Copy/pasted from the 3.7.2 release tag.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,8 +1,8 @@
 Changelog
 =========
 
-Changes in Version 3.8.0
-------------------------
+Changes in Version 3.8.0.dev0
+-----------------------------
 
 .. warning:: PyMongo no longer supports Python 2.6. RHEL 6 users should install
   Python 2.7 or newer from `Red Hat Software Collections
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
I'll merge this to master as well.

prashantmital:
Content looks fine but I think there is something weird about having a section for 3.7.2 changes in the changelog for 3.8. In my head, this is how changelogs are maintained:

- We are ready to release 3.7. We tag release 3.7 on the master branch and update the `3.7.dev` (or equivalent) section in the changelog to reflect all new features/fixes and rename it to 3.7.0.
- The next commit on `master` after the `3.7.0` tag, adds a new section in the changelog, say `3.8.dev` (which is empty for now). It also updates the version values in `setup.py` as appropriate.
- At some point, while work on the `3.8.0` release is ongoing, a critical bug is discovered and fixed and needs to be backported to the `3.7` release. We create a maintenance/release branch based on the `3.7.0` tag, let's call it `maintenance/3.7.x`. The commit introducing the critical bug fix (let's call it commit `abcdef`) is cherry-picked from `master` and applied to this branch. Version numbers are updated and an entry is added to the newly added `3.7.1` section in the changelog that describes the new bug fix. On master, the changelog entry for the same bug fix naturally ends up in the `3.8.dev` section.
- When `3.8.0` is ready for release, we rename the `3.8.dev` changelog section to `3.8.0` and tag the release on `master`.

Notice how the changelog entry corresponding to commit `abcdef` ends up in the `3.7.1` section on the maintenance branch, while showing up in the `3.8.0` section on `master`. I think this is the correct way to manage the changelog - for any given release `x.y.z`, the changelog will have entries for each of the previous versions `a.b.0`, (`a` and `b` such that `a<=x`, `b<y`), but only the latest major-minor version (i.e. the version that is most recently released) would have sections corresponding to the micro version. 

This becomes especially useful when maintenance branches diverge from master due to backports. I am thinking of a situation where we find a bug that must be fixed in `3.7.1`, but the same bug is not present in `3.8.x`. Why then, must the changelog of `3.8.0` contain a line-item for fixing this bug in `3.7.2` when it was never observed?

ShaneHarvey:
The master branch's changelog should include ALL previous releases. For example, see https://github.com/llvtt/mongo-python-driver/commit/69d22503110a4ee8e64a00cc322b4652efc4e7d8#diff-4b4cb2f6291a7768f6f3c32ab73a0b02R834 where we added changelog entries for 2.9.5 and 3.3.1.

> On master, the changelog entry for the same bug fix naturally ends up in the `3.8.dev` section.

The bug fix in your example would not be mentioned in the `3.8.dev` section. It should only be mentioned once in the `3.7.1` section. 

> Why then, must the changelog of `3.8.0` contain a line-item for fixing this bug in `3.7.2` when it was never observed?

As mentioned above it won't be duplicated in the `3.8.0` release notes.

PR #409, opened 2019-03-02
PYTHON-1579: Allow unencoded sub-delimiters in usernames and passwords

prashantmital:
Closes [PYTHON-1579](https://jira.mongodb.org/browse/PYTHON-1579).

### DIFF
--- a/pymongo/uri_parser.py
+++ b/pymongo/uri_parser.py
@@ -26,9 +26,9 @@ except ImportError:
 from bson.py3compat import abc, iteritems, string_type, PY3
 
 if PY3:
-    from urllib.parse import quote, quote_plus, unquote, unquote_plus
+    from urllib.parse import quote_plus, unquote, unquote_plus
 else:
-    from urllib import quote, quote_plus, unquote, unquote_plus
+    from urllib import quote_plus, unquote, unquote_plus
 
 from pymongo.common import (
     get_validated_options, URI_OPTIONS_DEPRECATION_MAP, INTERNAL_URI_OPTION_NAME_MAP)
--- a/pymongo/uri_parser.py
+++ b/pymongo/uri_parser.py
@@ -26,9 +26,9 @@ except ImportError:
 from bson.py3compat import abc, iteritems, string_type, PY3
 
 if PY3:
-    from urllib.parse import unquote_plus
+    from urllib.parse import quote, quote_plus, unquote, unquote_plus
 else:
-    from urllib import unquote_plus
+    from urllib import quote, quote_plus, unquote, unquote_plus
 
 from pymongo.common import (
     get_validated_options, URI_OPTIONS_DEPRECATION_MAP, INTERNAL_URI_OPTION_NAME_MAP)
@@ -40,6 +40,9 @@ SCHEME_LEN = len(SCHEME)
 SRV_SCHEME = 'mongodb+srv://'
 SRV_SCHEME_LEN = len(SRV_SCHEME)
 DEFAULT_PORT = 27017
+
+# List of sub-delimiters as defined in RFC 3986.
+SUBDELIMS = ["!", "$", "&", "'", "(", ")", "*", "+", ",", ";", "="]
 
 
 class _CaseInsensitiveDictionary(abc.MutableMapping):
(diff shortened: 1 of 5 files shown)


test/connection_string/test/valid-auth.json
```diff
@@ -240,6 +240,27 @@
         "authmechanism": "MONGODB-CR"
       }
     },
+    {
+      "description": "Subdelimiters in user/pass don't need escaping (MONGODB-CR)",
```
prashantmital: I am not sure if this test is correctly written but I'm guessing that it is, since a lot of other drivers have already implemented this spec change. The main reason I am skeptical is that the username/password here `!$&'()*+,;=` contains a `+` which, when `unquote`-ed with `unqoted_plus` method drops the `+` in favor of a space, but we expect the parsed username/password to be identical to what is in the input string. Consequently, I had to modfiy the `parse_userinfo` method to use the `unquote` method.

pymongo/uri_parser.py
```diff
@@ -141,7 +149,18 @@ def parse_userinfo(userinfo):
...
+    for value_name, quoted_value in zip(
+            ('username', 'password',), (user, passwd,)):
+        quoted_value_no_sdelims = "".join(
+            [ch for ch in quoted_value if ch not in SUBDELIMS])
+        unquoted_value = unquote_plus(quoted_value_no_sdelims)
+        if not quoted_value_no_sdelims == quote_plus(unquoted_value):
+            raise InvalidURI("%r is not a valid %s" % (
+                quoted_value, value_name))
+
+    return unquote(user), unquote(passwd)
```
ShaneHarvey: Isn't this a breaking change? A plus sign (`+`) used to turn into a space (` `) and now it does not?

pymongo/uri_parser.py
```diff
@@ -141,7 +149,18 @@ def parse_userinfo(userinfo):
     # No password is expected with GSSAPI authentication.
     if not user:
         raise InvalidURI("The empty string is not valid username.")
-    return unquote_plus(user), unquote_plus(passwd)
+
+    # If we cannot round-trip a value, it is not properly URI encoded.
```
prashantmital: This was done to ensure that we raise an exception when getting URI strings like the one in the test entitled "Username with password containing an unescaped percent sign". The URI is `"mongodb://alice%foo:bar@127.0.0.1"`.

pymongo/uri_parser.py
prashantmital: It seems that way. See my comment https://github.com/mongodb/mongo-python-driver/pull/409/files#r261802812. It seems like I MUST make this change in order to get the test entitled "Subdelimiters in user/pass don't need escaping (MONGODB-CR)" to pass.

test/connection_string/test/valid-auth.json
ShaneHarvey: Yeah we definitely can't make this change in 3.9. Technically we can do it in 4.0 since major versions allow breaking changes but I'm not sure we should do it ever. If we make this change then users will have to change all their URIs and code that already relies on `quote_plus`. That sounds very painful.

ShaneHarvey: I think we need to deviate from the spec a bit for backwards compatibility. We should change the "Username with password containing an unescaped percent sign" to `"valid": true` and change "Subdelimiters in user/pass don't need escaping (MONGODB-CR)" to:
```
        "username": "!$&'()* ,;=",
        "password": "!$&'()* ,;=",
```

This will be annoying for us to maintain because we'll need to reapply these patches every time we update the spec tests. Maybe there's a simpler way?

pymongo/uri_parser.py
behackett: I support this change (in PyMongo 4.0). We use (un)quote_plus because of a historical misunderstanding. Using just quote would have the side benefit of PyMongo matching the quoting rules supported by WinKerberos. We'll have to call this out boldly in the changelog for 4.0, however.

prashantmital:
Closing this PR for now since the work has been pushed back to 4.0. Keeping the branch around for resuscitation.

PR #411, opened 2019-03-05
PYTHON-1671 Add ClientSession.with_transaction API

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/PYTHON-1671

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -10,7 +10,7 @@ Version 3.9 adds support for MongoDB 4.2. Highlights include:
   the same API as replica set transactions. See :ref:`transactions-ref`.
 - New method :meth:`pymongo.client_session.ClientSession.with_transaction` to
   support conveniently running a transaction in a session with automatic
-  retries and at most once semantics.
+  retries and at-most-once semantics.
 - Implement the `URI options specification`_ in the
   :meth:`~pymongo.mongo_client.MongoClient` constructor. Consequently, there are
   a number of changes in connection options:
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -8,6 +8,9 @@ Version 3.9 adds support for MongoDB 4.2. Highlights include:
 
 - Support for MongoDB 4.2 sharded transactions. Sharded transactions have
   the same API as replica set transactions. See :ref:`transactions-ref`.
+- New method :meth:`pymongo.client_session.ClientSession.with_transaction` to
+  support conveniently running a transaction in a session with automatic
+  retries and at most once semantics.
 - Implement the `URI options specification`_ in the
   :meth:`~pymongo.mongo_client.MongoClient` constructor. Consequently, there are
   a number of changes in connection options:
--- a/pymongo/client_session.py
+++ b/pymongo/client_session.py
@@ -396,6 +396,16 @@ class ClientSession(object):
           with client.start_session() as session:
               session.with_transaction(callback)
 
+        To pass arbitrary arguments to the ``callback``, wrap your callable
+        with a ``lambda`` like this::
+
+          def callback(session, custom_arg, custom_kwarg=None):
+              # Transaction operations...
+
+          with client.start_session() as session:
+              session.with_transaction(
+                  lambda s: callback(s, "custom_arg", custom_kwarg=1))
+
         In the event an exception, `with_transaction` may retry the commit or
         entire transaction and thus ``callback`` may be invoked multiple times.
         Applications should take care when writing a ``callback`` that changes
--- a/pymongo/client_session.py
+++ b/pymongo/client_session.py
@@ -406,13 +406,14 @@ class ClientSession(object):
               session.with_transaction(
                   lambda s: callback(s, "custom_arg", custom_kwarg=1))
 
-        In the event an exception, `with_transaction` may retry the commit or
-        entire transaction and thus ``callback`` may be invoked multiple times.
-        Applications should take care when writing a ``callback`` that changes
-        application state because the ``callback`` may be executed multiple
-        times by a single call to `with_transaction`. Note, even when the
-        ``callback`` is invoked multiple times, ``with_transaction`` ensures
-        that the transaction will be committed at most once on the server.
+        In the event of an exception, ``with_transaction`` may retry the commit
+        or the entire transaction, therefore ``callback`` may be invoked
+        multiple times by a single call to ``with_transaction``. Developers
+        should be mindful of this possiblity when writing a ``callback`` that
+        modifies application state or has any other side-effects.
+        Note that even when the ``callback`` is invoked multiple times,
+        ``with_transaction`` ensures that the transaction will be committed
+        at-most-once on the server.
 
         The ``callback`` should not attempt to start new transactions, but
         should simply run operations meant to be contained within a
@@ -426,20 +427,19 @@ class ClientSession(object):
         automatically aborts the current transaction. When ``callback`` or
         :meth:`~ClientSession.commit_transaction` raises an exception that
         includes the ``"TransientTransactionError"`` error label,
-        ``with_transaction`` will start a new transaction and re-execute
-        ``callback``.
+        ``with_transaction`` starts a new transaction and re-executes
+        the ``callback``.
 
         When :meth:`~ClientSession.commit_transaction` raises an exception with
         the ``"UnknownTransactionCommitResult"`` error label,
-        ``with_transaction`` will retry the commit until the result of the
+        ``with_transaction`` retries the commit until the result of the
         transaction is known.
 
-        In order to safeguard applications from infinite retry loops,
-        ``with_transaction`` will cease retrying once it has exceeded a fixed
-        timeout period of 120 seconds. Any exception raised by the ``callback``
-        or by :meth:`ClientSession.commit_transaction` after the timeout is
-        reached will be re-raised. Applications that desire a different
-        timeout should not use this method.
+        This method will cease retrying after 120 seconds has elapsed. This
+        timeout is not configurable and any exception raised by the
+        ``callback`` or by :meth:`ClientSession.commit_transaction` after the
+        timeout is reached will be re-raised. Applications that desire a
+        different timeout duration should not use this method.
 
         :Parameters:
           - `callback`: The callable ``callback`` to run inside a transaction.
--- a/pymongo/client_session.py
+++ b/pymongo/client_session.py
@@ -109,8 +109,8 @@ from pymongo.errors import (ConfigurationError,
                             ConnectionFailure,
                             InvalidOperation,
                             OperationFailure,
+                            PyMongoError,
                             ServerSelectionTimeoutError,
-                            WriteConcernError,
                             WTimeoutError)
 from pymongo.helpers import _RETRYABLE_ERROR_CODES
 from pymongo.read_concern import ReadConcern
@@ -285,6 +285,15 @@ _UNKNOWN_COMMIT_ERROR_CODES = _RETRYABLE_ERROR_CODES | frozenset([
     64,    # WriteConcernFailed
 ])
 
+# From the Convenient API for Transactions spec, with_transaction must
+# halt retries after 120 seconds.
+_WITH_TRANSACTION_RETRY_TIME_LIMIT = 120
+
+
+def _within_time_limit(start_time):
+    """Are we within the with_transaction retry limit?"""
+    return monotonic.time() - start_time < _WITH_TRANSACTION_RETRY_TIME_LIMIT
+
 
 class ClientSession(object):
     """A session for ordering sequential operations."""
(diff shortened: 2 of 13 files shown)


pymongo/client_session.py
```diff
@@ -368,6 +377,97 @@ def _inherit_option(self, name, val):
...
+          - `read_concern` (optional): The
+            :class:`~pymongo.read_concern.ReadConcern` to use for this
+            transaction.
+          - `write_concern` (optional): The
+            :class:`~pymongo.write_concern.WriteConcern` to use for this
+            transaction.
+          - `read_preference` (optional): The read preference to use for this
+            transaction. If ``None`` (the default) the :attr:`read_preference`
+            of this :class:`Database` is used. See
+            :mod:`~pymongo.read_preferences` for options.
```
ShaneHarvey: Should we use a single `opts` argument for `TransactionOptions` instead of passing `read_concern`, `write_concern`, and `read_preference` individually?

Personally I would prefer `def with_transaction(self, callback, opts=None)` because it avoids the need to add new keyword arguments every time we add a new transaction option but pymongo usually exposes individually options. Also, `start_transaction` accepts the individual options.

pymongo/client_session.py
```diff
@@ -368,6 +377,97 @@ def _inherit_option(self, name, val):
             return val
         return getattr(self.client, name)
 
+    def with_transaction(self, callback, read_concern=None, write_concern=None,
```
ShaneHarvey: The spec allows some wiggle room in the API we expose for `callback`. For now, I've implemented the `callback` as a function that accepts a single argument, the session.

Should we add support for passing arbitrary options into the callback?:
```python
def callback(session, custom_arg, custom_kwarg=None):
    orders = session.client.db.orders
    inventory = session.client.db.inventory
    orders.insert_one({"sku": "abc123", "qty": 100}, session=session)
    inventory.update_one({"sku": "abc123", "qty": {"$gte": 100}},
                         {"$inc": {"qty": -100}}, session=session)


with client.start_session() as session:
    session.with_transaction(callback, 'custom_arg', custom_kwarg=10)
```

I think that would be handy but I'm not quite sure how to get such an API and also support the necessary transaction options. Also, if there is a way to do this, then we will have to worry about breaking compatibility anytime we introduce a new argument to `with_transaction`, for example when we add a new TransactionOption.

One way to have the "best" of both worlds would be:
```python
def with_transaction(self, callback, args=None, kwargs=None,
                     read_concern=None, write_concern=None,
                     read_preference=None):
     # ...
     args = (,) if args is None else args
     kwargs = {} is kwargs is None else kwargs
     ret = callback(self, *args, **kwargs)
     # ...

# Example:
with client.start_session() as session:
    session.with_transaction(callback, args=('custom_arg',),
                             kwargs={'custom_kwarg': 10})
```
But this doesn't seem very pythonic to me. 

After writing this out I think I'm somewhat convinced that the current API is good enough. If the user wants a callback to accept various arguments they can wrap the callback in a lambda like this:
```python
with client.start_session() as session:
    session.with_transaction(
        lambda s: callback(s, 'custom_arg', custom_kwarg=10))
```
In this case it's a bit cumbersome to pass the session (`s`) because it's already in scope but I think that's fine.

Any thoughts?

behackett: Your idea of accepting args and kwargs isn't totally unprecedented in Python -  https://docs.python.org/2/library/threading.html#threading.Thread

pymongo/client_session.py
behackett: Let's stick with the existing pattern.

pymongo/client_session.py
behackett: I like the args kwargs idea. Also, we'll hopefully get some beta feedback on this and can change our minds before GA.

prashantmital: Same goes for `multiprocessing.Process`: `class multiprocessing.Process(group=None, target=None, name=None, args=(), kwargs={})`. So it definitely seems like this is not an antipattern for Python.

doc/changelog.rst
```diff
@@ -8,6 +8,9 @@ Version 3.9 adds support for MongoDB 4.2. Highlights include:
 
 - Support for MongoDB 4.2 sharded transactions. Sharded transactions have
   the same API as replica set transactions. See :ref:`transactions-ref`.
+- New method :meth:`pymongo.client_session.ClientSession.with_transaction` to
+  support conveniently running a transaction in a session with automatic
+  retries and at most once semantics.
```
prashantmital: at most once --> at-most-once

pymongo/client_session.py
```diff
@@ -368,6 +377,97 @@ def _inherit_option(self, name, val):
...
+              orders = session.client.db.orders
+              inventory = session.client.db.inventory
+              orders.insert_one({"sku": "abc123", "qty": 100}, session=session)
+              inventory.update_one({"sku": "abc123", "qty": {"$gte": 100}},
+                                   {"$inc": {"qty": -100}}, session=session)
+
+          with client.start_session() as session:
+              session.with_transaction(callback)
+
+        In an ideal scenario, `with_transaction` will start a transaction,
```
prashantmital: The future tense is confusing here. Why not say

```
Ideally, `with_transaction` starts a transaction, executes the callback once, and commits the transaction....
```

pymongo/client_session.py
```diff
@@ -368,6 +377,97 @@ def _inherit_option(self, name, val):
...
+        execute the callback once, and commit the transaction. In the event of
```
prashantmital: Since callback is also the name of the argument, can we say

```
...execute `callback` once, and commit...
```

pymongo/client_session.py
```diff
@@ -368,6 +377,97 @@ def _inherit_option(self, name, val):
...
+        error, the commit or entire transaction may need to be retried and
+        thus the callback may be invoked multiple times. Even though the
+        callback may be invoked multiple times, `with_transaction` ensures
```
prashantmital:
```
Even though `callback` may be invoked multiple times...
```

pymongo/client_session.py
```diff
@@ -368,6 +377,97 @@ def _inherit_option(self, name, val):
...
+        that the transaction will be committed at most once on the server.
+
+        In order to safeguard applications from infinite retry loops,
+        `with_transaction` will cease retrying the callback and
+        `commit_transaction` if it has exceeded a fixed timeout period of 120
+        seconds. This limit is non-configurable and is intentionally twice the
```
prashantmital: Nit: I am not sure pointing out the fact that this unconfigurable limit is 2x the default timeout adds any value here.

pymongo/client_session.py
```diff
@@ -368,6 +377,97 @@ def _inherit_option(self, name, val):
...
+            try:
+                ret = callback(self)
+            except Exception as exc:
+                if self._in_transaction:
+                    self.abort_transaction()
+                if (isinstance(exc, PyMongoError) and
+                        exc.has_error_label("TransientTransactionError") and
+                        _within_time_limit(start_time)):
+                    # Retry the entire transaction.
+                    continue
```
prashantmital: I'm sure I'm missing something but it seems like if a transaction keeps encountering a `TransientTransactionError`, the 120 second timelimit would not be respected?

pymongo/client_session.py
```diff
@@ -368,6 +377,97 @@ def _inherit_option(self, name, val):
...
+                raise
+
+            if self._transaction.state in (
```
prashantmital: Should we mention in the docstring then that there is no restriction on the callback explicitly committing the transaction? Currently, the docstring gives the impression that the callback should comprise only of the operations and everything transaction-related is handled by the helper.

prashantmital: Ok, having read the [relevant section of the spec](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst#it-is-assumed-that-the-callback-will-not-start-a-new-transaction-on-the-clientsession) it seems like there IS a limitation on the callback but it doesn't have to do with ending a transaction, it has to do with starting it. I think we should very clearly state that users should not start new transactions within `callback` in the `with_transaction` docstring.

pymongo/client_session.py
prashantmital: Might be more useful to add this as a comment above the variable declaration for the max duration.

doc/changelog.rst

pymongo/client_session.py
ShaneHarvey: Interesting, I didn't realize this pattern was already in use in the standard library. We can easily add support for `args=(), kwargs={}` in a non-breaking way but we can't remove support once we release it. So, I propose we not add support for it now and wait to see if users think the current api is too restrictive.

pymongo/client_session.py
ShaneHarvey: Updated.

pymongo/client_session.py
ShaneHarvey: Thanks for pointing this out! I updated the docstring.

pymongo/client_session.py
ShaneHarvey: Sound good to me.

pymongo/client_session.py
```diff
@@ -381,7 +383,8 @@ def with_transaction(self, callback, read_concern=None, write_concern=None,
                          read_preference=None):
         """Execute a callback in a transaction.
 
-        .. code-block:: python
+        This method starts a transaction on this session, executes ``callback``
+        once, and then commits the transaction. For example::
```
prashantmital: Why remove the `.. code-block:: python` statement? I'm guessing you had that in there so the docs rendered with syntax-highlighting? Is there a reason we don't want that to happen now?

pymongo/client_session.py
```diff
@@ -393,26 +396,46 @@ def callback(session):
           with client.start_session() as session:
               session.with_transaction(callback)
 
-        In an ideal scenario, `with_transaction` will start a transaction,
-        execute the callback once, and commit the transaction. In the event of
-        error, the commit or entire transaction may need to be retried and
-        thus the callback may be invoked multiple times. Even though the
-        callback may be invoked multiple times, `with_transaction` ensures
+        In the event an exception, `with_transaction` may retry the commit or
```
prashantmital: In the event of* an exception...

pymongo/client_session.py
```diff
@@ -393,26 +396,46 @@ def callback(session):
...
+        entire transaction and thus ``callback`` may be invoked multiple times.
```
prashantmital: entire transaction, thereby invoking ``callback`` multiple times.

pymongo/client_session.py
```diff
@@ -393,26 +396,46 @@ def callback(session):
...
+        Applications should take care when writing a ``callback`` that changes
```
prashantmital: When writing ``callback``s that modify application state, developers should be mindful of the possibility that a ``callback`` may be executed multiple times by a single call to ``with_transaction``.

pymongo/client_session.py
```diff
@@ -393,26 +396,46 @@ def callback(session):
...
+        this is handled automatically by ``with_transaction``. If the
+        ``callback`` does commit or abort the transaction without error,
+        however, ``with_transaction`` will return without taking further
+        action.
+
+        When ``callback`` raises an exception, ``with_transaction``
+        automatically aborts the current transaction. When ``callback`` or
+        :meth:`~ClientSession.commit_transaction` raises an exception that
+        includes the ``"TransientTransactionError"`` error label,
+        ``with_transaction`` will start a new transaction and re-execute
```
prashantmital: no need to change tense here... ``with transaction`` starts a new transaction and re-executes ...

pymongo/client_session.py
```diff
@@ -393,26 +396,46 @@ def callback(session):
...
+        ``callback``.
+
+        When :meth:`~ClientSession.commit_transaction` raises an exception with
+        the ``"UnknownTransactionCommitResult"`` error label,
+        ``with_transaction`` will retry the commit until the result of the
```
prashantmital: Same as above... ``with_transaction`` retries the commit...

pymongo/client_session.py
```diff
@@ -393,26 +396,46 @@ def callback(session):
...
+        transaction is known.
+
         In order to safeguard applications from infinite retry loops,
```
prashantmital: How about replacing "from infinite retry loops" wiht "from retrying indefinitely"?

pymongo/client_session.py
```diff
@@ -393,26 +396,46 @@ def callback(session):
...
-        `commit_transaction` if it has exceeded a fixed timeout period of 120
-        seconds. This limit is non-configurable and is intentionally twice the
-        value of MongoDB's default for the  `transactionLifetimeLimitSeconds`
-        parameter (60 seconds). Applications that desire longer retry periods
-        may call `with_transaction` additional times as needed. Applications
-        that desire shorter retry periods should not use this method.
+        ``with_transaction`` will cease retrying once it has exceeded a fixed
+        timeout period of 120 seconds. Any exception raised by the ``callback``
+        or by :meth:`ClientSession.commit_transaction` after the timeout is
+        reached will be re-raised. Applications that desire a different
```
prashantmital: Were you trying to say ... Applications that desire to raise a different exception upon timing out...

pymongo/client_session.py
```diff
@@ -393,26 +396,46 @@ def callback(session):
...
+
         In order to safeguard applications from infinite retry loops,
-        `with_transaction` will cease retrying the callback and
-        `commit_transaction` if it has exceeded a fixed timeout period of 120
-        seconds. This limit is non-configurable and is intentionally twice the
-        value of MongoDB's default for the  `transactionLifetimeLimitSeconds`
-        parameter (60 seconds). Applications that desire longer retry periods
-        may call `with_transaction` additional times as needed. Applications
-        that desire shorter retry periods should not use this method.
+        ``with_transaction`` will cease retrying once it has exceeded a fixed
```
prashantmital: Might be clearer to say... will cease retrying after 120 seconds. This timeout period is fixed and not user-configurable.

pymongo/client_session.py
```diff
@@ -425,7 +448,7 @@ def callback(session):
             :mod:`~pymongo.read_preferences` for options.
 
         :Returns:
-          The return value of the callback.
+          The return value of the ``callback``.
```
prashantmital: Should we add a `:Raises:` section to this as well? Might be useful to have a quick-reference of all the failure modes.

pymongo/client_session.py
prashantmital: Good point about not being able to introduce the current API without a backward breaking change. But it also got me thinking that the standard library API is a superset of what is currently implemented so why not just implement it? Users are free to use `functools.partial` (or something similar) if they so desire. Just a thought!

pymongo/client_session.py
ShaneHarvey: A trailing `::` works the same and is shorter.

pymongo/client_session.py
ShaneHarvey: We could be it would just say `Exception` (from the callback) and `PyMongoError` so I'm not sure how useful that would be

pymongo/client_session.py
ShaneHarvey: I'll open a new ticket to add `args`, `kwargs`. I'm starting to like the idea but I still want to hold off until we get more feedback from real users.

pymongo/client_session.py
```diff
@@ -368,6 +379,128 @@ def _inherit_option(self, name, val):
...
+
+        .. versionadded:: 3.9
+        """
+        start_time = monotonic.time()
+        while True:
+            self.start_transaction(
+                read_concern, write_concern, read_preference)
+            try:
+                ret = callback(self)
```
ShaneHarvey: @ajdavis does this implementation work for motor? Will motor have to rewrite `with_transaction` no matter what because the `callback` will be an async method?

ajdavis: I think Motor can reuse this. Motor should accept a coroutine as a callback, wrap that in another callback which schedules the coroutine on the event loop and waits (blocking this thread) until the coroutine finishes, then passes back its return value or reraises its exception, unblocking this thread.

Once it's wrapped the coroutine in the wrapper callback, Motor will do its usual thing: schedule PyMongo's with_transaction() on the Motor thread pool and return a Future that's resolved when with_transaction() finishes. I can help whoever does this implementation.

ShaneHarvey: Perfect, that makes sense to me.

PR #414, opened 2019-03-06
Close sockets on object destruction

maxpv:
This would be useful when used in Jupyter notebook.
Or when using PyMongo in a generator in Keras (multi-threaded environment).

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -1961,6 +1961,9 @@ class MongoClient(common.BaseObject):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
+    def __del__(self):
+        self.close()
+         
     def __iter__(self):
         return self
 


behackett:
Can you describe the problem you are having that this patch attempts to solve? The __del__ method has a long history of issues in Python, and has caused no end of problems in the the one place we use it in PyMongo. The sockets created by PyMongo are cleaned up when MongoClient is deleted regardless.

maxpv:
Yes `__del__` isn't really a method of choice for ressources management, but it turns out that in IPython environement it is quite useful. Imagine that you execute twice a cell, where you instanciate an object, you'll end up with an object with no reference and should be collected by the garbage collector.

screenshot 2019-03-07 at 08 39 08

In my experience I end up having a decent amount of opened sockets that aren't closed until I restart the kernel or mongod.

In my opinion calling `__del__` is an extra safe layer and isn't meant to replace anything else.

ShaneHarvey:
@maxpv, what Bernie is saying is that it's intentional that MongoClient does not have a `__del__` and even without it, MongoClient will automatically close all sockets shortly after it is garbage collected. So there should be no need for a `__del__`. For example:
```python
>>> import subprocess
>>> import os
>>>
>>> client = MongoClient()
>>> client.admin.command('ping')
{'ok': 1.0, 'operationTime': Timestamp(1552424968, 1), '$clusterTime': {'clusterTime': Timestamp(1552424968, 1), 'signature': {'hash': b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 'keyId': 0}}}
>>>
>>> subprocess.run(['lsof', '-i', '4', '-a', '-p', str(os.getpid())])
COMMAND   PID  USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
Python  17058 shane    3u  IPv4 0xeb59d022e65425df      0t0  TCP localhost:52809->localhost:27017 (ESTABLISHED)
Python  17058 shane    6u  IPv4 0xeb59d023077619bf      0t0  TCP localhost:52810->localhost:27017 (ESTABLISHED)
CompletedProcess(args=['lsof', '-i', '4', '-a', '-p', '17058'], returncode=0)
>>>
>>> client = None
>>> # All sockets are closed:
>>> subprocess.run(['lsof', '-i', '4', '-a', '-p', str(os.getpid())])
CompletedProcess(args=['lsof', '-i', '4', '-a', '-p', '17058'], returncode=1)
```

maxpv:
@ShaneHarvey thanks for your answer. I understand this point of view. I didn't mean to replace the actual system but to add an extra layer for jupyter notebooks environment.

It is quite common to execute several times the same cell. If this cell happens to include a pymongo instantiation it"ll increase the number of connection just because the garbage collector isn't *always* reliable in IPython.

The same goes when using pymongo as a data loader in a multiprocessing environment where you don't control instances creations (like a generator in [keras](https://stanford.edu/~shervine/blog/keras-how-to-generate-data-on-the-fly)).

I understand that it is IPython is a particular environment and might represent a minority of the total users.

pymongo/mongo_client.py
```diff
@@ -1961,6 +1961,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
+    def __del__(self):
+        self.close()
```
ShaneHarvey: We definitely cannot call `close()` in `__del__` because the `close()` method will acquire locks, which can lead to deadlocking the garbage collector. A very similar deadlock situation is described in great detail in https://jira.mongodb.org/browse/PYTHON-799.

PR #416, opened 2019-03-11
PYTHON 1744: Re-sync APM spec tests

prashantmital:


behackett:
@prashantmital this PR got lost in the shuffle long ago. Does it need to be updated, or can @ShaneHarvey review it now?

prashantmital:
I feel like this should have been taken care of by a different PR. I'll double-check and update this thread accordingly.

PR #421, opened 2019-06-11
Crash in Cursor.__del__ when creating a cursor with invalid arguments

arthurdarcet:
If the constructor of Cursor is called with unexpected arguments (ie when calling a `find` wrongly), the `__init__` method does not run at all, but `__del__` is still called and will (of course) crash.

This leads to some noise in the logged error on top of the real TypeError:
```
>>> pymongo.cursor.Cursor(wrong=1)
Exception ignored in: <function Cursor.__del__ at 0x1048129d8>
  File "/Users/arthur/dev/.venvs/Users--arthur--dev--dragonstone/lib/python3.7/site-packages/pymongo/cursor.py", line 241, in __del__
    self.__die()
  File "/Users/arthur/dev/.venvs/Users--arthur--dev--dragonstone/lib/python3.7/site-packages/pymongo/cursor.py", line 298, in __die
    already_killed = self.__killed
AttributeError: 'Cursor' object has no attribute '_Cursor__killed'
TypeError: __init__() got an unexpected keyword argument 'wrong'
```

This PR might not be the best way to solve this, a `__new__` could also work for instance. Let me know if this looks acceptable and if there should also be a regression test.

Thanks!

### DIFF
--- a/pymongo/cursor.py
+++ b/pymongo/cursor.py
@@ -291,7 +291,13 @@ class Cursor(object):
     def __die(self, synchronous=False):
         """Closes this cursor.
         """
-        already_killed = self.__killed
+        # the __init__ might not have run completely here (or at all)
+        try:
+            already_killed = self.__killed
+        except AttributeError:
+            # __init__ did not run to completion (or at all)
+            return
+
         self.__killed = True
         if self.__id and not already_killed:
             if self.__exhaust and self.__exhaust_mgr:
--- a/test/test_cursor.py
+++ b/test/test_cursor.py
@@ -35,7 +35,7 @@ from pymongo import (monitoring,
                      ALL,
                      OFF)
 from pymongo.collation import Collation
-from pymongo.cursor import CursorType
+from pymongo.cursor import Cursor, CursorType
 from pymongo.errors import (ConfigurationError,
                             ExecutionTimeout,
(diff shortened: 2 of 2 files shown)


pymongo/cursor.py
```diff
@@ -291,14 +284,20 @@ def _clone_base(self, session):
     def __die(self, synchronous=False):
         """Closes this cursor.
         """
-        already_killed = self.__killed
+        # the __init__ might not have run completely here (or at all)
```
ShaneHarvey: I think we can just do:
```python
try:
    already_killed = self.__killed
except AttributeError:
    # __init__ failed to run because of an invalid argument.
    return
```
We know that `__killed`, `__id`, `__exhaust`, and `__exhaust_mgr` will always be defined if `__init__` actually runs.

arthurdarcet: yes much better, thanks! I have amended the commit to use this approach

arthurdarcet:
Thanks for the review @ShaneHarvey . I just push a regression test.

ShaneHarvey:
Thanks! Merged in https://github.com/mongodb/mongo-python-driver/commit/876db9939e14c2115f9fdab5dcfc8fa3891d6200.

PR #423, opened 2019-09-06
Fix removed argument 'lifetime'

joackobengochea:
The argument 'lifetime' was removed from resolver.query(), srv_resolver.py crashed when trying to sent it.

### DIFF
--- a/pymongo/srv_resolver.py
+++ b/pymongo/srv_resolver.py
@@ -54,8 +54,7 @@ class _SrvResolver(object):
 
     def get_options(self):
         try:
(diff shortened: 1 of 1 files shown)


behackett:
Lifetime wasn't removed. It was added in dnspython 1.16:

https://github.com/rthalley/dnspython/commit/b782778e1a9374b160d00a3d67370e2be48bc688

See the release notes for PyMongo 3.9, that mention you have to upgrade dnspython to 1.16:

https://api.mongodb.com/python/current/changelog.html#changes-in-version-3-9-0

PR #424, opened 2019-09-15
GridFS.delete returns DeleteResult

drhagen:
`pymongo.Collection.delete_one` returns a `DeleteResult` object which, among other things, provides the `deleted_count` attribute telling whether or not anything was actually deleted by the call. However, `gridfs.GridFS.delete` returns nothing regardless if a document is deleted. This PR changes the behavior of GridFS to more closely follow `Collection` by returning the `DeleteResult` from the underlying call to delete the document index.

### DIFF
--- a/gridfs/__init__.py
+++ b/gridfs/__init__.py
@@ -243,14 +243,18 @@ class GridFS(object):
           - `session` (optional): a
             :class:`~pymongo.client_session.ClientSession`
 
+        :Returns:
+          - An instance of :class:`~pymongo.results.DeleteResult`.
+
         .. versionchanged:: 3.6
            Added ``session`` parameter.
 
         .. versionchanged:: 3.1
            ``delete`` no longer ensures indexes.
         """
-        self.__files.delete_one({"_id": file_id}, session=session)
+        res = self.__files.delete_one({"_id": file_id}, session=session)
         self.__chunks.delete_many({"files_id": file_id}, session=session)
+        return res
 
     def list(self, session=None):
         """List the names of all files stored in this instance of


ShaneHarvey:
Thanks for the pull. The implementation seems fine but [GridFs](https://pymongo.readthedocs.io/en/stable/api/gridfs/index.html#gridfs.GridFS) is pymongo's old API for gridfs. [GridFSBucket](https://pymongo.readthedocs.io/en/stable/api/gridfs/index.html#gridfs.GridFSBucket) is the new API which is implemented according to the [GridFS spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst).

The newer [GridFSBucket.delete](https://pymongo.readthedocs.io/en/stable/api/gridfs/index.html#gridfs.GridFSBucket.delete) method raises a [NoFile](https://pymongo.readthedocs.io/en/stable/api/gridfs/errors.html#gridfs.errors.NoFile) exception when the file does not exist. May I suggest using the newer API instead because it already supports your use-case?

rathisekaran:
Tracking ticket - [PYTHON-2004](https://jira.mongodb.org/browse/PYTHON-2004)

ShaneHarvey:
Hi @drhagen, we have decided not to add this feature to the old GridFS api. The newer [GridFSBucket.delete](https://pymongo.readthedocs.io/en/stable/api/gridfs/index.html#gridfs.GridFSBucket.delete) method already supports this feature. Please let us know if you have any issues using GridFSBucket:
```python
import gridfs
my_db = MongoClient().test
fs = gridfs.GridFSBucket(my_db)
try:
    fs.delete(file_id)
except gridfs.errors.NoFile:
    # file_id did not exist.
    pass
```

PR #425, opened 2019-10-22
Fix Python 3.8 SyntaxWarning: "is not" with a literal

andersk:
Fixes this warning from Python 3.8:

```
bson/json_util.py:702: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if doc['$minKey'] is not 1:
bson/json_util.py:711: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if doc['$maxKey'] is not 1:
```

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -699,7 +699,7 @@ def _parse_canonical_decimal128(doc):
 
 def _parse_canonical_minkey(doc):
     """Decode a JSON MinKey to bson.min_key.MinKey."""
-    if doc['$minKey'] is not 1:
+    if type(doc['$minKey']) is not int or doc['$minKey'] != 1:
         raise TypeError('$minKey value must be 1: %s' % (doc,))
     if len(doc) != 1:
         raise TypeError('Bad $minKey, extra field(s): %s' % (doc,))
@@ -708,7 +708,7 @@ def _parse_canonical_minkey(doc):
 
 def _parse_canonical_maxkey(doc):
     """Decode a JSON MaxKey to bson.max_key.MaxKey."""
-    if doc['$maxKey'] is not 1:
+    if type(doc['$maxKey']) is not int or doc['$maxKey'] != 1:
         raise TypeError('$maxKey value must be 1: %s', (doc,))
     if len(doc) != 1:
         raise TypeError('Bad $minKey, extra field(s): %s' % (doc,))


ShaneHarvey:
Thanks @andersk! I think the use of "is not" was intentional here as otherwise our json parser will incorrectly parse these values as MinKey:
```
>>> json_util.loads('{"a": {"$minKey": {"$numberLong": 1}}}')
{'a': MinKey()}
>>> json_util.loads('{"a": {"$minKey": {"$numberDouble": "1"}}}')
{'a': MinKey()}
```

Currently this raises an error:
```
>>> json_util.loads('{"a": {"$minKey": {"$numberLong": 1}}}')
  File "bson/json_util.py", line 410, in loads
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 351, in loads
    return cls(encoding=encoding, **kw).decode(s)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 382, in raw_decode
    obj, end = self.scan_once(s, idx)
  File "bson/json_util.py", line 409, in <lambda>
    pairs, json_options)
  File "bson/json_util.py", line 429, in object_pairs_hook
    return object_hook(json_options.document_class(pairs), json_options)
  File "bson/json_util.py", line 442, in object_hook
    return _parse_canonical_minkey(dct)
  File "bson/json_util.py", line 703, in _parse_canonical_minkey
    raise TypeError('$minKey value must be 1: %s' % (doc,))
TypeError: $minKey value must be 1: {u'$minKey': 1L}
```

Although I wonder if this matters given our general philosophy of being loose when parsing JSON but strict when generating JSON. For example we already accept:
```
>>> json_util.loads('{"a": {"$minKey": {"$numberInt": "1"}}}')
{u'a': MinKey()}
```

I will discuss this with the team and report back.

andersk:
The `is not` operator isn’t a “strict” version of `!=`. It’s an object identity test, not an equality test. Its behavior on integers is an unspecified implementation detail that one [should never rely on](https://bugs.python.org/issue34850).

```pycon
>>> x = 100
>>> x is not 100
False
>>> x = 1000
>>> x is not 1000
True
>>> x = 1000; x is not 1000
False
```

Here’s a version that passes the existing tests.

prashantmital:
>  It’s an object identity test

This is my understanding as well.

PR #426, opened 2019-10-30
Test against Python 3.7 on Travis

rooterkyberian:


.travis.yml
```diff
@@ -5,6 +5,7 @@ python:
   - 3.4
   - 3.5
   - 3.6
+  - 3.7
```
ShaneHarvey: Can we add Python 3.8 as well? https://docs.travis-ci.com/user/languages/python/#specifying-python-versions

rooterkyberian:
reverted 3.8 change - lets keep this PR simple - i.e. to orginal and working scope of working testing against Python 3.7, other change requests can be done in other PRs

ShaneHarvey:
Sure that's fine. Also, FYI we do already test against Python 3.7. Our main CI/testing tool is Evergreen and we have a very, very extensive test matrix which includes all versions of Python that we support (and even some we don't support) on all platforms that we support. You can check it out here: https://evergreen.mongodb.com/waterfall/mongo-python-driver

We only use Travis to make it easier for external contributors who don't have access submit jobs to Evergreen.

PR #429, opened 2019-11-21
[PYTHON-2061] bson: check for negative entry size in read_file_iter.

thetorpedodog:
This is a cut at [bug 2061](https://jira.mongodb.org/projects/PYTHON/issues/PYTHON-2061). I’m a little unsure about whether the test change I’ve done is the best way to do it—the solution I have seems…OK, but not ideal.

I’m open to any feedback about this. Thanks!

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -1166,6 +1166,9 @@ def decode_file_iter(file_obj, codec_options=DEFAULT_CODEC_OPTIONS):
         elif len(size_data) != 4:
             raise InvalidBSON("cut off in middle of objsize")
         obj_size = _UNPACK_INT_FROM(size_data, 0)[0] - 4
+        if obj_size < 0:
+            raise InvalidBSON(
+                "invalid objsize: {!r} -> {}".format(size_data, obj_size))
         elements = size_data + file_obj.read(obj_size)
         yield _bson_to_dict(elements, codec_options)
 
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -1166,10 +1166,7 @@ def decode_file_iter(file_obj, codec_options=DEFAULT_CODEC_OPTIONS):
         elif len(size_data) != 4:
             raise InvalidBSON("cut off in middle of objsize")
         obj_size = _UNPACK_INT_FROM(size_data, 0)[0] - 4
-        if obj_size < 0:
-            raise InvalidBSON(
-                "invalid objsize: {!r} -> {}".format(size_data, obj_size))
-        elements = size_data + file_obj.read(obj_size)
+        elements = size_data + file_obj.read(max(0, obj_size))
         yield _bson_to_dict(elements, codec_options)
 
 
--- a/test/test_bson.py
+++ b/test/test_bson.py
@@ -18,8 +18,10 @@
 
 import collections
 import datetime
+import os
 import re
 import sys
+import tempfile
 import uuid
 
 sys.path[0:0] = [""]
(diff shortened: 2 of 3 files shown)


bson/__init__.py
```diff
@@ -1166,6 +1166,9 @@ def decode_file_iter(file_obj, codec_options=DEFAULT_CODEC_OPTIONS):
         elif len(size_data) != 4:
             raise InvalidBSON("cut off in middle of objsize")
         obj_size = _UNPACK_INT_FROM(size_data, 0)[0] - 4
+        if obj_size < 0:
+            raise InvalidBSON(
+                "invalid objsize: {!r} -> {}".format(size_data, obj_size))
```
ShaneHarvey: What if instead of raising InvalidBSON explicitly, we let _bson_to_dict handle the error?:
```python
if obj_size <= 0:
    # Invalid object size, continue and let _bson_to_dict raise an error.
    elements = size_data
else:
    elements = size_data + file_obj.read(obj_size)
yield _bson_to_dict(elements, codec_options)
```

Then the following two files would raise the error from the same place (`_bson_to_dict`):
```python
import bson
with open('example1.bson', 'wb') as outfile:
  outfile.write(b'\x00\x00\x00\x00')
with open('example1.bson', 'rb') as infile:
    list(bson.decode_file_iter(infile))
```
```python
import bson
with open('example2.bson', 'wb') as outfile:
  outfile.write(b'\x05\x00\x00\x00')
with open('example2.bson', 'rb') as infile:
    list(bson.decode_file_iter(infile))

```

test/test_bson.py
```diff
@@ -58,6 +58,21 @@
     long = int
 
 
+class StrictStringIO(object):
```
ShaneHarvey: Can we add a new test with your repro (ie using a real file returned by `open()`)? That would help catch future changes to Python's api.

bson/__init__.py
thetorpedodog: I like this idea. I'm using min() here since `read(0)` will do what it Should™ (that is, read zero bytes and return `b''`).

test/test_bson.py
thetorpedodog: Done.  I've also turned the test into a more Go-like loop-based test to reduce repetition.

thetorpedodog: A couple follow-up thoughts:

- Would it be worth returning the `StrictStringIO` tests to just use `StringIO`?
- Or maybe get rid of them entirely?

bson/__init__.py
```diff
@@ -1166,7 +1166,7 @@ def decode_file_iter(file_obj, codec_options=DEFAULT_CODEC_OPTIONS):
         elif len(size_data) != 4:
             raise InvalidBSON("cut off in middle of objsize")
         obj_size = _UNPACK_INT_FROM(size_data, 0)[0] - 4
-        elements = size_data + file_obj.read(obj_size)
+        elements = size_data + file_obj.read(max(0, obj_size))
```
ShaneHarvey: Neat!

test/test_bson.py
ShaneHarvey: Thanks! Yeah I think we can remove `StrictStringIO` and keep the `StringIO` ones as they were before.

thetorpedodog: Done. Also fixed the formatting issues I introduced (so used to 2-space indents…).

Also added a message to the `assertRaises`es so that if it does fail it tells which input caused it.

PR #432, opened 2019-12-07
Fix simple typo: overidden -> overridden

timgates42:
There is a small typo in doc/examples/authentication.rst.
Should read overridden rather than overidden.


ShaneHarvey:
Thanks!

PR #436, opened 2020-01-29
adds note to gridfs.get function that the id needs to be an ObjectId

dskleingeld:
While working on an distributed sorting system (as a class assignment) I could not get the gridfs [get](https://api.mongodb.com/python/current/api/gridfs/index.html#gridfs.GridFS.get) function to work. After a while I found out the problem was that the `id` I was using (send over from another machine) was a `str` while it should be an `ObjectId`.

Currently this is not documented in `GridFS`. I propose to add a small note to the inline documentation of the `get` function. 

An alternative solution would be a type check on the id using `isinstance` and constucting a ObjectId from the `id` if possible:

```python
def get(self, file_id, session=None):
    if not isinstance(id, bson.objectid.ObjectId):
        id = bson.objectid.ObjectId(id)

```


ShaneHarvey:
A GridFS file's `_id` is not required to be an ObjectId. Custom file `_id` types are fully supported so we can't merge this change. 

The [GridFS spec "Terms section"](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#terms) defines a file's `_id` as:
>  a unique ID for this document. Usually this will be of type ObjectId, but a custom _id value provided by the application may be of any type.

dskleingeld:
Maybe a warning/note would be useful at the end of the `get` function documentation and a type description in the `put` function? I was thinking something along the lines of:

```
The id is type sensitive, ensure the same type is used for putting and getting data from GridFS. If the id was generated by put it should (still) be an instance of :class: `~bson.objectid.ObjectId`
```

If that seems useful I will update the files. Otherwise feel free to close this PR.

ShaneHarvey:
Thanks for the pull. However, we're not going to merge it because we already document this behavior in two places:

- https://pymongo.readthedocs.io/en/stable/faq.html#when-i-query-for-a-document-by-objectid-in-my-web-application-i-get-no-result
- https://pymongo.readthedocs.io/en/stable/tutorial.html#querying-by-objectid

yingshaoxo:
Clearly, I think you guys just have to add a tip at here, it'll be enough: https://api.mongodb.com/python/current/examples/gridfs.html?highlight=gridfs#saving-and-retrieving-data


If all of you are busy and can't make a pull about it, I'd like to help, to make a little bit of changes to that doc and make a pull request.

PR #439, opened 2020-04-14
Move sock.settimeout() call to before the SSL handshake occurs

slatermethuen:
On a heavily loaded virtual machine running multiple docker containers I have been encountering an intermittent SSL handshake timeout error.   The error was was being raised in `pool.py` in the `_configured_socket()` function on the line where `ssl_context.wrap_socket()` is called.

This patch moves the `sock.settimeout()` call to before the `wrap_socket()` calls so that the SSL handshake process uses the configured timeout.

### DIFF
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -882,6 +882,7 @@ def _configured_socket(address, options):
     Sets socket's SSL and timeout options.
     """
     sock = _create_connection(address, options)
+    sock.settimeout(options.socket_timeout)
     ssl_context = options.ssl_context
 
     if ssl_context is not None:
@@ -918,7 +919,6 @@ def _configured_socket(address, options):
                 sock.close()
                 raise
 
-    sock.settimeout(options.socket_timeout)
     return sock
 
 


ShaneHarvey:
The `_create_connection` function already sets the socket's timeout to `options.connect_timeout` before the TLS handshake [on line 856](https://github.com/mongodb/mongo-python-driver/blob/3.11.0b0/pymongo/pool.py#L856). `options.connect_timeout` corresponds to the [connectTimeoutMS option](https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient):

> connectTimeoutMS: (integer or None) Controls how long (in milliseconds) the driver will wait during server monitoring when connecting a new socket to a server before concluding the server is unavailable. Defaults to 20000 (20 seconds).

`options.socket_timeout` (socketTimeoutMS) is the timeout to use for regular database operations which is why it is only set _after_ connecting the socket. 

In summary I think pymongo is working as intended here. To avoid the TLS handshake timeouts your app can increase connectTimeoutMS like this:
```python
client = MongoClient('mongodb://host/?connectTimeoutMS=60000')  # 60 second connect timeout
# or via keyword argument:
client = MongoClient('mongodb://host/', connectTimeoutMS=60000)  # 60 second connect timeout
```

Okay to close?

slatermethuen:
Looks like I need to do some further investigation!  I was already setting `connectTimeoutMS` to 30000.   I had convinced myself that this patch did fix the issue, but from what you are saying it sounds like there must be something else going on.

I will do some more tests today with extra instrumentation and let you know the results.

ShaneHarvey:
This change might seem to "fix" the issue because the default `options.socket_timeout` (socketTimeoutMS) is None, which means never timeout:
> socketTimeoutMS: (integer or None) Controls how long (in milliseconds) the driver will wait for a response after sending an ordinary (non-monitoring) database operation before concluding that a network error has occurred. Defaults to None (no timeout).

I suspect you can increase `connectTimeoutMS` to > 30 seconds (or put less load on the VM of course).

PR #444, opened 2020-06-24
PYTHON-2136: added more diagnostic information for ServerSelectionTimeoutError

juliusgeo:


juliusgeo:
> Can you provide an example of what the new error message output looks like?

Here is an example:
```
  File "/Users/julius/mongo-python-driver/pymongo/database.py", line 737, in command
    read_preference, session) as (sock_info, slave_ok):
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/Users/julius/mongo-python-driver/pymongo/mongo_client.py", line 1317, in _socket_for_reads
    server = self._select_server(read_preference, session)
  File "/Users/julius/mongo-python-driver/pymongo/mongo_client.py", line 1270, in _select_server
    server = topology.select_server(server_selector)
  File "/Users/julius/mongo-python-driver/pymongo/topology.py", line 242, in select_server
    address))
  File "/Users/julius/mongo-python-driver/pymongo/topology.py", line 200, in select_servers
    selector, server_timeout, address)
  File "/Users/julius/mongo-python-driver/pymongo/topology.py", line 216, in _select_servers_loop
    self._error_message(selector)+repr(self.description))
pymongo.errors.ServerSelectionTimeoutError: Timeout: 1.0s SSL handshake failed: localhost:27017: [Errno 0] Error<TopologyDescription id: 5ef3568c01dcd8c679625393, topology_type: Single, servers: [<ServerDescription ('localhost', 27017) server_type: Unknown, rtt: None, error=AutoReconnect('SSL handshake failed: localhost:27017: [Errno 0] Error')>]>
```

pymongo/topology.py
```diff
@@ -213,7 +213,7 @@ def _select_servers_loop(self, selector, timeout, address):
             # No suitable servers.
             if timeout == 0 or now > end_time:
                 raise ServerSelectionTimeoutError(
-                    self._error_message(selector))
+                    self._error_message(selector)+", Timeout: "+str(timeout)+"s, "+repr(self.description))
```
prashantmital, line 216: I think we should prefix the `self.description` repr with something like `Topology Description: ` so the overall error message has a structure like this: `ServerSelectionTimeoutError: <reason>, Timeout: <duration>s, Topology Description: <topology repr>`.

pymongo/topology.py
```diff
@@ -213,7 +213,7 @@ def _select_servers_loop(self, selector, timeout, address):
             # No suitable servers.
             if timeout == 0 or now > end_time:
                 raise ServerSelectionTimeoutError(
-                    self._error_message(selector))
+                    self._error_message(selector)+", Timeout: "+str(timeout)+"s, Topology Description: "+repr(self.description))
```
ShaneHarvey, line 216: A couple suggestions:

1. We try to keep our lines limited to 80 characters. Do you mind wrapping? (Eventually we'll start using black/pep8 for auto-formatting.)
2. Usually we use format strings instead of using `+`. Do you mind reformatting to `"%s, Timeout: ...." % (self._error_message(selector), timeout, self.description)`?

pymongo/topology.py
```diff
@@ -213,7 +213,9 @@ def _select_servers_loop(self, selector, timeout, address):
             # No suitable servers.
             if timeout == 0 or now > end_time:
                 raise ServerSelectionTimeoutError(
-                    self._error_message(selector))
+                    "%s, Timeout: %ss, Topology Description: %s" %
+                    (self._error_message(selector), str(timeout),
+                     repr(self.description)))
```
ShaneHarvey, line 218: nit: it would be cleaner to remove the str() and repr() calls and rely on the format string (`%s` for str() and `%r` for repr()):
```
                    "%s, Timeout: %ss, Topology Description: %r" %
                    (self._error_message(selector), timeout, self.description))
```

test/test_replica_set_client.py
```diff
@@ -343,10 +343,9 @@ def test_connect_with_internal_ips(self):
         with self.assertRaises(AutoReconnect) as context:
             connected(client)
 
-        self.assertEqual(
-            "Could not reach any servers in [('internal-ip', 27017)]."
-            " Replica set is configured with internal hostnames or IPs?",
-            str(context.exception))
+        self.assertTrue(
+            str(context.exception).startswith("Could not reach any servers in [('internal-ip', 27017)]."
+            " Replica set is configured with internal hostnames or IPs?"))
```
ShaneHarvey, line 348: Can you also wrap this line at 80 (like it was before)? Also it might be nicer to use self.assertIn (https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn) because it will give a better error message than assertTrue if/when this assertion fails: `self.assertIn("Could ...", str(context.exception))`

test/test_topology.py
```diff
@@ -766,7 +766,7 @@ def assertMessage(self, message, topology, selector=any_server_selector):
         with self.assertRaises(ConnectionFailure) as context:
             topology.select_server(selector, server_selection_timeout=0)
 
-        self.assertEqual(message, str(context.exception))
+        self.assertTrue(str(context.exception).startswith(message))
```
ShaneHarvey, line 769: Same comment about using assertIn instead of assertTrue.

PR #445, opened 2020-06-25
PYTHON-2143 Use an allow-list to determine resumable change stream errors

prashantmital:


pymongo/change_stream.py
```diff
@@ -283,12 +298,15 @@ def try_next(self):
         # one resume attempt.
         try:
             change = self._cursor._try_next(True)
-        except ConnectionFailure:
+        except (ConnectionFailure, CursorNotFound):
```
prashantmital, line 301: This addition makes it so that we resume on `CursorNotFound` exceptions. This is a deviation from the current spec, but I believe this to be the appropriate behavior. 
Note that this deviation from the spec allows a lot of PyMongo's tests (non-spec) to pass. Without this change, all tests that use `kill_cursors` to force a change stream resumption start failing.

test/test_change_stream.py
```diff
@@ -1086,14 +1050,66 @@ def setUp(self):
...
+            return
+
+        fail_cmd = SON([('configureFailPoint', 'failCommand')])
+        fail_cmd.update(fail_point)
+        client_context.client.admin.command(fail_cmd)
+        self.addCleanup(
+            client_context.client.admin.command,
+            'configureFailPoint', fail_cmd['configureFailPoint'], mode='off')
+
+    def assert_list_contents_are_subset(self, superlist, sublist):
```
prashantmital, line 1073: I added these methods to the class because now we can raise much better error messages when the equality/subset checks herein fail. Earlier, everything just raised `AssertionError` and print the same code where the error happened which wasn't helpful.

pymongo/change_stream.py
prashantmital: With https://jira.mongodb.org/browse/PYTHON-2302 having been filed, we can proceed with merging this as-is and come back to deal with `CursorNotFound` errors at a later time when the spec has been finalized.

ShaneHarvey: Nice find! For the record, I think we should keep resuming on CursorNotFound.

pymongo/errors.py
```diff
@@ -167,6 +170,15 @@ def details(self):
         """
         return self.__details
 
+    @property
+    def max_wire_version(self):
```
ShaneHarvey, line 174: Can we make this private+undocumented (`_ max_wire_version`)? Do you see any reason to make it public?

pymongo/helpers.py
```diff
@@ -103,14 +103,16 @@ def _index_document(index_list):
 
 
 def _check_command_response(response, msg=None, allowable_errors=None,
-                            parse_write_concern_error=False):
+                            parse_write_concern_error=False,
+                            max_wire_version=None):
```
ShaneHarvey, line 107: I suggest we make `max_wire_version` a required argument to ensure we don't forget to pass it.

test/test_change_stream.py
ShaneHarvey: Acknowledged.

test/test_change_stream.py
```diff
@@ -1202,17 +1193,17 @@ def run_scenario(self):
...
-                assert_dict_is_subset(change, expected_changes)
+                self.assert_dict_is_subset(change, expected_changes)
             self.assertEqual(len(changes), len(test["result"]["success"]))
         
         finally:
             # Check for expected events
             results = self.listener.results
-            for expectation in test.get("expectations", []):
-                for idx, (event_type, event_desc) in enumerate(iteritems(expectation)):
+            for idx, expectation in enumerate(test.get("expectations", [])):
```
ShaneHarvey, line 1202: Was the `idx` change a bugfix?

pymongo/change_stream.py
```diff
@@ -283,12 +298,15 @@ def try_next(self):
...
             self._resume()
             change = self._cursor._try_next(False)
         except OperationFailure as exc:
-            if (exc.code in _NON_RESUMABLE_GETMORE_ERRORS or
-                    exc.has_error_label("NonResumableChangeStreamError")):
+            is_resumable = ((exc.max_wire_version >= 9 and
+                             exc.has_error_label("ResumableChangeStreamError")) or
+                            (exc.max_wire_version < 9 and
+                             exc.code in _RESUMABLE_GETMORE_ERRORS))
+            if not is_resumable:
```
ShaneHarvey, line 309: Neat!

test/test_change_stream.py
prashantmital: Yes.

pymongo/change_stream.py
prashantmital: Agreed. Rachelle also agreed with this. We can break this in 4.0 but not before.

pymongo/helpers.py
prashantmital: Done. Note that I used None as the `max_wire_version` value in several tests that directly use this method.

pymongo/change_stream.py
```diff
@@ -283,12 +298,15 @@ def try_next(self):
...
         try:
             change = self._cursor._try_next(True)
-        except ConnectionFailure:
+        except (ConnectionFailure, CursorNotFound):
             self._resume()
             change = self._cursor._try_next(False)
         except OperationFailure as exc:
-            if (exc.code in _NON_RESUMABLE_GETMORE_ERRORS or
-                    exc.has_error_label("NonResumableChangeStreamError")):
+            is_resumable = ((exc._max_wire_version >= 9 and
```
ShaneHarvey, line 305: Is there any chance that `_max_wire_version` could be None here?

prashantmital: AFAICT this is not possible. The only place we seem to be raising `OperationFailure` in this code path is via _check_command_response which always populates `_max_wire_version`

pymongo/errors.py
```diff
@@ -136,18 +136,25 @@ class ConfigurationError(PyMongoError):
 class OperationFailure(PyMongoError):
     """Raised when a database operation fails.
 
+    .. versionadded:: 3.11
+       The :attr:`max_wire_version` attribute.
```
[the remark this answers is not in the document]
prashantmital: Done.

pymongo/change_stream.py
ShaneHarvey: I don't think this is necessarily the case. Auth failures might raise OperationFailure without  `_check_command_response` for example: https://github.com/mongodb/mongo-python-driver/blob/e608ff41662c694f7ba84e9d1a004463677886cf/pymongo/auth.py#L301

Can we guard against the None case?

pymongo/errors.py
ShaneHarvey: Can you make this private (again)? Somehow it got re-added.

pymongo/network.py
```diff
@@ -40,7 +40,7 @@
 _UNPACK_HEADER = struct.Struct("<iiii").unpack
 
 
-def command(sock_info, dbname, spec, slave_ok, is_mongos,
+def command(sock_info, max_wire_version, dbname, spec, slave_ok, is_mongos,
```
ShaneHarvey, line 43: Please remove max_wire_version and use sock_info.max_wire_version instead.

pymongo/pool.py
```diff
@@ -680,11 +680,11 @@ def command(self, dbname, spec, slave_ok=False,
         if self.op_msg_enabled:
             self._raise_if_not_writable(unacknowledged)
         try:
-            return command(self, dbname, spec, slave_ok,
-                           self.is_mongos, read_preference, codec_options,
-                           session, client, check, allowable_errors,
-                           self.address, check_keys, listeners,
-                           self.max_bson_size, read_concern,
```
ShaneHarvey, line 687: Can you undo these changes?

pymongo/errors.py
```diff
@@ -171,11 +182,6 @@ def details(self):
         """
         return self.__details
 
-    def __str__(self):
-        output_str = "%s, full error: %s" % (self._message, self.__details)
-        if sys.version_info[0] == 2 and isinstance(output_str, unicode):
-            return output_str.encode('utf-8', errors='replace')
-        return output_str
```
ShaneHarvey, line 178: Can you move `__str__` back to where it was so that it's not changed in this PR?

PR #446, opened 2020-06-26
PYTHON-2295 MongoClient with multiple hosts directConnection=True should raise an error

prashantmital:


pymongo/mongo_client.py
```diff
@@ -624,7 +624,7 @@ def __init__(
         username = None
         password = None
         dbase = None
-        opts = {}
+        opts = common._CaseInsensitiveDictionary()
```
ShaneHarvey, line 627: Is this needed? We already seem to use _CaseInsensitiveDictionary for `keyword_opts` down on line 663.

test/test_client.py
```diff
@@ -1566,12 +1566,21 @@ def test_direct_connection(self):
...
                        TOPOLOGY_TYPE.ReplicaSetWithPrimary])
         client.close()
 
+        # directConnection=True, should error with multiple hosts as a list.
+        if len(nodes) > 1:
+            hosts = []
+            for h, p in nodes:
+                hosts.append("%s:%s" % (h, p))
+            with self.assertRaises(ConfigurationError):
+                MongoClient(hosts, directConnection=True)
```
ShaneHarvey, line 1582: I think we can use multiple hosts without caring whether they exist or not (since we'll raise an error before attempting to connect). So we could reduce this to:
```
            with self.assertRaises(ConfigurationError):
                MongoClient(['host1', 'host2'], directConnection=True)
```

PR #447, opened 2020-06-26
PYTHON-2272 Explicitly specify UuidRepresentation in tests

prashantmital:


test/__init__.py
```diff
@@ -200,6 +201,9 @@ def ismaster(self):
         return self.client.admin.command('isMaster')
 
     def _connect(self, host, port, **kwargs):
+        # Use STANDARD representation unless one is specified.
+        if 'uuidRepresentation' not in kwargs:
+            kwargs['uuidRepresentation'] = "unspecified"
```
ShaneHarvey, line 206: You can add uuidRepresentation  to `self.default_client_options = {}` instead. PS I think we should use "standard" since "unspecified" may break a lot of tests.

ShaneHarvey:
Can we close this now?

PR #448, opened 2020-06-26
PYTHON-1787: add details to OperationFailure exception

juliusgeo:


pymongo/errors.py
```diff
@@ -146,9 +146,16 @@ def __init__(self, error, code=None, details=None):
             error_labels = details.get('errorLabels')
         super(OperationFailure, self).__init__(
             error, error_labels=error_labels)
+        self.__error = error
```
ShaneHarvey, line 149: What do you think about using  `self._message` from the super class (PyMongoError) instead of adding a new attribute?

pymongo/errors.py
```diff
@@ -146,9 +146,16 @@ def __init__(self, error, code=None, details=None):
...
         self.__code = code
         self.__details = details
 
+    @property
+    def error(self):
+        """The error string returned by the server, if any.
+        """
+        return self.__error
```
ShaneHarvey, line 157: Same question about `self._message`.

pymongo/errors.py
```diff
@@ -167,6 +174,8 @@ def details(self):
         """
         return self.__details
 
+    def __str__(self):
+        return "%s, full error: %s" % (self.__error, self.__details)
```
ShaneHarvey, line 178: I think we'll need some more work here to specially handle unicode errors on Python 2. Check out the `PyMongoError.__str__/__unicode__` methods for some inspiration. It might be easier to explain in person but [PYTHON-1966](https://jira.mongodb.org/browse/PYTHON-1966) has some more context. FYI addressing this issue will resolve the remaining `test_write_error_unicode` failure.

pymongo/errors.py
juliusgeo: Ah, yeah that would probably be a better method. Let me try that.

pymongo/errors.py
```diff
@@ -167,6 +167,11 @@ def details(self):
         """
         return self.__details
 
+    def __str__(self):
+        output_str = "%s, full error: %s" % (self._message, self.__details)
+        if isinstance(output_str, unicode):
```
ShaneHarvey, line 172: Since `unicode` does not exist in Python 3 we'll need to restrict this to Python 2 only (`sys.version_info[0] == 2`).

pymongo/errors.py
```diff
@@ -167,6 +167,11 @@ def details(self):
         """
         return self.__details
 
+    def __str__(self):
+        output_str = "%s, full error: %s" % (self._message, self.__details)
+        if sys.version_info[0] == 2 and isinstance(output_str, unicode):
+            return output_str.encode('utf-8', errors='replace')
+        return output_str
```
ShaneHarvey, line 174: Great! Can you add the same functionality to NotMasterError as well?

PR #450, opened 2020-06-30
PYTHON-1787: fix NotMasterError wrong attribute error

juliusgeo:
This is fixing the issue from the previous pull request which was caused by NotMasterError failing

https://jira.mongodb.org/browse/PYTHON-1787

### DIFF
--- a/pymongo/errors.py
+++ b/pymongo/errors.py
@@ -114,7 +114,7 @@ class NotMasterError(AutoReconnect):
     Subclass of :exc:`~pymongo.errors.AutoReconnect`.
     """
     def __str__(self):
-        output_str = "%s, full error: %s" % (self._message, self.__details)
+        output_str = "%s, full error: %s" % (self._message, self.details)
         if sys.version_info[0] == 2 and isinstance(output_str, unicode):
             return output_str.encode('utf-8', errors='replace')
         return output_str
--- a/pymongo/errors.py
+++ b/pymongo/errors.py
@@ -100,6 +100,14 @@ class NetworkTimeout(AutoReconnect):
     """
 
 
+def _format_detailed_error(message, details):
+    if details is not None:
+        message = "%s, full error: %s" % (message, details)
+        if sys.version_info[0] == 2 and isinstance(message, unicode):
+            message = message.encode('utf-8', errors='replace')
+    return message
+
+
 class NotMasterError(AutoReconnect):
     """The server responded "not master" or "node is recovering".
 
@@ -114,11 +122,8 @@ class NotMasterError(AutoReconnect):
     Subclass of :exc:`~pymongo.errors.AutoReconnect`.
     """
     def __init__(self, message='', errors=None):
-        if errors is not None:
-            message = "%s, full error: %s" % (message, errors)
-            if sys.version_info[0] == 2 and isinstance(message, unicode):
-                message = message.encode('utf-8', errors='replace')
-        super(NotMasterError, self).__init__(message, errors=errors)
+        super(NotMasterError, self).__init__(
+            _format_detailed_error(message, errors), errors=errors)
 
 
 class ServerSelectionTimeoutError(AutoReconnect):
@@ -150,11 +155,8 @@ class OperationFailure(PyMongoError):
         error_labels = None
         if details is not None:
             error_labels = details.get('errorLabels')
-            error = "%s, full error: %s" % (error, details)
-            if sys.version_info[0] == 2 and isinstance(error, unicode):
-                error = error.encode('utf-8', errors='replace')
         super(OperationFailure, self).__init__(
-            error, error_labels=error_labels)
+            _format_detailed_error(error, details), error_labels=error_labels)
         self.__code = code
         self.__details = details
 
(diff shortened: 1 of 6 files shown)


pymongo/errors.py
```diff
@@ -114,7 +114,7 @@ class NotMasterError(AutoReconnect):
     Subclass of :exc:`~pymongo.errors.AutoReconnect`.
     """
     def __str__(self):
```
ShaneHarvey, line 116: As discussed, can you add a test that asserts that the "full error" string is included in str(exc) and [traceback.format_exc()](https://docs.python.org/3/library/traceback.html#traceback.format_exc?

test/test_monitoring.py
```diff
@@ -439,6 +439,9 @@ def test_not_master_error(self):
             client.pymongo_test.test.find_one_and_delete({})
         except NotMasterError as exc:
             error = exc.errors
+            import traceback
+            self.assertIn("full error", str(exc))
+            self.assertIn("full error", traceback.format_exc())
```
ShaneHarvey, line 444: This is exactly what I had in mind. What do you think about moving these assertions to their own test cases (maybe to a new file `test_errors.py`)? You can use the error classes directly, like:
```
exc = NotMasterError(...)
self.assertIn("full error", str(exc))
try:
    raise exc
except NotMasterError:
    self.assertIn("full error", traceback.format_exc())
```

test/test_errors.py
```diff
@@ -0,0 +1,18 @@
...
+                  unittest)
+
+class TestErrors(PyMongoTestCase):
+    def test_not_master_error(self):
+        exc = NotMasterError("not master test", "details")
+        self.assertIn("full error", str(exc))
+        try:
+            raise exc
+        except NotMasterError:
+            import traceback
```
ShaneHarvey, line 17: Can you move this import to the top of the file?

test/test_errors.py
```diff
@@ -0,0 +1,18 @@
...
+                            OperationFailure)
+from test import (client_context,
+                  client_knobs,
+                  PyMongoTestCase,
+                  sanitize_cmd,
+                  unittest)
+
+class TestErrors(PyMongoTestCase):
+    def test_not_master_error(self):
+        exc = NotMasterError("not master test", "details")
```
ShaneHarvey, line 12: Suggest using something like `{ok: 0, errmsg:"error"}` instead of "details".

test/test_errors.py
```diff
@@ -0,0 +1,18 @@
...
+                            NotMasterError,
+                            OperationFailure)
+from test import (client_context,
+                  client_knobs,
+                  PyMongoTestCase,
+                  sanitize_cmd,
+                  unittest)
+
+class TestErrors(PyMongoTestCase):
+    def test_not_master_error(self):
```
ShaneHarvey, line 11: Can you also add a similar test for OperationFailure?

test/test_errors.py
```diff
@@ -0,0 +1,18 @@
+from pymongo.errors import (AutoReconnect,
```
ShaneHarvey, line 1: Add the copyright notice `# Copyright 2020-present MongoDB, Inc. ....`

ShaneHarvey, line 1: After importing standard python libraries but before importing pymongo can you add `sys.path[0:0] = [""]`. For example: https://github.com/mongodb/mongo-python-driver/blob/4760d0781505c0f13d5d1bc061ff89075927eb2f/test/test_bson.py#L27

test/test_errors.py
```diff
@@ -0,0 +1,18 @@
...
+
+class TestErrors(PyMongoTestCase):
+    def test_not_master_error(self):
+        exc = NotMasterError("not master test", "details")
+        self.assertIn("full error", str(exc))
+        try:
+            raise exc
+        except NotMasterError:
+            import traceback
+            self.assertIn("full error", traceback.format_exc())
```
ShaneHarvey, line 18: Can you add the `if __name__ == "__main__":` logic at the bottom of this file? https://github.com/mongodb/mongo-python-driver/blob/4760d0781505c0f13d5d1bc061ff89075927eb2f/test/test_bson.py#L1088-L1089

test/test_errors.py
```diff
@@ -0,0 +1,18 @@
+from pymongo.errors import (AutoReconnect,
+                            NotMasterError,
+                            OperationFailure)
+from test import (client_context,
+                  client_knobs,
+                  PyMongoTestCase,
+                  sanitize_cmd,
+                  unittest)
```
ShaneHarvey, line 8: Suggest removing any unused imports.

pymongo/errors.py
juliusgeo: Done.

test/test_errors.py
juliusgeo: Done.

test/test_errors.py
juliusgeo: Done.

test/test_errors.py
juliusgeo: Done.

PR #456, opened 2020-07-08
PYTHON-2305 Fix postBatchResumeToken caching logic

prashantmital:


prashantmital:
TL;DR: Our resume logic was broken in PyMongo. It is a small wonder we didn't catch this earlier!

The main issue here was two-fold:
1. We were not always setting the `resumeAfter` parameter in the aggregation query when resuming change streams. We should have been setting that value whenever there was a cached resume token and `_uses_start_after` was `False`. Instead, we were only setting it when `_uses_resume_after` was `True` which is not the same thing.
2. We were not caching the resume token appropriately in the situation when the `aggregate` command returned an empty batch containing a `postBatchResumeToken`. 

The issue has now been fixed and we now follow the change stream spec appropriately.

pymongo/change_stream.py
```diff
@@ -157,15 +157,19 @@ def _process_result(self, result, session, server, sock_info, slave_ok):
...
+                    self._uses_start_after is False and
                     sock_info.max_wire_version >= 7):
                 self._start_at_operation_time = result.get("operationTime")
                 # PYTHON-2181: informative error on missing operationTime.
                 if self._start_at_operation_time is None:
                     raise OperationFailure(
                         "Expected field 'operationTime' missing from command "
                         "response : %r" % (result, ))
+            elif 'postBatchResumeToken' in result['cursor']:
+                self._resume_token = result['cursor']['postBatchResumeToken']
```
ShaneHarvey, line 172: I suggest moving this case (`'postBatchResumeToken' in result['cursor']`) to be the first `if` so we don't need to repeat the `postBatchResumeToken` check in both.

prashantmital:
Addressed your comments @ShaneHarvey. Ready for another look.

ShaneHarvey:
Can you fix the 3.6 test failures?

prashantmital:
Done. @ShaneHarvey

PR #462, opened 2020-07-09
fix check_keys param - just ignore it

Bezbran:
Since BSON library is a general purpose library and so must follow BSON spec and the spec permits such keys, we cannot have this flag in BSON library.

Actually, everyone who need to store Kubernets YAML file in MongoDB will have problem. Once they have label / attribute like "app.kubernetes.io/managed-by: helm", they will encounter this issue as well.

If MongoDB guys has some needs to verify keys, they has to do it in pymongo code and not in BSON library.

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -1213,7 +1213,7 @@ class BSON(bytes):
         .. versionchanged:: 3.0
            Replaced `uuid_subtype` option with `codec_options`.
         """
-        return cls(encode(document, check_keys, codec_options))
+        return cls(encode(document, False, codec_options))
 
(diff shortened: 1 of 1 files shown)


behackett:
Hi. Thanks for your interest and contribution. The check_keys option (which defaults to False) exists because MongoDB doesn't currently support unambiguously querying on keys that begin with dollar signs or keys that contain dots. The most efficient way for PyMongo to filter out such keys is when iterating over the documents to encode. The option is necessary, and since it defaults to False your code change has no practical effect, so we're not going to take it.

PR #472, opened 2020-07-27
PYTHON-2334: Fix gevent race condition

TylerWilley:
Gevent patches the Lock class such that it can raise Timeout exceptions while waiting in .acquire()

When these raise, it can cause the pymongo thread_util.py:Semaphore to exit the acquire() function without releasing self._cond's lock. Changing to `with self._cond` guarantees release even if an exception is raised by gevent.

Also, in pool.py, when waiting for the lock for active sockets, this same condition could occur and cause self._socket_semaphore's lock to be lost.

### DIFF
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -1258,20 +1258,14 @@ class Pool:
                 True, self.opts.wait_queue_timeout):
             self._raise_wait_queue_timeout()
 
+        # We've now acquired the semaphore and must release it on error.
+        sock_info = None
+        incremented = False
         try:
             with self.lock:
                 self.active_sockets += 1
-        except Exception:
-            self._socket_semaphore.release()
+                incremented = True
 
-            if self.enabled_for_cmap:
-                self.opts.event_listeners.publish_connection_check_out_failed(
-                    self.address, ConnectionCheckOutFailedReason.CONN_ERROR)
-            raise
-
-        # We've now acquired the semaphore and must release it on error.
-        sock_info = None
-        try:
             while sock_info is None:
                 try:
                     with self.lock:
@@ -1288,8 +1282,10 @@ class Pool:
                 # We checked out a socket but authentication failed.
                 sock_info.close_socket(ConnectionClosedReason.ERROR)
             self._socket_semaphore.release()
-            with self.lock:
-                self.active_sockets -= 1
+
+            if incremented:
+                with self.lock:
+                    self.active_sockets -= 1
 
             if self.enabled_for_cmap:
                 self.opts.event_listeners.publish_connection_check_out_failed(
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -1257,8 +1257,17 @@ class Pool:
         if not self._socket_semaphore.acquire(
                 True, self.opts.wait_queue_timeout):
             self._raise_wait_queue_timeout()
-        with self.lock:
-            self.active_sockets += 1
+
+        try:
+            with self.lock:
+                self.active_sockets += 1
+        except Exception:
+            self._socket_semaphore.release()
+
+            if self.enabled_for_cmap:
+                self.opts.event_listeners.publish_connection_check_out_failed(
+                    self.address, ConnectionCheckOutFailedReason.CONN_ERROR)
+            raise
 
         # We've now acquired the semaphore and must release it on error.
         sock_info = None
--- a/pymongo/thread_util.py
+++ b/pymongo/thread_util.py
@@ -40,22 +40,21 @@ class Semaphore:
             raise ValueError("can't specify timeout for non-blocking acquire")
         rc = False
         endtime = None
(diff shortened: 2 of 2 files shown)


pymongo/thread_util.py
```diff
@@ -40,22 +40,21 @@ def acquire(self, blocking=True, timeout=None):
...
+                if not blocking:
+                    break
+                if timeout is not None:
+                    if endtime is None:
+                        endtime = _time() + timeout
+                    else:
+                        timeout = endtime - _time()
+                        if timeout <= 0:
+                            break
+                self._cond.wait(timeout)
```
TylerWilley, line 54: This will release, and re-lock self._cond

Condition has 2 acquires in it, If _waiter.acquire() raises, Condition will re-acquire in the finally clause.

If _acquire_restore raises, then Condition won't re-acquire and a double-release is possible. I can't think of a good way to solve that. One possible solution would be to update gevent to monkey patch Condition with a version that guarantees self._lock re-acquisition prior to raising. A double free here is somewhat less concerning though, as it would only cause one waiter to hang until the next notify, instead of leaking a lock.

Reference:
https://github.com/python/cpython/blob/v3.6.9/Lib/threading.py#L295-L304

pymongo/pool.py
```diff
@@ -1257,8 +1257,17 @@ def _get_socket(self, all_credentials):
         if not self._socket_semaphore.acquire(
                 True, self.opts.wait_queue_timeout):
             self._raise_wait_queue_timeout()
-        with self.lock:
-            self.active_sockets += 1
+
+        try:
+            with self.lock:
```
TylerWilley, line 1262: If gevent raises on the acquire here, self._socket_semaphore's lock will be permanently decremented.

ShaneHarvey:
Hi @TylerWilley, can you provide some sample code that reproduces this issue?

pymongo/pool.py
ShaneHarvey: I think it would be simpler to move the `active_sockets` modification down below inside the existing try/except:
```python
# We've now acquired the semaphore and must release it on error.
sock_info = None
try:
    with self.lock:
        self.active_sockets += 1
    while sock_info is None:...
```

TylerWilley:
```import time
from gevent import monkey, Greenlet, Timeout
monkey.patch_all()

import threading
from pymongo.thread_util import Semaphore

x = Semaphore()

def BrokenThread():
    global x
    with Timeout(1):
        try:
            x.acquire()
            print("Greenlet acquired lock")
        except Timeout:
            print("Greenlet timedout")

def ContentiousThread():
    x.acquire()
    time.sleep(2)
    x.release()

ct = Greenlet.spawn(ContentiousThread)
bt = Greenlet.spawn(BrokenThread)

ct.join()
bt.join()

print("Condition Lock State:", x._cond._lock.locked())```

Expected:
```Greenlet timedout
Condition Lock State: False```

Actual
```Greenlet timedout
  File "test_thread_utils.py", line 27, in <module>
    ct.join()
  File "/Users/twilley56/.pyenv/versions/3.6.9/lib/python3.6/site-packages/gevent/greenlet.py", line 812, in join
    result = get_my_hub(self).switch()
  File "/Users/twilley56/.pyenv/versions/3.6.9/lib/python3.6/site-packages/gevent/_greenlet_primitives.py", line 65, in switch
    return _greenlet_switch(self) # pylint:disable=undefined-variable
gevent.exceptions.LoopExit: This operation would block forever
	Hub: <Hub '' at 0x108a31d58 select default pending=0 ref=0 thread_ident=0x10a4555c0>
	Handles:
[]
```

pymongo/pool.py
TylerWilley: I thought of that, however the second try/except block will decrement the active sockets if an exception is raised.

If you raise within the first with self.lock, you will not have first incremented active_sockets, so you will end up permanently decreasing active_sockets.

ShaneHarvey: Good point. What about adding a flag:
```python
# We've now acquired the semaphore and must release it on error.
sock_info = None
incremented = False
try:
    with self.lock:
        self.active_sockets += 1
        incremented = True
    while sock_info is None:...
except Exception:
    if sock_info:
        # We checked out a socket but authentication failed.	                
        sock_info.close_socket(ConnectionClosedReason.ERROR)
    self._socket_semaphore.release()
    if incremented:
        with self.lock:
            self.active_sockets -= 1
```

TylerWilley: That would work... Should I add that?

TylerWilley: Updated, seems to still work

ShaneHarvey:
Thanks @TylerWilley! I've opened https://github.com/mongodb/mongo-python-driver/pull/475 to add a regression test for gevent.Timeout and fix Semaphore.release.

PR #473, opened 2020-07-27
PYTHON-1631 Automate release wheels for Windows and manylinux

ShaneHarvey:


.evergreen/config.yml
```diff
@@ -2519,6 +2555,13 @@ buildvariants:
     - name: "aws-auth-test-4.4"
     - name: "aws-auth-test-latest"
 
+- matrix_name: "Release"
+  matrix_spec:
+    platform: [ubuntu-16.04, windows-64-vsMulti-small]
```
behackett, line 2560: Why not Ubuntu 18? 16 goes EOL in April.

.evergreen/build-mac.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -ex
```
behackett, line 1: Why are you adding this now since we can't actually do it?

.evergreen/build-manylinux.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash -ex
+
+docker version
+
+# 2020-03-20-2fda31c Was the last release to include Python 3.4.
+images=(quay.io/pypa/manylinux1_x86_64:2020-03-20-2fda31c \
```
behackett, line 6: Do all all versions other than 3.4 get built twice?

.evergreen/build-mac.sh
ShaneHarvey: We can still use it to build macos wheels manually.

.evergreen/build-manylinux.sh
ShaneHarvey: Yes. They get built first with these old images and then rebuilt and overwritten with the latest manylinux1_x86_64/manylinux1_i686 images.

.evergreen/config.yml
ShaneHarvey: Changed to 18. They both have Docker so either one works.

PR #477, opened 2020-08-11
PYTHON-2332 Skip threaded SDAM tests when cdecimal is monkey patched

ShaneHarvey:
Add 60 second timeout for joining threads in SDAM tests.


test/test_discovery_and_monitoring.py
```diff
@@ -400,9 +410,11 @@ def wait_for_thread(self, name):
         """Run the 'waitForThread' operation."""
         thread = self.targets[name]
         thread.stop()
-        thread.join()
+        thread.join(60)
```
prashantmital, line 413: Was this timing out?

ShaneHarvey: Yes, see https://jira.mongodb.org/browse/PYTHON-2332.

PR #479, opened 2020-08-16
PYTHON-1932: replace_one does not validate keys for dots

shreybatra:


shreybatra:
@ShaneHarvey @prashantmital Please check this PR. I am unable to assign it to you guys and not able to tag you guys on the JIRA platform.

prashantmital:
Hi @shreybatra. First of all, thanks for your PR.

Unfortunately, we cannot currently merge this PR as this change would cause a backward-breaking change in driver behavior (consider, e.g. apps that currently use keys with dots - these will break after they update to PyMongo 3.12). Furthermore, we already have a mechanism for checking BSON keys for dots that we would use for this purpose (instead of the new validation function you have written in this PR) - see e.g. the `check_keys` kwarg of [BSON.encode](https://pymongo.readthedocs.io/en/stable/api/bson/index.html#bson.BSON.encode).

shreybatra:
Great thanks. Just for a head's up, for codebase of PyMongo v4.0, which branch should I see? The master is supposed to be for v3.12 I guess.

prashantmital:
Correct, master is currently tracking 3.12. We don't have a branch for 4.0. Once 3.12 is released, `master` will become the working branch for 4.0.

behackett:
We've disabled checking for dots and dollar signs across the board. The change will be released in PyMongo 3.12 and 4.0. See  https://github.com/mongodb/mongo-python-driver/commit/640fee9d5d7c937cb39dedd3218178c76b927631 and https://github.com/mongodb/mongo-python-driver/commit/ccb62d4c1bed0817f635bd2e0aaa822adf606027

PR #480, opened 2020-08-16
PYTHON-1915: Prohibit copying ClientSession objects

ishmum123:


pymongo/client_session.py
```diff
@@ -827,6 +827,9 @@ def _start_retryable_write(self):
         self._check_ended()
         self._server_session.inc_transaction_id()
 
+    def __copy__(self):
+        raise RuntimeError('A session cannot be copied (simply)!')
```
ShaneHarvey, line 831: I think [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) would be a better fit here:
> This exception may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be.

Also, can you change the message to:
```
'A ClientSession cannot be copied, create a new session instead'
```

ishmum123: Thanks for the feedback... Done!

prashantmital:
@ishmum123 instead of adding a mention in the commit message, perhaps you'd like to add your name to https://github.com/mongodb/mongo-python-driver/blob/master/doc/contributors.rst ? Just adding yourself to the tail end of that list will suffice. These names are visible in our documentation here: https://pymongo.readthedocs.io/en/stable/contributors.html

ishmum123:
Thank you very much, I would be delighted to...

PR #484, opened 2020-09-10
PYTHON-2362 Test with dnspython<2.0 to avoid timeouts

ShaneHarvey:
There seems to be a bug in dnspython 2.0 related to SRV lookups. For now we should downgrade to `dnspython<2.0` to get the test suite green.

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -201,6 +201,12 @@ if [ -n "$COVERAGE" -a $PYTHON_IMPL = "CPython" ]; then
     fi
 fi
 
+if $PYTHON -c 'import dns'; then
+  # Trying with/without --user to avoid:
+  # ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
+  $PYTHON -m pip install --upgrade --user 'dnspython<2.0.0' || $PYTHON -m pip install --upgrade 'dnspython<2.0.0'
+fi
+
 $PYTHON setup.py clean
 if [ -z "$GREEN_FRAMEWORK" ]; then
     if [ -z "$C_EXTENSIONS" -a $PYTHON_IMPL = "CPython" ]; then


.evergreen/run-tests.sh
```diff
@@ -201,6 +201,12 @@ if [ -n "$COVERAGE" -a $PYTHON_IMPL = "CPython" ]; then
     fi
 fi
 
+if $PYTHON -c 'import dns'; then
+  # Trying with/without --user to avoid:
+  # ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
+  $PYTHON -m pip install --upgrade --user 'dnspython<2.0.0' || $PYTHON -m pip install --upgrade 'dnspython<2.0.0'
```
prashantmital, line 207: Why not always install without `--user`?

prashantmital (commented):
Should we cut a 3.11.1 release that pins DNSPython to <2 so that new installs of PyMongo can still talk to Atlas? This seems like a massive issue...

behackett:
The problem is with Eventlet, right? Probably why no one has complained.

ShaneHarvey:
Yes this is an eventlet problem. dnspython 2.0 works fine up until eventlet is imported:
```python
>>> from dns import resolver
>>> resolver.query('_mongodb._tcp.test5.test.build.10gen.cc', 'SRV', lifetime=5)
<stdin>:1: DeprecationWarning: please use dns.resolver.resolve() instead
<dns.resolver.Answer object at 0x7fb575599940>
>>> import eventlet
>>> resolver.query('_mongodb._tcp.test5.test.build.10gen.cc', 'SRV', lifetime=5)
  File "/home/ubuntu/venv/lib/python3.8/site-packages/dns/resolver.py", line 1221, in query
    return resolve(qname, rdtype, rdclass, tcp, source,
  File "/home/ubuntu/venv/lib/python3.8/site-packages/dns/resolver.py", line 1205, in resolve
    return get_default_resolver().resolve(qname, rdtype, rdclass, tcp, source,
  File "/home/ubuntu/venv/lib/python3.8/site-packages/dns/resolver.py", line 1043, in resolve
    timeout = self._compute_timeout(start, lifetime)
  File "/home/ubuntu/venv/lib/python3.8/site-packages/dns/resolver.py", line 950, in _compute_timeout
    raise Timeout(timeout=duration)
dns.exception.Timeout: The DNS operation timed out after 5.10618257522583 seconds
```

This should be fixed by eventlet: https://github.com/eventlet/eventlet/issues/619.

ShaneHarvey:
> Should we cut a 3.11.1 release that pins DNSPython to <2 so that new installs of PyMongo can still talk to Atlas? This seems like a massive issue...

I don't think this is a massive issue and I'm not sure we should do anything else. eventlet>=0.26.1 was released which pins to `dnspython<2.0` and pymongo already specifies `dnspython<2.0` [here](https://github.com/mongodb/mongo-python-driver/blob/add995feb4793129b4971c4e29670f35a4f7619b/setup.py#L368):
```
    extras_require.update({'srv': ["dnspython>=1.16.0,<2.0.0"]})
```

.evergreen/run-tests.sh
ShaneHarvey: Because then we might need `sudo`. This will just be a temporary workaround until the toolchain is updated.

prashantmital (approved):

PR #492, opened 2020-09-28
PYTHON-2342 Prefer checking error codes over error messages

ShaneHarvey:
I've performed the following requests in the drivers ticket:

> - Audit the driver for uses of error messages to make decisions
> - For each error message, determine the server's related error code and check that first, falling back to error message matching to avoid breaking changes with old server versions.
>
> For example, the Python driver does an exact match on the error message "ns not found". It should attempt to match error code 26 first, matching the error message as a fallback. The fallback is necessary to avoid a backward breaking change with MongoDB 2.6 and 3.0 which do not return an error code.

I also removed the "db assertion failure" check which was removed by the server in MongoDB 2.5.

### DIFF
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -120,18 +120,8 @@ def _check_command_response(response, max_wire_version,
     if response["ok"]:
         return
 
-    details = response
-    # Mongos returns the error details in a 'raw' object
-    # for some errors.
-    if "raw" in response:
-        for shard in itervalues(response["raw"]):
-            # Grab the first non-empty raw error from a shard.
-            if shard.get("errmsg") and not shard.get("ok"):
-                details = shard
-                break
-
-    errmsg = details["errmsg"]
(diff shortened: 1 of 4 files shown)


pymongo/helpers.py
```diff
@@ -102,7 +102,7 @@ def _index_document(index_list):
     return index
 
 
-def _check_command_response(response, max_wire_version, msg=None,
```
ShaneHarvey, line 105: `msg` was never used.

ShaneHarvey (commented):
Enabling "Hide whitespace changes" (`w=1` in the url) makes this change easier to review: https://github.com/mongodb/mongo-python-driver/pull/492/files?diff=split&w=1

ShaneHarvey:
This is ready for review. I had to revert the PYTHON-2381 because of backwards compatibility concerns with 2.6/3.0.

prashantmital (approved):

PR #493, opened 2020-09-30
PYTHON-2360 Ensure ConnectionCreatedEvents are emitted before ConnectionReadyEvents

ShaneHarvey:
This uncovered a minor oversight in our minPoolSize connection creation logic. We did create the connection but we did not ensure that the connection was authenticated. Connections created in the background (for minPoolSize) are now authenticated. This is a minor performance improvement (lower latency on first connection use).

### DIFF
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -1150,7 +1150,6 @@ class Pool:
                 break
             try:
                 sock_info = self.connect(all_credentials)
-                sock_info.check_auth(all_credentials)
                 with self.lock:
                     # Close connection and return if the pool was reset during
                     # socket creation or while acquiring the pool lock.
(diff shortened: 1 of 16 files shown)


prashantmital (commented):
Can you explain how this results in improved performance? It seems to me that with this change `check_auth` will be called twice - once in the background on socket creation and then again when the socket is checked out (i.e. in `Pool._get_socket()`). I'm guessing that the second call would result in a no-op but it is not clear to me how?

ShaneHarvey:
> Can you explain how this results in improved performance?

Before this change the background thread which creates minPoolSize connections would only create the connection and run the handshake. This means that the first operation which using that connection needs to authenticate which incurs added latency.

After this change, the background thread creates the connection and authenticates it too.

prashantmital (approved):

PR #497, opened 2020-10-06
PYTHON-2390 Retryable read attempts should use the same implicit session

ShaneHarvey:
My testing approach uses the `assertSameLsidOnLastTwoCommands` spec test operation from the session spec test to assert that both attempts use the same implicit session. For now I added these tests to the `retryable_reads_custom` directory but I'm also planning to make push these changes upstream into the spec repo.

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -1440,8 +1440,8 @@ class MongoClient(common.BaseObject):
                     retrying = True
                 last_error = exc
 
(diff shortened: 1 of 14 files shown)


ShaneHarvey:
I'm going to close this PR and move this ticket into the backlog. The fix here introduces bugs for methods that return CommandCursors (aggregate_raw_batches, list_collections, etc.). A real fix needs to ensure that cursor methods save the implicit session for use later on in getMore commands.

PR #499, opened 2020-10-06
PYTHON-2308 Test that we exclusively depend on existence of logicalSessionsTimeoutMinutes for sessions support

ShaneHarvey:
This change resyncs the session spec tests and updates the prose tests to ensure we comply with the updated spec.


test/test_session.py
```diff
@@ -863,7 +866,6 @@ def _test_reads(self, op, exception=None):
                 'readConcern', {}).get('afterClusterTime')
             self.assertEqual(operation_time, act)
 
-    @client_context.require_no_standalone
```
ShaneHarvey, line 866: I made these changes because this line was removed from the prose tests:
> The test plan SHOULD be run against both replica set and sharded cluster	
topologies.  It MUST NOT be run against a standalone server.

 https://github.com/mongodb/specifications/commit/cdbc6ca01a00e7a0b9046e41ed29701271afd6f0#diff-042f65298ff835d62143081f7356a4ceL981

prashantmital (approved):

prashantmital (changes requested):
Errors in evergreen seem to be related to this change on 4.2 standalone?

test/test_session.py
ShaneHarvey: Turns out we already run the session prose tests (`TestSession`) against standalone clusters. I reverted the changes I made here because they were for the `TestCausalConsistency` class which implements the Causal Consistency spec prose tests. The [causal prose test plan](https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst#test-plan) still says to skip standalones so there's nothing to be done here.

PR #502, opened 2020-10-14
Added the mincePy ODM library to documented tools

muhrin:
MincePy is an ODM built on top of pymongo using the data-mapper pattern.

Be great if you guys would be willing to include this in the list as I think it offers something quite different from most (if not all) of the other ODMs in that it doesn't require that users subclass from a common model class, which means that any arbitrary Python object can be made storable.  This is invaluable in situations when you don't own (i.e. they're not in your codebase) the types you want to save.

### DIFF
--- a/doc/tools.rst
+++ b/doc/tools.rst
@@ -49,7 +49,7 @@ Humongolus
    examples are also available at GitHub.
    
 MincePy
-   `MincePy <https://https://mincepy.readthedocs.io/en/latest/>`_ is an 
+   `MincePy <https://mincepy.readthedocs.io/en/latest/>`_ is an 
    object-document mapper (ODM) designed to make any Python object storable 
    and queryable in a MongoDB database. It is designed with machine learning 
    and big-data computational and experimental science applications in mind 
--- a/doc/tools.rst
+++ b/doc/tools.rst
@@ -47,6 +47,15 @@ Humongolus
    possible. The code is available for download
    `at GitHub <https://github.com/entone/Humongolus>`_. Tutorials and usage
    examples are also available at GitHub.
+   
+MincePy
+   `MincePy <https://https://mincepy.readthedocs.io/en/latest/>`_ is an 
+   object-document mapper (ODM) designed to make any Python object storable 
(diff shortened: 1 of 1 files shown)


doc/tools.rst
```diff
@@ -47,6 +47,15 @@ Humongolus
    possible. The code is available for download
    `at GitHub <https://github.com/entone/Humongolus>`_. Tutorials and usage
    examples are also available at GitHub.
+   
+MincePy
+   `MincePy <https://https://mincepy.readthedocs.io/en/latest/>`_ is an
```
prashantmital, line 52: The `https://` is repeated so this link won't actually work.

prashantmital (changes requested):
Thank you for the PR! We'd be happy to include MincePy to this list. I wasn't aware of this tool and it looks pretty cool so thank you for bringing it to the attention of the PyMongo development team 😄 .
We can merge this once the issue with the link has been rectified.

doc/tools.rst
muhrin: Ah, thanks for spotting this!

muhrin:
Very welcome.  And thanks for your enthusiasm, I've certainly found mincePy to be very useful for my work so hopefully there are others out there who could benefit too :)

prashantmital (approved):

prashantmital:
@muhrin can you share what use-case this library has helped you out with? Is it a big-data/analytics workflow? I'd love to find out more about how you use MongoDB if you're available for a call sometime.

muhrin:
@prashantmital, fantastic, thanks a lot for merging this.

In short, I'm a physist at the Danish Technical University working on the discovery of new battery materials.  We run high-throughput quantum mechanical calculations, the results of which get used as training data for various machine learning models to accelerate the discovery process.  Through mincePy/pymongo MongoDB allows us to store all of this data, including the running workflows in once place where we can all collaborate together.  For me it's a total game changer over the old way of sending around tarballs or pointing people to shared paths.

Do get in touch if you'd like to hear more, I'm always open to a call: maruh AT dtu DOT dk

PR #503, opened 2020-10-14
Changing maxIdleTime to maxIdleTimeMS

nloadholtes:
While chasing down a configuration issue I noticed the docs refer to `maxIdleTime` but that doesn't seem to exist in the code, but `maxIdleTimeMS` does. This just updates the FAQ so no one else trips on this.


ShaneHarvey (approved):
Thanks!

PR #504, opened 2020-10-16
PYTHON-2344 Update TLS examples to use unified TLS URI options

ShaneHarvey:


doc/examples/tls.rst
```diff
@@ -147,28 +147,29 @@ Client certificates
...
+``tlsCertificateKeyFile`` option::
 
   >>> client = pymongo.MongoClient('example.com',
-  ...                              ssl=True,
-  ...                              ssl_certfile='/path/to/client.pem')
+  ...                              tls=True,
+  ...                              tlsCertificateKeyFile='/path/to/client.pem')
 
 If the private key for the client certificate is stored in a separate file use
-the `ssl_keyfile` option::
```
ShaneHarvey, line 157: I've kept in `ssl_keyfile` because we currently still support it. https://jira.mongodb.org/browse/PYTHON-2396 will figure out if we want to remove it in PyMongo 4.0.

prashantmital (approved):
This LGTM but I couldn't find a ticket where we would rename the internal names for these options to the `TLS*`-equivalents. Currently, the `ssl*` names are used everywhere internally in the PyMongo codebase. Do you think we should have a ticket for changing that slated for PyMongo 4.0?

ShaneHarvey:
> Do you think we should have a ticket for changing that slated for PyMongo 4.0?

Sure. We could even do it sooner since it's all internal only names.

PR #520, opened 2020-11-20
PYTHON-2433 Fix Python 3 ServerDescription/Exception memory leak

ShaneHarvey:
When the SDAM monitor check fails, a ServerDescription is created from
the exception. This exception is kept alive via the
ServerDescription.error field. Unfortunately, the exception's traceback
contains a reference to the previous ServerDescription. Altogether this
means that each consecutively failing check leaks memory by building an
ever growing chain of ServerDescription -> Exception -> Traceback ->
Frame -> ServerDescription -> ... objects.

This change breaks the chain and prevents the memory leak by clearing
the Exception's __traceback__, __context__, and __cause__ fields.

### DIFF
--- a/pymongo/server_description.py
+++ b/pymongo/server_description.py
@@ -15,7 +15,6 @@
 """Represent one server the driver is connected to."""
 
 from bson import EPOCH_NAIVE
-from bson.py3compat import PY3
 from pymongo.server_type import SERVER_TYPE
 from pymongo.ismaster import IsMaster
 from pymongo.monotonic import time as _time
@@ -70,11 +69,6 @@ class ServerDescription(object):
         self._me = ismaster.me
         self._last_update_time = _time()
         self._error = error
-        # PYTHON-2433 Clear error traceback info.
-        if error and PY3:
-            error.__traceback__ = None
-            error.__context__ = None
-            error.__cause__ = None
         self._topology_version = ismaster.topology_version
         if error:
             if hasattr(error, 'details') and isinstance(error.details, dict):
--- a/pymongo/server_description.py
+++ b/pymongo/server_description.py
@@ -15,6 +15,7 @@
 """Represent one server the driver is connected to."""
 
 from bson import EPOCH_NAIVE
+from bson.py3compat import PY3
 from pymongo.server_type import SERVER_TYPE
 from pymongo.ismaster import IsMaster
 from pymongo.monotonic import time as _time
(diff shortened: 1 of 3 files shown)


ShaneHarvey:
Rebased to fix test failures caused by PYTHON-2436.

test/test_client.py
```diff
@@ -1614,6 +1615,33 @@ def test_direct_connection(self):
...
+            initial_count = server_description_count()
+            self.addCleanup(client.close)
+            with self.assertRaises(ServerSelectionTimeoutError):
+                client.test.test.find_one()
+            gc.collect()
+            final_count = server_description_count()
+            # If a bug like PYTHON-2433 is reintroduced then too many
+            # ServerDescriptions will be kept alive and this test will fail:
+            # AssertionError: 4 != 22 within 5 delta (18 difference)
+            self.assertAlmostEqual(initial_count, final_count, delta=5)
```
prashantmital, line 1643: Nice test. What is behind the choice of using `delta=5`? Is that number `5` liable to change?

pymongo/server_description.py
```diff
@@ -69,6 +70,11 @@ def __init__(
         self._me = ismaster.me
         self._last_update_time = _time()
         self._error = error
+        # PYTHON-2433 Clear error traceback info.
```
prashantmital, line 73: Any particular reason you are purging the traceback, etc. in this location as opposed to after emitting the heartbeat failed event in `Monitor._check_server`? Why preclude any exception on a ServerDescription from having a traceback?

test/test_client.py
ShaneHarvey: 5 is arbitrary but it allows for some tolerance for race conditions, like having several clients open in the background (creating and destroying ServerDescriptions).

pymongo/server_description.py
ShaneHarvey: Yeah it is a bit odd to mutate the exception in the ServerDescription constructor. I moved this logic to the Monitor.

prashantmital (approved):
Awesome job identifying this so quickly. It would have been a real doozy to figure out independently!

ShaneHarvey:
Alternative fixes would be:
- refactor Monitor's code such that exceptions won't hold references to previous ServerDescriptions. I rejected this approach because it seems too brittle.
- Use [traceback.clear_frames()](https://docs.python.org/3/library/traceback.html#traceback.clear_frames) to clear the references in the traceback. For some reason this doesn't fix the bug.
- Change ServerDescription.error to be a string representation of the error (`repr(error)`). This is probably the most robust approach but it's a backwards breaking change and it breaks a few tests.

PR #514, opened 2020-11-06
PYTHON-2415 Support pickle in BulkWriteError

ShaneHarvey:
For some background see my comment on the Jira ticket here: https://jira.mongodb.org/browse/PYTHON-2415?focusedCommentId=3476301&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3476301

PR #522, opened 2020-11-22
PYTHON-2438 Fix str representation of BulkWriteError

pcorpet:


pymongo/errors.py
```diff
@@ -243,6 +243,10 @@ def __init__(self, results):
         # For pickle support
         self.args = (results,)
 
+    if sys.version_info[0] != 2:
+        def __str__(self):
+            return str(self._message)
+
```
ShaneHarvey, line 249: The fact that we have to add `__str__` to fix this bug makes me a little uncomfortable. It actually makes me think that using `self.args = (results,)` to "fix" the original pickle issue (in https://github.com/mongodb/mongo-python-driver/pull/514) was the wrong choice.

Looking again, I think the best option may be to use `__reduce__` like this:

```suggestion

    def __reduce__(self):
        return self.__class__, (self.details,)

```

Could you make the above change?

ShaneHarvey (changes requested):

pymongo/errors.py
pcorpet: Way better. I took your recommendation.

PR #524, opened 2020-11-24
PYTHON-2457 Test that clients wait 500ms between failed heartbeat checks

ShaneHarvey:
Implements: https://github.com/mongodb/specifications/pull/879/

### DIFF
--- a/test/test_streaming_protocol.py
+++ b/test/test_streaming_protocol.py
@@ -130,54 +130,28 @@ class TestStreamingProtocol(IntegrationTest):
         self.assertEqual(1, len(events))
         self.assertGreater(events[0].new_description.round_trip_time, 0)
 
+    @client_context.require_version_min(4, 9, -1)
     @client_context.require_failCommand_appName
     def test_monitor_waits_after_server_check_error(self):
-        hb_listener = HeartbeatEventListener()
-        client = rs_or_single_client(
-            event_listeners=[hb_listener], heartbeatFrequencyMS=500,
-            appName='waitAfterErrorTest')
-        self.addCleanup(client.close)
-        # Force a connection.
-        client.admin.command('ping')
-        address = client.address
-
         fail_ismaster = {
-            'mode': {'times': 50},
+            'mode': {'times': 5},
             'data': {
                 'failCommands': ['isMaster'],
-                'closeConnection': False,
-                'errorCode': 91,
-                # This can be uncommented after SERVER-49220 is fixed.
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Patch: https://evergreen.mongodb.com/version/5fbc764b0ae606058eff4c97

ShaneHarvey:
This is ready for review although I still need to create a PYTHON ticket to track the work.

patrickfreed (approved):

test/test_streaming_protocol.py
```diff
@@ -130,54 +130,28 @@ def changed_event(event):
...
-        self.assertLess(len(hb_started_events), 40)
+            start = time.time()
+            client = single_client(
+                appName='SDAMMinHeartbeatFrequencyTest',
+                serverSelectionTimeoutMS=5000)
+            self.addCleanup(client.close)
+            # Force a connection.
+            client.admin.command('ping')
+            duration = time.time() - start
+            self.assertGreaterEqual(duration, 2)
```
prashantmital, line 153: Can you add comments here that explain the math? E.g.
```
# isMaster attempted 5 times with 500 ms between checks => duration > 4 * 500ms 
self.assertGreaterEqual(duration, 2)
```

test/test_streaming_protocol.py
```diff
@@ -130,54 +130,28 @@ def changed_event(event):
...
+            self.assertLessEqual(duration, 3.5)
```
prashantmital, line 154: How do you arrive at 3.5s?

prashantmital (changes requested):

test/test_streaming_protocol.py
ShaneHarvey: Added a link to the spec and an explanation like the previous version of the test.

test/test_streaming_protocol.py
ShaneHarvey: That's what the spec says to use: https://github.com/mongodb/specifications/blob/6c5b2ac/source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst#monitors-sleep-at-least-minheartbeatfreqencyms-between-checks

patrickfreed: I added it as a generous upper bound to account for any timing differences in the various test runner implementatinos. Drivers should really take between 2 and 2.5 seconds (depending on when the timer was started and how fast the monitors start up).

prashantmital (approved):

PR #530, opened 2020-12-07
PYTHON-2452 Ensure command-responses with RetryableWriteError label are retried on MongoDB 4.4+

prashantmital:
This change will be backported to 3.11 and 3.12 branches

### DIFF
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -115,7 +115,10 @@ def _check_command_response(response, max_wire_version,
                                max_wire_version)
 
     if parse_write_concern_error and 'writeConcernError' in response:
-        _raise_write_concern_error(response['writeConcernError'])
+        _error = response["writeConcernError"]
+        _labels = response.get("errorLabels", [])
+        _error.update({'errorLabels': _labels})
+        _raise_write_concern_error(_error)
 
     if response["ok"]:
         return
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -117,7 +117,8 @@ def _check_command_response(response, max_wire_version,
     if parse_write_concern_error and 'writeConcernError' in response:
         _error = response["writeConcernError"]
         _labels = response.get("errorLabels", [])
-        _error.update({'errorLabels': _labels})
+        if _labels:
+            _error.update({'errorLabels': _labels})
         _raise_write_concern_error(_error)
 
     if response["ok"]:
@@ -225,9 +226,10 @@ def _check_write_command_response(result):
         _raise_last_write_error(write_errors)
 
     error = result.get("writeConcernError", {})
-    error_labels = result.get("errorLabels", [])
-    error.update({'errorLabels': error_labels})
     if error:
+        error_labels = result.get("errorLabels", [])
+        if error_labels:
+            error.update({'errorLabels': error_labels})
         _raise_write_concern_error(error)
 
 
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -116,7 +116,7 @@ def _check_command_response(response, max_wire_version,
 
     if parse_write_concern_error and 'writeConcernError' in response:
         _error = response["writeConcernError"]
-        _labels = response.get("errorLabels", [])
+        _labels = response.get("errorLabels")
         if _labels:
             _error.update({'errorLabels': _labels})
         _raise_write_concern_error(_error)
@@ -225,9 +225,9 @@ def _check_write_command_response(result):
     if write_errors:
         _raise_last_write_error(write_errors)
 
-    error = result.get("writeConcernError", {})
+    error = result.get("writeConcernError")
     if error:
-        error_labels = result.get("errorLabels", [])
+        error_labels = result.get("errorLabels")
         if error_labels:
             error.update({'errorLabels': error_labels})
         _raise_write_concern_error(error)
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -221,7 +221,9 @@ def _check_write_command_response(result):
     if write_errors:
         _raise_last_write_error(write_errors)
 
-    error = result.get("writeConcernError")
+    error = result.get("writeConcernError", {})
+    error_labels = result.get("errorLabels", [])
+    error.update({'errorLabels': error_labels})
     if error:
         _raise_write_concern_error(error)
 
--- a/test/test_retryable_writes.py
+++ b/test/test_retryable_writes.py
@@ -44,6 +44,7 @@ from test.utils import (rs_or_single_client,
                         OvertCommandListener,
                         TestCreator)
 from test.utils_spec_runner import SpecRunner
+from test.version import Version
 
 # Location of JSON test specifications.
 _TEST_PATH = os.path.join(
@@ -454,9 +455,13 @@ class TestRetryableWrites(IgnoreDeprecationsTest):
             self.assertEqual(final_txn, expected_txn)
         self.assertEqual(coll.find_one(projection={'_id': True}), {'_id': 1})
 
-    @client_context.require_version_min(4, 4)
+
+class TestWriteConcernError(IntegrationTest):
+    @client_context.require_version_min(3, 6)
+    @client_context.require_no_standalone
+    @client_context.require_no_mmap
     @client_context.require_failCommand_fail_point
-    def test_retryable_write_error_label_is_propagated(self):
+    def test_RetryableWriteError_error_label(self):
         listener = OvertCommandListener()
         client = rs_or_single_client(
             retryWrites=True, event_listeners=[listener])
@@ -478,9 +483,11 @@ class TestRetryableWrites(IgnoreDeprecationsTest):
             self.assertIn('RetryableWriteError',
                           cm.exception._error_labels)
 
-        self.assertIn(
-            'RetryableWriteError',
-            listener.results['succeeded'][-1].reply['errorLabels'])
+        if client_context.version >= Version(4, 4):
+            # In MongoDB 4.4+ we rely on the server returning the error label.
+            self.assertIn(
+                'RetryableWriteError',
+                listener.results['succeeded'][-1].reply['errorLabels'])
 
 
 # TODO: Make this a real integration test where we stepdown the primary.
--- a/test/test_retryable_writes.py
+++ b/test/test_retryable_writes.py
@@ -457,7 +457,7 @@ class TestRetryableWrites(IgnoreDeprecationsTest):
 
 
 class TestWriteConcernError(IntegrationTest):
-    @client_context.require_version_min(3, 6)
+    @client_context.require_version_min(4, 0)
     @client_context.require_no_standalone
     @client_context.require_no_mmap
     @client_context.require_failCommand_fail_point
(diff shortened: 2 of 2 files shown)


test/test_retryable_writes.py
```diff
@@ -453,6 +454,27 @@ def test_batch_splitting_retry_fails(self):
...
+            'configureFailPoint': 'failCommand',
+            'mode': {'times': 2},
+            'data': {
+                'failCommands': ['insert'],
+                'writeConcernError': {
+                    'code': 91,
+                    'errmsg': 'Replication is being shut down'}}}
+
+        with self.fail_point(fail_insert):
+            with self.assertRaises(WriteConcernError) as exc:
```
ShaneHarvey, line 473: nit: `exc` should be `cm` since it's not an exception but a context manager. `cm` is what the python doc examples use: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

test/test_retryable_writes.py
```diff
@@ -453,6 +454,27 @@ def test_batch_splitting_retry_fails(self):
             self.assertEqual(final_txn, expected_txn)
         self.assertEqual(coll.find_one(projection={'_id': True}), {'_id': 1})
 
+    @client_context.require_version_min(4, 4)
```
ShaneHarvey, line 457: This also needs `@require_failCommand_fail_point`

test/test_retryable_writes.py
```diff
@@ -453,6 +454,27 @@ def test_batch_splitting_retry_fails(self):
...
+    def test_retryable_write_error_label_is_propagated(self):
+        listener = OvertCommandListener()
+        client = rs_or_single_client(
+            retryWrites=True, event_listeners=[listener])
```
ShaneHarvey, line 461: Can we require the client_context client instead of creating a new one?

pymongo/helpers.py
```diff
@@ -221,7 +221,9 @@ def _check_write_command_response(result):
     if write_errors:
         _raise_last_write_error(write_errors)
 
-    error = result.get("writeConcernError")
+    error = result.get("writeConcernError", {})
+    error_labels = result.get("errorLabels", [])
+    error.update({'errorLabels': error_labels})
```
ShaneHarvey, line 226: We should ensure that `_check_command_response` above has the same behavior. Also is it ever possible for `error` to be  RawBSONDocument (where `.update` would fail)?

ShaneHarvey (changes requested):

pymongo/helpers.py
prashantmital: Added the behavior to `_check_command_response`. The question about `RawBSONDocument` needs further investigation.

test/test_retryable_writes.py
prashantmital: Done.

test/test_retryable_writes.py
prashantmital: How do I add an event listener to the existing client?

test/test_retryable_writes.py
prashantmital: Changed to `cm`.

pymongo/helpers.py
```diff
@@ -115,7 +115,10 @@ def _check_command_response(response, max_wire_version,
                                max_wire_version)
 
     if parse_write_concern_error and 'writeConcernError' in response:
-        _raise_write_concern_error(response['writeConcernError'])
+        _error = response["writeConcernError"]
+        _labels = response.get("errorLabels", [])
+        _error.update({'errorLabels': _labels})
```
ShaneHarvey, line 120: Should be:
```
        if _labels:
            _error.update({'errorLabels': _labels})
```

So that we don't add an empty errorLabels to the error when the response doesn't have errorLabels.

pymongo/helpers.py
ShaneHarvey: This logic is flawed because it ends up always making `error` true. It should be:
```python
    error = result.get("writeConcernError", {})
    if error:
        error_labels = result.get("errorLabels", [])
        if error_labels:
            error.update({'errorLabels': error_labels})
        _raise_write_concern_error(error)
```

test/test_retryable_writes.py
ShaneHarvey: Why require 4.4? Shouldn't this test also work the same on 4.2?

test/test_retryable_writes.py
```diff
@@ -453,6 +454,28 @@ def test_batch_splitting_retry_fails(self):
...
+            'mode': {'times': 2},
+            'data': {
+                'failCommands': ['insert'],
+                'writeConcernError': {
+                    'code': 91,
+                    'errmsg': 'Replication is being shut down'}}}
+
+        with self.fail_point(fail_insert):
+            with self.assertRaises(WriteConcernError) as cm:
+                client.testdb.testcoll.insert_one({})
```
ShaneHarvey, line 475: I suggest using the standard `pymongo_test`  database name here.

test/test_retryable_writes.py
prashantmital: This section https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#determining-retryable-errors seems to suggest that only MongoDB 4.4+ will add the `RetryableWritesError` to its command response. So the command assertion wouldn't work with MongoDB 4.2

pymongo/helpers.py
prashantmital: Good catch. Thanks.

test/test_retryable_writes.py
prashantmital: Done.

test/test_retryable_writes.py
ShaneHarvey: That's correct but if I understand correctly, the spec says that on <4.4 the driver adds the RetryableWritesError label to retryable errors. So shouldn't the label be there either way?

test/test_retryable_writes.py
```diff
@@ -453,6 +454,34 @@ def test_batch_splitting_retry_fails(self):
...
+                'writeConcernError': {
+                    'code': 91,
+                    'errmsg': 'Replication is being shut down'},
+            }}
+
+        with self.fail_point(fail_insert):
+            with self.assertRaises(WriteConcernError) as cm:
+                client.pymongo_test.testcoll.insert_one({})
+            self.assertIn('RetryableWriteError',
+                          cm.exception._error_labels)
```
ShaneHarvey, line 479: This should use the has_error_label api.

test/test_retryable_writes.py
ShaneHarvey: Please update this to `client_context.require_version_min(3, 6)` to make sure we add the error label correctly on <4.4. We can then make the final assertion server version specific:
```
if version >= 4.4:
        self.assertIn(
            'RetryableWriteError',
            listener.results['succeeded'][-1].reply['errorLabels'])
```

ShaneHarvey (changes requested):

test/test_retryable_writes.py
prashantmital: Done.

test/test_retryable_writes.py
```diff
@@ -454,6 +456,40 @@ def test_batch_splitting_retry_fails(self):
         self.assertEqual(coll.find_one(projection={'_id': True}), {'_id': 1})
 
 
+class TestWriteConcernError(IntegrationTest):
+    @client_context.require_version_min(3, 6)
```
ShaneHarvey, line 460: This needs 4.0 because the writeConcernError argument to failCommand was added in 4.0: https://jira.mongodb.org/browse/SERVER-35083

test/test_retryable_writes.py
```diff
@@ -454,6 +456,40 @@ def test_batch_splitting_retry_fails(self):
...
+    @client_context.require_no_standalone
+    @client_context.require_no_mmap
+    @client_context.require_failCommand_fail_point
+    def test_RetryableWriteError_error_label(self):
+        listener = OvertCommandListener()
+        client = rs_or_single_client(
+            retryWrites=True, event_listeners=[listener])
+
+        fail_insert = {
```
ShaneHarvey, line 469: Can perform a successful insert before testing the failpoint behavior to ensure the collection is created first:
```
client.pymongo_test.testcoll.insert_one({})
```
AFAIK this shouldn't matter but its best practice to create the collection first.

prashantmital: Done.

test/test_retryable_writes.py
prashantmital: Done.

prashantmital:
Updated the test to account for https://jira.mongodb.org/browse/SERVER-39292 in MongoDB 4.0.x

pymongo/helpers.py
```diff
@@ -221,8 +225,11 @@ def _check_write_command_response(result):
     if write_errors:
         _raise_last_write_error(write_errors)
 
-    error = result.get("writeConcernError")
+    error = result.get("writeConcernError", {})
```
ShaneHarvey, line 228: Minor nit: we can remove the `, {}` here because None is okay.

ShaneHarvey (commented):
Did you get clarification from the server/retryable writes team on whether it's possible for errorLabels to appear in two places?

test/test_retryable_writes.py
```diff
@@ -454,6 +456,48 @@ def test_batch_splitting_retry_fails(self):
...
+                'closeConnection': False,
+                'writeConcernError': {
+                    'code': 91,
+                    'errmsg': 'Replication is being shut down'},
+            }}
+
+        if client_context.version < Version(4, 2):
+            # SERVER-39292: specifying closeConnection on MongoDB 4.0+,<4.2
+            # causes the failPoint to fire twice so we remove it.
+            fail_insert['data'].pop('closeConnection')
```
ShaneHarvey, line 483: Can we just skip adding `closeConnection` for all server versions?

pymongo/helpers.py
prashantmital: Done. Changed this pattern elsewhere as well.

test/test_retryable_writes.py
prashantmital: Done.

pymongo/helpers.py
ShaneHarvey: This is not resolved because we still need to answer the RawBSONDocument question.

ShaneHarvey:
Still waiting on the two questions on expected server behavior and RawBSONDocument.

prashantmital:
`RawBSONDocument` will not break `db.command` because we set `check_command_response` to `False` 
 [here](https://github.com/mongodb/mongo-python-driver/blob/61232b7f21d8f0ad9023c8ef0912017d062fafc0/pymongo/database.py#L738) which gets passed down [here](https://github.com/mongodb/mongo-python-driver/blob/61232b7f21d8f0ad9023c8ef0912017d062fafc0/pymongo/helpers.py#L105) and ensures that we never check for the `writeConcernError` document within responses to `db.command`.

Also, I confirmed that the server only returns `errorLabels` as a top-level field. See the HELP ticket linked to the JIRA ticket for details.

ShaneHarvey:
Hmm, thanks for doing that investigation.

> I also quickly tested this by setting a failpoint that returned writeConcernError on insert and then triggering it via db.command.

Can you add this as a test so that we don't regress on this behavior?

prashantmital:
Added a test.

ShaneHarvey (approved):
LGTM

PR #533, opened 2020-12-14
PYTHON-2466 Make pymongo client, database and collection objects hashable.

pcorpet:
…able.

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -303,6 +303,9 @@ class Collection(common.BaseObject):
     def __ne__(self, other):
         return not self == other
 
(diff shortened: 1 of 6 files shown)


prashantmital (approved):
I think in the future we might want to consider adding some configuration information to the `__eq__` and `__hash__` methods for these classes - the present implementation will result in two DBs with different readPreferences, for example, to hash to the same value. 
For the time being, this implementation LGTM since `__hash__` must be the same for values that compare equal using `__eq__`.

prashantmital:
Thank you for your contribution @pcorpet !

PR #534, opened 2020-12-15
PYTHON-2366 Test OCSP+FLE with Python 3.9

ShaneHarvey:
This change does the following:
- PYTHON-2366  Adds Python 3.9 testing for OCSP+FLE.
- PYTHON-2449 Fixes PyPy testing by migrating from RHEL 6.2/Ubuntu 16 to Debian 9.2. The former have OpenSSL 1.0.1/1.0.2 while debian has 1.1.0. This works around the pyopenssl connect errors described in https://github.com/pyca/pyopenssl/issues/981 .
- PYTHON-2449 Fixes Windows test setup failures caused by outdated pip versions being unable to install cryptography wheels. The fix is to upgrade pip.

I'm planning to backport these changes which is why I had to change PyPy 2 testing. 

Tested here: https://evergreen.mongodb.com/version/5fd929ef57e85a0b36c0ec64

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1561,6 +1561,7 @@ axes:
         run_on: debian92-test
         batchtime: 10080  # 7 days
         variables:
+          python3_binary: "/opt/python/3.8/bin/python3"
           libmongocrypt_url: https://s3.amazonaws.com/mciuploads/libmongocrypt/debian92/master/latest/libmongocrypt.tar.gz
       - id: macos-1014
         display_name: "macOS 10.14"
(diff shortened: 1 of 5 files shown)


prashantmital (approved):

.evergreen/config.yml
```diff
@@ -2177,6 +2178,19 @@ buildvariants:
...
+  matrix_spec:
+    platform: debian92
+    python-version: ["pypy", "pypy3.5", "pypy3.6"]
+    auth: "auth"
+    ssl: "ssl"
+    pyopenssl: "*"
+  display_name: "PyOpenSSL ${platform} ${python-version} ${auth}"
+  tasks:
+    - '.replica_set !.2.6 !.3.0 !.3.2 !.3.4'
+    # Test standalone and sharded only on 4.4.
```
prashantmital, line 2191: Wouldn't this also test RS?

prashantmital: Can you also explain why we want to only run this on RS for 3.6, 4.0, 4.2 ?

ShaneHarvey: This config tests:
- 3.6+ with replica sets
- 4.4 with standalone, repl, sharded.

We don't test standalone+sharded on other server versions because it would be overkill.

PR #537, opened 2020-12-22
PYTHON-2474 Fix non-disabled client_knobs bug in Data Lake tests

ShaneHarvey:


test/test_data_lake.py
```diff
@@ -35,11 +35,10 @@ class DataLakeTestSpec(TestCrudV2):
     TEST_COLLECTION = 'driverdata'
 
     @classmethod
+    @unittest.skipUnless(client_context.is_data_lake,
+                         'Not connected to Atlas Data Lake')
     def setUpClass(cls):
         super(DataLakeTestSpec, cls).setUpClass()
-        # Skip these tests unless connected to data lake.
-        if not client_context.is_data_lake:
-            raise unittest.SkipTest('Not connected to Atlas Data Lake')
```
ShaneHarvey, line 42: Explanation:

1. The super class' setUpClass method enabled client_knobs.
2. Here we raise SkipTest
3. tearDownClass is never called because setUpClass raised an error.
4. The side-effects of client_knobs are still present.

The change here fixes the bug by raising SkipTest _before_ calling setUpClass.

prashantmital (approved):
Nice work diagnosing this!

PR #538, opened 2021-01-05
PYTHON-1878 Add mongodb+srv URIs to Atlas Connectivity tests

ShaneHarvey:


.evergreen/config.yml
```diff
@@ -444,8 +444,29 @@ functions:
...
+      type: test
+      params:
+        working_dir: "src"
+        script: |
+          # Disable xtrace (just in case it was accidentally set).
+          set +x
+          . ./prepare_atlas_connectivity.sh
+          rm -f ./prepare_atlas_connectivity.sh
+
+          PYTHON_BINARY=${PYTHON_BINARY} bash ${PROJECT_DIRECTORY}/.evergreen/run-atlas-tests.sh
```
ShaneHarvey, line 469: This is a more user friendly way to manage secrets. We load the env variables in secret (with `silent:true`) and then we're free to use xtrace and view test output as usual. Now we can actually see why a task failed :)

.evergreen/run-atlas-tests.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-# Don't trace to avoid secrets showing up in the logs
+# Exit on error and enable trace.
 set -o errexit
+set -o xtrace
```
ShaneHarvey, line 5: We can do this safely. We just need to remember not to print or use the secret variables in the shell script.

.evergreen/run-atlas-tests.sh
```diff
@@ -15,27 +16,37 @@ if [ -z "$PYTHON_BINARY" ]; then
...
-        PYTHON=$PYTHON_BINARY
     fi
 fi
 
-echo "Running tests"
-$PYTHON test/atlas/test_connection.py
+echo "Running tests without dnspython"
+python test/atlas/test_connection.py
+
+# dnspython is incompatible with Jython so don't test that combination.
```
ShaneHarvey, line 47: See my comment in the Jira ticket: https://jira.mongodb.org/browse/PYTHON-1878?focusedCommentId=3548124&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3548124

.evergreen/config.yml
```diff
@@ -444,8 +444,29 @@ functions:
         silent: true
         working_dir: "src"
         script: |
-          # DO NOT ECHO WITH XTRACE (which PREPARE_SHELL does)
-          PYTHON_BINARY=${PYTHON_BINARY} ATLAS_REPL='${atlas_repl}' ATLAS_SHRD='${atlas_shrd}' ATLAS_FREE='${atlas_free}' ATLAS_TLS11='${atlas_tls11}' ATLAS_TLS12='${atlas_tls12}'  sh ${PROJECT_DIRECTORY}/.evergreen/run-atlas-tests.sh
+          cat <<EOT > prepare_atlas_connectivity.sh
+          export ATLAS_FREE='${atlas_free}'
```
prashantmital, line 448: I think we can further optimize some of these invocations by switching to using `subprocess.exec` instead of `shell.exec` and then setting either the`include_expansions_in_env` option or the `add_expansions_to_env` option to automatically set the appropriate environment variables. I suppose we don't need to do it right now, but being able to eventually eliminate the use of `$PREPARE_SHELL` would be a huge improvement IMO.

.evergreen/config.yml
prashantmital: <3 it!

test/atlas/test_connection.py
```diff
@@ -40,29 +55,47 @@ def _connect(uri):
...
+
+    @unittest.skipUnless(HAS_DNS, 'SRV requires dnspython')
+    def test_srv_tls_11(self):
+        _connect(ATLAS_SRV_TLS11)
+
+    @unittest.skipUnless(HAS_DNS, 'SRV requires dnspython')
+    def test_srv_tls_12(self):
+        _connect(ATLAS_SRV_TLS12)
+
+    @unittest.skipUnless(PYMONGO_MUST_HAVE_DNS, 'dnspython is optional')
```
prashantmital, line 95: Nice! Can you add a comment/docstring detailing what this test achieves? Specifically, we don't end up silently skipping all tests if `dnspython` is not installed (instead we get a failure due to this test).

.evergreen/config.yml
prashantmital: Just to clarify, I am just suggesting this as a future optimization/refactor. Not saying that we need to do it in this PR. Would love to hear your thoughts on it @ShaneHarvey

ShaneHarvey: Cool idea! Can you open a ticket with this comment?

test/atlas/test_connection.py
ShaneHarvey: I changed this up a bit. PTAL at the new approach:
> `@unittest.skipUnless(HAS_DNS or MUST_TEST_SRV, 'SRV requires dnspython')`

ShaneHarvey (commented):
I updated the tests in response to your detailed review and added an extra test to ensure that we don't accidentally duplicate the test URIs in our evergreen project configs (because I'm paranoid). Please take another look.

.evergreen/config.yml
prashantmital: https://jira.mongodb.org/browse/PYTHON-2481

prashantmital (approved):

PR #541, opened 2021-01-11
PYTHON-2455 Add DOCS ticket step to release checklist

ShaneHarvey:


RELEASE.rst
```diff
@@ -102,3 +102,7 @@ Doing a Release
 
 15. Announce the release on:
     https://developer.mongodb.com/community/forums/c/community/release-notes/
+
+16. File a ticket for DOCS highlighting changes in server version and Python
```
prashantmital, line 106: Is there an example of one such ticket that you can point to? Might be useful to link to that here so we don't have to go searching everytime this needs to be done....

ShaneHarvey: Done.

prashantmital (approved):

PR #542, opened 2021-01-11
PYTHON-1316 Remove Database.eval and database.SystemJS

ShaneHarvey:
Depends on https://github.com/mongodb/mongo-python-driver/pull/540.

### DIFF
--- a/doc/migrate-to-pymongo4.rst
+++ b/doc/migrate-to-pymongo4.rst
@@ -50,27 +50,23 @@ Warnings can also be changed to errors::
 .. note:: Not all deprecated features raise :exc:`DeprecationWarning` when
   used. See `Removed features with no migration path`_.
 
-Database
---------
+Removed features with no migration path
+---------------------------------------
 
-The "eval" method and SystemJS class are removed
-................................................
+Database.eval, Database.system_js, and SystemJS are removed
+...........................................................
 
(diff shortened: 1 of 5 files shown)


doc/migrate-to-pymongo4.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Removed :meth:`~pymongo.database.Database.eval` and
+:class:`~pymongo.database.SystemJS`. The eval command was deprecated in
+MongoDB 3.0 and removed in MongoDB 4.2.
+
+Code like this::
+
+  >>> result = database.eval('function (x) {return x;}', 3)
+
+can be changed to this with MongoDB <= 4.0::
```
behackett, line 60: You should say explicitly that there is no replacement for eval when using MongoDB 4.2+.

behackett (changes requested):

prashantmital (commented):
Does this not count as a removed feature with no migration path?

ShaneHarvey:
> Does this not count as a removed feature with no migration path for MongoDB 4.2+? Should we add a line or two addressing this to that section?

I didn't really think it was a good fit for that category because `eval` will not work on MongoDB 4.2+ even on PyMongo 3.x.

prashantmital (approved):
LGTM

PR #555, opened 2021-01-21
PYTHON-1319 Remove Collection.ensure_index and reindex

ShaneHarvey:


doc/migrate-to-pymongo4.rst
```diff
@@ -136,6 +136,47 @@ can be changed to this::
...
+
+Collection.ensure_index is removed
+..................................
+
+Removed :meth:`pymongo.collection.Collection.ensure_index`. Use
+:meth:`~pymongo.collection.Collection.create_index` or
+:meth:`~pymongo.collection.Collection.create_indexes` instead. Note that
+``ensure_index`` maintained an in memory cache of recently created indexes
+whereas the newer methods do not. Applications should avoid frequent calls
+to :meth:`~pymongo.collection.Collection.create_index` or
```
behackett, line 151: IIRC, createIndexes is a noop when creating the same index more than one time. Is all this text necessary?

pymongo/mongo_client.py
```diff
@@ -705,7 +705,6 @@ def __init__(
 
         self._event_listeners = options.pool_options.event_listeners
 
-        # Cache of existing indexes used by ensure_index ops.
         self.__index_cache = {}
```
behackett, line 708: Why are you leaving the cache?

doc/migrate-to-pymongo4.rst
ShaneHarvey: Yes, even if it's a no-op on the server side it is still potentially very expensive since it means an extra network roundtrip. mongoengine is a real example where performance suffers greatly by naively replacing ensure_index with create_index. See https://github.com/MongoEngine/mongoengine/issues/1446

behackett: OK, I relent. I think my real quibble is with the idea that your application or ODM would almost arbitrarily create indexes. I don't want to put that idea into any developer's head.

behackett (approved):

doc/migrate-to-pymongo4.rst
ShaneHarvey: Unfortunately that idea is already out there in pretty much all mongodb ODMs including PyMODM and MongoEngine. We also use the same pattern in GridFS.

PR #556, opened 2021-01-22
PYTHON-1320 Remove legacy CRUD methods

ShaneHarvey:
Remove save, insert, update, remove, and find_and_modify.


test/test_bulk.py
```diff
@@ -295,8 +295,8 @@ def test_upsert(self):
         self.assertEqual(self.coll.count_documents({'foo': 'bar'}), 1)
 
     def test_numerous_inserts(self):
-        # Ensure we don't exceed server's 1000-document batch size limit.
-        n_docs = 2100
```
behackett, line 299: I don't remember why this was hard coded. Do we not get it from the server on some old version like 2.6?

test/test_collection.py
```diff
@@ -1941,13 +1957,79 @@ def test_min_query(self):
         self.assertEqual(2, docs[0]["x"])
 
     def test_numerous_inserts(self):
-        # Ensure we don't exceed server's 1000-document batch size limit.
+        # Ensure we don't exceed server's maxWriteBatchSize size limit.
         self.db.test.drop()
-        n_docs = 2100
```
behackett, line 1946: Same question here.

test/test_monitoring.py
```diff
@@ -1088,49 +865,46 @@ def test_insert_many(self):
 
     def test_legacy_insert_many(self):
```
behackett, line 866: Seems like you should remove "legacy" from the name.

tools/benchmark.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2009-2015 MongoDB, Inc.
```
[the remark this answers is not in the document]
behackett: This benchmark isn't great and never really has been. Let's just get rid of it.

behackett (changes requested):

prashantmital (approved):

tools/benchmark.py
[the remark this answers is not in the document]
ShaneHarvey: Removed.

test/test_bulk.py
ShaneHarvey: We always use max_write_batch_size internally, this test was never updated when the server updated the batch size from 1000 to 100000.

behackett (approved):

PR #561, opened 2021-01-26
PYTHON-1312 Remove Database.add_user and Database.remove_user

ShaneHarvey:


doc/migrate-to-pymongo4.rst
```diff
@@ -136,6 +136,39 @@ can be changed to this::
     names = client.list_collection_names()
     non_system_names = client.list_collection_names(filter={"name": {"$regex": r"^(?!system\\.)"}})
 
+Database.add_user is removed
```
behackett, line 139: I think you should provide some motivation for the removals here. Especially about the problems with sessions.

behackett (changes requested):

prashantmital (approved):

doc/migrate-to-pymongo4.rst
behackett: I'm actually getting ahead of myself here. This comment is for the upcoming change to remove authenticate and logout. Sorry about that.

behackett (approved):

PR #565, opened 2021-02-03
PYTHON-2537 Fix benchmark when using Python 3

ShaneHarvey:
Tested here: https://spruce.mongodb.com/version/601aeb54d1fe0705740e9e26

Depends on: https://github.com/mongodb-labs/drivers-evergreen-tools/pull/139

### DIFF
--- a/.evergreen/run-perf-tests.sh
+++ b/.evergreen/run-perf-tests.sh
@@ -22,7 +22,7 @@ pip install simplejson
 
 python setup.py build_ext -i
 start_time=$(date +%s)
-python test/performance/perf_test.py
+python test/performance/perf_test.py --locals
 end_time=$(date +%s)
 elapsed_secs=$((end_time-start_time))
 
--- a/test/performance/perf_test.py
+++ b/test/performance/perf_test.py
@@ -399,7 +399,7 @@ def insert_json_file_with_file_id(filename):
 
 def read_json_file(filename):
     coll = proc_client.perftest.corpus
-    temp = tempfile.TemporaryFile()
+    temp = tempfile.TemporaryFile(mode='w')
     try:
         temp.writelines(
             [json.dumps(doc) + '\n' for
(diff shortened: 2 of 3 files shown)


.evergreen/run-perf-tests.sh
```diff
@@ -22,7 +22,7 @@ pip install simplejson
 
 python setup.py build_ext -i
 start_time=$(date +%s)
-python test/performance/perf_test.py
+python test/performance/perf_test.py --locals
```
ShaneHarvey, line 25: `--locals` is a unittest option: https://docs.python.org/3/library/unittest.html#cmdoption-unittest-locals
> Show local variables in tracebacks

This improves the tracebacks displayed when a test fails by including the local variables in each frame.

test/performance/perf_test.py
```diff
@@ -399,7 +399,7 @@ def insert_json_file_with_file_id(filename):
 
 def read_json_file(filename):
     coll = proc_client.perftest.corpus
-    temp = tempfile.TemporaryFile()
+    temp = tempfile.TemporaryFile(mode='w')
```
ShaneHarvey, line 402: This is required to fix the python 3 failure because TemporaryFile defaults to binary mode (https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) but we are writing non-binary strings.

.evergreen/run-perf-tests.sh
behackett: Will this slow down the benchmarks?

ShaneHarvey: Nope. Here's the cpython change which added it: https://github.com/python/cpython/commit/f0c819acd0f#diff-ba8a6f32f39f6047ea998da6ddc7356a30eff06dd2b84d19f9a285c3162d2a93R187

The flag only changes the behavior of how tracebacks are printed after a test fails. Since we don't expect any tests to fail in the normal case there won't be any perf impact.

behackett (approved):

PR #566, opened 2021-02-03
PYTHON-1974 Remove manylinux containers only needed for 3.4

ShaneHarvey:
Minor cleanup I noticed while doing the 3.11.3 release.

### DIFF
--- a/.evergreen/build-manylinux.sh
+++ b/.evergreen/build-manylinux.sh
@@ -2,10 +2,7 @@
 
 docker version
 
-# 2020-03-20-2fda31c Was the last release to include Python 3.4.
-images=(quay.io/pypa/manylinux1_x86_64:2020-03-20-2fda31c \
-        quay.io/pypa/manylinux1_i686:2020-03-20-2fda31c \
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Note the task is failing due to an unrelated issue (https://jira.mongodb.org/browse/PYTHON-2541) but it's fine to merge this change as is.

behackett:
This seems fine since it's a change for master. What about building wheels for PyMongo 3.x?

ShaneHarvey:
Not sure I understand the question. This change can't be backported to since 3.x still supports Python 3.4.

Although the manylinux release task is failing on evergreen it didn't hold up the 3.11.3 release. Instead, I ran the manylinux build script locally on my own machine and it worked fine. It's as simple as this:
```
./.evergreen/build-manylinux.sh
```

behackett:
Great. I guess my question was will we be able to ship wheels for PyMongo 3.x and Python 3.4 in the future. Sounds like the answer is yes.

behackett (approved):

PR #568, opened 2021-02-12
PYTHON-1314 Remove Database.authenticate and Database.logout

ShaneHarvey:
This change resolves: PYTHON-1314  and PYTHON-1601


prashantmital (commented):
We seem to retain the ability to authenticate sockets with multiple mechanisms/creds via a single Pool - is there a reason we need to have this ability since we are moving to one set of creds per client?

pymongo/pool.py
```diff
@@ -815,10 +808,6 @@ def validate_session(self, client, session):
                 raise InvalidOperation(
                     'Can only use session with the MongoClient that'
                     ' started it')
-            if session._authset != self.authset:
```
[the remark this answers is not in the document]
ShaneHarvey: Done.

ShaneHarvey (commented):
> We seem to retain the ability to authenticate sockets with multiple mechanisms/creds via a single Pool - is there a reason we need to have this ability since we are moving to one set of creds per client?

I left that in for the sake of time and to make this review smaller. I'll open a new ticket to cleanup the rest of the multi-auth code later on.

prashantmital (approved):
Thanks for clarifying. LGTM but I'll defer to @behackett for a final 👍  since he is more familiar with this code.

doc/migrate-to-pymongo4.rst
```diff
@@ -121,6 +121,27 @@ can be changed to this::
...
+    client = MongoClient()
+    client.admin.authenticate('user1', 'pass1')
+    client.admin.authenticate('user2', 'pass2')
+
+can be changed to this::
+
+    client1 = MongoClient(username='user1', password='pass1')
+    client2 = MongoClient(username='user2', password='pass2')
+
+Alternatively, create a single user that contains all the authentication privileges
```
prashantmital, line 142: This sounds like a much better pattern. Do you think it is feasible for us to say loudly that this is what we **recommend**?

ShaneHarvey: I think both options are valid. Sometimes you would want to use a separate client entirely, for example if the app needs to run maintenance tasks in addition to regular CRUD. It would not be ideal (security-wise) to give the entire CRUD app all the maintenance task's privileges. In the security world this is called the principle of least privilege: https://en.wikipedia.org/wiki/Principle_of_least_privilege

behackett (approved):
LGTM. It will be nice when we can clean up __all_credentials in the future.

ShaneHarvey:
I opened https://jira.mongodb.org/browse/PYTHON-2585 for the follow up work.

PR #572, opened 2021-03-02
Python 2548/add update description.truncated arrays field

prashantmital:


ShaneHarvey:
Does this feature require any documentation changes?

prashantmital:
AFAICT we don't document the various change event documents that the server could return and we don't wrap them in a class while returning them to the user so I don't see why we would need to document this?

ShaneHarvey (approved):
LGTM

PR #573, opened 2021-03-02
PYTHON-2591 Use server toolchain Python on Ubuntu-16 with x86-64, arm64, power8

prashantmital:


.evergreen/run-tests.sh
```diff
@@ -58,8 +58,13 @@ fi
 
 if [ -z "$PYTHON_BINARY" ]; then
     # Use Python 3 from the server toolchain to test on ARM, POWER or zSeries if a
-    # system python3 doesn't exist. This seems to only be an issue on RHEL 7.x.
+    # system python3 doesn't exist or exists but is older than 3.6.
     PYTHON=$(command -v python3 || command -v /opt/mongodbtoolchain/v2/bin/python3) || true
```
behackett, line 62: I think it might be more clear to split this line up into two checks. The first checks if a new enough python 3 version is provided by the OS. If not, see if a new enough python version is in the server toolchain.

ShaneHarvey:
Can you also test this change on RHEL 7.2 zSeries (PYTHON-2590)? This might fix the problem there too.

.evergreen/run-tests.sh
prashantmital: Done.

.evergreen/run-tests.sh
```diff
@@ -58,8 +58,13 @@ fi
...
+    if $PYTHON -c "import sys; exit(sys.version_info[1] > 5)"; then
```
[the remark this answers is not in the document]
prashantmital: Done.

prashantmital:
It does indeed fix the issue on Zseries as well... The failure on Amazon Linux seems to be unrelated.

ShaneHarvey (approved):
LGTM

behackett (approved):
Nice

PR #574, opened 2021-03-03
PYTHON-2544 Do not check error messages when an error code is present

ShaneHarvey:
Add 10058 as a "not master" error code to account for MongoDB<=3.2 errors.

This change implements: https://github.com/mongodb/specifications/pull/920


prashantmital (approved):
I am loving the close correspondence between these code changes and the pseudocode included in the spec. LGTM

PR #578, opened 2021-03-16
PYTHON-1359: Add Example for RawBSONDocument

guanlinzhou:
Adds an Example for usage of RawBSONDocuments with the use case of moving documents across different collections

### DIFF
--- a/bson/raw_bson.py
+++ b/bson/raw_bson.py
@@ -30,7 +30,7 @@ Example: Moving a document between different databases/collections
   >>> import bson
   >>> from pymongo import MongoClient
   >>> from bson.raw_bson import RawBSONDocument
-  >>> client = MongoClient("localhost", 27017, document_class=bson.)
+  >>> client = MongoClient("localhost", 27017, document_class=RawBSONDocument)
   >>> db = client.db
   >>> result = db.test.insert_many([
   ...  {'a': 1},
@@ -42,8 +42,7 @@ Example: Moving a document between different databases/collections
   >>> for doc in db.test.find():
   ...    print(f"raw document: {doc.raw}")
   ...    print(f"decoded document: {bson.decode(doc.raw)}")
-  ...    result = replica_db.test.insert_one(doc)
-  ...    assert result.acknowledged
+  ...    replica_db.test.insert_one(doc)
   raw document: b'...'
   decoded document: {'_id': ObjectId('...'), 'a': 1}
   raw document: b'...'
@@ -67,8 +66,6 @@ Example: Moving a document between different databases/collections
 For use cases like moving documents across different databases or writing binary
 blobs to disk, using raw BSON documents provides better speed and avoids the
 overhead of decoding or encoding BSON.
-
-.. versionadded:: 3.12
 """
 
 from collections.abc import Mapping as _Mapping
--- a/bson/raw_bson.py
+++ b/bson/raw_bson.py
@@ -18,41 +18,57 @@ Inserting and Retrieving RawBSONDocuments
 =========================================
 
 Example: Moving a document between different databases/collections
-.. code-block:: python
 
-  import bsonjs
+.. testsetup::
   from pymongo import MongoClient
-  from bson.raw_bson import RawBSONDocument
+  client = MongoClient("localhost", 27017, document_class=RawBSONDocument)
+  client.drop_database('db')
+  client.drop_database('replica_db')
 
-  client = MongoClient("Localhost", 27017, document_class=RawBSONDocument)
-  db = client.db
-  doc = {"_id": 1, "test": "1"}
-  doc_bson = bsonjs.loads('{"_id": 1, "test": "1"}')
+.. doctest::
 
-  # add original document to collection
-  result = db.collection.insert_one(doc)
-  assert result.acknowledged
-
-  # retrieve doc from collection
-  retrieved_doc = db.collection.find_one({"test":"1"})
-  assert retrieved_doc.raw == doc_bson
-
-  # insert raw BSON into replica db/collection
-  replica_db = client.replica_db
-  result = replica_db.collection.insert_one(retrieved_doc)
-  assert result.acknowledged
-
-  # retrieve doc from replica db/collection
-  retrieved_replica_doc = replica_db.collection.find_one({"test":"1"})
-  assert retrieved_replica_doc.raw == doc_bson
-  assert bsonjs.dumps(retrieved_replica_doc.raw) == '{ "_id" : 1, "test" : "1" }'
+  >>> import bson
+  >>> from pymongo import MongoClient
+  >>> from bson.raw_bson import RawBSONDocument
+  >>> client = MongoClient("localhost", 27017, document_class=bson.)
+  >>> db = client.db
+  >>> docs = [
+  ...  {'a': 1},
+  ...  {'b': 1},
+  ...  {'c': 1},
+  ...  {'d': 1}]
+  >>> result = db.test.insert_many(docs)
+  >>> assert result.acknowledged
+  >>> replica_db = client.replica_db
+  >>> for doc in db.test.find():
+  ...    print(f"raw document: {doc.raw}")
+  ...    print(f"decoded document: {bson.decode(doc.raw)}")
+  ...    replica_db.test.insert_one(doc)
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'a': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'b': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'c': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'd': 1}
+  >>> for doc in replica_db.test.find():
+  ...    print(f"raw document: {doc.raw}")
+  ...    print(f"decoded document: {bson.decode(doc.raw)}")  
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'a': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'b': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'c': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'd': 1}
 
 For use cases like moving documents across different databases or writing binary
 blobs to disk, using raw BSON documents provides better speed and avoids the
-overhead of decoding BSON to JSON.
+overhead of decoding or encoding BSON.
 
 .. versionadded:: 3.12
-
 """
 
 from collections.abc import Mapping as _Mapping
--- a/bson/raw_bson.py
+++ b/bson/raw_bson.py
@@ -13,6 +13,46 @@
 # limitations under the License.
 
 """Tools for representing raw BSON documents.
+
+Inserting and Retrieving RawBSONDocuments
+=========================================
+
+Example: Moving a document between different databases/collections
+.. code-block:: python
+
+  import bsonjs
+  from pymongo import MongoClient
+  from bson.raw_bson import RawBSONDocument
+
+  client = MongoClient("Localhost", 27017, document_class=RawBSONDocument)
+  db = client.db
+  doc = {"_id": 1, "test": "1"}
+  doc_bson = bsonjs.loads('{"_id": 1, "test": "1"}')
+
+  # add original document to collection
+  result = db.collection.insert_one(doc)
+  assert result.acknowledged
+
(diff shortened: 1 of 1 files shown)


bson/raw_bson.py
```diff
@@ -13,6 +13,46 @@
 # limitations under the License.
 
 """Tools for representing raw BSON documents.
+
+Inserting and Retrieving RawBSONDocuments
+=========================================
+
+Example: Moving a document between different databases/collections
+.. code-block:: python
```
prashantmital, line 21: Does this render correctly? I feel like you might need to have a newline before `.. code-block::` to get it to render correctly?

You can build the docs locally by installing `sphinx` and running `python setup.py doc`

bson/raw_bson.py
```diff
@@ -13,6 +13,46 @@
...
+
+  import bsonjs
```
prashantmital, line 23: We shouldn't be using `bsonjs` here because we don't know if the user is familiar with that package. It is also not needed for this particular usecase.
To be clear, `bsonjs` is used to directly convert between JSON (which is a string) and BSON (which is binary `bytes`). It saves us the trouble of having to convert everything to Python dicts when converting BSON data into JSON. E.g. the following snippet would retrieve a document and print it as JSON:
```
doc = db.coll.find_one()
print(json.dumps(doc))
```
However, in doing so we actually end up doing 2x the work we need to. `doc = db.coll.find_one()` gets the document in BSON format from the server and converts it to a Python dict. `json.dumps(doc)` converts that Python dict to a JSON string. With `bsonjs.dumps` we can directly convert BSON bytes to a JSON string hence saving us CPU time. This is the main usecase for `bsonjs` as people often want to server query results as JSON. However, this is not needed to demonstrate usage of `RawBSONDocument`.

bson/raw_bson.py
```diff
@@ -13,6 +13,46 @@
...
+  from pymongo import MongoClient
+  from bson.raw_bson import RawBSONDocument
+
+  client = MongoClient("Localhost", 27017, document_class=RawBSONDocument)
+  db = client.db
+  doc = {"_id": 1, "test": "1"}
+  doc_bson = bsonjs.loads('{"_id": 1, "test": "1"}')
```
prashantmital, line 30: This is not needed.

bson/raw_bson.py
```diff
@@ -13,6 +13,46 @@
...
+=========================================
+
+Example: Moving a document between different databases/collections
+.. code-block:: python
+
+  import bsonjs
+  from pymongo import MongoClient
+  from bson.raw_bson import RawBSONDocument
+
+  client = MongoClient("Localhost", 27017, document_class=RawBSONDocument)
```
prashantmital, line 27: Does this actually work with `Localhost` instead of `localhost`?

bson/raw_bson.py
```diff
@@ -13,6 +13,46 @@
...
+  db = client.db
+  doc = {"_id": 1, "test": "1"}
+  doc_bson = bsonjs.loads('{"_id": 1, "test": "1"}')
+
+  # add original document to collection
```
prashantmital, line 32: Suggest changing this to something like:

```
  # add some documents to collection 'test'
  result = db.test.insert_many([
      {'a': 1},
      {'b': 1},
      {'c': 1},
      {'d': 1}])
  assert result.acknowledged

  # duplicate all documents in collection 'test' in collection 'test_copy'
  for doc in db.test.find():
      print(f"raw document: {doc.raw}\ndecoded document: {bson.decode(doc.raw)}")
      db.test_copy.insert_one(doc)
```

prashantmital: You should also include the sample output generated from the `print` above in your code sample to illustrate how raw BSON documents work. 
This is just a suggestion - you should write the example the way it makes most sense to you.

bson/raw_bson.py
```diff
@@ -13,6 +13,46 @@
...
+  assert result.acknowledged
+
+  # retrieve doc from replica db/collection
+  retrieved_replica_doc = replica_db.collection.find_one({"test":"1"})
+  assert retrieved_replica_doc.raw == doc_bson
+  assert bsonjs.dumps(retrieved_replica_doc.raw) == '{ "_id" : 1, "test" : "1" }'
+
+For use cases like moving documents across different databases or writing binary
+blobs to disk, using raw BSON documents provides better speed and avoids the
+overhead of decoding BSON to JSON.
```
prashantmital, line 52: We don't decode BSON to JSON, we decode it to a Python dict.

bson/raw_bson.py
ShaneHarvey, line 23: Let's remove bsonjs since most people are not familiar with it and it's not needed to work with RawBSONDocuments.

bson/raw_bson.py
ShaneHarvey, line 30: Let's replace `bsonjs.loads(...)` with `bson.encode(doc)`.

bson/raw_bson.py
ShaneHarvey, line 52: "decoding BSON to JSON." -> "decoding or encoding BSON."

prashantmital (changes requested):

bson/raw_bson.py
```diff
@@ -13,6 +13,46 @@
...
+
+  # insert raw BSON into replica db/collection
+  replica_db = client.replica_db
+  result = replica_db.collection.insert_one(retrieved_doc)
+  assert result.acknowledged
+
+  # retrieve doc from replica db/collection
+  retrieved_replica_doc = replica_db.collection.find_one({"test":"1"})
+  assert retrieved_replica_doc.raw == doc_bson
+  assert bsonjs.dumps(retrieved_replica_doc.raw) == '{ "_id" : 1, "test" : "1" }'
```
ShaneHarvey, line 48: Let's remove this line since it uses bsonjs.

ShaneHarvey (commented):
Great start!

bson/raw_bson.py
guanlinzhou: Ah you're right, thanks for noticing!

bson/raw_bson.py
guanlinzhou: rm'd

bson/raw_bson.py
guanlinzhou: It did work, but I changed it to lowercase anyways

bson/raw_bson.py
guanlinzhou: yep i updated it to use multiple docs

bson/raw_bson.py
guanlinzhou: changed!

bson/raw_bson.py
guanlinzhou: rm'd

bson/raw_bson.py
guanlinzhou: sounds good, rm'd

bson/raw_bson.py
guanlinzhou: I removed this entirely to change the test to something else

bson/raw_bson.py
guanlinzhou: rm'd

bson/raw_bson.py
guanlinzhou: changed!

guanlinzhou:
Attaching Screenshot of Built Example
image

bson/raw_bson.py
```diff
@@ -13,6 +13,62 @@
...
+  ...  {'b': 1},
+  ...  {'c': 1},
+  ...  {'d': 1}])
+  >>> assert result.acknowledged
+  >>> replica_db = client.replica_db
+  >>> for doc in db.test.find():
+  ...    print(f"raw document: {doc.raw}")
+  ...    print(f"decoded document: {bson.decode(doc.raw)}")
+  ...    result = replica_db.test.insert_one(doc)
+  ...    assert result.acknowledged
```
ShaneHarvey, line 46: I think we can ignore result here:
```
replica_db.test.insert_one(doc)
```

The `acknowledged` property isn't expected to be checked very often and it's not very relevant to this example.

bson/raw_bson.py
```diff
@@ -13,6 +13,62 @@
...
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'a': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'b': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'c': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'd': 1}
+  >>> for doc in replica_db.test.find():
```
ShaneHarvey, line 55: How much value does iterating the `replica_db.test` collection in this example provide?

bson/raw_bson.py
```diff
@@ -13,6 +13,62 @@
...
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'c': 1}
+  raw document: b'...'
+  decoded document: {'_id': ObjectId('...'), 'd': 1}
+
+For use cases like moving documents across different databases or writing binary
+blobs to disk, using raw BSON documents provides better speed and avoids the
+overhead of decoding or encoding BSON.
+
+.. versionadded:: 3.12
```
ShaneHarvey, line 71: We should remove `versionadded`. We use that to document new features or new APIs whereas this change is improving the documentation for an existing feature.

bson/raw_bson.py
guanlinzhou: removed!

bson/raw_bson.py
guanlinzhou: removed

bson/raw_bson.py
guanlinzhou: It was mostly just to show that the documents in the new collection match the ones in the old one. I'm not sure how I'd change it aside from just straight removing it (which I'd be ok with doing). Let me know if you have any suggestions!

bson/raw_bson.py
```diff
@@ -13,6 +13,59 @@
...
 """Tools for representing raw BSON documents.
+
+Inserting and Retrieving RawBSONDocuments
+=========================================
+
+Example: Moving a document between different databases/collections
+
+.. testsetup::
+  from pymongo import MongoClient
+  client = MongoClient("localhost", 27017, document_class=RawBSONDocument)
```
prashantmital, line 24: Let's remove `localhost` and `27017` as these are the defaults used by PyMongo anyway.

bson/raw_bson.py
```diff
@@ -13,6 +13,59 @@
...
+  client.drop_database('db')
+  client.drop_database('replica_db')
+
+.. doctest::
+
+  >>> import bson
+  >>> from pymongo import MongoClient
+  >>> from bson.raw_bson import RawBSONDocument
+  >>> client = MongoClient("localhost", 27017, document_class=RawBSONDocument)
```
prashantmital, line 33: Same comment about the `host` and `port` parameters here... I am not sure they add value to the example?

bson/raw_bson.py
```diff
@@ -13,6 +13,59 @@
...
+  >>> db = client.db
+  >>> result = db.test.insert_many([
+  ...  {'a': 1},
```
prashantmital, line 36: This indentation seems a bit weird to me... Can we align the documents the same way as https://pymongo.readthedocs.io/en/stable/examples/aggregation.html

bson/raw_bson.py
```diff
@@ -13,6 +13,59 @@
...
+  ...  {'b': 1},
+  ...  {'c': 1},
+  ...  {'d': 1}])
+  >>> assert result.acknowledged
```
prashantmital, line 40: @ShaneHarvey is this necessary?

bson/raw_bson.py
prashantmital: I'd be fine with removing this entirely. @ShaneHarvey can make the final call.

prashantmital (changes requested):

bson/raw_bson.py
ShaneHarvey: Let's remove it.

bson/raw_bson.py
ShaneHarvey: Not needed. We can remove `result` entirely. I made a similar comment here: https://github.com/mongodb/mongo-python-driver/pull/578/files/1bea157142595b2f7b1472bb42035d958a3875ae#r596222833

bson/raw_bson.py
guanlinzhou: didn't know that these were defaults, removing~

prashantmital:
Can we run the doctest task on the Evergreen patch build to make sure this passes?

prashantmital (approved):

ShaneHarvey (approved):

PR #580, opened 2021-03-18
PYTHON-1690: Fix error message when insert_many is given a single RawBSONDocument instead of a list

guanlinzhou:
Changed type checking from `abc.Iterable` to `abc.Sequence` in order to better match 'list-like' types

```
TypeError                                 
Traceback (most recent call last)
<ipython-input-6-9084eaf77999> in <module>
----> 1 client.test.test.insert_many(RawBSONDocument(bson.BSON.encode({'_id':2})))

~/mongo-python-driver/pymongo/collection.py in insert_many(self, documents, ordered, bypass_document_validation, session)
    688         """
    689         if not isinstance(documents, abc.Sequence) or not documents:
--> 690             raise TypeError("documents must be a non-empty list")
    691         inserted_ids = []
    692         def gen():

TypeError: documents must be a non-empty list
```

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -687,7 +687,7 @@ class Collection(common.BaseObject):
         .. versionadded:: 3.0
         """
         if (not isinstance(documents, abc.Iterable)
-                or isinstance(documents, abc.Mapping) 
+                or isinstance(documents, abc.Mapping)
                 or not documents):
             raise TypeError("documents must be a non-empty list")
         inserted_ids = []
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -686,7 +686,7 @@ class Collection(common.BaseObject):
 
         .. versionadded:: 3.0
         """
-        if not isinstance(documents, abc.Sequence) or not documents:
+        if not isinstance(documents, abc.Iterable) or isinstance(documents, abc.Mapping) or not documents:
             raise TypeError("documents must be a non-empty list")
         inserted_ids = []
         def gen():
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -686,7 +686,7 @@ class Collection(common.BaseObject):
 
         .. versionadded:: 3.0
         """
-        if not isinstance(documents, abc.Iterable) or not documents:
+        if not isinstance(documents, abc.Sequence) or not documents:
             raise TypeError("documents must be a non-empty list")
         inserted_ids = []
         def gen():
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -686,9 +686,9 @@ class Collection(common.BaseObject):
 
         .. versionadded:: 3.0
         """
-        if not isinstance(documents, abc.Iterable) \
-                or isinstance(documents, abc.Mapping) \
-                or not documents:
+        if (not isinstance(documents, abc.Iterable)
+                or isinstance(documents, abc.Mapping) 
+                or not documents):
             raise TypeError("documents must be a non-empty list")
         inserted_ids = []
         def gen():
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -686,7 +686,9 @@ class Collection(common.BaseObject):
 
         .. versionadded:: 3.0
         """
-        if not isinstance(documents, abc.Iterable) or isinstance(documents, abc.Mapping) or not documents:
+        if not isinstance(documents, abc.Iterable) \
+                or isinstance(documents, abc.Mapping) \
+                or not documents:
             raise TypeError("documents must be a non-empty list")
         inserted_ids = []
         def gen():
--- a/test/test_collection.py
+++ b/test/test_collection.py
@@ -23,6 +23,8 @@ import sys
 from codecs import utf_8_decode
 from collections import defaultdict
 
+import bson
+
 sys.path[0:0] = [""]
 
 from bson import encode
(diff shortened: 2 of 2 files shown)


test/test_collection.py
```diff
@@ -876,7 +876,7 @@ def test_bad_dbref(self):
         id_only = {'ref': {'$id': ObjectId()}}
         self.assertRaises(InvalidDocument, self.db.test.insert_one, ref_only)
         self.assertRaises(InvalidDocument, self.db.test.insert_one, id_only)
-
+x
```
ShaneHarvey, line 879: Looks like a typo? Let's remove it.

guanlinzhou: oops not sure how that got there, removed!

guanlinzhou:
@prashantmital @ShaneHarvey thoughts on whether this requires a testing change? I was looking for tests for invalid input types and couldn't find any existing ones

ShaneHarvey:
Yeah it would be great to add a test in `test_collection.py` which uses `with self.assertRaisesRegex(...` on various invalid types.

pymongo/collection.py
```diff
@@ -686,7 +686,7 @@ def insert_many(self, documents, ordered=True,
 
         .. versionadded:: 3.0
         """
-        if not isinstance(documents, abc.Iterable) or not documents:
+        if not isinstance(documents, abc.Iterable) or isinstance(documents, abc.Mapping) or not documents:
```
prashantmital, line 689: Can we make sure this to 80 characters or less?

guanlinzhou: done

guanlinzhou:
> Yeah it would be great to add a test in `test_collection.py` which uses `with self.assertRaisesRegex(...` on various invalid types.

added tests in the most recent commit

test/test_collection.py
```diff
@@ -23,6 +23,8 @@
 from codecs import utf_8_decode
 from collections import defaultdict
 
+import bson
+
 sys.path[0:0] = [""]
 
 from bson import encode
```
prashantmital, line 30: `encode` is already imported here so you can just use that on L809 instead of re-importing the entire `bson` module.

test/test_collection.py
```diff
@@ -786,6 +788,26 @@ def test_insert_many(self):
         self.assertFalse(result.acknowledged)
         self.assertEqual(20, db.test.count_documents({}))
 
+    def test_insert_many_invalid(self):
+        db = self.db
+        db.test.drop()
```
ShaneHarvey, line 793: Let's remove this `db.test.drop()` since in this test it doesn't actually matter what the database looks like.

pymongo/collection.py
```diff
@@ -686,7 +686,9 @@ def insert_many(self, documents, ordered=True,
 
         .. versionadded:: 3.0
         """
-        if not isinstance(documents, abc.Iterable) or not documents:
+        if not isinstance(documents, abc.Iterable) \
```
prashantmital, line 689: While this isn't incorrect, it is bad style to use backslash. As per the [style guide](https://www.python.org/dev/peps/pep-0008/#maximum-line-length):

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

Can you modify this so the conditional check is enclosed in `()` instead?

test/test_collection.py
```diff
@@ -23,6 +23,8 @@
 from codecs import utf_8_decode
 from collections import defaultdict
 
+import bson
```
prashantmital, line 26: This isn't needed - see comment below.

test/test_collection.py
prashantmital, line 793: I see that many tests in this file do this at the top which, I assume is why you included it here. However, there is also the fact that the class is wired to do this before _and_ after every test:

```
    def setUp(self):
        self.db.test.drop()

    def tearDown(self):
        self.db.test.drop()
```

Seems like overkill to me. I think we can remove the drop() from this newly added test at least.

test/test_collection.py
```diff
@@ -786,6 +788,26 @@ def test_insert_many(self):
         self.assertFalse(result.acknowledged)
         self.assertEqual(20, db.test.count_documents({}))
 
+    def test_insert_many_invalid(self):
+        db = self.db
+        db.test.drop()
+
+        with self.assertRaisesRegex(
```
prashantmital, line 795: Nice!

pymongo/collection.py
```diff
@@ -686,7 +686,9 @@ def insert_many(self, documents, ordered=True,
...
+                or isinstance(documents, abc.Mapping) \
+                or not documents:
```
ShaneHarvey, line 691: We prefer to use parentheses instead of ending lines with `\`. Do you mind changing it to:
```python
        if (not isinstance(documents, abc.Iterable)
                or isinstance(documents, abc.Mapping)
                or not documents):
 ```
 
 (And yes we should be using a code formatter to make this easier. One day soon we'll get to it: https://jira.mongodb.org/browse/PYTHON-1834)

test/test_collection.py
guanlinzhou: ok sounds good!

pymongo/collection.py
guanlinzhou: fixed!

prashantmital (approved):
LGTM 🥳 

Note that normally if you would have backported this commit to a branch that tested against Python 2.7, your tests would have failed because you are using `assertRaisesRegex` where Py 2 called that same method `assertRaisesRegexp`. Fortunately, our test setup code has a nifty tweak in place to prevent this from becoming a problem - see https://github.com/mongodb/mongo-python-driver/blob/c40a4554c3de42439cdfb07d9166d5b619423c29/test/__init__.py#L822

ShaneHarvey (approved):

ShaneHarvey:
Please run the tests before pushing a backport to v3.12. I think the `abc.Mapping` usage may be different in that branch.

PR #581, opened 2021-03-19
PYTHON-2615 Reinstate TLS network timeout workaround due to eventlet

ShaneHarvey:
PYTHON-2616 Fix test_network_error_message when TLS is enabled.

### DIFF
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -237,6 +237,12 @@ def _raise_connection_failure(address, error, msg_prefix=None):
         msg = msg_prefix + msg
     if isinstance(error, socket.timeout):
         raise NetworkTimeout(msg) from error
+    elif isinstance(error, _SSLError) and 'timed out' in str(error):
+        # Eventlet does not distinguish TLS network timeouts from other
+        # SSLErrors (https://github.com/eventlet/eventlet/issues/692).
+        # Luckily, we can work around this limitation because the phrase
+        # 'timed out' appears in all the timeout related SSLErrors raised.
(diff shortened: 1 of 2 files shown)


pymongo/pool.py
```diff
@@ -237,6 +237,12 @@ def _raise_connection_failure(address, error, msg_prefix=None):
         msg = msg_prefix + msg
     if isinstance(error, socket.timeout):
         raise NetworkTimeout(msg) from error
+    elif isinstance(error, _SSLError) and 'timed out' in str(error):
+        # Eventlet does not distinguish TLS network timeouts from other
```
prashantmital, line 241: I am wondering since this is only occurring when `eventlet` is installed should we check whether `eventlet` is installed (I don't know if its possible - I feel like it should be) inside this `elif`?

ShaneHarvey: No we can't check if eventlet is installed. Merely importing eventlet can have effects like we safe in the dnspython 2.0 vs eventlet bug.

prashantmital: I see.
Using `__import__` will actually eliminate any side effects as it doesn't mess with the current global/local namespaces (unless instructed to do so) but I take your point that checking `eventlet` in any form would present a risk with minimal reward.

prashantmital (approved):

pymongo/pool.py
ShaneHarvey: I don't think `__import__` can prevent the types of issues we saw in https://github.com/mongodb/mongo-python-driver/pull/484#issuecomment-691224196

PR #583, opened 2021-03-22
PYTHON-2480: Add MongoClient helper to access the current TopologyDescription

guanlinzhou:
Add a property for accessing current topology description in MongoClient.

Open Questions:
How to test these changes/run locally?

### DIFF
--- a/doc/api/pymongo/index.rst
+++ b/doc/api/pymongo/index.rst
@@ -48,6 +48,7 @@ Sub-modules:
    read_preferences
    results
    server_api
+   server_description
    topology_description
    uri_parser
    write_concern
--- a/doc/api/pymongo/index.rst
+++ b/doc/api/pymongo/index.rst
@@ -48,6 +48,7 @@ Sub-modules:
    read_preferences
    results
    server_api
+   topology_description
    uri_parser
    write_concern
    event_loggers
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -11,7 +11,7 @@ changes. For example, all APIs deprecated in PyMongo 3.X have been removed.
 Be sure to read the changes listed below and the :doc:`migrate-to-pymongo4`
 before upgrading from PyMongo 3.x.
 
-- Added :attr:`pymongo.mongo_client.MongoClient.topology_description`
+- Added :attr:`pymongo.mongo_client.MongoClient.topology_description`.
 
 Breaking Changes in 4.0
 .......................
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -11,6 +11,8 @@ changes. For example, all APIs deprecated in PyMongo 3.X have been removed.
 Be sure to read the changes listed below and the :doc:`migrate-to-pymongo4`
 before upgrading from PyMongo 3.x.
 
+- Added :attr:`pymongo.mongo_client.MongoClient.topology_description`
+
 Breaking Changes in 4.0
 .......................
 
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -894,8 +894,8 @@ class MongoClient(common.BaseObject):
     @property
     def topology_description(self):
         """Returns the current :class:`~pymongo.topology_description.TopologyDescription`.
-        The returned object is immutable, and should be called periodically/as needed
-        to avoid using stale information.
+        The returned object is immutable, and should be accessed periodically/as needed
+        to avoid seeing stale topology information.
 
         .. versionadded:: 4.0
         """
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -897,7 +897,7 @@ class MongoClient(common.BaseObject):
         The returned object is immutable, and should be called periodically/as needed
         to avoid using stale information.
 
-        .. versionadded:: 3.12
+        .. versionadded:: 4.0
         """
         return self._topology.description
 
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -893,7 +893,9 @@ class MongoClient(common.BaseObject):
 
     @property
     def topology_description(self):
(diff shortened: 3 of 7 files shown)


prashantmital (changes requested):
Since we are adding a public property we also want to add something to `test_client.py` that would result in a failure if we ever inadvertently removed this property from `MongoClient`.  The simplest( and most naive) approach to testing this would be to add a test to `ClientUnitTest` that checks that `client.topology_description` returns the same object as `client._topology._description`.

pymongo/mongo_client.py
```diff
@@ -891,6 +891,14 @@ def event_listeners(self):
         """
         return self._event_listeners.event_listeners
 
+    @property
+    def topology_description(self):
+        """(id, topology_type, servers, etc.) of current topology
```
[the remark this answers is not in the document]
ShaneHarvey: TopologyDescription is already a publicly documented type so we can link to it directly: https://pymongo.readthedocs.io/en/stable/api/pymongo/topology_description.html#pymongo.topology_description.TopologyDescription

We should probably improve the TopologyDescription/ServerDescription docs as well but let's file a new issue for that.

doc/api/pymongo/index.rst
```diff
@@ -48,6 +48,7 @@ Sub-modules:
    read_preferences
    results
    server_api
+   topology_description
```
ShaneHarvey, line 51: Can you also add `server_description` which is a related module.

ShaneHarvey (commented):
Great work. Just one quick ask.

pymongo/mongo_client.py
```diff
@@ -893,7 +893,9 @@ def event_listeners(self):
 
     @property
     def topology_description(self):
-        """(id, topology_type, servers, etc.) of current topology
+        """Returns the current :class:`~pymongo.topology_description.TopologyDescription`.
+        The returned object is immutable, and should be called periodically/as needed
+        to avoid using stale information.
```
prashantmital, line 898: called --> accessed 
to avoid using stale information --> to avoid seeing stale topology information

prashantmital: @ShaneHarvey should we also tell users that it is impossible to guarantee that the `topology_description` won't be changed out from under them while they are accessing this property? Since we are not exposing the topology lock (or even exposing it to users for that matter) it seems impossible to guarantee that accessing this object will provide the latest view of the topology -- conceivably, a monitoring thread could acquire the lock and update the topology while the user holds onto a stale reference. I am not saying we need to provide a way for users to get the 'current'/'in-use' topology description, but it seems like something worth documenting?

doc/changelog.rst
```diff
@@ -11,6 +11,8 @@ changes. For example, all APIs deprecated in PyMongo 3.X have been removed.
 Be sure to read the changes listed below and the :doc:`migrate-to-pymongo4`
 before upgrading from PyMongo 3.x.
 
+- Added :attr:`pymongo.mongo_client.MongoClient.topology_description`
```
prashantmital, line 14: We usually end changelog entries with a period - `.`

prashantmital (changes requested):

pymongo/mongo_client.py
ShaneHarvey: Good question. Yes the topology can be updated directly after the user calls this method. That will always be the case even if we acquire/release the Topology lock here. I don't think we need to call this out besides what is already written in the docstring.

ShaneHarvey: Perhaps "current" is the confusing part. Maybe "Returns the most recent :class:...".

The issue is that the description is stale as soon as it's returned from this method.

pymongo/mongo_client.py
```diff
@@ -891,6 +891,16 @@ def event_listeners(self):
         """
         return self._event_listeners.event_listeners
 
+    @property
+    def topology_description(self):
+        """Returns the current :class:`~pymongo.topology_description.TopologyDescription`.
+        The returned object is immutable, and should be accessed periodically/as needed
+        to avoid seeing stale topology information.
```
ShaneHarvey, line 898: This might be clearer:

```python
        """The description of the connected MongoDB deployment.

        >>> client.topology_description
        <TopologyDescription id: 605a7b04e76489833a7c6113, topology_type: ReplicaSetWithPrimary, servers: [<ServerDescription ('localhost', 27017) server_type: RSPrimary, rtt: 0.0007973677999995488>, <ServerDescription ('localhost', 27018) server_type: RSSecondary, rtt: 0.0005540556000003249>, <ServerDescription ('localhost', 27019) server_type: RSSecondary, rtt: 0.0010367483999999649>]>
        >>> client.topology_description.topology_type_name
        'ReplicaSetWithPrimary'

        Note that the description is periodically updated in the background
        but the returned object itself is immutable. Access this property again
        to get a more recent
        :class:`~pymongo.topology_description.TopologyDescription`.

        :Returns:
          An instance of
          :class:`~pymongo.topology_description.TopologyDescription`.

        .. versionadded:: 3.12
        """
```

What do you think?

pymongo/mongo_client.py
prashantmital: +1 to changing this to 'most recent'. WE can always expound the precise behavior to any users who end up needing it.

pymongo/mongo_client.py
prashantmital: Nice! I love it.

test/test_client.py
```diff
@@ -597,6 +598,8 @@ def test_init_disconnected(self):
         self.assertFalse(c.secondaries)
         c = rs_or_single_client(connect=False)
         self.assertIsInstance(c.max_write_batch_size, int)
+        self.assertIsInstance(c.topology_description, TopologyDescription)
```
prashantmital, line 601: Nicely done.

prashantmital (approved):

ShaneHarvey (approved):

PR #586, opened 2021-03-31
BUMP 3.12.0b0

prashantmital:


doc/changelog.rst
```diff
@@ -1,21 +1,58 @@
...
+  :class:`~pymongo.collection.Collection` (`PYTHON-2466`_).
+
+Bug fixes
+.........
+
+- Fixed a bug that could cause the driver to deadlock during automatic
+  client side field level encryption (`PYTHON-2472`_).
+- Fixed a bug that prevented PyMongo from retrying writes after
+  a ``writeConcernError`` on MongoDB 4.4+ (`PYTHON-2452`_).
+- Fixed a memory leak caused by failing SDAM monitor checks on Python 3 (`PYTHON-2433`_).
```
ShaneHarvey, line 29: Almost all of these were fixed in 3.11.3 (or earlier 3.11 releases). Can we remove them?

ShaneHarvey: In fact, can you add the 3.11.2 and 3.11.3 Changelog entries?

prashantmital: OK. I was a bit confused about how we want this to be reflected... Since the git histories diverged and are not common between 3.11.x and 3.12.x I thought duplicating everything in 3.12.x was the way to go... Doesn't adding the 3.11.x sections here seem to imply that 3.12.0b0 builds on the source tree of 3.11.3?

ShaneHarvey: The changelog shouldn't need to match the git branch history all the time. If users want a full breakdown of all the tickets they can look at the jira issues or git commits but I don't think we need to duplicate changelog entries for bugs fixed log ago. It would be different if 3.12.0 was released and we needed to fix the same bug in 3.11.4 and 3.11.1. In that case the bug fix would go under both releases.

prashantmital: Ok sounds good.

ShaneHarvey (approved):

PR #589, opened 2021-04-01
PYTHON-2631 Add missing error message to InvalidBSON error

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-2631

### DIFF
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -2342,7 +2342,7 @@ static int _element_to_dict(PyObject* self, const char* string,
     if (name_length > BSON_MAX_SIZE || position + name_length >= max) {
         PyObject* InvalidBSON = _error("InvalidBSON");
         if (InvalidBSON) {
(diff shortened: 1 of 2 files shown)


bson/_cbsonmodule.c
```diff
@@ -2342,7 +2342,7 @@ static int _element_to_dict(PyObject* self, const char* string,
     if (name_length > BSON_MAX_SIZE || position + name_length >= max) {
         PyObject* InvalidBSON = _error("InvalidBSON");
         if (InvalidBSON) {
-            PyErr_SetNone(InvalidBSON);
+            PyErr_SetString(InvalidBSON, "invalid field name");
```
behackett, line 2345: Say why it's invalid?

ShaneHarvey: Done. I went with "field name too large". This can happen when the BSON itself gets corrupted or mangled. For example the null terminator was overwritten or the field was internally truncated like in the test case below.

behackett (approved):

PR #592, opened 2021-04-10
PYTHON-2397: MongoClient(ssl=True, tls=False) fails with an AttributeError

guanlinzhou:
Switch from common dictionary to CaseInsensitiveDictionary to fix `cased_key` attribute error.


ShaneHarvey (commented):
Could you add a test to ensure the `MongoClient(ssl=True, tls=False)` bug is fixed?

ShaneHarvey (approved):

prashantmital (approved):

PR #593, opened 2021-04-13
[PYTHON-2643] Requirements.txt

DuckMasterAl:
Adds a requirements.txt file that tells pip to install dnspython, a required package to connect to a MongoDB server.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+dnspython


ShaneHarvey:
Thanks but pymongo does not have any strict dependencies so we don't need a requirements.txt. To install pymongo with the optional dnspython dependency needed to use 'mongodb+srv://' connection strings you can install pymongo with the "srv" extra like this:
```
$ python -m pip install 'pymongo[srv]'
```
See https://pymongo.readthedocs.io/en/stable/installation.html#dependencies

PR #594, opened 2021-04-20
PYTHON-1880: Raise a warning when no_cursor_timeout is used with an implicit session

Kn99HN:
# Description
- Add a informative UserWarning when no_cursor_timeout is set

### DIFF
--- a/pymongo/cursor.py
+++ b/pymongo/cursor.py
@@ -144,10 +144,13 @@ class Cursor(object):
         if not isinstance(limit, int):
             raise TypeError("limit must be an instance of int")
         validate_boolean("no_cursor_timeout", no_cursor_timeout)
-        if no_cursor_timeout:
-            warnings.warn("use an explicit session with no_cursor_timeout=True"\
-                          "otherwise the cursor may still timeout after 30 minutes,"\
-                          "for more info see DOCS-11255",
+        if no_cursor_timeout and not self.__explicit_session:
+            warnings.warn("use an explicit session with no_cursor_timeout=True "
+                          "otherwise the cursor may still timeout after "
+                          "30 minutes, for more info see "
+                          "https://docs.mongodb.com/v4.4/reference/method/"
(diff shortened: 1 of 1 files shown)


pymongo/cursor.py
```diff
@@ -144,6 +144,11 @@ def __init__(self, collection, filter=None, projection=None, skip=0,
         if not isinstance(limit, int):
             raise TypeError("limit must be an instance of int")
         validate_boolean("no_cursor_timeout", no_cursor_timeout)
+        if no_cursor_timeout:
```
ShaneHarvey, line 147: suggest: `if no_cursor_timeout and not self.__explicit_session:`

This way users can avoid seeing the warning by applying the workaround described here: https://docs.mongodb.com/v4.4/reference/method/cursor.noCursorTimeout/#session-idle-timeout-overrides-nocursortimeout

pymongo/cursor.py
```diff
@@ -144,6 +144,11 @@ def __init__(self, collection, filter=None, projection=None, skip=0,
...
+            warnings.warn("use an explicit session with no_cursor_timeout=True"\
+                          "otherwise the cursor may still timeout after 30 minutes,"\
+                          "for more info see DOCS-11255",
```
ShaneHarvey, line 150: A few suggestions:
- replace DOCS-11255 with https://docs.mongodb.com/v4.4/reference/method/cursor.noCursorTimeout/#session-idle-timeout-overrides-nocursortimeout
- limit lines to 80 characters
- remove the trailing `\` since it's not needed.

ShaneHarvey (changes requested):
Thanks for taking this one @Kn99HN!

pymongo/cursor.py
Kn99HN: Thank you for pointing this out!

ShaneHarvey (approved):

ShaneHarvey:
Thanks @Kn99HN!

PR #595, opened 2021-04-20
PYTHON-2584 Use large localThresholdMS to fix test_load_balancing

ShaneHarvey:
Tested here: https://evergreen.mongodb.com/version/607a1187d1fe0724ae3b6f3c

[no ci]

### DIFF
--- a/test/test_server_selection_in_window.py
+++ b/test/test_server_selection_in_window.py
@@ -21,7 +21,7 @@ from pymongo.common import clean_node
 from pymongo.read_preferences import ReadPreference
 from test import client_context, IntegrationTest, unittest
 from test.utils_selection_tests import create_topology
-from test.utils import TestCreator, rs_client, OvertCommandListener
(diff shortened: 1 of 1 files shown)


test/test_server_selection_in_window.py
```diff
@@ -127,10 +127,13 @@ def frequencies(self, client, listener):
     @client_context.require_multiple_mongoses
     def test_load_balancing(self):
         listener = OvertCommandListener()
+        # Use a large localThresholdMS to avoid the impact of varying RTTs.
```
prashantmital, line 130: Minor nit: can we add the ticket number here? The ticket description has a great summary of what the exact problem is. Having it in the comment is a bit friendlier than need to check the blame in my book...

prashantmital (approved):

test/test_server_selection_in_window.py
ShaneHarvey: Done.

PR #602, opened 2021-04-23
PYTHON-2605: Improve mongodb+srv:// error message when dnspython is not installed

Kn99HN:
# Description
- Generate installation instruction for user to copy paste when dnspython is not installed. 

# Note for reviewers 
- I wasn't sure about getting path for Python so I think this is the most straightforward but having to import `os` only for the path seems a bit excessive. But let me know if you would like me to change it :)

### DIFF
--- a/pymongo/uri_parser.py
+++ b/pymongo/uri_parser.py
@@ -16,7 +16,7 @@
 """Tools to parse and validate a MongoDB URI."""
 import re
 import warnings
-import os
+import sys
 
 from urllib.parse import unquote_plus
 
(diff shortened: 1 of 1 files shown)


pymongo/uri_parser.py
```diff
@@ -416,8 +417,13 @@ def parse_uri(uri, default_port=DEFAULT_PORT, validate=True, warn=False,
         scheme_free = uri[SCHEME_LEN:]
     elif uri.startswith(SRV_SCHEME):
         if not _HAVE_DNSPYTHON:
+            python_path = os.getenv('_', "python")
```
ShaneHarvey, line 420: Instead of `os.getenv` let's use `sys.executable` like this:
```python
python_path = sys.executable or 'python'
```

https://docs.python.org/3/library/sys.html#sys.executable

pymongo/uri_parser.py
```diff
@@ -416,8 +417,13 @@ def parse_uri(uri, default_port=DEFAULT_PORT, validate=True, warn=False,
...
             raise ConfigurationError('The "dnspython" module must be '
-                                     'installed to use mongodb+srv:// URIs')
+                                     'installed to use mongodb+srv:// URIs. '
+                                     'To fix this error install pymongo '
+                                     'with the srv extra:\n   '
+                                     '%s -m pip install "pymongo[srv]"'
+                                     % (python_path))
```
ShaneHarvey, line 426: We could make this more readable by formatting like this:
```python
            raise ConfigurationError(
                'The "dnspython" module must be installed to use '
                'mongodb+srv:// URIs. To fix this error...'
```

ShaneHarvey (changes requested):

pymongo/uri_parser.py
Kn99HN: Thanks for this!

ShaneHarvey (approved):
Thanks @Kn99HN!

PR #603, opened 2021-04-23
PYTHON-2141: PyMongo's URI parser should be more permissive when handling boolean values

Kn99HN:
# Description
- Allow `validate_boolean_or_string` to accept `'True'` and `'False'` in addition to `'true'` and `'false'`

### DIFF
--- a/pymongo/common.py
+++ b/pymongo/common.py
@@ -162,12 +162,12 @@ def validate_boolean(option, value):
 
 
 def validate_boolean_or_string(option, value):
-    """Validates that value is True, False, 'true', or 'false'."""
+    """Validates that value is True, False, 'True', 'False', 'true', 'false'."""
     if isinstance(value, str):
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Thanks but we're not sure we want to make this change yet. The issue is that URIs are supposed to be portable across all the langauges we support (python, node, java, etc...) and I don't know if all the other langauges support "True" or "False". I'll  consult the drivers team and get back to you.

Kn99HN:
Thanks for letting me know! I can close this PR if you'd like

ShaneHarvey:
Thanks @Kn99HN but I couldn't convince the other teams to make this change. Boolean values in URIs will need to continue to use lowercase "true" and "false" which is portable across different language APIs.

Kn99HN:
Makes sense. Thanks so much for letting me know!

PR #605, opened 2021-04-23
PYTHON-2533 Add support for sample_rate and filter in set_profiling_level

ShaneHarvey:
See the docs here: https://docs.mongodb.com/manual/reference/command/profile/

`sampleRate` was added in 3.6.
`filter` was added in 4.4.2.

Filter is a bit odd in that it accepts a query document OR the string "unset" which resets and disables the filter. For this reason I decided not to perform client-side type validation on the filter parameter.

I also decided not to add `**kwargs`. Adding `**kwargs` let's users do strange things like pass arbitrary command arguments (txnNumber, lsid, writeConcern, etc..) and makes the code more confusing since we then need to define (or at least think about) precedence in cases like this: `db.set_profiling_level(ALL, slow_ms=50, slowms=55)`

### DIFF
--- a/pymongo/database.py
+++ b/pymongo/database.py
@@ -914,7 +914,8 @@ class Database(common.BaseObject):
         assert result["was"] >= 0 and result["was"] <= 2
         return result["was"]
 
-    def set_profiling_level(self, level, slow_ms=None, session=None):
+    def set_profiling_level(self, level, slow_ms=None, session=None,
+                            sample_rate=None, filter=None):
         """Set the database's profiling level.
 
         :Parameters:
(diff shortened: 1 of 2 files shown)


prashantmital (approved):
Changes LGTM but given that there is no driver spec for this command, is this something we want to remove in PyMongo 4.0? If so, can we:
- Change the base branch to merge this change to `v3.12`
- Add a deprecation warning to this helper either here or open a ticket to deprecate it in 3.12
- Open a ticket to remove the profiling helper in PyMongo 4.0

ShaneHarvey:
I opened https://jira.mongodb.org/browse/PYTHON-2662 to investigate the deprecation idea.

PR #607, opened 2021-04-25
PYTHON-2656 Mongodb allows storing negative Dates, but pymongo cannot…

seamys:
… parse this value


ShaneHarvey:
Thanks @seamys, I'm closing this PR since we are planning to use a different approach to fix this issue. 

Just a note, pymongo does not support negative years but we do support negative milliseconds between datetime.datetime.min (year 1) and the UNIX epoch (year 1970). This change would break applications that use dates in this range.

PR #609, opened 2021-04-26
PYTHON-2635 Unpin on abortTransaction

ShaneHarvey:
Resolves the following related tickets:
- PYTHON-2635 Test that session is unpinned when executing a non-transaction operation
- PYTHON-2654 Add unpinning transaction test for read operations
- PYTHON-2657 Unpin sessions after all abortTransaction attempts

Note we already unpin on any retryable or transient errors so the only thing left to do in pymongo is to unpin before returning from the abort_transaction() method.

This change also introduces `unified/` spec tests for transactions.


ShaneHarvey:
Note the one test failure is unrelated to this change and is already a known (rare) issue that appears to be back from the grave: https://jira.mongodb.org/browse/PYTHON-2279

test/test_transactions.py
```diff
@@ -474,5 +477,12 @@ def run_scenario(self):
             TestTransactionsConvenientAPI.TEST_PATH).create_tests()
 
 
+UNIFIED_TEST_PATH = os.path.join(
```
prashantmital, line 480: Cruft?

prashantmital (approved):

test/test_transactions.py
ShaneHarvey: Fixed.

PR #610, opened 2021-04-27
PYTHON-2629 Use hello command when API Version is declared

ShaneHarvey:


pymongo/pool.py
```diff
@@ -536,7 +536,10 @@ def ismaster(self, all_credentials=None):
 
     def _ismaster(self, cluster_time, topology_version,
                   heartbeat_frequency, all_credentials):
-        cmd = SON([('ismaster', 1)])
+        if self.opts.server_api:
```
prashantmital, line 539: Does server_api get auto populated with a value above some `max_wire_version`? I am a bit confused as to how/if this change actually causes us to the hit the `hello` failPoint set in the updated spec tests?

ShaneHarvey: No, server_api is only set when it's passed explicitly to MongoClient. We have a test suite job that runs the entire test suite with server_api enabled. That is when we hit the "hello" failpoint. Grep for MONGODB_API_VERSION if you're curious.

ShaneHarvey (commented):
Note, I'm going to make some changes to the implementation heres so please hold of on reviewing for now. I'll re-request a review when it's ready.

pymongo/pool.py
prashantmital: Ah, forgot about that. Thanks for clarifying!

PR #616, opened 2021-05-04
PYTHON-2396 Deprecate ssl_keyfile and ssl_certfile URI options

prashantmital:
Should I create a new section in the PyMongo 4 migration guide for deprecated/remove URI options?

### DIFF
--- a/doc/examples/tls.rst
+++ b/doc/examples/tls.rst
@@ -120,25 +120,26 @@ PyMongo can be configured to present a client certificate using the
 
 If the private key for the client certificate is stored in a separate file,
 it should be concatenated with the certificate file. For example, to
-concatenate PEM-formatted certificate file ``cert.pem`` and keyfile ``key.pem``
-into a single file ``concat_client.pem``, on *nix systems users can run::
+concatenate a PEM-formatted certificate file ``cert.pem`` and a PEM-formatted
+keyfile ``key.pem`` into a single file ``combined.pem``, on *nix systems users
+can run::
 
-  $ cat key.pem cert.pem > concat_client.pem
+  $ cat key.pem cert.pem > combined.pem
 
 PyMongo can be configured with the concatenated certificate keyfile using the
 ``tlsCertificateKeyFile`` option::
 
   >>> client = pymongo.MongoClient('example.com',
   ...                              tls=True,
-  ...                              tlsCertificateKeyFile='/path/to/concat_client.pem')
+  ...                              tlsCertificateKeyFile='/path/to/combined.pem')
 
-Python supports providing a password or passphrase to decrypt encrypted
-private keys. Use the ``tlsCertificateKeyFilePassword`` option::
+If the private key contained in the certificate keyfile is encrypted, users
+can provide a password or passphrase to decrypt the encrypted private keys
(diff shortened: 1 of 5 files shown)


doc/examples/tls.rst
```diff
@@ -118,13 +118,19 @@ PyMongo can be configured to present a client certificate using the
...
+``tlsCertificateKeyFile`` option::
 
   >>> client = pymongo.MongoClient('example.com',
   ...                              tls=True,
-  ...                              tlsCertificateKeyFile='/path/to/client.pem',
-  ...                              ssl_keyfile='/path/to/key.pem')
+  ...                              tlsCertificateKeyFile='/path/to/concat_client.pem')
 
 Python supports providing a password or passphrase to decrypt encrypted
 private keys. Use the ``tlsCertificateKeyFilePassword`` option::
```
ShaneHarvey, line 136: Hmmm, what's the upgrade path for users of this example? Can we remove `ssl_keyfile` here?

ShaneHarvey (changes requested):
> Should I create a new section in the PyMongo 4 migration guide for deprecated/remove URI options?

Yes, you can add it under the MongoClient section.

Can you also replace Deprecate with Remove in the PR title?

doc/examples/tls.rst
prashantmital: I am not sure I follow? Users will need to concat the files...

prashantmital:
Ready for another look @ShaneHarvey. I have removed the migration notes as they will be added in https://jira.mongodb.org/browse/PYTHON-2162 
This change is intended to be merged into both v3.12 and master branches.

doc/examples/tls.rst
```diff
@@ -118,24 +118,30 @@ PyMongo can be configured to present a client certificate using the
   ...                              tls=True,
   ...                              tlsCertificateKeyFile='/path/to/client.pem')
 
-If the private key for the client certificate is stored in a separate file use
-the ``ssl_keyfile`` option::
+If the private key for the client certificate is stored in a separate file,
+it should be concatenated with the certificate file. For example, to
+concatenate a PEM-formatted certificate file ``cert.pem`` and a PEM-formatted
+keyfile ``key.pem`` into a single file ``combined.pem``, on *nix systems users
```
ShaneHarvey, line 124: `*` means italics is RST. Does this render? Also everywhere else in the docs we just say "Unix" so maybe just change to that?

ShaneHarvey (changes requested):
Please adjust the test suite to ignore deprecation warnings when using the old option names.

doc/examples/tls.rst
prashantmital: Switched to Unix.

prashantmital:
I've ignored deprecations in `test_ssl` but noticed that we were ignorning already deprecated options that were in that file like `ssl_cert_reqs` so I am not sure why we are doing so specifically for these newly deprecated options? 
Is the intention to ignore the deprecations in all other test modules other than SSL?

ShaneHarvey (approved):
Not sure I fully understand your question. In general we should use `ignore_deprecations` whenever we're testing a deprecating feature otherwise the test output is cluttered with DeprecationWarnings.

PR #619, opened 2021-05-11
Deprecate profiler helpers

prashantmital:
Patch: https://spruce.mongodb.com/version/6099cc480305b97fd5038662/


prashantmital:
See the link in the description. We have no default tasks configured for the PR integration on 3.12 so it always says that AFAICT

ShaneHarvey:
Ah I tracked down the issue. We changed the EVG tasks such that the project's Github PR regex no longer matched any tasks. I've updated it to match some 3.9+2.7 tasks on amazon1-2018.

PR #620, opened 2021-05-11
PYTHON-2666 Unified test runner should not close the global client

ShaneHarvey:
I suspect this will fix the "error: can't start new thread" we're seeing on SUSE 12 with Python 2.7.


ShaneHarvey:
This doesn't fix the "can't start new thread" errors. Closing.

PR #623, opened 2021-05-13
PYTHON-2676 Unified Test Runner changes in preparation for Load Balancer Support

ShaneHarvey:
This PR contains Unified Test Runner changes in preparation for Load Balancer Support, including:
- resyncing various spec tests
- Adding support for CRUD unified tests and removing crud_v2.
- Splitting out the transactions unified tests into test_transactions_unified.
- PYTHON-2348 Correctly express lack of event assertions in change stream tests


ShaneHarvey:
I think this is in a good state to review. This PR contains Unified Test Runner changes in preparation for Load Balancer Support, including:
- resyncing various spec tests
- Adding support for CRUD unified tests and removing crud_v2.
- Splitting out the transactions unified tests into test_transactions_unified.
- PYTHON-2348 Correctly express lack of event assertions in change stream tests

test/test_change_stream.py
```diff
@@ -1205,7 +1205,7 @@ def run_scenario(self):
         finally:
             # Check for expected events
             results = self.listener.results
-            for idx, expectation in enumerate(test.get("expectations", [])):
+            for idx, expectation in enumerate(test.get("expectations") or []):
```
prashantmital, line 1208: Ah, this is subtle. Can you add a comment saying `expectations` key can be a list or `None`/`null` when there are no assertions?

test/unified_format.py
```diff
@@ -440,7 +440,11 @@ def _match_document(self, expectation, actual, is_root):
         if self._evaluate_if_special_operation(expectation, actual):
             return
 
-        self._test_class.assertIsInstance(actual, abc.Mapping)
+        try:
+            self._test_class.assertIsInstance(actual, abc.Mapping)
+        except AssertionError:
+            print()
```
prashantmital, line 446: What's going on here?

test/unified_format.py
```diff
@@ -688,34 +706,11 @@ def _databaseOperation_aggregate(self, target, *args, **kwargs):
...
     def _collectionOperation_find(self, target, *args, **kwargs):
         self.__raise_if_unsupported('find', target, Collection)
         find_cursor = target.find(*args, **kwargs)
         return list(find_cursor)
 
-    def _collectionOperation_findOneAndReplace(self, target, *args, **kwargs):
-        self.__raise_if_unsupported('findOneAndReplace', target, Collection)
-        return target.find_one_and_replace(*args, **kwargs)
-
-    def _collectionOperation_findOneAndUpdate(self, target, *args, **kwargs):
```
prashantmital, line 705: Hmm not sure why I had these wrappers in place... They don't seem to do anything. 
I think it might have been just because these ops are explicitly called out in the Unified Test Spec but not sure. Thanks for removing.

test/unified_format.py
```diff
@@ -511,6 +515,25 @@ def match_event(self, expectation, actual):
                 'Unsupported event type %s' % (event_type,))
 
 
+def coerce_result(opname, result):
```
prashantmital, line 518: Should this be a method of `MatchEvaluatorUtil` class?

test/utils_selection_tests.py
```diff
@@ -59,7 +59,7 @@ def make_last_write_date(server):
 def make_server_description(server, hosts):
     """Make a ServerDescription from server info in a JSON test."""
     server_type = server['type']
-    if server_type == "Unknown":
+    if server_type in ("Unknown", "PossiblePrimary"):
```
prashantmital, line 62: Which spec change is this coming from?

prashantmital (changes requested):

test/test_change_stream.py
ShaneHarvey: Done.

test/unified_format.py
ShaneHarvey: Debug code. Removed.

test/unified_format.py
ShaneHarvey: I see no benefit to move it there.

test/unified_format.py
ShaneHarvey: Yeah I see no reason to keep them either.

test/utils_selection_tests.py
ShaneHarvey: The `server_selection/ReplicaSetNoPrimary/read/PossiblePrimary.json` and `server_selection/ReplicaSetNoPrimary/read/PossiblePrimaryNearest.json` tests which have exists forever. Previously we just removed those test files manually after resyncing. Pymongo doesn't actually support a PossiblePrimary and doesn't need to support it. Now we run the tests because it's easy enough to convert PossiblePrimary to Unknown.

TLDR: this makes it easier to resync the server-selection spec tests.

prashantmital (approved):

PR #627, opened 2021-05-24
Added dnspython to the install_requires

NeilShah2026:


ShaneHarvey:
Closing since we intentionally do not have a hard dependency here. dnspython is an optional dependency for users that want to use "mongodb+srv://" URIs. To install pymongo with dnspython you can use the "srv" extra:
```
$ python -m pip install 'pymongo[srv]'
```
You can also put `pymongo[srv]` in a requirements.txt file.

https://pymongo.readthedocs.io/en/stable/installation.html#dependencies

prashantmital:
Thanks for your pull request @NeilShah2026 
PyMongo is already installable with `dnspython` using the `srv` extra as detailed [here](https://pymongo.readthedocs.io/en/stable/installation.html#dependencies):
```
$ pip install 'pymongo[srv]'
```

PyMongo purposefully doesn't have any mandatory/strict dependencies since many of our users operate in environments where installing packages from PyPI is either not possible or adds a lot of overhead (this is most common in security-sensitive industries). Consequently, we list `dnspython` as an optional extra that can be installed either manually or using the snippet supplied above.

NeilShah2026:
Oh, sorry, did not know.

PR #628, opened 2021-05-24
PYTHON-2674 Pool.reset only clears connections to the given serviceId

ShaneHarvey:
Maintain individual generations for each serviceId.


ShaneHarvey:
I just had an idea which simplify the _PoolGeneration class a good amount. Please hold off reviewing this until I update.

pymongo/pool.py
```diff
@@ -1044,6 +1046,35 @@ class _PoolClosedError(PyMongoError):
...
+        # Overall pool generation.
+        self._generations[None] = 0
+
+    def get(self, service_id):
+        """Get the generation for the given service_id."""
+        return self._generations[service_id]
+
+    def get_overall(self):
+        """Get the Pool's overall generation."""
+        return self._generations[None]
```
ShaneHarvey, line 1062: The simplification I was going to make was to combine the `get` and `get_overall` methods since `get(None)` is equivalent to `get_overall()` but I decided that actually makes more cryptic. Instead I think we should keep them separate.  `get_overall()` is called to get the "overall pool generation" before a connection is created. `get(service_id)` is called after a service_id is known (eg in SocketInfo._ismaster and Pool.stale_generation).

prashantmital: When we are in LB mode what is the significance of the overall generation ID? Aren't generation numbers only relevant on a per-service ID basis in that case?

pymongo/pool.py
```diff
@@ -1044,6 +1046,35 @@ class _PoolClosedError(PyMongoError):
     pass
 
 
+class _PoolGeneration(object):
+    def __init__(self):
+        # Maps service_id to generation.
+        self._generations = collections.defaultdict(int)
+        # Overall pool generation.
+        self._generations[None] = 0
```
prashantmital, line 1054: Can we use a better sentinel than `None` for tracking overall generation ID? Does this even need to be in the `_generations` map? I think we can just track it as an integer class attribute.

prashantmital (changes requested):

pymongo/pool.py
prashantmital: Thanks.

pymongo/pool.py
prashantmital: LGTM!

prashantmital (approved):

PR #630, opened 2021-05-27
PYTHON-2673 Connection pinning behavior for load balanced clusters

ShaneHarvey:
- [X] Connection pinning for transactions. In a transaction the first connection checked out of the pool becomes pinned to the session. The session follows the normal rules for unpinning outlined in the transaction spec.
- [X] Connection pinning for find cursors (outside a transaction). The connection gets pinned using the "SocketManager" class.
- [x] Connection pinning for CommandCursors (EG aggregate, list collections, etc..). Should work similarly to the Cursor class but the initial aggregate command uses a different execution path.

### DIFF
--- a/pymongo/aggregation.py
+++ b/pymongo/aggregation.py
@@ -161,11 +161,13 @@ class _AggregationCommand(object):
             }
 
         # Create and return cursor instance.
-        return self._cursor_class(
+        cmd_cursor = self._cursor_class(
             self._cursor_collection(cursor), cursor, sock_info.address,
             batch_size=self._batch_size or 0,
             max_await_time_ms=self._max_await_time_ms,
             session=session, explicit_session=self._explicit_session)
+        cmd_cursor._maybe_pin_connection(sock_info)
+        return cmd_cursor
 
 
 class _CollectionAggregationCommand(_AggregationCommand):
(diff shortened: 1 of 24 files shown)


ShaneHarvey:
This is ready for review. Note there are some minor TODO left in the code which I will work on resolving next week. Please ignore them for now.

ShaneHarvey:
Updated to significantly simplify the ClientSession connection pinning and unpinning logic (hopefully). The session._pin() method is now only called in one place.

I also fixed an oversight where command events were not being published with the service_id field and fixed the hasServiceId unified test runner logic.

Finally I've resolved all the remaining misc. TODOs.

ShaneHarvey:
Updated to workaround the CPython GC bug described in https://jira.mongodb.org/browse/PYTHON-2751. New patch:
https://spruce.mongodb.com/version/60c2ba243e8e8642a9d2e319/tasks

prashantmital (approved):

ShaneHarvey:
Merged in two commits:
- The first resyncs the spec tests: https://github.com/mongodb/mongo-python-driver/commit/7a48831124015d5636f5478817727a7cd4be8d8c
- The second has the actually code changes and spec test tweaks: https://github.com/mongodb/mongo-python-driver/commit/c8f32a7a376e027cbe30dcb8df9bf05778e61248

PR #633, opened 2021-05-28
PYTHON-2743 Fix compatibility with gevent.Timeout

TylerWilley:
gevent.Timeout extends BaseException, so we still sometimes lose the lock.

Perhaps we should instead have this as an empty except?


pymongo/pool.py
```diff
@@ -1415,7 +1415,7 @@ def _get_socket(self, all_credentials):
                             self._pending -= 1
                             self._max_connecting_cond.notify()
             sock_info.check_auth(all_credentials)
-        except Exception:
+        except BaseException:
```
ShaneHarvey, line 1418: This doesn't really make sense to me. Sure we now catch gevent.Timeout, but won't gevent again raise _another_ timeout when we try to acquire the lock below on line 1422? In that case we might as well have not caught the exception anyway.

TylerWilley: You know what, I was on the old 3.11 branch...

The latest master uses contexts instead for acquiring / releasing, which should prevent the lock from remaining acquired... Let me try reproducing on master, and if it's fixed on master, we can just use an RC build from there.

TylerWilley: (then we can just close this)

TylerWilley: So as followup, it still causes problems on 3.11.4 because the except block isn't executed. `BaseException`, or a bare `except:` still fix this, because we don't nest timeouts currently.

Considering a worst case scenario where you have infinite timeouts waiting to fire, you would have a timeout fire at each:
- close_socket
- with self.size_cond
-  self.size_cond.notify (not 100%, this may not block and therefore not be an eligible point to fire a timeout)
- event_listeners.publish_connection_check_out_failed, if the listener executes blocking code that would call the hub's switch() method

Given we only have 1 active at a time, this isn't an issue for us in particular, but would be if someone did use stacked timeouts and they all fired in a worst-case scenario.

TylerWilley: It seems like it might be necessary to subclass Pool as GeventPool, I see that custom Pool classes can be provided, so that's what I'll try next. Thanks for being responsive though :)

TylerWilley:
Ok, hopefully this makes sense....

Instead I simply sub-classed Pool, as GeventPool. When the exception handler is entered, I start a greenlet to run the exception handling. This makes the exception handling code un-interruptible from gevent's perspective so that even multiple successive Timeouts being raised in application code will not prevent releasing the resources properly.

I don't know if you'd want to integrate this somehow... But since I can put this into our application as-is without updating pymongo using the `_pool_class=GeventPool` argument to the MongoClient, it's not necessary.

If you don't, we can close this. Thanks!

TylerWilley:
https://www.gevent.org/api/gevent.timeout.html

see the `Caution` in the `Use As A Context Manager` section

TylerWilley:
I could also see _pool_class being useful for implementing an asyncio pool to run queries from as well, which would be the more built-in way of running coroutines in python.

ShaneHarvey:
Thanks for the link. While I do think the thread-cleanup idea is clever, I am not a fan of adding a special pool class just for Gevent support. Looking into this more, in [PYTHON-1663](https://jira.mongodb.org/browse/PYTHON-1663) (PyMongo 3.8) the pool was changed from catching all exceptions using a bare-except to catching only Exception:
```
-    except:
+    except Exception:
```
I am open to changing it back and adding a comment on why we have to use bare except so the next maintain doesn't accidentally make the same mistake. This would at least deal with the single Timeout issue while we consider how (and if) we want to support multiple timeouts.

agolin95:
Tracked in: https://jira.mongodb.org/browse/PYTHON-2743

TylerWilley:
I added a unit test that will break if we add Exception back to the except clause.

In order for it to function, the Timeout has to raise at a specific point, I used mock to do this, I don't know if you have a more standard "pymongo" way of doing that... It was previously raising at the condition check prior to entering the try, thus not incrementing / decrementing active_count at all.

ShaneHarvey:
Thanks @TylerWilley! I've merged this fix in https://github.com/mongodb/mongo-python-driver/commit/9c1ff6ad9da88692014848a5fed01f964138de4d and https://github.com/mongodb/mongo-python-driver/commit/a2d687b4bb2110c868d406a03734e1a18bf8fa5e.

TylerWilley:
Thank you @ShaneHarvey ! :)

PR #634, opened 2021-06-07
PYTHON-2734 Document that find_raw_batches now sends user-specified R…

prashantmital:
…eadConcern to the server instead of raising ConfigurationError

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -1485,7 +1485,7 @@ class Collection(common.BaseObject):
 
         .. versionchanged:: 3.12
            Instead of ignoring the user-specified read concern, this method
-           now sends it to the server when connected to MongoDB 3.2+.
+           now sends it to the server when connected to MongoDB 3.6+.
 
         .. versionadded:: 3.6
         """
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -1484,8 +1484,8 @@ class Collection(common.BaseObject):
            encryption.
 
         .. versionchanged:: 3.12
-           Instead of raising ConfigurationError, this method now sends the
-           user-specified read concern to the server.
+           Instead of ignoring the user-specified read concern, this method
+           now sends it to the server.
 
         .. versionadded:: 3.6
         """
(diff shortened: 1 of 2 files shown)


pymongo/collection.py
```diff
@@ -1483,6 +1483,10 @@ def find_raw_batches(self, *args, **kwargs):
         .. note:: find_raw_batches does not support sessions or auto
            encryption.
 
+        .. versionchanged:: 3.12
+           Instead of raising ConfigurationError, this method now sends the
```
ShaneHarvey, line 1487: `Instead of raising ConfigurationError` did we actually raise ConfigurationError before? I thought that the read concern was just ignored.

prashantmital: Ah my bad. Got confused while writing this.

pymongo/collection.py
```diff
@@ -1483,6 +1483,10 @@ def find_raw_batches(self, *args, **kwargs):
         .. note:: find_raw_batches does not support sessions or auto
            encryption.
 
+        .. versionchanged:: 3.12
+           Instead of ignoring the user-specified read concern, this method
+           now sends it to the server.
```
ShaneHarvey, line 1488: "when connected to MongoDB 3.6+"?

ShaneHarvey (changes requested):

pymongo/collection.py
prashantmital: Its actually 3.2+ AFAICT. changed.

ShaneHarvey: Can you explain? Judging from the test failures it's 3.6+: https://evergreen.mongodb.com/build/mongo_python_driver_tests_storage_engines__platform~awslinux_python_version~3.6_storage_engine~mmapv1_c36ec37281be1d663f8eba1169b29b09fffef451_21_06_07_19_12_49

prashantmital: Never mind, you are right... Its only when OP_MSG is available.

prashantmital:
Ok to merge this? @ShaneHarvey

ShaneHarvey (approved):

PR #635, opened 2021-06-07
Backport/python 2727

prashantmital:


ShaneHarvey (approved):

prashantmital:
Python 3.4 failures are being caused due to the fact that Ubuntu 18 does not have Python 3.4 installed. Some older pypy versions are also having trouble building the cryptography module which is causing OCSP + PyPy tests to fail.

PR #637, opened 2021-06-08
PYTHON-2718 Test redaction of security sensitive command monitoring events

prashantmital:
Ready for review. Failures are unrelated to the change.


prashantmital:
@ShaneHarvey `monitoring._SENSITIVE_COMMANDS` is used in many places as a source of truth for what is a sensitive command (e.g. in `compression_support.py`, in `test_auth.py`). Should I create a new method that accepts command name + payload and returns a boolean if the command is sensitive to use in these places instead?

test/unified_format.py
```diff
@@ -124,8 +124,16 @@ def is_run_on_requirement_satisfied(requirement):
             elif client_context.server_parameters[param] != val:
                 params_satisfied = False
 
+    auth = requirement.get('auth')
```
ShaneHarvey, line 127: I already made this change. Please revert and rebase.

test/test_command_monitoring_spec.py
```diff
@@ -239,5 +241,11 @@ def create_tests():
 
 create_tests()
 
+
+globals().update(generate_test_classes(
+    os.path.join(_TEST_PATH, 'unified'),
+    module=__name__,))
```
ShaneHarvey, line 247: Please move this to a new file `test/test_command_monitoring_unified.py` and add a corresponding `test/load_balancer/test_command_monitoring_unified.py`.

ShaneHarvey (commented):
> Should I create a new method that accepts command name + payload and returns a boolean if the command is sensitive to use in these places instead?

The tests are all passing now correct? If so, I don't see a reason to make that change.

test/unified_format.py
prashantmital: 👍🏽

test/test_command_monitoring_spec.py
prashantmital: Ah... Thanks for pointing this out. Done.

ShaneHarvey (approved):
LGTM!

PR #639, opened 2021-06-11
PYTHON-2677 Better wait queue timeout errors for load balanced clusters

ShaneHarvey:
See https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#connection-tracking

### DIFF
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -551,6 +551,14 @@ class SocketInfo(object):
         self.pinned_txn = False
         self.pinned_cursor = False
 
+    def pin_txn(self):
+        self.pinned_txn = True
+        assert not self.pinned_cursor
+
+    def pin_cursor(self):
+        self.pinned_cursor = True
+        assert not self.pinned_txn
+
     def unpin(self):
         pool = self.pool_ref()
         if pool:
@@ -1356,7 +1364,7 @@ class Pool:
         return sock_info
 
     @contextlib.contextmanager
-    def get_socket(self, all_credentials, checkout=False, handler=None):
+    def get_socket(self, all_credentials, handler=None):
         """Get a socket from the pool. Use with a "with" statement.
 
         Returns a :class:`SocketInfo` object wrapping a connected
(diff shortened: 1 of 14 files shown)


pymongo/pool.py
```diff
@@ -1378,7 +1380,10 @@ def get_socket(self, all_credentials, checkout=False, handler=None):
 
         sock_info = self._get_socket(all_credentials)
         if checkout:
-            self.__pinned_sockets.add(sock_info)
+            with self.lock:
+                sock_info.pinned_cursor = True
+                self.__pinned_sockets.add(sock_info)
+                self.ncursors += 1
```
prashantmital, line 1386: Can you explain how `checkout=True` implies that the connection is for a cursor? Why can't it be for a txn?

ShaneHarvey: When a connection is pinned for a cursor operation, we always set checkout=True ahead of time. When a connection is pinned for a transaction, we never set checkout=True. Instead the checkout happens normally, then ClientSession.pin() does `sock_info.pinned_txn = True` to indicate that the session has been pinned and should not be checked back into the pool.  Does that seem reasonable?

prashantmital: Though it obviously works, I find it particularly hard to follow... Can we change how we do checkouts in general?
How about adding a `SockInfo.checkout(reason=SocketCheckoutReason.Unknown)` method to sock info which we can call after getting a socket from the pool that we don't intend to check back in. `SocketCheckoutReason` can be an enum that specifies the reason for the checkout which you can use for tracking sockets for txns/cursors/other ops... 
If you think it'll be too much work, we can leave it be. I think it'll be great to do it because the code is really hard to follow right now and its not clear what the various flags like `pinned_txn` are being used for exactly...

ShaneHarvey: Updated to remove the "checkout" flag. Instead, the socket is pinned after being checked out like this:
```python
with self.pool.get_socket({}) as sock:
    sock.pin_cursor()  # socket will not be checked back in by get_socket.
sock.unpin()  # Returned to the pool.
```
Or for transactions:
```python
with self.pool.get_socket({}) as sock:
    sock.pin_txn()  # socket will not be checked back in by get_socket.
sock.unpin()  # Returned to the pool.
```

prashantmital (approved):

PR #646, opened 2021-06-22
PYTHON-2572 Introduce NotPrimaryError and deprecate NotMasterError

prashantmital:
We _could_ have raised a DeprecationWarning in `NotMasterError.__init__` if we didn't need to continue supporting application code that uses `NotMasterError` - specifically, we want `except NotMasterError` to catch raising `NotPrimaryError` which requires that `NotPrimaryError` inherit from `NotMasterError`. Consequently, every time `NotPrimaryError` is instantiated it would end up raising `NotMasterError`'s `DeprecationWarning`.

### DIFF
--- a/pymongo/errors.py
+++ b/pymongo/errors.py
@@ -106,9 +106,6 @@ class NotMasterError(AutoReconnect):
     Deprecated. Use :exc:`~pymongo.errors.NotPrimaryError` instead.
     """
     def __init__(self, message='', errors=None):
-        warnings.warn(
-            "NotMasterError is deprecated. Use NotPrimaryError instead.",
-            DeprecationWarning, stacklevel=2)
         super(NotMasterError, self).__init__(
             _format_detailed_error(message, errors), errors=errors)
 
@@ -129,8 +126,7 @@ class NotPrimaryError(NotMasterError):
     .. versionadded:: 3.12
     """
     def __init__(self, message='', errors=None):
-        super(NotPrimaryError, self).__init__(
-            _format_detailed_error(message, errors), errors=errors)
+        super(NotPrimaryError, self).__init__(message, errors=errors)
 
 
 class ServerSelectionTimeoutError(AutoReconnect):
--- a/pymongo/errors.py
+++ b/pymongo/errors.py
@@ -14,6 +14,8 @@
 
 """Exceptions raised by PyMongo."""
 
+import warnings
+
 from bson.errors import *
 
 try:
@@ -104,6 +106,9 @@ class NotMasterError(AutoReconnect):
     Deprecated. Use :exc:`~pymongo.errors.NotPrimaryError` instead.
     """
     def __init__(self, message='', errors=None):
+        warnings.warn(
+            "NotMasterError is deprecated. Use NotPrimaryError instead.",
+            DeprecationWarning, stacklevel=2)
         super(NotMasterError, self).__init__(
             _format_detailed_error(message, errors), errors=errors)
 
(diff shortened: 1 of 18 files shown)


pymongo/pool.py
```diff
@@ -770,13 +770,13 @@ def receive_message(self, request_id):
             self._raise_connection_failure(error)
 
     def _raise_if_not_writable(self, unacknowledged):
-        """Raise NotMasterError on unacknowledged write if this socket is not
+        """Raise NotPrimaryError on unacknowledged write if this socket is not
         writable.
         """
         if unacknowledged and not self.is_writable:
             # Write won't succeed, bail as if we'd received a not master error.
```
ShaneHarvey, line 777: not master  -> not primary

test/test_connections_survive_primary_stepdown_spec.py
```diff
@@ -96,7 +96,7 @@ def test_get_more_iteration(self):
         # notMaster error on the subsequent operation.
```
ShaneHarvey, line 96: notMaster -> not primary

pymongo/errors.py
```diff
@@ -14,6 +14,8 @@
 
 """Exceptions raised by PyMongo."""
 
+import warnings
```
ShaneHarvey, line 17: Not needed.

test/test_errors.py
```diff
@@ -20,19 +20,19 @@
...
 from pymongo.errors import (BulkWriteError,
                             EncryptionError,
-                            NotMasterError,
+                            NotPrimaryError,
                             OperationFailure)
 from test import (PyMongoTestCase,
                   unittest)
 
 
 class TestErrors(PyMongoTestCase):
```
ShaneHarvey, line 29: Please add a test that raises NotMasterError and excepts NotPrimaryError.

ShaneHarvey (changes requested):
Is there already a ticket for 4.0 to remove NotMasterError? If not can you create one?

test/test_errors.py
prashantmital: Done.

prashantmital:
> Is there already a ticket for 4.0 to remove NotMasterError? If not can you create one?

Yes. [PYTHON-2571](https://jira.mongodb.org/browse/PYTHON-2571).

@ShaneHarvey do we need to do anything for places where 'not master' is hardcoded? E.g.: 
- https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/message.py#L1539
- https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/helpers.py#L155
- https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/helpers.py#L187

test/test_errors.py
```diff
@@ -100,6 +101,13 @@ def test_pickle_EncryptionError(self):
         self.assertPyMongoErrorEqual(exc, exc2)
         self.assertOperationFailureEqual(cause, exc2.cause)
 
+    def test_NotMasterError_catches_NotPrimaryError(self):
+        def callback():
+            raise NotPrimaryError("not primary test", {"errmsg": "error"})
+        with self.assertRaises(NotMasterError) as exc:
+            callback()
```
ShaneHarvey, line 108: Nit: there's no need for the `callback`.

ShaneHarvey (approved):
LGTM. 

> do we need to do anything for places where 'not master' is hardcoded?

No, those are intentional and required for compatibility with older server versions.

PR #647, opened 2021-06-22
PYTHON-2737 Run Load Balancer test suite with Python 3.4

ShaneHarvey:


behackett (approved):

prashantmital:
Looks like HAProxy might not be on the path if it is even installed on these hosts?

ShaneHarvey:
Sorry I opened this just to test if the BUILD change was deployed. It is not so we can't merge this just yet.

PR #649, opened 2021-06-22
PYTHON-2762 Avoid duplicating unified test files for LB testing

ShaneHarvey:


ShaneHarvey:
Failures are because Ubuntu 18 no longer has <3.6 Pythons on it. We're once again blocked on the build team installing haproxy on Amazon Linux.

ShaneHarvey:
I worked around the PyPy test failures by creating a new client for each of the failing tests above. Test are now green.

test/unified_format.py
```diff
@@ -635,6 +635,7 @@ class UnifiedSpecTestMixinV1(IntegrationTest):
     a class attribute ``TEST_SPEC``.
     """
     SCHEMA_VERSION = Version.from_string('1.5')
+    RUN_ON_LOAD_BALANCER = True
```
prashantmital, line 638: Is it certain that all Unified format tests added in the future will need to run against LoadBalanced topologies? If not, we might have a problem when some unified format tests have to not be run against LB topologies with this pattern. One way to keep this configurable would be to add a kwarg to `generate_test_classes` that allows us to set `RUN_ON_LOAD_BALANCER=False` when desired.

prashantmital (commented):
With these changes, are we now running _more_ tests against LB topology than we were previously? IIUC all tests in the suite will be run against LB clusters unless explicitly decorated with `require_no_load_balancer`. This seems to be a qualify a much larger set of tests to run against LB topologies than were previously present in the `tests/load_balancer` directory. 

Am I getting this right?

test/unified_format.py
ShaneHarvey: > Is it certain that all Unified format tests added in the future will need to run against LoadBalanced topologies?
Yes, see https://github.com/mongodb/specifications/tree/master/source/load-balancers/tests#tests.

ShaneHarvey (commented):
> With these changes, are we now running more tests against LB topology than we were previously? IIUC all tests in the suite will be run against LB clusters unless explicitly decorated with require_no_load_balancer. 

Not quite. The updated IntegrationTest class will skip all tests unless the RUN_ON_LOAD_BALANCER is explicitly set to True:
```python
class IntegrationTest(PyMongoTestCase):
    """Base class for TestCases that need a connection to MongoDB to pass."""
    @classmethod
    @client_context.require_connection
    def setUpClass(cls):
        if (client_context.load_balancer and
                not getattr(cls, 'RUN_ON_LOAD_BALANCER', False)):
            raise SkipTest('this test does not support load balancers')
```

So while it's true that we are running more tests now, the only extra tests we're running are unittests which don't communicate with the server.

prashantmital (approved):

PR #653, opened 2021-06-23
PYTHON-2765 Fix test_exhaust failure due to OP_MSG and __del__ changes

ShaneHarvey:
This test started failing because of the combination of OP_MSG exhaust (PYTHON-1636) and Cursor.__del__ changes (PYTHON-1272). With OP_MSG exhaust the more_to_come flag is not set until after the first getMore. If the exhaust cursor is closed before that happens then we don't have to close the connection. I've updated the test to ensure the first getMore runs so that the cursor will always be pinned and more_to_come will always be true.

### DIFF
--- a/test/test_collection.py
+++ b/test/test_collection.py
@@ -1851,29 +1851,33 @@ class TestCollection(IntegrationTest):
         client = rs_or_single_client(maxPoolSize=1)
         self.addCleanup(client.close)
         pool = get_pool(client)
-        socks = get_pool(client).sockets
 
         # Make sure the socket is returned after exhaustion.
         cur = client[self.db.name].test.find(cursor_type=CursorType.EXHAUST)
         next(cur)
-        self.assertEqual(0, len(socks))
+        self.assertEqual(0, len(pool.sockets))
         for _ in cur:
             pass
-        self.assertEqual(1, len(socks))
+        self.assertEqual(1, len(pool.sockets))
 
         # Same as previous but don't call next()
         for _ in client[self.db.name].test.find(cursor_type=CursorType.EXHAUST):
             pass
-        self.assertEqual(1, len(socks))
+        self.assertEqual(1, len(pool.sockets))
 
-        # If the Cursor instance is discarded before being
-        # completely iterated we have to close and
-        # discard the socket.
+        # If the Cursor instance is discarded before being completely iterated
+        # and the socket has pending data (more_to_come=True) we have to close
+        # and discard the socket.
         cur = client[self.db.name].test.find(cursor_type=CursorType.EXHAUST,
                                              batch_size=2)
-        for _ in range(3):
+        if client_context.version.at_least(4, 2):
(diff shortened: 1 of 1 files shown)


test/test_collection.py
```diff
@@ -1849,6 +1849,8 @@ def test_exhaust(self):
         self.db.test.insert_many([{'i': i} for i in range(150)])
 
         client = rs_or_single_client(maxPoolSize=1)
+        self.addCleanup(client.close)
+        pool = get_pool(client)
         socks = get_pool(client).sockets
```
prashantmital, line 1854: We can probably remove this now and just use `pool.sockets` everywhere?

prashantmital (changes requested):
> If the exhaust cursor is closed before that happens then we don't have to close the connection.

Is there as per the SDAM spec? Where is this behavior defined?

test/test_collection.py
ShaneHarvey: Done.

ShaneHarvey:
> Is there as per the SDAM spec? Where is this behavior defined?

There is no exhaust cursor spec so we define it ourselves. When an exhaust cursor has pending data on the connection when the cursor is closed or garbage collected we have to close and discard the connection. Otherwise the next operation using that connection will read the incorrect command response. 

I've update the comments in the test. PTAL.

ShaneHarvey:
The one test failure is an unrelated issue PYTHON-2534:
```
 [2021/06/23 19:13:06.383] FAIL [0.522s]: test_pool_paused_error_is_retryable (test_cmap.TestCMAP)
 [2021/06/23 19:13:06.383] ----------------------------------------------------------------------
 [2021/06/23 19:13:06.383] Traceback (most recent call last):
 [2021/06/23 19:13:06.383]   File "/data/mci/d8acf3d21e677f7d2af8a61e96111ea5/src/test/__init__.py", line 513, in wrap
 [2021/06/23 19:13:06.383]     return f(*args, **kwargs)
 [2021/06/23 19:13:06.383]   File "/data/mci/d8acf3d21e677f7d2af8a61e96111ea5/src/test/__init__.py", line 513, in wrap
 [2021/06/23 19:13:06.383]     return f(*args, **kwargs)
 [2021/06/23 19:13:06.383]   File "/data/mci/d8acf3d21e677f7d2af8a61e96111ea5/src/test/__init__.py", line 513, in wrap
 [2021/06/23 19:13:06.383]     return f(*args, **kwargs)
 [2021/06/23 19:13:06.383]   File "/data/mci/d8acf3d21e677f7d2af8a61e96111ea5/src/test/test_cmap.py", line 477, in test_pool_paused_error_is_retryable
 [2021/06/23 19:13:06.383]     2, cmap_listener.event_count(ConnectionCheckOutFailedEvent))
 [2021/06/23 19:13:06.383] AssertionError: 2 != 1
```

prashantmital (approved):

PR #654, opened 2021-06-23
PYTHON-2715 Use hello command for monitoring when supported

prashantmital:
We should wait for https://github.com/mongodb/specifications/pull/990 to be merged before merging this but the code changes can be reviewed.

Note that I have not yet synced the `legacy_hello` tests as they seem to still be in flux.


pymongo/pool.py
```diff
@@ -571,7 +572,9 @@ def hello_cmd(self):
         if self.opts.server_api:
             return SON([(HelloCompat.CMD, 1)])
         else:
-            return SON([(HelloCompat.LEGACY_CMD, 1)])
+            if self.hello_ok:
+                return SON([(HelloCompat.CMD, 1)])
```
ShaneHarvey, line 576: Suggest merging this with the server_api check above:
```
if self.opts.server_api or self.hello_ok:
...
```

pymongo/pool.py
```diff
@@ -571,7 +572,9 @@ def hello_cmd(self):
...
+            return SON([(HelloCompat.LEGACY_CMD, 1), ('helloOk', True)])
```
ShaneHarvey, line 577: Is helloOk only supposed to be sent with the initial handshake? If so we should move it to `_ismaster()` under the `if performing_handshake` check.

ShaneHarvey (changes requested):
Hard to tell from the PR UI but did you remove the old SDAM tests before resyncing? This makes sure we delete tests that are no longer in the specs repo.

pymongo/pool.py
prashantmital: Not needed anymore since `helloOk` is only sent on initial handshake.

pymongo/pool.py
prashantmital: Thanks for catching that. Done!

ShaneHarvey (approved):
LGTM

test/discovery_and_monitoring_integration/minPoolSize-error.json
```diff
@@ -17,7 +17,6 @@
         },
         "data": {
           "failCommands": [
-            "isMaster",
```
ShaneHarvey, line 20: This file needs to be reverted to fix the test failures.

ShaneHarvey (changes requested):

ShaneHarvey:
Can you schedule all the "Versioned API" tests too?

test/discovery_and_monitoring_integration/minPoolSize-error.json
ShaneHarvey: This file needs both isMaster and hello as mentioned in the spec PR.

prashantmital:
@ShaneHarvey updated the spec test so this is ready for your LGTM. I will wait until the spec PR is merged before merging this.

ShaneHarvey (approved):
LGTM

pymongo/pool.py
ShaneHarvey: I brought up this issue in the spec PR and James says we should send helloOk on all isMaster commands, not just the initial handshake.

pymongo/pool.py
```diff
@@ -583,6 +584,7 @@ def _ismaster(self, cluster_time, topology_version,
         awaitable = False
         if performing_handshake:
             self.performed_handshake = True
+            cmd['helloOk'] = True
```
ShaneHarvey, line 587: The the Spec PR James says we should send helloOk on all isMaster commands, not just the initial handshake: https://github.com/mongodb/specifications/pull/990#issuecomment-870792241

ShaneHarvey (changes requested):

pymongo/pool.py
prashantmital: Done.

ShaneHarvey (approved):
LGTM

PR #657, opened 2021-06-28
PYTHON-2779 Fix topologies field in snapshot reads test

ShaneHarvey:
See: https://github.com/mongodb/specifications/pull/1030

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -673,9 +673,10 @@ class ClientContext(object):
                              func=func)
 
     def is_topology_type(self, topologies):
-        for t in topologies:
-            assert t in {'load-balanced', 'single', 'replicaset', 'sharded',
-                         'sharded-replicaset'}
+        unknown = set(topologies) - {'single', 'replicaset', 'sharded',
+                                     'sharded-replicaset', 'load-balanced'}
+        if unknown:
+            raise AssertionError('Unknown topologies: %r' % (unknown,))
         if self.load_balancer:
             if 'load-balanced' in topologies:
                 return True
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -673,6 +673,9 @@ class ClientContext(object):
                              func=func)
 
     def is_topology_type(self, topologies):
+        for t in topologies:
+            assert t in {'load-balanced', 'single', 'replicaset', 'sharded',
+                         'sharded-replicaset'}
         if self.load_balancer:
             if 'load-balanced' in topologies:
                 return True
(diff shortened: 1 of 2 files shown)


test/__init__.py
```diff
@@ -673,6 +673,9 @@ def check_auth_with_sharding(self, func):
                              func=func)
 
     def is_topology_type(self, topologies):
+        for t in topologies:
```
prashantmital, line 676: Why is this needed? Was there an error in the original spec test that this would have helped catch? Can we change this to raise a more informative message when it fails than assertion error? Something like:

```
unsupported_topologies = set(topologies) - {'load-balanced', 'single', 'replicaset', 'sharded', 'sharded-replicaset'}
if unsupported_topologies:
    raise ValueError(f"Specified topology types {unsupported_topologies} are not supported")
```

ShaneHarvey:
Yep, fixed.

test/__init__.py
ShaneHarvey: Yes the test accidentally had `["replicaset, sharded"]` instead of `["replicaset", "sharded"]`. Here's an example of the new behavior:
```
ERROR: setUpClass (__main__.TestUnifiedSnapshotSessionsNotSupportedServerError)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 678, in setUpClass
    if not cls.should_run_on(run_on_spec):
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 650, in should_run_on
    if is_run_on_requirement_satisfied(req):
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 103, in is_run_on_requirement_satisfied
    topology_satisfied = client_context.is_topology_type(
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 679, in is_topology_type
    raise AssertionError('Unknown topologies: %r' % (unknown,))
AssertionError: Unknown topologies: {'replicaset, sharded'}
```

test/__init__.py
```diff
@@ -673,6 +673,10 @@ def check_auth_with_sharding(self, func):
                              func=func)
 
     def is_topology_type(self, topologies):
+        unknown = set(topologies) - {'single', 'replicaset', 'sharded',
+                                     'sharded-replicaset', 'load-balanced'}
+        if unknown:
+            raise AssertionError('Unknown topologies: %r' % (unknown,))
```
prashantmital, line 679: Not sure why this should be an AssertionError but it doesn't really matter.

prashantmital (approved):

PR #659, opened 2021-06-28
PYTHON-2777 Test all snapshot read commands on unsupported servers

ShaneHarvey:


pymongo/client_session.py
```diff
@@ -900,12 +900,15 @@ def _start_retryable_write(self):
...
 
-    def _update_read_concern(self, cmd):
+    def _update_read_concern(self, cmd, sock_info):
         if (self.options.causal_consistency
                 and self.operation_time is not None):
             cmd.setdefault('readConcern', {})[
                 'afterClusterTime'] = self.operation_time
         if self.options.snapshot:
+            if sock_info.max_wire_version < 13:
+                raise ConfigurationError(
```
ShaneHarvey, line 910: This could be InvalidOperation or ConfigurationError. We seem to use both in this file for various "this thing is not supported" errors. Do you have a preference?

prashantmital: `ConfigurationError` makes more sense to me.

prashantmital (approved):

PR #661, opened 2021-06-30
added support for datetime.date obj in bson

tusharvickey1999:
silently converting datetime.date obj to datetime.datetime obj. Now date obj can also be given in insert query


behackett:
PyMongo doesn't support the date type because there is no corresponding type in MongoDB. This is covered in the docs here:
https://pymongo.readthedocs.io/en/stable/faq.html#how-can-i-save-a-datetime-date-instance

The problem with automatically coercing types like this to some type MongoDB understands is that when you query you get back a different type (in this case you will always get a datetime.datetime). That might be fine for your application, but it may not be fine for other applications. Therefore we leave it up to individual applications to do this coercion

PR #662, opened 2021-06-30
PYTHON-2775 Add docs for snapshot reads

ShaneHarvey:


pymongo/client_session.py
```diff
@@ -76,23 +78,55 @@
...
+majority-committed snapshot timestamp when executing the first read operation
+using the session. Subsequent reads on this session read from the same
+snapshot timestamp. Snapshot reads are also supported when reading from
+replica set secondaries.
+
+.. code-block:: python
+
+  # Each read using this session reads data from the same point in time.
+  with client.start_session(snapshot=True) as session:
+      order = orders.find_one({"sku": "abc123"}, session=session)
```
prashantmital, line 115: Should this be more illustrative of what this functionally means? E.g. 
```
  with client.start_session(snapshot=True) as session:
      order_snapshot_1 = orders.find_one({"sku": "abc123"}, session=session)
      # Update operations cannot use snapshot session
      orders.find_one_and_update({"sku": "abc123"}, {"$set": {"sku": "def456"}})
      # Reads not using the snapshot session see the updated value
      order_new = orders.find_one({"sku": "def456"})
      assert order_new["sku"] == "def456"
      # Reads using the snapshot session continue to see the original value
      order_snapshot_2 = orders.find_one({"sku": "abc123"}, session=session)
      assert order_snapshot_1 == order_snapshot_2
```

prashantmital (changes requested):

pymongo/client_session.py
ShaneHarvey: I'd prefer to wait for some guidance from the DOCS team before making the example more complicated.

prashantmital (approved):

PR #663, opened 2021-06-30
PYTHON-2791 Ignore erroneous serviceId field for non-LB connections

ShaneHarvey:
The serviceId field is not supposed to be returned for non-LB mode. This change fixes the pymongo bug in the ticket by removing the field (if it exists) when in non-LB mode.

### DIFF
--- a/pymongo/topology_description.py
+++ b/pymongo/topology_description.py
@@ -361,7 +361,7 @@ def updated_topology_description(topology_description, server_description):
             topology_description._topology_settings)
 
     if topology_type == TOPOLOGY_TYPE.Unknown:
-        if server_type == SERVER_TYPE.Standalone:
+        if server_type in (SERVER_TYPE.Standalone, SERVER_TYPE.LoadBalancer):
             if len(topology_description._topology_settings.seeds) == 1:
                 topology_type = TOPOLOGY_TYPE.Single
             else:
--- a/test/test_topology.py
+++ b/test/test_topology.py
@@ -18,6 +18,8 @@ import sys
 
 sys.path[0:0] = [""]
 
+from bson.objectid import ObjectId
+
 from pymongo import common
 from pymongo.read_preferences import ReadPreference, Secondary
 from pymongo.server_type import SERVER_TYPE
@@ -276,7 +278,7 @@ class TestMultiServerTopology(TopologyTest):
             'setName': 'rs',
             'hosts': ['a', 'b']})
 
-        self.assertTrue(
+        self.assertEqual(
             t.description.topology_type_name, 'ReplicaSetWithPrimary')
         self.assertTrue(t.description.has_writable_server())
         self.assertTrue(t.description.has_readable_server())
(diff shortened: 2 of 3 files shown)


prashantmital:
If this is a serverless bug should add a comment that the special-casing needs to be removed once serverless fixes the problem upstream?

prashantmital:
On a second thought, lets leave this open for the time being. The code change is small enough that I can manually copy it to get unblocked on serverless testing.

ShaneHarvey:
I'd like to merge this instead of wait for serverless testing. I've added a test. We can remove the pool logic in PYTHON-2712 when we remove the other serviceId related code.

pymongo/topology_description.py
```diff
@@ -361,7 +361,7 @@ def updated_topology_description(topology_description, server_description):
             topology_description._topology_settings)
 
     if topology_type == TOPOLOGY_TYPE.Unknown:
-        if server_type == SERVER_TYPE.Standalone:
+        if server_type in (SERVER_TYPE.Standalone, SERVER_TYPE.LoadBalancer):
```
prashantmital, line 364: So with multiple seeds + LB we will now end up removing hosts until we have only one left and then connect to that as a Single?
Is this what the spec says to do or is this behavior undefined?

test/test_topology.py
```diff
@@ -621,6 +623,26 @@ def test_topology_repr(self):
...
+        mock_lb_response = {'ok': 1, 'msg': 'isdbgrid',
+                            'serviceId': ObjectId(), 'maxWireVersion': 13}
+        got_ismaster(t, ('a', 27017), mock_lb_response)
+        sds = t.description.server_descriptions()
+        self.assertIn(('a', 27017), sds)
+        self.assertEqual(sds[('a', 27017)].server_type_name, 'LoadBalancer')
+        self.assertEqual(t.description.topology_type_name, 'Single')
+        self.assertTrue(t.description.has_writable_server())
+
+        # Load balancers are removed from a topology with multiple seeds.
```
prashantmital, line 640: What if both `a` and `b` advertise a serviceId?

prashantmital (changes requested):

pymongo/topology_description.py
ShaneHarvey: See the comment. This is unreachable code in practice. It's just a failsafe in case the "serviceId" popping logic doesn't work or gets removed prematurely.

test/test_topology.py
ShaneHarvey: They'd both get removed. No need to test it though since this is unreachable in practice. We don't need to test every combination IMO.

prashantmital (approved):
LGTM assuming tests pass.

test/test_topology.py
prashantmital: My reading of the code tells me that if the serviceId popping fails, we'll be left with one LB to which we'll connect as a standalone but I take your point that this is unreachable - we can leave it at that.

prashantmital:
To clarify, is the plan to revert this change entirely as part of [PYTHON-2712](https://jira.mongodb.org/browse/PYTHON-2712)? If so, can you make a note about the same on the ticket?

PR #665, opened 2021-07-07
PYTHON-2475 Implement ADL prose tests

prashantmital:


test/test_data_lake.py
```diff
@@ -19,26 +19,91 @@
...
+        self.assertIn(cursor_id, started.command["cursors"])
+        target_ns = ".".join([started.command['$db'],
+                              started.command['killCursors']])
+        self.assertEqual(cursor_ns, target_ns)
+
+        self.assertIn(cursor_id, succeeded.reply["cursorsKilled"])
+
+    # Test no auth
+    def test_2(self):
+        client = _mongo_client(None, None, authenticate=False)
```
ShaneHarvey, line 87: I think you want to use `rs_client_noauth`.

test/test_data_lake.py
```diff
@@ -19,26 +19,91 @@
...
+        client.admin.command('ping')
+
+    # Test with auth
+    def test_3(self):
+        for mechanism in ['SCRAM-SHA-1', 'SCRAM-SHA-256']:
+            client = rs_or_single_client(authMechanism=mechanism)
+            client.admin.command('ping')
```
ShaneHarvey, line 94: I see the spec says to use ping but it should really be something that requires auth. Can you change it to findOne?

test/test_data_lake.py
```diff
@@ -19,26 +19,91 @@
...
+
+
 class DataLakeTestSpec(TestCrudV2):
     # Default test database and collection names.
     TEST_DB = 'test'
     TEST_COLLECTION = 'driverdata'
 
     @classmethod
-    @unittest.skipUnless(client_context.is_data_lake,
```
ShaneHarvey, line 38: Please use `client_context._require` as a decorator instead. Consider renaming to `client_context.require`. 

The problem with the new code is that it leaves client_knobs enabled because when setUpClass errors with an exception then tearDownClass is never called.

test/test_data_lake.py
prashantmital: Done.

test/test_data_lake.py
prashantmital: Done. Will it work regardless of auth mechanism?

test/test_data_lake.py
prashantmital: Ah, right... Added a new helper `client_context.require_data_lake`

test/test_data_lake.py
ShaneHarvey: It should. Privileges are assigned to a user. Signing in to the same user with different authMechanisms will result in the same privileges.

ShaneHarvey (approved):
LGTM assuming the tests still pass.

PR #667, opened 2021-07-08
PYTHON-2608 Test that KMS TLS connections verify peer certificates

ShaneHarvey:
Implements the tests described here: https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#kms-tls-tests

Patch (Note the Windows failures are unrelated see PYTHON-2798): https://spruce.mongodb.com/version/60e643e47742ae7acd922fe3/tasks

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -97,7 +97,7 @@ functions:
             # If this was a patch build, doing a fresh clone would not actually test the patch
             cp -R ${PROJECT_DIRECTORY}/ $DRIVERS_TOOLS
           else
-            git clone git://github.com/mongodb-labs/drivers-evergreen-tools.git $DRIVERS_TOOLS
+            git clone git://github.com/ShaneHarvey/drivers-evergreen-tools.git --branch DRIVERS-1560-fix-self-signed-cert $DRIVERS_TOOLS
           fi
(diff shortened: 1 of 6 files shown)


.evergreen/config.yml
```diff
@@ -361,6 +361,32 @@ functions:
            PYTHON_BINARY=${PYTHON_BINARY} sh ${PROJECT_DIRECTORY}/.evergreen/run-doctests.sh
 
   "run tests":
+    # If testing encryption start the mock KMS servers.
+    # First create the virtualenv and install dependencies.
+    - command: shell.exec
```
prashantmital, line 366: Instead of doing this here, can't we do this in `run-tests.sh` gated under a check on `TEST_ENCRYPTION`? In config.yml under `"run tests"` we seem to limit ourselves to setting up environment variables which then get used by `run-tests.sh`. It'll be nice to stick to that...

prashantmital (changes requested):

.evergreen/config.yml
ShaneHarvey: Done. That's way simpler.

ShaneHarvey:
Updated to start the mock KMS servers in run-test.sh. I also move all our scripts to bash instead of sh. The remaining windows failures are due to PYTHON-2798.

prashantmital (approved):

PR #670, opened 2021-07-09
PYTHON-2798 Workaround windows cert issue with SSL_CERT_FILE

ShaneHarvey:
Apparently SSL_CERT_FILE does work on Windows. We can use it to workaround the issue in BUILD-13574.

### DIFF
--- a/test/test_encryption.py
+++ b/test/test_encryption.py
@@ -19,7 +19,6 @@ import copy
 import os
 import traceback
 import socket
-import ssl
 import sys
 import textwrap
 import uuid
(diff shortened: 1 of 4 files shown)


ShaneHarvey:
Note that the Windows 3.10 failures are expected.

test/test_encryption.py
```diff
@@ -62,7 +65,26 @@
...
+try:
+    import certifi
+    HAVE_CERTIFI = True
+except ImportError:
+    HAVE_CERTIFI = False
+
+patcher = None
+
+
+def setUpModule():
```
prashantmital, line 78: Is this a standard interface expected by unittest?

prashantmital: Also, why is this needed if we can just call `self.patch_system_certs` wherever required?

ShaneHarvey: Yes. https://docs.python.org/3/library/unittest.html#setupmodule-and-teardownmodule

It was easier to implement this for the entire module than it was to try do it vie setUpClass() or setUp().

prashantmital (approved):

PR #671, opened 2021-07-12
PYTHON-2608 Fix KMS TLS testing on Python <3.5

ShaneHarvey:
When I backported PYTHON-2608 I didn't realize that the mock KMS script requires Python 3.5+. For <3.5 we need to create yet another virtualenv. This fixes the encryption tests on v3.12.

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -187,11 +187,7 @@ if [ -n "$TEST_ENCRYPTION" ]; then
     # Restore the test virtualenv.
     if [ $IS_PRE_35 = "1" ]; then
         deactivate
-        if [ "Windows_NT" = "$OS" ]; then
-            . venv-encryption/Scripts/activate
-        else
-            . venv-encryption/bin/activate
-        fi
+        activatevritualenv venv-encryption
     fi
 fi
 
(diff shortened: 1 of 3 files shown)


.evergreen/run-tests.sh
```diff
@@ -170,11 +172,27 @@ if [ -n "$TEST_ENCRYPTION" ]; then
...
+        createvirtualenv $PYTHON3_BINARY venv-kms
+        python -m pip install boto3
+    fi
     pushd ${DRIVERS_TOOLS}/.evergreen/csfle
     python -u lib/kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/expired.pem --port 8000 &
     python -u lib/kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/wrong-host.pem --port 8001 &
     trap 'kill $(jobs -p)' EXIT HUP
     popd
+    # Restore the test virtualenv.
+    if [ $IS_PRE_35 = "1" ]; then
```
prashantmital, line 188: Can we add an `activatevritualenv` function to `utils.sh` and just call that from both `createvirtualenv` and also here? It'll be much cleaner.

prashantmital (changes requested):

.evergreen/run-tests.sh
ShaneHarvey: Done.

prashantmital (approved):

ShaneHarvey:
I had to revert the activatevritualenv change because it doesn't work. After that change the tests no longer ran in the venv where pymongocrypt was installed and all the encrypt tests were skipped. Not sure what exactly the issue was but it's simpler to revert.

PR #673, opened 2021-07-12
PYTHON-2806 Fix test_aggregate_raw_transaction

ShaneHarvey:


test/test_cursor.py
```diff
@@ -1483,10 +1483,13 @@ def test_find_raw_transaction(self):
                     session=session).sort('_id'))
                 cmd = listener.results['started'][0]
                 self.assertEqual(cmd.command_name, 'find')
-                self.assertEqual(cmd.command['$clusterTime'],
-                                 decode_all(session.cluster_time.raw)[0])
+                self.assertIn('$clusterTime', cmd.command)
```
ShaneHarvey, line 1486: Here we only care that $clusterTime was added to the outgoing command.

test/test_cursor.py
```diff
@@ -1483,10 +1483,13 @@ def test_find_raw_transaction(self):
...
                 self.assertEqual(cmd.command['startTransaction'], True)
                 self.assertEqual(cmd.command['txnNumber'], 1)
+                # Ensure we update $clusterTime from the command response.
+                last_cmd = listener.results['succeeded'][-1]
+                self.assertEqual(last_cmd.reply['$clusterTime']['clusterTime'],
+                                 session.cluster_time['clusterTime'])
```
ShaneHarvey, line 1492: Here we check that the most recent command response using the session is equal to the session's $clusterTime. I used `['clusterTime']` to workaround the RawBSONDocument issue; `['clusterTime']` will always be a bson Timestamp.

prashantmital (approved):
Not sure how this wasn't caught when the test was initially merged?

ShaneHarvey:
> Not sure how this wasn't caught when the test was initially merged?

I'm not sure. Perhaps it was just missed?

PR #677, opened 2021-07-18
PYTHON-2824 Make GridOut implement full io.IOBase spec

henrifroese:


gridfs/grid_file.py
```diff
@@ -421,7 +421,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         return False
 
 
-class GridOut(object):
+class GridOut(io.IOBase):
```
ShaneHarvey, line 424: In order to subclass io.IOBase we need to resolve one more compatibility issue. The problem is that `IOBase.__iter__` yields _lines_ in the file while `GridOut.__iter__`(GridOutIterator) yields the entire chunk. In order to support subclassing IOBase we'll need to change GridOutIterator to return lines, delimitated by `b'\n'`.

gridfs/grid_file.py
```diff
@@ -611,6 +612,33 @@ def readline(self, size=-1):
         data.seek(0)
         return data.read(size)
 
+    def readlines(self, hint=-1):
```
ShaneHarvey, line 615: IOBase implements  readlines() so we get it for free by subclassing. Well after we fix the `GridOut.__iter__` chunk vs line problem described above we will get the correct behavior for free.

ShaneHarvey (changes requested):

gridfs/grid_file.py
```diff
@@ -465,6 +465,7 @@ def __init__(self, root_collection, file_id=None, file_document=None,
 
         root_collection = _clear_entity_type_registry(root_collection)
 
+        self.__closed = False
```
ShaneHarvey, line 468: Do we need a `super().__init__()` call here now that we inherit from IOBase?

gridfs/grid_file.py
```diff
@@ -679,10 +707,21 @@ def close(self):
         if self.__chunk_iter:
             self.__chunk_iter.close()
             self.__chunk_iter = None
+            self.__closed = True
```
ShaneHarvey, line 710: I think we need a super().close() call here now.

gridfs/grid_file.py
henrifroese: Replied on jira to discuss approach to __iter__ :ok_hand:

gridfs/grid_file.py
henrifroese: Removed the method :+1:

gridfs/grid_file.py
henrifroese: Fixed

gridfs/grid_file.py
henrifroese: Fixed

gridfs/grid_file.py
henrifroese: See comment below

gridfs/grid_file.py
```diff
@@ -656,33 +658,40 @@ def seekable(self):
...
            chunk.
+
+        .. versionchanged:: 4.0
+            The iterator now iterates over *lines* in the file, instead
+            of chunks, to conform to the base class :py:class:`io.IOBase`.
+            Use :meth:`GridOut.readchunk` to read chunk by chunk instead
+            of line by line.
         """
-        return GridOutIterator(self, self.__chunks, self._session)
+        return self
```
henrifroese, line 677: The existing IOBase iterator implementation already does all the work for the iterator by using `readline` in `__next__`, so `GridOutIterator` is not needed anymore as far as I can tell, IOBase already does all the work.

test/test_grid_file.py
```diff
@@ -447,27 +462,85 @@ def test_readline(self):
         self.assertEqual(b"e", g.readline(1))
         self.assertEqual(b"llo world,\n", g.readline())
 
+    def test_readlines(self):
```
henrifroese, line 465: Not sure if this test should be removed; it tests `readlines` which is just the IOBase implementation of `readlines` that internally uses GridOut `readline`, and `readline` is already tested in `test_readline`.

test/test_grid_file.py
```diff
@@ -447,27 +462,85 @@ def test_readline(self):
...
+
     def test_iterator(self):
         f = GridIn(self.db.fs)
         f.close()
         g = GridOut(self.db.fs, f._id)
         self.assertEqual([], list(g))
 
         f = GridIn(self.db.fs)
-        f.write(b"hello world")
+        f.write(b"hello world\nhere are\nsome lines.")
```
henrifroese, line 524: This test surprised me when I made changes to it. It previously tested:

```python
g = GridOut(self.db.fs, f._id)
self.assertEqual([b"hello world"], list(g))
self.assertEqual(b"hello", g.read(5))
self.assertEqual([b"hello world"], list(g))
self.assertEqual(b" worl", g.read(5))
```

so iterating with GridOutIterator previously did not actually move the "file pointer" ahead and doing
`next(iter(g))` always gives the same result (first chunk of the file).

The IOBase implementation moves the file pointer ahead when iterating, so expected behaviour is e.g.

```python
>>> with open("test.txt", "w") as f_out:
...     f_out.write("here are\nsome lines\nin a file")
>>> with open("test.txt", "r") as f_in:
...     print(next(iter(f_in)))
...     print(f_in.read(5))
...     print(list(f_in))
... 
here are

some 
['lines\n', 'in a file']
```

and that's also the behaviour that GridOut iter has with this PR. Is there something I'm missing as to why the iterator previously always started from the top and didn't move the "file pointer" ahead?

gridfs/grid_file.py
```diff
@@ -656,33 +658,40 @@ def seekable(self):
...
+        using a webserver that handles such an iterator efficiently.
 
         .. versionchanged:: 3.8
            The iterator now raises :class:`CorruptGridFile` when encountering
            any truncated, missing, or extra chunk in a file. The previous
            behavior was to only raise :class:`CorruptGridFile` on a missing
            chunk.
+
+        .. versionchanged:: 4.0
+            The iterator now iterates over *lines* in the file, instead
```
ShaneHarvey, line 672: nit: can you deindent these lines by one space to match the `versionchanged:: 3.8` indentation level above?

test/test_grid_file.py
ShaneHarvey: Great catch. I don't understand why we had that behavior. The new behavior seems much more reasonable. I'm running the test suite to see if any other regression tests break. If the tests pass then I think we're good to go. We can make a minor note in the changelog that users can call `seek(0)` to reset the iterator.

test/test_grid_file.py
ShaneHarvey: Let's keep this test. It proves that readlines() actually works.

gridfs/grid_file.py
```diff
@@ -656,33 +658,40 @@ def seekable(self):
...
         """
-        return GridOutIterator(self, self.__chunks, self._session)
+        return self
 
     def close(self):
         """Make GridOut more generically file-like."""
         if self.__chunk_iter:
             self.__chunk_iter.close()
             self.__chunk_iter = None
+            super().close()
```
ShaneHarvey, line 684: I think we need to call `super().close()` even if `self.__chunk_iter` is none. Like this:
```python
        if not self.closed:
            if self.__chunk_iter:
                self.__chunk_iter.close()
                self.__chunk_iter = None
            super().close()
```

ShaneHarvey (changes requested):
Great work! A few minor comments. I'm looking into the test failures. It's possible they are unrelated to your changes. Feel free to add yourself to `doc/contributors.rst` as well.

ShaneHarvey (commented):
I've narrowed down the test failures to be caused by this change:
```
- class GridOut(object):
+ class GridOut(io.IOBase):
```
Reverting this line causes the test_retryable_reads tests to pass again but I have no idea why. This needs more investigation. 

The other test failure (`test_collection.TestCollection.test_manual_last_error`) will be fixed if you rebase your changes using `git pull --rebase upstream master`.

test/test_grid_file.py
henrifroese: Will add a note to the changelog

henrifroese:
Interesting; I've made the other changes and will now have a look at the tests. (As an aside, it took some time to get the `test_retryable_reads.py` running locally, would be great if there was some docker-compose somewhere in the repo to make it easier for others.)

henrifroese:
So the first failure (test `gridfs_download_Download_fails_on_first_attempt`) seems to be due to this :mag_right: :

`IOBase` has a destructor (see code here: https://github.com/python/cpython/blob/main/Lib/_pyio.py#L432)
that accesses attribute `_IOBase__closed`. The attribute access however actually calls `GridOut.__getattr__`,
which in turn calls `GridOut._ensure_file`, and ensure_file notices `self._file` is not set (as the first setting failed and is not retried - that's exactly what the test is testing: first find throws an error and retryReads is disabled).

Overriding `__del__`:

```python
def __del__(self):
    pass
```

makes all tests pass.

As far as I can see, we don't need the destructor - there're no system resources or anything we want to release come garbage collection

henrifroese:
One possibility would now be to just override the destructor to `pass`. Another would be to change `__getattr__` to not call `_ensure_file` for every argument.

ShaneHarvey:
Great find! I am still confused as to why `self.__closed` would trigger `__getattr__`. Normally `__getattr__` is called only if the name isn't part of the instance/class already (ie only for "unknown" attribute names). But `self.__closed` should be an attribute that's already defined. Can you explain why `__getattr__` is called at all in this case?

> Another would be to change __getattr__ to not call _ensure_file for every argument.

We could but this would be yet another subtle breaking change and I'd like to avoid adding more of these.

henrifroese:
Actually not sure about that. Minimal example to reproduce would be this:

```python
>>> import io
>>> class A(io.IOBase):
...     def __getattr__(self, item):
...             print(f"__getattr__ called with {item=}")
... 
>>> 
>>> a = A()
>>> a.closed
__getattr__ called with item='__IOBase_closed'
True
>>> 
>>> del a
__getattr__ called with item='__IOBase_closed'
```

ShaneHarvey:
Thanks @henrifroese, apologies for the delay. My plan is to create a Python bug report for the IOBase.__closed + getattr issue but we don't need to hold up this PR for it. I suggest going with your override `__del__` approach with an explanatory comment:
```
    # Override IOBase.__del__ otherwise it will lead to __getattr__ on
    # __IOBase_closed which calls _ensure_file and potentially performs I/O.
    # We cannot do I/O in __del__ since it can lead to a deadlock.
    def __del__(self):
        pass
```

henrifroese:
Makes sense, done :ok_hand:

ShaneHarvey:
Thanks @henrifroese! I opened https://jira.mongodb.org/browse/PYTHON-2874 to make sure GridIn also implements io.IOBase.

PR #679, opened 2021-07-26
PYTHON-2816 Generate pip < 20.3 compatible manylinux wheels

ShaneHarvey:
See: https://evergreen.mongodb.com/version/60ff3bd31e2d176fd57d25d7

I decided to split the manylinux release process into 2 tasks to reduce the runtime. Each task takes ~25 minutes, whereas combined into one task would take ~50 minutes.


prashantmital (approved):
Nice!

PR #687, opened 2021-07-28
PYTHON-2827 Versioned API migration example for ecosystem docs

prashantmital:


test/test_examples.py
```diff
@@ -1137,6 +1137,42 @@ def test_versioned_api(self):
             uri, server_api=ServerApi("1", deprecation_errors=True))
         # End Versioned API Example 4
 
+    @client_context.require_version_min(4, 7)
```
ShaneHarvey, line 1140: Do we still need to skip sharded clusters?

prashantmital: The issue has been fixed in 5.0.2 - https://jira.mongodb.org/browse/SERVER-58794 but seems like we are still using 5.0. 🤔 
Should I just skip it on sharded or we can wait for this to pass before merging?

ShaneHarvey: Do you mind updating drivers evergreen tools to use 5.0.2 real quick?

prashantmital: How does one figure out if it is available yet? I only see 5.0.1 on `m`...

prashantmital: Yeah its definitely not yet available.

ShaneHarvey: It's not out yet: https://docs.mongodb.com/manual/release-notes/5.0/

I suggest putting something like :
```
if sharded and 5.0 and <5.0.2:
    self.skip()
```

test/test_examples.py
```diff
@@ -1137,6 +1137,42 @@ def test_versioned_api(self):
...
+        # pymongo.errors.OperationFailure: Provided apiStrict:true, but the command count is not in API Version 1, full error: {'ok': 0.0, 'errmsg': 'Provided apiStrict:true, but the command count is not in API Version 1', 'code': 323, 'codeName': 'APIStrictError'}
+        # End Versioned API Example 6
+
+        # Start Versioned API Example 7
+        client.db.sales.count_documents({})
+        # End Versioned API Example 7
+
+        # Start Versioned API Example 8
+        # 8
+        # End Versioned API Example 7
```
benjirewis, line 1174: Drive-by fix. I think this typo is also in motor [here](https://github.com/mongodb/motor/blob/master/test/asyncio_tests/test_examples.py#L1180).
```suggestion
        # End Versioned API Example 8
```

ShaneHarvey (approved):

PR #689, opened 2021-07-28
PYTHON-1363 Remove unused retrieved parameter from CommandCursor

prashantmital:
Seeing as this class isn't really meant for end user consumption I don't think there is any point of documenting this?


ShaneHarvey (approved):
LGTM assuming the tests pass.

PR #690, opened 2021-07-28
PYTHON-2288 Remove IsMaster

prashantmital:


pymongo/hello.py
```diff
@@ -12,11 +12,192 @@
...
+
+
+class Hello(object):
+    __slots__ = ('_doc', '_server_type', '_is_writable', '_is_readable',
+                 '_awaitable')
+
+    def __init__(self, doc, awaitable=False):
+        """Parse a hello response from the server.
+
+        .. versionadded:: 4.0
```
ShaneHarvey, line 36: This should say 3.12. Is this not a copy/paste from v3.12?

ShaneHarvey (commented):
I think Bernie may have been working on this. Please check with him so we don't double book anyone.

behackett:
> I think Bernie may have been working on this. Please check with him so we don't double book anyone.

I have no objections to this. It will save me some work.

pymongo/hello.py
prashantmital: It wasn't a copy-paste. I actually did a refactor. However, even what is in v3.12 doesn't have a `..versionadded` in the docstring.

I have updated the PR to use what is 3.12 verbatim as much as possible.

ShaneHarvey (approved):
LGTM!

PR #696, opened 2021-08-06
PYTHON-2532 Remove modifiers option from find methods

ShaneHarvey:


pymongo/cursor.py
```diff
@@ -136,10 +136,10 @@ def __init__(self, collection, filter=None, projection=None, skip=0,
                  limit=0, no_cursor_timeout=False,
                  cursor_type=CursorType.NON_TAILABLE,
                  sort=None, allow_partial_results=False, oplog_replay=False,
-                 modifiers=None, batch_size=0,
+                 batch_size=0,
                  collation=None, hint=None, max_scan=None, max_time_ms=None,
-                 max=None, min=None, return_key=False, show_record_id=False,
-                 snapshot=False, comment=None, session=None,
+                 max=None, min=None, return_key=None, show_record_id=None,
+                 snapshot=None, comment=None, session=None,
```
ShaneHarvey, line 142: This change was necessary to get the spec test to pass. Before we would only send the return_key/show_record_id/snapshot flag if it was True. We now send it when it's explicitly passed as False as well.

PR #700, opened 2021-08-09
PYTHON-2525 Remove Collection.count and Cursor.count

ShaneHarvey:


doc/migrate-to-pymongo4.rst
```diff
@@ -365,6 +365,46 @@ Can be changed to this::
   replaced_doc = collection.find_one_and_replace({'b': 1}, {'c': 1})
   deleted_doc = collection.find_one_and_delete({'c': 1})
 
+Collection.count and Cursor.count is removed
```
prashantmital, line 368: This is great! Just out of curiosity, if users don't read this migration guide, how else might they figure out the operator replacements? Is this transition also documented somewhere on docs.mongodb.com?

prashantmital (approved):

doc/migrate-to-pymongo4.rst
ShaneHarvey: Yes it's documented here: https://docs.mongodb.com/v4.4/reference/method/db.collection.countDocuments/#query-restrictions

PR #705, opened 2021-08-10
PYTHON-2038 Remove pymongo.errors.CertificateError

ShaneHarvey:


pymongo/errors.py
```diff
@@ -18,14 +18,15 @@
 
 try:
```
behackett, line 19: How about moving this to ssl_support instead?

ShaneHarvey: It can't go there because CertificateError because then there would be a circular import ssl_support -> pyopenssl_context -> ssl_support

prashantmital (approved):

PR #706, opened 2021-08-12
PYTHON-2162 Remove support for ssl* URI options

prashantmital:


test/test_ssl.py
```diff
@@ -91,71 +92,38 @@ def test_no_ssl_module(self):
...
         self.assertRaises(ConfigurationError,
-                          MongoClient, ssl_certfile=CLIENT_PEM)
+                          MongoClient, tlsCertificateKeyFile=CLIENT_PEM)
 
     @unittest.skipUnless(HAVE_SSL, "The ssl module is not available.")
     @ignore_deprecations
     def test_config_ssl(self):
         # Tests various ssl configurations
         self.assertRaises(ValueError, MongoClient, ssl='foo')
-        self.assertRaises(ConfigurationError,
```
prashantmital, line 101: The spec does not mandate that ssl/tls=False + one of the tls* options implying a conflicting value of ssl/tls should error. It mandates only that multiple conflicting values of ssl/tls should error and we do adhere to that. Consequently, setting an option like tlsCAFile implicitly sets tls=True and a conflicting tls=False in the URI is simply ignored. 
@ShaneHarvey what is your opinion here? Do you think we should error? I think it is not required since we seem to do the more secure thing and since its 4.0 a breaking change is fine.

ShaneHarvey: I think it makes more since for us to error on things like this:
```python
MongoClient('mongodb://host/?tls=false&tlsCAFile=<...>')  # Conflict
MongoClient('mongodb://host/?tls=false&tlsInsecure=false')  # Conflict
```

pymongo/common.py
```diff
@@ -585,18 +584,11 @@ def validate_tzinfo(dummy, value):
 
 
 # Dictionary where keys are the names of public URI options, and values
-# are lists of aliases for that option. Aliases of option names are assumed
-# to have been deprecated.
+# are lists of aliases for that option.
 URI_OPTIONS_ALIAS_MAP = {
     'journal': ['j'],
     'wtimeoutms': ['wtimeout'],
     'tls': ['ssl'],
```
ShaneHarvey, line 590: Should we remove ssl here (and INTERNAL_URI_OPTION_NAME_MAP) as well?

pymongo/uri_parser.py
```diff
@@ -122,12 +122,10 @@ def parse_host(entity, default_port=DEFAULT_PORT):
...
     "tlsallowinvalidcertificates",
     "tlsallowinvalidhostnames",
-    "tlsdisableocspendpointcheck",}
+    "tlsdisableocspendpointcheck"}
 
 # Options that cannot be specified when tlsInsecure is also specified.
-_TLSINSECURE_EXCLUDE_OPTS = (
-        {k for k in _IMPLICIT_TLSINSECURE_OPTS} |
-        {INTERNAL_URI_OPTION_NAME_MAP[k] for k in _IMPLICIT_TLSINSECURE_OPTS})
+_TLSINSECURE_EXCLUDE_OPTS = _IMPLICIT_TLSINSECURE_OPTS
```
ShaneHarvey, line 128: Is this needed now?

prashantmital: Not really. I just left it in, in case the group of implicit options ever diverges from the list of options that cannot be specified concurrently with tlsInsecure. I'll remove it.

pymongo/common.py
prashantmital: I think the decision was made to keep `ssl` indefinitely since we never deprecated it? Do you think we should deprecate it in the next 3.x release and remove?

prashantmital: Note also that the follow up PR to this will be a better place remove this kind of bookkeeping code as that PR will get rid of all remaining options that have been deprecated in 3.x

ShaneHarvey: > Do you think we should deprecate it in the next 3.x release and remove?

I think you're right that we should keep it. The URI-options spec still has `ssl` and doesn't say to remove it. It might be helpful to ask other drivers what their plan is.

test/test_ssl.py
```diff
@@ -26,7 +26,8 @@
 from pymongo.errors import (ConfigurationError,
                             ConnectionFailure,
                             OperationFailure)
-from pymongo.ssl_support import HAVE_SSL, get_ssl_context, validate_cert_reqs, _ssl
+from pymongo.ssl_support import (
+    HAVE_SSL, get_ssl_context, validate_allow_invalid_certs, _ssl)
```
ShaneHarvey, line 30: validate_allow_invalid_certs doesn't exist.

test/test_ssl.py
prashantmital: Agreed... Reinstated this behavior.

pymongo/client_options.py
```diff
@@ -69,43 +69,51 @@ def _parse_read_concern(options):
...
+        if use_tls is False:
+            raise ConfigurationError("TLS has not been enabled but the "
+                                     "following tls parameters have been set: "
+                                     "%s. Please set `tls=True` or remove."
+                                     % ', '.join(enabled_tls_opts))
+        elif use_tls is None:
+            # Implicitly enable TLS when one of the tls* options is set.
+            use_tls = True
+
+    if use_tls is True:
```
ShaneHarvey, line 106: `use_tls is True` -> `use_tls`

pymongo/ssl_support.py
```diff
@@ -39,51 +39,29 @@
...
          ca_certs,
-         cert_reqs,
+         allow_invalid_certificates,
          crlfile,
-         match_hostname,
-         check_ocsp_endpoint) = args
-        verify_mode = CERT_REQUIRED if cert_reqs is None else cert_reqs
+         allow_invalid_hostnames,
+         disable_ocsp_endpoint_check) = args
+        if allow_invalid_certificates is True:
```
ShaneHarvey, line 52: `allow_invalid_certificates is True` -> `allow_invalid_certificates`

pymongo/client_options.py
```diff
@@ -69,43 +69,51 @@ def _parse_read_concern(options):
...
+    enabled_tls_opts = []
+    for opt in ('tlscertificatekeyfile', 'tlscertificatekeyfilepassword',
+                'tlscafile', 'tlscrlfile'):
+        # Any non-null value of these options implies tls=True.
+        if opt in options and options[opt]:
+            enabled_tls_opts.append(opt)
+    for opt in ('tlsallowinvalidcertificates', 'tlsallowinvalidhostnames',
+                'tlsdisableocspendpointcheck'):
+        # A value of False for these options implies tls=True.
+        if opt in options and options[opt] is False:
```
ShaneHarvey, line 93: Is `is False` needed or can this be `and not options[opt]`?

pymongo/client_options.py
```diff
@@ -69,43 +69,51 @@ def _parse_read_concern(options):
...
+            enabled_tls_opts.append(opt)
+
+    if enabled_tls_opts:
+        if use_tls is False:
```
ShaneHarvey, line 97: Suggest switching the order of the this `if` so we can replace the `use_tls is False` with `not use_tls`.

ShaneHarvey:
Did you run the pyopenssl tests as well?

ShaneHarvey (approved):

PR #711, opened 2021-08-23
PYTHON-2046 Change default JSONMode from LEGACY to RELAXED

ShaneHarvey:
TODO: 

- [x] Decide if we should change the behavior and make strict_number_long/datetime_representation/strict_uuid configurable even with JSONMode.RELAXED/JSONMode.CANONICAL. 
  The options are to 1) ignore these options (keep the current behavior), 2) raise an error when an incompatible option is specified, or 3) allow users to configure these options with any json_mode. For now I've gone with option 2).
- [x] Update changelog
- [x] Update migration guide
- [x] Update any other misc docs

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -238,23 +238,27 @@ class JSONOptions(CodecOptions):
 
     .. seealso:: The specification for Relaxed and Canonical `Extended JSON`_.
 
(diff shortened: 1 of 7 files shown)


juliusgeo (approved):
LGTM!

PR #713, opened 2021-08-24
PYTHON-2879 Fix get_ssl_context for CSFLE and ocsptest.py

ShaneHarvey:
This PR fixes the refactor in PYTHON-2162. I've changed get_ssl_context in all places and double checked the parameter translation.

### DIFF
--- a/test/test_ssl.py
+++ b/test/test_ssl.py
@@ -291,17 +291,13 @@ class TestSSL(IntegrationTest):
         #
         #   --sslPEMKeyFile=/path/to/pymongo/test/certificates/server.pem
         #   --sslCAFile=/path/to/pymongo/test/certificates/ca.pem
-        ctx = get_ssl_context(
-            None, None, None, True, None, True, False)
+        ctx = get_ssl_context(None, None, None, None, True, True, False)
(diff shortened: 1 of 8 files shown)


test/test_ssl.py
```diff
@@ -291,17 +291,13 @@ def test_cert_ssl_validation_hostname_matching(self):
         #
         #   --sslPEMKeyFile=/path/to/pymongo/test/certificates/server.pem
         #   --sslCAFile=/path/to/pymongo/test/certificates/ca.pem
-        ctx = get_ssl_context(
-            None, None, None, True, None, True, False)
+        ctx = get_ssl_context(None, None, None, None, True, True, False)
```
prashantmital, line 294: Just to be sure I understand this correctly, these function calls were already correct but are included in this diff simply because you moved around the arguments in the `get_ssl_context` signature, right?

prashantmital (approved):
Thanks for fixing this. I never ran CSFLE tests while doing this refactor which was a my bad!

test/test_ssl.py
ShaneHarvey: This test was okay but a lot of the other calls in this file were incorrect.

PR #721, opened 2021-09-03
PYTHON-2883 Regex decoding error tests in top.json have unexpected, invalid syntax

juliusgeo:
https://jira.mongodb.org/browse/PYTHON-2883

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -634,13 +634,13 @@ def _parse_canonical_code(doc):
 def _parse_canonical_regex(doc):
     """Decode a JSON regex to bson.regex.Regex."""
     regex = doc['$regularExpression']
-    if not isinstance(regex['options'], str):
-        raise TypeError('Bad $regularExpression options, options must be string')
     if len(doc) != 1:
         raise TypeError('Bad $regularExpression, extra field(s): %s' % (doc,))
     if len(regex) != 2:
         raise TypeError('Bad $regularExpression must include only "pattern"'
                         'and "options" components: %s' % (doc,))
+    if not isinstance(regex['options'], str):
+        raise TypeError('Bad $regularExpression options, options must be string, was type %s' % (type(regex['options'])))
     return Regex(regex['pattern'], regex['options'])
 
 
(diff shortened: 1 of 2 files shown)


bson/json_util.py
```diff
@@ -634,6 +634,8 @@ def _parse_canonical_code(doc):
 def _parse_canonical_regex(doc):
     """Decode a JSON regex to bson.regex.Regex."""
     regex = doc['$regularExpression']
+    if not isinstance(regex['options'], str):
+        raise TypeError('Bad $regularExpression options, options must be string')
```
ShaneHarvey, line 638: Please move this check to below the len(doc) and len(regex) checks to ensure we give good error messages in all cases. For the same reason can you also add the type() of regex['options'] to the error message?

ShaneHarvey (changes requested):

bson/json_util.py
juliusgeo: Done.

bson/json_util.py
```diff
@@ -639,6 +639,8 @@ def _parse_canonical_regex(doc):
     if len(regex) != 2:
         raise TypeError('Bad $regularExpression must include only "pattern"'
                         'and "options" components: %s' % (doc,))
+    if not isinstance(regex['options'], str):
+        raise TypeError('Bad $regularExpression options, options must be string, was type %s' % (type(regex['options'])))
```
ShaneHarvey, line 643: We limit our lines to 79 chars according to pep 8: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

ShaneHarvey (changes requested):

bson/json_util.py
juliusgeo: Done.

bson/json_util.py
```diff
@@ -639,6 +639,9 @@ def _parse_canonical_regex(doc):
     if len(regex) != 2:
         raise TypeError('Bad $regularExpression must include only "pattern"'
                         'and "options" components: %s' % (doc,))
+    if not isinstance(regex['options'], str):
+        raise TypeError('Bad $regularExpression options, options must be '
+                        'string, was type %s' % (type(regex['options'])))
     return Regex(regex['pattern'], regex['options'])
```
ShaneHarvey, line 645: I suggest refactoring to `regex['options']` to a local variable to avoid looking it up repeatedly and improve readability.

ShaneHarvey (commented):
One minor comment

bson/json_util.py
juliusgeo: Done.

ShaneHarvey (approved):
LGTM

PR #723, opened 2021-09-09
PYTHON-2820 Test serialization of BSON with embedded null bytes in strings

ShaneHarvey:


test/test_bson_corpus.py
```diff
@@ -51,6 +51,8 @@
     # This variant of $numberLong may have been generated by an old version
     # of mongoexport.
     'Bad $numberLong (number, not string)',
+    # We parse Regex flags with extra characters, including nulls.
+    'Null byte in $regularExpression options',
```
ShaneHarvey, line 55: Regex allows extra characters in "flags":
```
>>> Regex('', '1234567qwertyasdfvbn\x00')
Regex('', re.DOTALL)
```

behackett (approved):

PR #728, opened 2021-09-15
PYTHON-808 Prevent use of Database and Collection in boolean expressions

juliusgeo:


pymongo/collection.py
```diff
@@ -289,6 +289,10 @@ def __ne__(self, other):
     def __hash__(self):
         return hash((self.__database, self.__name))
 
+    def __bool__(self):
+        raise NotImplementedError("Collection does not implement Boolean "
```
ShaneHarvey, line 293: This error message here should be more helpful. Try to put yourself in the user's shoes. Suppose you had an app that was relying on this feature:
```python
if some_condition():
    coll = client.test.test
else:
    coll = None

if coll:  # This line worked in Pymongo 3 but will raise an error in PyMongo 4
    coll.find_one()
```

What would you want the error message to say? It should be something that makes it clear what the developer should do to resolve the error.

ShaneHarvey (changes requested):
This change needs tests, a changelog entry, and a pymongo 4 migration entry.

pymongo/collection.py
juliusgeo: I tried to improve the error message.

ShaneHarvey: Great, it's much better now! I suggest a few tweaks to better match the language used in [Python's documentation for `__bool__`](https://docs.python.org/3/reference/datamodel.html#object.__bool__):
```
Collection objects do not implement truth value testing or bool(). Please compare with None instead: collection is not None
```

test/test_collection.py
```diff
@@ -2265,6 +2265,10 @@ def test_find_command_generation(self):
                  ('$dumb', 2),
                  ('filter', {'foo': 1})]).to_dict())
 
+    def test_bool(self):
+        with self.assertRaises(NotImplementedError):
+            Collection(self.db, 'test').__bool__()
```
ShaneHarvey, line 2270: I think it would be more "Pythonic" to call bool() instead of `.__bool__()`

doc/changelog.rst
```diff
@@ -134,8 +134,13 @@ decodes datetime as naive by default.
 - The ``hint`` option is now required when using ``min`` or ``max`` queries
   with :meth:`~pymongo.collection.Collection.find`.
 - ``name`` is now a required argument for the :class:`pymongo.driver_info.DriverInfo` class.
+- :class:`~pymongo.collection.Collection` and :class:`~pymongo.database.Database`
+now raises an error upon evaluating as a Boolean, please use the
+syntax ``if collection is not None:`` or ``if database is not None:`` as
+opposed to
+the previous syntax which was simply ``if collection:`` or ``if database:``.
+You must now explicitly compare with None.
```
ShaneHarvey, line 142: These lines won't render properly because of the indentation.

doc/migrate-to-pymongo4.rst
```diff
@@ -321,6 +321,13 @@ Can be changed to this::
 
 .. _'system.profile' collection: https://docs.mongodb.com/manual/reference/database-profiler/
 
+Database.__bool__ raises NotImplementedError
+..............................................
+:class:`~pymongo.database.Database` now raises an error upon evaluating as a
+Boolean, please use the syntax ``if database is not None:`` as opposed to
+the previous syntax which was simply ``if database:``. You must now
+explicitly compare with None.
```
ShaneHarvey, line 329: Can you use the same "Code like this::...Can be changed to this::.." pattern as the other entries in this file? :
```
Code like this::

...

Can be changed to this::

...

```

ShaneHarvey (changes requested):

pymongo/collection.py
juliusgeo: Done.

test/test_collection.py
juliusgeo: Done.

doc/changelog.rst

doc/migrate-to-pymongo4.rst
juliusgeo: Done. Also changed it for collection.

pymongo/database.py
```diff
@@ -840,6 +840,11 @@ def __next__(self):
 
     next = __next__
 
+    def __bool__(self):
+        raise NotImplementedError("Database objects do not implement truth "
+                                  "value testing or bool(). Please compare "
+                                  "with None instead: collection is not None")
```
ShaneHarvey, line 846: "collection is not None" -> "database is not None"

ShaneHarvey (commented):
One minor tweak, otherwise looks good!

pymongo/database.py
juliusgeo: Done.

ShaneHarvey (approved):
LGTM

shaoran:
I recently made an update of my dependencies and pymongo got updated from version 3 to 4.5.0.

My application stopped working because of this and I don't want to spend my time looking for all the places where I have `if not database` instead of `if database is None`, so I ended up monkey patching this.

However, I'd like to understand why you thought this change is that important that you don't mind breaking the API, in which case I would invest time into changing my source code, because at the moment I don't see the benefit from changing `if not database` to `if database is None`, when the context is pretty clear. Sadly this pull request has no description. In the pull request title there is PYTHON-808 but I have no idea what it is referring to, a google search did not help here

Thanks.

shaoran:
Oh, I see, after clicking on *comment*, the `PYTHON-808` text turned into a Jira link.

ShaneHarvey:
MongoClient has various boolean or Truthy properties like `client.primary`. We decided that this behavior was too error prone because the old behavior would hide bugs like this:
```python
if client.pirmary:  # Oops "pirmary" is a Database, not the "primary" property.
    print("I'm connected!")
```

A real example of this bug is here: https://github.com/mongodb/mongo-python-driver/commit/73edd2c134f43d58531e3dc26499b777354fb7ea

PR #730, opened 2021-09-15
PYTHON-2518 SON class should be compatible with Python 3's OrderedDict API

juliusgeo:


bson/son.py
```diff
@@ -63,7 +63,7 @@ def __delitem__(self, key):
         dict.__delitem__(self, key)
 
     def keys(self):
-        return list(self.__keys)
+        return OrderedDict.fromkeys(self.__keys)
```
ShaneHarvey, line 66: We want `keys` to return something compatible with what dict().keys() and OrderedDict().keys() return. We don't want to create a new OrderedDict here.

ShaneHarvey (changes requested):

bson/son.py
juliusgeo: Fixed.

ShaneHarvey (approved):

PR #731, opened 2021-09-16
PYTHON-2501 Remove iteritems from son.SON

juliusgeo:
…function

### DIFF
--- a/bson/son.py
+++ b/bson/son.py
@@ -80,21 +80,16 @@ class SON(dict):
     def has_key(self, key):
         return key in self.__keys
 
-    # third level takes advantage of second level definitions
-    def iteritems(self):
-        for k in self:
-            yield (k, self[k])
-
     def iterkeys(self):
         return self.__iter__()
 
     # fourth level uses definitions from lower levels
     def itervalues(self):
-        for _, v in self.iteritems():
+        for _, v in self.items():
             yield v
 
     def values(self):
-        return [v for _, v in self.iteritems()]
+        return [v for _, v in self.items()]
 
     def items(self):
         return [(key, self[key]) for key in self]
@@ -125,7 +120,7 @@ class SON(dict):
 
     def popitem(self):
         try:
-            k, v = next(self.iteritems())
+            k, v = next(self.items())
         except StopIteration:
             raise KeyError('container is empty')
         del self[k]
@@ -135,8 +130,8 @@ class SON(dict):
         # Make progressively weaker assumptions about "other"
         if other is None:
             pass
-        elif hasattr(other, 'iteritems'):  # iteritems saves memory and lookups
-            for k, v in other.iteritems():
+        elif hasattr(other, 'items'):
+            for k, v in other.items():
                 self[k] = v
         elif hasattr(other, 'keys'):
             for k in other.keys():
@@ -192,7 +187,7 @@ class SON(dict):
         if val_id in memo:
             return memo.get(val_id)
         memo[val_id] = out
-        for k, v in self.iteritems():
+        for k, v in self.items():
             if not isinstance(v, RE_TYPE):
                 v = copy.deepcopy(v, memo)
             out[k] = v


bson/son.py
```diff
@@ -80,21 +80,16 @@ def __iter__(self):
...
-        for _, v in self.iteritems():
+        for _, v in self.items():
             yield v
 
     def values(self):
-        return [v for _, v in self.iteritems()]
+        return [v for _, v in self.items()]
 
     def items(self):
         return [(key, self[key]) for key in self]
```
ShaneHarvey, line 95: Using items instead of iteritems is going to be a performance problem because items() returns a fully inflated list. If we want to make this change we have to avoid that.

ShaneHarvey (changes requested):

bson/son.py
juliusgeo: Done.

test/test_collection.py
```diff
@@ -390,26 +390,28 @@ def test_index_info(self):
 
         db.test.create_index("hello")
         self.assertEqual(len(db.test.index_information()), 2)
-        self.assertEqual(db.test.index_information()["hello_1"]["key"],
+        self.assertEqual(list(db.test.index_information()["hello_1"]["key"]),
```
ShaneHarvey, line 393: Why are these changes needed? The output of `index_information()` should not change in this PR.

pymongo/helpers.py
```diff
@@ -72,6 +72,8 @@ def _index_list(key_or_list, direction=None):
     else:
         if isinstance(key_or_list, str):
             return [(key_or_list, ASCENDING)]
+        if isinstance(key_or_list, abc.ItemsView):
+            return list(key_or_list)
```
ShaneHarvey, line 76: Why is this change needed?

doc/migrate-to-pymongo4.rst
```diff
@@ -621,6 +621,7 @@ can be changed to this::
 
   cursor = coll.find({}, min={'x', min_value}, hint=[('x', ASCENDING)])
 
+
```
ShaneHarvey, line 624: Extra space here.

doc/changelog.rst
```diff
@@ -134,8 +134,9 @@ decodes datetime as naive by default.
 - The ``hint`` option is now required when using ``min`` or ``max`` queries
   with :meth:`~pymongo.collection.Collection.find`.
 - ``name`` is now a required argument for the :class:`pymongo.driver_info.DriverInfo` class.
+- :meth:`~bson.son.SON.items` now returns a ``dict_items`` object rather
+  than a list.
```
ShaneHarvey, line 138: We should also mention that SON.iteritems is removed.

ShaneHarvey (changes requested):

doc/changelog.rst
juliusgeo: Done.

doc/migrate-to-pymongo4.rst
juliusgeo: Done.

pymongo/helpers.py
juliusgeo: This is necessary so that when a dict_items object is passed into _index_list it will return a list rather than a dict_items object.

test/test_collection.py
juliusgeo: Those changes are needed to get the tests to pass, because otherwise it tries to compare a dict_items object with a list. I am not sure why the output of index_information is changing but it appears it is.

ShaneHarvey: Please investigate why `index_information` has changed. It should not be returning a dict_items object. We'll need to revert these test changes too.

ShaneHarvey (changes requested):
It's looking good so far. Do you mind fixing the merge conflict?

ShaneHarvey (approved):
LGTM!

PR #737, opened 2021-09-17
PYTHON-2463 Do not allow a MongoClient to be reused after it is closed

juliusgeo:


test/test_client.py
```diff
@@ -107,6 +108,7 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         cls.client.close()
+        cls.client = rs_or_single_client()
```
ShaneHarvey, line 111: Unneeded?

test/test_client.py
```diff
@@ -775,14 +777,8 @@ def test_close(self):
...
-        self.client.close()
-
-        coll.count_documents({})
-
-        self.client.close()
-        self.client.close()
-
-        coll.count_documents({})
+        self.assertRaises(InvalidOperation, coll.count_documents, {})
+        self.client = rs_or_single_client()
```
ShaneHarvey, line 781: Here and in the other tests let's create and close a _new_ client instead of closing `self.client`. Closing self.client can be problematic when there are already other objects that reference it (like self.db).

pymongo/mongo_client.py
```diff
@@ -1192,6 +1193,9 @@ def _select_server(self, server_selector, session, address=None):
           - `address` (optional): Address when sending a message
             to a specific server, used for getMore.
         """
+        if self._is_closed:
+            raise InvalidOperation("Once a MongoClient is closed, "
+                                   "all operations raise an error")
```
ShaneHarvey, line 1198: It might make more sense to put this logic in `Topology.open()` that way we will never reopen a closed Topology.

test/test_client.py
juliusgeo: Done.

test/test_client.py
juliusgeo: Done.

pymongo/topology.py
```diff
@@ -237,6 +240,9 @@ def select_server(self,
                       server_selection_timeout=None,
                       address=None):
         """Like select_servers, but choose a random server if several match."""
+        if self._is_closed:
+            raise InvalidOperation("Once a Topology is closed, "
+                                   "all operations raise an error")
```
ShaneHarvey, line 245: I suggest moving this to the Topology._ensure_opened for a few reasons:
- this is the only way to ensure that we won't re-open the topology.
- we should be holding the lock when checking _is_closed.

ShaneHarvey (changes requested):
Don't forget to update the changelog and migration guide because this is a breaking change. Also please grep through the code base to look for "reopen" and "re-open" and remove any docs or comments that mention the old re-opening behavior.

pymongo/topology.py
```diff
@@ -77,6 +78,8 @@ def __init__(self, topology_settings):
         self._publish_server = pub and self._listeners.enabled_for_server
         self._publish_tp = pub and self._listeners.enabled_for_topology
 
+        self._is_closed = False
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/topology.py
juliusgeo: That makes sense. Fixed.

test/test_client.py
```diff
@@ -828,13 +823,8 @@ def test_close_stops_kill_cursors_thread(self):
...
 
-        # Reusing the closed client should restart the thread.
-        client.admin.command('ping')
-        self.assertFalse(client._kill_cursors_executor._stopped)
-
-        # Again, closing the client should stop the thread.
-        client.close()
-        self.assertTrue(client._kill_cursors_executor._stopped)
+        # Reusing the closed client should raise an InvalidOperation error.
+        self.assertRaises(InvalidOperation, client.admin.command, 'ping')
```
juliusgeo, line 827: I'm not sure we need this if we are already testing that it raises an InvalidOperation exception in test_client.test_close

pymongo/mongo_client.py
```diff
@@ -1127,10 +1126,9 @@ def close(self):
         sending one or more endSessions commands.
 
         Close all sockets in the connection pools and stop the monitor threads.
-        If this instance is used again it will be automatically re-opened and
-        the threads restarted unless auto encryption is enabled. A client
-        enabled with auto encryption cannot be used again after being closed;
-        any attempt will raise :exc:`~.errors.InvalidOperation`.
+        If this instance is used again it will NOT automatically be
+        re-opened and the threads restarted. No client can be used again after
+        being closed; any attempt will raise :exc:`~.errors.InvalidOperation`.
```
ShaneHarvey, line 1131: I think we can remove the  "NOT automatically be re-opened and the threads restarted". Instead let's say:
"Once closed, the client cannot be used again and any attempt will raise  :exc:`~pymongo.errors.InvalidOperation`."

Please also add a versionchanged to this method.

pymongo/topology.py
```diff
@@ -247,6 +249,7 @@ def select_server(self,
         else:
             return server2
 
+
```
ShaneHarvey, line 252: Extra newline.

pymongo/topology.py
```diff
@@ -550,6 +556,9 @@ def _ensure_opened(self):
 
         Hold the lock when calling this.
         """
+        if self._closed:
+            raise InvalidOperation("Once a Topology is closed, "
+                                   "all operations raise an error")
```
ShaneHarvey, line 561: Let's come up with a more informative error message for the user.

test/test_client.py
```diff
@@ -106,7 +107,7 @@ def setUpClass(cls):
 
     @classmethod
     def tearDownClass(cls):
-        cls.client.close()
```
ShaneHarvey, line 109: I think this close should still be here, why did you choose to remove it?

test/test_client.py
ShaneHarvey: This line becomes meaningful once we add this:
```python
self.assertRaises(InvalidOperation, client.admin.command, 'ping')
# Thread is still stopped.
self.assertTrue(client._kill_cursors_executor._stopped)
```

ShaneHarvey (changes requested):

doc/changelog.rst
```diff
@@ -140,6 +140,9 @@ decodes datetime as naive by default.
   opposed to
   the previous syntax which was simply ``if collection:`` or ``if database:``.
   You must now explicitly compare with None.
+- :class:`~pymongo.topology.Topology` cannot execute any operations after
```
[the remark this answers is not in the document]

pymongo/mongo_client.py

pymongo/topology.py
juliusgeo: Done.

pymongo/topology.py
juliusgeo: Tried my best to make it a bit more clear.

test/test_client.py
juliusgeo: I think you are right--I reverted it.

test/test_client.py
juliusgeo: Done.

test/test_client.py
```diff
@@ -1132,11 +1122,16 @@ def test_contextlib(self):
         with contextlib.closing(client):
             self.assertEqual("bar", client.pymongo_test.test.find_one()["foo"])
             self.assertEqual(1, len(get_pool(client).sockets))
-        self.assertEqual(0, len(get_pool(client).sockets))
-
+            wait_until(lambda: len(client.nodes) > 0, "find all nodes")
+            self.assertGreater(len(client.nodes), 0)
+        self.assertEqual(0, len(client.nodes))
```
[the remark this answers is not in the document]
juliusgeo: Don't we need to check client.nodes to make sure that we're fully connected to the server?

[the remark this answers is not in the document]
juliusgeo: Nevermind, I think I see what you mean. We already know the collection is opened because of the assertEqual comparing "bar" and find_one output.

test/unified_format.py
```diff
@@ -977,7 +977,6 @@ def _testOperation_targetedFailPoint(self, spec):
         client = single_client('%s:%s' % session._pinned_address)
         self.__set_fail_point(
             client=client, command_args=spec['failPoint'])
-        self.addCleanup(client.close)
```
ShaneHarvey, line 980: Instead of removing this, we should move it above `self.__set_fail_point`.

doc/migrate-to-pymongo4.rst
```diff
@@ -181,6 +181,14 @@ now defaults to ``False`` instead of ``True``. ``json_util.loads`` now
...
 
 
+MongoClient cannot execute operations after ``close()``
+.......................................................
+
+:class:`~pymongo.mongo_client.MongoClient` cannot execute any operations
+after being closed. The previous behavior would simply reconnect. However,
+now you must create a new instance.
+
+
```
ShaneHarvey, line 191: There's an extra new line here.

ShaneHarvey (changes requested):
Two minor comments about the latest changes.

doc/migrate-to-pymongo4.rst
juliusgeo: Done.

test/unified_format.py
juliusgeo: Done.

ShaneHarvey (approved):
LGTM. However, there's one last thing we need to address here that I just remembered. MongoClient starts a background thread that periodically calls MongoClient._process_periodic_tasks(). Right now there is a race between the thread and client.close() that could cause the thread to fail with InvalidOperation.

To proceed, can you describe this issue in a comment on the jira ticket? Then we can merge this PR and fix this above issue in a new PR rather than prolong this one.

PR #743, opened 2021-09-24
PYTHON-2915 Fix bug when starting a transaction with a large bulk write

ShaneHarvey:


pymongo/bulk.py
```diff
@@ -267,17 +267,18 @@ def _execute_command(self, generator, write_concern, session,
         # sock_info.write_command.
         sock_info.validate_session(client, session)
         while run:
-            cmd = SON([(_COMMANDS[run.op_type], self.collection.name),
```
ShaneHarvey, line 270: Let me know if there are any questions. The gist of the bug is that we were reusing this shared SON object for all batches in a run. Usually that works fine but when starting a multi-doc transaction only the first command can have the startTransaction=True field. My fix refactors the code path to create a new SON object for each batch sent to the server.

juliusgeo (approved):
LGTM!

PR #748, opened 2021-09-27
PYTHON-2793 Stop testing TLS with Python 3.10 and MongoDB <= 3.4

ShaneHarvey:


.evergreen/config.yml
```diff
@@ -2431,6 +2431,20 @@ buildvariants:
   display_name: "PyOpenSSL ${platform} ${python-version-windows} ${auth}"
   tasks:
     - '.replica_set'
+  rules:
+    # Python 3.10 does not support TLS on MongoDB <= 3.4.
```
behackett, line 2435: It's more like "On Windows, Python 3.10+ and MongoDB <= 3.4 don't have any ciphers in common."

.evergreen/config.yml
```diff
@@ -2431,6 +2431,20 @@ buildvariants:
   display_name: "PyOpenSSL ${platform} ${python-version-windows} ${auth}"
   tasks:
     - '.replica_set'
+  rules:
```
behackett, line 2434: Is this actually a problem when using PyOpenSSL, or is it only a problem with CPython's ssl module?

.evergreen/config.yml
```diff
@@ -2533,6 +2547,18 @@ buildvariants:
     auth-ssl: "*"
   display_name: "${platform} ${python-version-windows} ${auth-ssl}"
   tasks: *all-server-versions
+  rules:
+    # Python 3.10 does not support TLS on MongoDB <= 3.4.
+    - if:
+        platform: "*"
+        python-version-windows: "3.10"
+        auth-ssl: "auth-ssl"
```
behackett, line 2555: I think we still want to test auth, right? Doesn't this skip tls and auth?

.evergreen/config.yml
```diff
@@ -2541,6 +2567,18 @@ buildvariants:
     auth-ssl: "*"
   display_name: "${platform} ${python-version-windows-32} ${auth-ssl}"
   tasks: *all-server-versions
+  rules:
+    # Python 3.10 does not support TLS on MongoDB <= 3.4.
+    - if:
+        platform: "*"
+        python-version-windows-32: "3.10"
+        auth-ssl: "auth-ssl"
```
behackett, line 2575: Same comment here.

doc/examples/tls.rst
```diff
@@ -249,3 +249,14 @@ revocation checking failed::
   [('SSL routines', 'tls_process_initial_server_flight', 'invalid status response')]
 
 See :ref:`OCSP` for more details.
+
+Python 3.10+ is incompatible with TLS/SSL on MongoDB <= 3.4
```
behackett, line 253: On Windows. It seems to work just fine everywhere else.

doc/examples/tls.rst
```diff
@@ -249,3 +249,14 @@ revocation checking failed::
...
+...........................................................
+
+Note that Python 3.10+ is incompatible with TLS/SSL on MongoDB <= 3.4. The
+following are some example errors that may occur with this combination::
+
+  SSL handshake failed: localhost:27017: [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:997)
```
behackett, line 259: I think there is also an error about cipher mismatch, which is the real problem.

behackett: From the ticket description, we have this in the server log:
```SSL: error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher```

.evergreen/config.yml
ShaneHarvey: Done.

.evergreen/config.yml
ShaneHarvey: Oh right, I was confused by the other test failure. PyOpenSSL+3.10+MongoDB 2.6 works fine.

doc/examples/tls.rst
ShaneHarvey: Done.

.evergreen/config.yml
ShaneHarvey: Done.

.evergreen/config.yml
ShaneHarvey: Done.

doc/examples/tls.rst
ShaneHarvey: I tested locally on macos and it's the same error behavior due to no shared ciphers.

behackett: Thanks for checking.

behackett (approved):

PR #749, opened 2021-09-29
PYTHON-2823 Allow custom service names with srvServiceName URI option

juliusgeo:


pymongo/srv_resolver.py
```diff
@@ -83,8 +85,8 @@ def get_options(self):
 
     def _resolve_uri(self, encapsulate_errors):
         try:
-            results = _resolve('_mongodb._tcp.' + self.__fqdn, 'SRV',
-                               lifetime=self.__connect_timeout)
+            results = _resolve('_'+str(self.__srv)+'._tcp.' + self.__fqdn,
```
behackett, line 88: Add spaces around the '+' characters. Also, why do you have to wrap self.__srv in str? Is it being converted to a bytes before it gets here?

pymongo/uri_parser.py
```diff
@@ -514,6 +519,9 @@ def parse_uri(uri, default_port=DEFAULT_PORT, validate=True, warn=False,
                     options[opt] = val
         if "tls" not in options and "ssl" not in options:
             options["tls"] = True if validate else 'true'
+    #elif not is_srv and options.get("srvServiceName", None) is not None:
```
behackett, line 522: Remove the dead code.

behackett (changes requested):
You need to add documentation for the new option in the docstring for MongoClient, and a mention in the changelog.

pymongo/srv_resolver.py
juliusgeo: I think you are right and we can remove that.

pymongo/uri_parser.py
juliusgeo: This isn't dead code, it was to fix the test that was testing for a configuration to be raised under those conditions. So I restored it.

test/srv_seedlist/replica-set/srv-service-name.json
```diff
@@ -0,0 +1,16 @@
...
+    "localhost.test.build.10gen.cc:27018"
+  ],
+  "hosts": [
+    "localhost:27017",
+    "localhost:27018",
+    "localhost:27019"
+  ],
+  "options": {
+    "ssl": true,
+    "srvServiceName": "customname"
```
ShaneHarvey, line 14: Does the srv_seedlist test suite actually check the `srvServiceName` option here? EG would the test fail with 
`"srvServiceName": "wrongname"`

test/uri_options/srv-service-name-option.json
```diff
@@ -0,0 +1,24 @@
...
+  "tests": [
+    {
+      "description": "SRV URI with custom srvServiceName",
+      "uri": "mongodb+srv://test22.test.build.10gen.cc/?srvServiceName=customname",
+      "valid": true,
+      "warning": false,
+      "hosts": null,
+      "auth": null,
+      "options": {
+        "srvServiceName": "customname"
```
ShaneHarvey, line 11: Does the uri_options test suite actually check the srvServiceName option here? EG would the test fail with 
`"srvServiceName": "wrongname"`

ShaneHarvey (changes requested):
From a grep through the codebase we also need to pass the new option to the _SrvResolver in pymongo/monitor.py. It's somewhat concerning that the tests pass without this change (but not your fault at all of course). Does the spec change have any test coverage for mongos SRV polling?

pymongo/srv_resolver.py
```diff
@@ -47,8 +47,10 @@ def _resolve(*args, **kwargs):
     "Did you mean to use 'mongodb://'?")
 
 class _SrvResolver(object):
-    def __init__(self, fqdn, connect_timeout=None):
+    def __init__(self, fqdn,
+                 connect_timeout=None, srv_service_name="mongodb"):
```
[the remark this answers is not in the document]
juliusgeo: Nevermind I figured out the plumbing. However, I don't think that connect_timeout is always passed when creating a _SrvResolver.

test/uri_options/srv-service-name-option.json
juliusgeo: Yes it does, it compares the value parsed from the uri with the expected value.

pymongo/mongo_client.py
```diff
@@ -503,6 +507,10 @@ def __init__(
            arguments.
            The default for `uuidRepresentation` was changed from
            ``pythonLegacy`` to ``unspecified``.
+           There is a new parameter for SRV URIs ``srvServiceName`` which
+           can be used like so::
+
+                MongoClient("mongodb+srv://test22.test.build.10gen.cc/?srvServiceName=customname")
```
ShaneHarvey, line 513: We should not repeat the docs here. We should just say:
```
Added the ``srvServiceName`` URI and keyword argument.
```

pymongo/mongo_client.py
```diff
@@ -201,6 +201,10 @@ def __init__(
             and decoding of custom types.
 
           | **Other optional parameters can be passed as keyword arguments:**
+          - Starting in version 4.0 you can pass a custom SRV service name by
+            using the ``srvServiceName`` option like so::
+
+                MongoClient("mongodb+srv://test22.test.build.10gen.cc/?srvServiceName=customname")
```
ShaneHarvey, line 207: Let's move this to the bottom of this list (under unicode_decode_error_handler). Also please match the doc formatting and style of the other options here.

pymongo/mongo_client.py
```diff
@@ -730,6 +740,7 @@ def __init__(
             server_selector=options.server_selector,
             heartbeat_frequency=options.heartbeat_frequency,
             fqdn=fqdn,
+            srv_service_name=srv_service_name,
```
ShaneHarvey, line 743: Please add a test to make sure that srvServiceName is supported via kwargs as well as in the URI. For example:
```python
client = MongoClient('mongodb+srv://...', srvServiceName='namefromkwarg')
self.assertEqual(client's srvServiceName, 'namefromkwarg')
```

pymongo/monitor.py
```diff
@@ -316,7 +317,9 @@ def _get_seedlist(self):
         Returns a list of ServerDescriptions.
         """
         try:
-            seedlist, ttl = _SrvResolver(self._fqdn).get_hosts_and_min_ttl()
+            seedlist, ttl = _SrvResolver(self._fqdn,
+                                         srv_service_name=self._srv_service_name
+                                         ).get_hosts_and_min_ttl()
```
ShaneHarvey, line 322: The style here would be cleaner like this:
```python
        try:
            resolver = _SrvResolver(
                self._fqdn, 
                srv_service_name=self._srv_service_name)
            seedlist, ttl = resolver.get_hosts_and_min_ttl()
```

pymongo/srv_resolver.py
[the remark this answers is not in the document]
ShaneHarvey: connect_timeout needs to always be passed. The fact that it's not already is an oversight. Please update SrvMonitor to get the timeout from topology_settings (self._settings).

pymongo/srv_resolver.py
```diff
@@ -83,8 +85,8 @@ def get_options(self):
 
     def _resolve_uri(self, encapsulate_errors):
         try:
-            results = _resolve('_mongodb._tcp.' + self.__fqdn, 'SRV',
-                               lifetime=self.__connect_timeout)
+            results = _resolve('_' + self.__srv + '._tcp.' + self.__fqdn,
+                                'SRV', lifetime=self.__connect_timeout)
```
ShaneHarvey, line 89: The indentation here still looks off, should be:
```python
            results = _resolve('_' + self.__srv + '._tcp.' + self.__fqdn,
                               'SRV', lifetime=self.__connect_timeout)
```

doc/changelog.rst
```diff
@@ -134,8 +134,10 @@ decodes datetime as naive by default.
 - The ``hint`` option is now required when using ``min`` or ``max`` queries
   with :meth:`~pymongo.collection.Collection.find`.
 - ``name`` is now a required argument for the :class:`pymongo.driver_info.DriverInfo` class.
+- When providing a URI to the :class:`~pymongo.mongo_client.MongoClient`
```
ShaneHarvey, line 137: Suggest: `When providing a "mongodb+srv://" URI to :class:...`

ShaneHarvey (changes requested):

pymongo/mongo_client.py

pymongo/mongo_client.py

pymongo/mongo_client.py
```diff
@@ -667,6 +676,7 @@ def __init__(
                 dbase = res["database"] or dbase
                 opts = res["options"]
                 fqdn = res["fqdn"]
+                srv_service_name = opts.get("srvServiceName", "mongodb")
```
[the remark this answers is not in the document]

pymongo/srv_resolver.py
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/monitor.py
juliusgeo: done.

pymongo/mongo_client.py
juliusgeo: Done.

pymongo/srv_resolver.py
juliusgeo: Done.

pymongo/common.py
```diff
@@ -113,6 +113,9 @@
 # From the driver sessions spec.
 _MAX_END_SESSIONS = 10000
 
+# Default value for srvServiceName
+SRV_SERVICE_NAME_DEFAULT = "mongodb"
```
ShaneHarvey, line 117: I'd omit the `_DEFAULT` because none of the other constants have it.

pymongo/mongo_client.py
```diff
@@ -201,8 +201,7 @@ def __init__(
             and decoding of custom types.
 
           | **Other optional parameters can be passed as keyword arguments:**
-
-          - `directConnection` (optional): if ``True``, forces this client to
```
ShaneHarvey, line 205: Please undo this change.

pymongo/mongo_client.py
```diff
@@ -333,6 +332,11 @@ def __init__(
             a Unicode-related error occurs during BSON decoding that would
             otherwise raise :exc:`UnicodeDecodeError`. Valid options include
             'strict', 'replace', and 'ignore'. Defaults to 'strict'.
+          - Starting in version 4.0 you can pass a custom SRV service name by
+            using the ``srvServiceName`` option like so::
```
ShaneHarvey, line 336: This does not match how we document the other options. The line needs to start with:
```
- `<option name>`: description.
```

There's also no need to mention the `Starting in version 4.0` part.

pymongo/uri_parser.py
```diff
@@ -20,7 +20,7 @@
 
 from urllib.parse import unquote_plus
 
-from pymongo.common import (
+from pymongo.common import (SRV_SERVICE_NAME_DEFAULT,
```
ShaneHarvey, line 23: This import should be on a new line to match the existing style.

test/test_client.py
```diff
@@ -1601,6 +1601,12 @@ def test_network_error_message(self):
             with self.assertRaisesRegex(AutoReconnect, expected):
                 client.pymongo_test.test.find_one({})
 
+    def test_service_name_from_kwargs(self):
+        client = MongoClient('mongodb+srv://test22.test.build.10gen.cc',
+                             srvServiceName='namefromkwarg')
+        self.assertEqual(client._topology_settings._srv_service_name,
+        'namefromkwarg')
```
ShaneHarvey, line 1608: Indentation.

test/test_client.py
```diff
@@ -1601,6 +1601,12 @@ def test_network_error_message(self):
             with self.assertRaisesRegex(AutoReconnect, expected):
                 client.pymongo_test.test.find_one({})
 
+    def test_service_name_from_kwargs(self):
+        client = MongoClient('mongodb+srv://test22.test.build.10gen.cc',
+                             srvServiceName='namefromkwarg')
```
ShaneHarvey, line 1606: Please add connect=False since this test doesn't need to open the topology.

test/test_client.py
```diff
@@ -1601,6 +1601,19 @@ def test_network_error_message(self):
...
+            'mongodb+srv://user:password@test22.test.build.10gen.cc',
+            srvServiceName='customname', connect=False)
+        self.assertEqual(client._topology_settings._srv_service_name,
+                         'customname')
+        client = MongoClient(
+             'mongodb+srv://user:password@test22.test.build.10gen.cc'
+             '/?srvServiceName=customname',
+             connect=False)
+        self.assertEqual(client._topology_settings._srv_service_name,
+                         'customname')
```
ShaneHarvey, line 1615: We need a test which actually checks that the kwarg overrides the URI. Like this:
```python
        client = MongoClient(
             'mongodb+srv://user:password@test22.test.build.10gen.cc'
             '/?srvServiceName=shouldBeOverridden',
             srvServiceName='customname',
             connect=False)
        self.assertEqual(client._topology_settings._srv_service_name,
                         'customname')
```

pymongo/uri_parser.py
```diff
@@ -373,7 +374,7 @@ def _check_options(nodes, options):
 
 
 def parse_uri(uri, default_port=DEFAULT_PORT, validate=True, warn=False,
-              normalize=True, connect_timeout=None):
+              normalize=True, connect_timeout=None, srv_service_name=None):
```
ShaneHarvey, line 377: Please update the docstring below to reflect the new srv_service_name param.

pymongo/mongo_client.py
```diff
@@ -197,7 +197,6 @@ def __init__(
             and decoding of custom types.
 
           | **Other optional parameters can be passed as keyword arguments:**
-
```
ShaneHarvey, line 200: Please undo this change.

pymongo/mongo_client.py
```diff
@@ -329,6 +328,10 @@ def __init__(
             a Unicode-related error occurs during BSON decoding that would
             otherwise raise :exc:`UnicodeDecodeError`. Valid options include
             'strict', 'replace', and 'ignore'. Defaults to 'strict'.
+          - `srvServiceName`: A custom SRV service name. Use it like so::
```
ShaneHarvey, line 331: I suggest:
```
`srvServiceName`: (string) The SRV service name to use for "mongodb+srv://" URIs. Defaults to "mongodb". Use it like so::
```

ShaneHarvey:
> From a grep through the codebase we also need to pass the new option to the _SrvResolver in pymongo/monitor.py. It's somewhat concerning that the tests pass without this change (but not your fault at all of course). Does the spec change have any test coverage for mongos SRV polling?

I know you've already addressed the option plumbing but do you have any thoughts on the test coverage issue?

pymongo/mongo_client.py
juliusgeo: Done.

pymongo/mongo_client.py
juliusgeo: Done.

ShaneHarvey (approved):
LGTM! 

> I'm not sure about the test coverage. ...

Please file a DRIVERS ticket to request test coverage here. The problem is that there is no test coverage for SRV polling with a custom service name.

PR #750, opened 2021-10-01
PYTHON-2928 Fix mod_wsgi test failures

juliusgeo:


test/mod_wsgi_test/mod_wsgi_test.wsgi
```diff
@@ -41,7 +41,7 @@ ndocs = 20
 
 collection.drop()
 collection.insert_many([{'i': i} for i in range(ndocs)])
-client.close()  # Discard main thread's request socket.
```
ShaneHarvey, line 44: I understand why removing this line fixes the test failure but is this line important? It claims to be discarding the "main thread's request socket" which seems like it might be important for the test. This was added in https://jira.mongodb.org/browse/PYTHON-353. Reading through that issue might give a clue.

I suggest we err on the side of caution and do this:
```python
client = MongoClient()
collection = client.test.test
ndocs = 20
collection.drop()
collection.insert_many([{'i': i} for i in range(ndocs)])
client.close()  # Discard main thread's request socket.
client = MongoClient()
collection = client.test.test
```

ShaneHarvey (changes requested):

test/mod_wsgi_test/mod_wsgi_test.wsgi
juliusgeo: Done.

juliusgeo: Done.

ShaneHarvey:
Please schedule the mod_wsgi test tasks in the latest patch build (https://spruce.mongodb.com/version/615ce0f31e2d175df43cc025)

ShaneHarvey (approved):
LGTM

PR #752, opened 2021-10-06
PYTHON-2938 Fix race condition caused by MongoClient._process_periodic_tasks(client)

juliusgeo:


pymongo/mongo_client.py
```diff
@@ -738,7 +738,13 @@ def target():
...
             if client is None:
                 return False  # Stop the executor.
-            MongoClient._process_periodic_tasks(client)
+            try:
+                MongoClient._process_periodic_tasks(client)
+            except InvalidOperation as e:
+                if client._get_topology._closed:
+                    return False
+                else:
+                    raise e
```
ShaneHarvey, line 747: I don't think this will actually work. The problem is that `_process_periodic_tasks` already catches and logs the InvalidOperation error which is want we want to prevent. We need to refactor this kind of logic so that we can ignore InvalidOperation without logging it at all:
```python
            try:
                self._cleanup_cursor(True, cursor_id, address, sock_mgr,
                                     None, False)
            except Exception:
                helpers._handle_exception()
```

It would also be good to add a test for this. Perhaps something like this would work but it requires some experimentation:
```python
coll.insert_many([{} for _ in range(5])
cursor = coll.find(batchSize=2)
cursor.next()
assert cursor.cursor_id
client.close()
# Add cursor to kill cursors queue 
del cursor
wait_until(cursor_id is added to client's kill cursors queue)
client._process_periodic_tasks()   # This must not raise or print any exceptions
```

test/test_client.py
```diff
@@ -1591,6 +1591,25 @@ def test_network_error_message(self):
...
+        # Add cursor to kill cursors queue
+        del cursor
+        wait_until(lambda: c_id in [c for _, c, _ in
+                                    client._MongoClient__kill_cursors_queue],
+                   "waited for cursor to be added to queue")
+        try:
+            client._process_periodic_tasks()  # This must not raise or print any exceptions
+        except Exception:
+            self.fail("client._process_periodic_tasks() raised an exception")
+
```
ShaneHarvey, line 1613: We don't need the try/except/fail here. Just call client._process_periodic_tasks(). If that raises unexpectedly the teat will fail automatically.

test/test_client.py
```diff
@@ -1591,6 +1591,25 @@ def test_network_error_message(self):
...
                 client.pymongo_test.test.find_one({})
 
+    def test_process_periodic_tasks(self):
+        client = rs_or_single_client()
+        coll = client.db.collection
+        coll.insert_many([{} for _ in range(5)])
+        cursor = coll.find(batch_size=2)
+        cursor.next()
+        c_id = cursor.cursor_id
+        assert c_id
```
ShaneHarvey, line 1601: Should use self.assertIsNotNone here.

test/test_client.py
```diff
@@ -1591,6 +1591,25 @@ def test_network_error_message(self):
             with self.assertRaisesRegex(AutoReconnect, expected):
                 client.pymongo_test.test.find_one({})
 
+    def test_process_periodic_tasks(self):
```
ShaneHarvey, line 1594: I suspect this test will fail on PyPy because garbage collection works slightly differently. Let's skip this test entirely when running with PyPy (see my current PR for how to do that)

pymongo/mongo_client.py
```diff
@@ -1583,6 +1583,8 @@ def _process_kill_cursors(self):
             try:
                 self._cleanup_cursor(True, cursor_id, address, sock_mgr,
                                      None, False)
+            except InvalidOperation as e:
+                raise e
             except Exception:
                 helpers._handle_exception()
```
ShaneHarvey, line 1589: There's another try/except thats needs to be updated below too.

pymongo/mongo_client.py
```diff
@@ -1600,9 +1602,11 @@ def _process_kill_cursors(self):
     def _process_periodic_tasks(self):
         """Process any pending kill cursors requests and
         maintain connection pool parameters."""
-        self._process_kill_cursors()
         try:
+            self._process_kill_cursors()
             self._topology.update_pool(self.__all_credentials)
+        except InvalidOperation:
+            pass
```
ShaneHarvey, line 1609: We only want to ignore this error if the client was closed.

ShaneHarvey (changes requested):

pymongo/mongo_client.py
juliusgeo: Done.

test/test_client.py
juliusgeo: Done.

pymongo/mongo_client.py
juliusgeo: Done.

test/test_client.py
juliusgeo: Done.

pymongo/mongo_client.py
```diff
@@ -1583,6 +1583,11 @@ def _process_kill_cursors(self):
             try:
                 self._cleanup_cursor(True, cursor_id, address, sock_mgr,
                                      None, False)
+            except InvalidOperation as e:
+                if self._topology._closed:
```
ShaneHarvey, line 1587: It's worth adding a comment here to explain why we're raising the error when the client is closed.

pymongo/mongo_client.py
```diff
@@ -1583,6 +1583,11 @@ def _process_kill_cursors(self):
...
+                    raise e
```
ShaneHarvey, line 1588: You should use bare `raise` instead of `raise e`.

pymongo/mongo_client.py
```diff
@@ -1600,9 +1605,14 @@ def _process_kill_cursors(self):
...
-        self._process_kill_cursors()
         try:
+            self._process_kill_cursors()
             self._topology.update_pool(self.__all_credentials)
+        except InvalidOperation:
+            if self._topology._closed:
+                pass
+            else:
+                helpers._handle_exception()
         except Exception:
```
ShaneHarvey, line 1616: This might be cleaner:
```python
except Exception as exc:
    if isinstance(exc, InvalidOperation) and self._topology._closed:
        return # Ignore when client has been closed. 
```

pymongo/mongo_client.py
ShaneHarvey: It's still missing (line 1602).

pymongo/mongo_client.py

pymongo/mongo_client.py

pymongo/mongo_client.py

juliusgeo:
> We need to fix this bug:
> 
> ```
>  [2021/10/08 02:08:52.536] ERROR: MongoClient has no attribute '_MongoClient__kill_cursors_queuee'. To access the _MongoClient__kill_cursors_queuee database, use client['_MongoClient__kill_cursors_queuee']. (AttributeError)
>  [2021/10/08 02:08:52.536] Traceback (most recent call last):
>  [2021/10/08 02:08:52.536]   File "/data/mci/4133ce1e64692bb1a00e7fbf98b2478b/src/test/test_client.py", line 1607, in test_process_periodic_tasks
>  [2021/10/08 02:08:52.536]     "waited for cursor to be added to queue")
>  [2021/10/08 02:08:52.536]   File "/data/mci/4133ce1e64692bb1a00e7fbf98b2478b/src/test/utils.py", line 750, in wait_until
>  [2021/10/08 02:08:52.536]     retval = predicate()
>  [2021/10/08 02:08:52.536]   File "/data/mci/4133ce1e64692bb1a00e7fbf98b2478b/src/test/test_client.py", line 1606, in <lambda>
>  [2021/10/08 02:08:52.536]     wait_until(lambda: client._MongoClient__kill_cursors_queuee,
>  [2021/10/08 02:08:52.536]   File "/data/mci/4133ce1e64692bb1a00e7fbf98b2478b/src/pymongo/mongo_client.py", line 1469, in __getattr__
>  [2021/10/08 02:08:52.536]     " database, use client[%r]." % (name, name, name))
>  [2021/10/08 02:08:52.536] AttributeError: MongoClient has no attribute '_MongoClient__kill_cursors_queuee'. To access the _MongoClient__kill_cursors_queuee database, use client['_MongoClient__kill_cursors_queuee'].
> ```

This typo has been fixed.

pymongo/mongo_client.py
```diff
@@ -1607,17 +1612,24 @@ def _process_kill_cursors(self):
                     self._kill_cursors(
                         cursor_ids, address, topology, session=None)
                 except Exception:
-                    helpers._handle_exception()
+                    if (isinstance(e,InvalidOperation) and
```
ShaneHarvey, line 1615: Missing a space between `e,` and InvalidOperation.

pymongo/mongo_client.py
```diff
@@ -1596,8 +1596,13 @@ def _process_kill_cursors(self):
             try:
                 self._cleanup_cursor(True, cursor_id, address, sock_mgr,
                                      None, False)
-            except Exception:
-                helpers._handle_exception()
+            except Exception as e:
```
ShaneHarvey, line 1599: Please rename e to exc everywhere. We usually use exc as the name for exceptions.

ShaneHarvey (changes requested):
Two minor comments

ShaneHarvey (approved):
LGTM

PR #754, opened 2021-10-12
PYTHON-2832 Provide options to limit number of mongos servers used in connecting to sharded clusters

juliusgeo:


pymongo/mongo_client.py
```diff
@@ -651,7 +651,7 @@ def __init__(
         opts = common._CaseInsensitiveDictionary()
         fqdn = None
         srv_service_name = keyword_opts.get("srvservicename", None)
```
ShaneHarvey, line 653: Please replace `keyword_opts.get("srvservicename", None)` with `keyword_opts.get("srvservicename")`

pymongo/monitor.py
```diff
@@ -319,7 +320,7 @@ def _get_seedlist(self):
         try:
             resolver = _SrvResolver(self._fqdn,
                                     self._settings.pool_options.connect_timeout,
-                                    self._srv_service_name)
+                                    self._settings._srv_max_hosts)
```
ShaneHarvey, line 323: This looks off. Don't we need to pass *both* `srv_service_name` and `srv_max_hosts`?

pymongo/srv_resolver.py
```diff
@@ -24,7 +24,7 @@
 
 from pymongo.common import CONNECT_TIMEOUT
 from pymongo.errors import ConfigurationError
-
+from random import sample
```
ShaneHarvey, line 27: Imports from the std lib should go before pymongo imports and before 3rd party (dns) imports.

pymongo/srv_resolver.py
```diff
@@ -48,11 +48,11 @@ def _resolve(*args, **kwargs):
 
 class _SrvResolver(object):
     def __init__(self, fqdn,
-                 connect_timeout, srv_service_name):
+                 connect_timeout, srv_service_name, srv_max_hosts):
         self.__fqdn = fqdn
         self.__srv = srv_service_name
         self.__connect_timeout = connect_timeout or CONNECT_TIMEOUT
-
+        self.__srv_max_hosts = srv_max_hosts
```
ShaneHarvey, line 55: This should be ` srv_max_hosts or 0` so that we don't have to deal with `self.__srv_max_hosts is not None ` checks.

pymongo/srv_resolver.py
```diff
@@ -97,6 +97,9 @@ def _resolve_uri(self, encapsulate_errors):
 
     def _get_srv_response_and_hosts(self, encapsulate_errors):
         results = self._resolve_uri(encapsulate_errors)
+        if (self.__srv_max_hosts != 0 and self.__srv_max_hosts is not None) \
+                and self.__srv_max_hosts <= len(results):
+            results = sample(results, self.__srv_max_hosts)
```
ShaneHarvey, line 102: Is this what the spec says to do? The way this PR is currently written, the driver will completely replace all the hosts with a new random set of hosts every time the SRV record gets polled (every 60 seconds I think). That does not sound like a good idea to me.

test/test_srv_polling.py
```diff
@@ -108,8 +108,11 @@ def assert_nodelist_change(self, expected_nodelist, client):
         """Check if the client._topology eventually sees all nodes in the
         expected_nodelist.
         """
+
         def predicate():
             nodelist = self.get_nodelist(client)
+            print("nodelist seen", nodelist)
+            print("nodelist expected", expected_nodelist)
```
ShaneHarvey, line 115: Is this needed?

pymongo/mongo_client.py

pymongo/monitor.py
juliusgeo: Oh oops, yeah that is a typo. Fixed.

pymongo/srv_resolver.py
juliusgeo: Done.

pymongo/srv_resolver.py
juliusgeo: Done.

pymongo/srv_resolver.py
juliusgeo: I'm not sure how to find the relevant section of the spec. I took a look around in Jira but could not find any reference to a specification for this behavior.

test/test_srv_polling.py
juliusgeo: Nope, removed.

pymongo/topology_description.py
```diff
@@ -447,9 +447,21 @@ def _updated_topology_description_srv_polling(topology_description, seedlist):
...
     # Add SDs corresponding to servers recently added to the SRV record.
-    for address in seedlist:
-        if address not in sds:
+    if topology_description._srv_max_hosts != 0:
+        available_new_hosts = set(seedlist) - set(sds.keys())
+        new_hosts_needed = topology_description._srv_max_hosts - len(sds)
+        if new_hosts_needed > 0:
+            seedlist = random.sample(available_new_hosts, min(new_hosts_needed,
+                                                              len(available_new_hosts)))
+        elif new_hosts_needed < 0:
```
ShaneHarvey, line 456: Since srv_max_hosts is set once and never changes, shouldn't it be impossible to have more than that many hosts? If so this is dead code that we can just remove.

pymongo/topology_description.py
```diff
@@ -447,9 +447,21 @@ def _updated_topology_description_srv_polling(topology_description, seedlist):
...
+            while len(sds) > topology_description._srv_max_hosts:
+                sds.pop()
+        for address in seedlist:
             sds[address] = ServerDescription(address)
+    else:
+        for address in seedlist:
+            if address not in sds:
+                sds[address] = ServerDescription(address)
```
ShaneHarvey, line 464: We should refactor this to avoid duplicating the for loop which does: `sds[address] = ServerDescription(address)`.

pymongo/monitor.py
```diff
@@ -300,6 +300,7 @@ def __init__(self, topology, topology_settings):
         self._seedlist = self._settings._seeds
         self._fqdn = self._settings.fqdn
         self._srv_service_name = self._settings._srv_service_name
+        self._srv_max_hosts = self._settings._srv_max_hosts
```
ShaneHarvey, line 303: We no longer need to pass srv_max_hosts here do we?

pymongo/topology_description.py
juliusgeo: srv_max_hosts never changes but I don't understand how that means that it's impossible to have more than that number of hosts returned. This is the code section that enforces the number of hosts. Seedlist should just be every discovered seed host right? There's nothing about the seedlist that enforces a limit on the number of hosts that it gets from _SrvResolver, we have to impose that limit ourselves by popping off seeds or adding seeds as necessary. I might have misunderstood, though.

pymongo/topology_description.py
juliusgeo: Done.

pymongo/monitor.py
juliusgeo: You're right, done.

pymongo/topology_description.py
```diff
@@ -447,6 +447,15 @@ def _updated_topology_description_srv_polling(topology_description, seedlist):
         return topology_description
 
     # Add SDs corresponding to servers recently added to the SRV record.
+    if topology_description._srv_max_hosts != 0:
```
ShaneHarvey, line 450: This logic needs to happen *after* the `Remove SDs corresponding to servers no longer part of the SRV record` logic. Otherwise old hosts might never be removed.

ShaneHarvey (changes requested):

pymongo/srv_resolver.py
```diff
@@ -111,9 +113,12 @@ def _get_srv_response_and_hosts(self, encapsulate_errors):
                 raise ConfigurationError("Invalid SRV host: %s" % (node[0],))
             if self.__plist != nlist:
                 raise ConfigurationError("Invalid SRV host: %s" % (node[0],))
-
+        if(self.__srv_max_hosts != 0):
+            ret = list(zip(results, nodes))
+            random.shuffle(ret)
```
ShaneHarvey, line 118: Why isn't this using random.sample? Ideally we can refactor this logic into a single place that's shared between the initial discovery phase and _updated_topology_description_srv_polling.

test/test_srv_polling.py
```diff
@@ -108,6 +108,7 @@ def assert_nodelist_change(self, expected_nodelist, client):
         """Check if the client._topology eventually sees all nodes in the
         expected_nodelist.
         """
+
```
ShaneHarvey, line 111: unneeded.

test/test_client.py
```diff
@@ -1629,7 +1629,23 @@ def test_service_name_from_kwargs(self):
              connect=False)
         self.assertEqual(client._topology_settings._srv_service_name,
                          'customname')
-
+    def test_srv_max_hosts(self):
```
ShaneHarvey, line 1632: What about all the other new tests added in https://github.com/mongodb/specifications/pull/1069?

ShaneHarvey (changes requested):

pymongo/srv_resolver.py
juliusgeo: So actually I think we want to use the shuffle function in both places because the shuffle function basically does exactly what the go driver implements (https://stackoverflow.com/questions/12798363/what-is-the-difference-between-random-sample-and-random-shuffle-in-python), and I also think it looks cleaner and is easier to read. But I wanted to make sure that this code made sense to you before changing both of them.

test/test_client.py
juliusgeo: I will resync spec tests.

pymongo/srv_resolver.py
```diff
@@ -111,9 +113,11 @@ def _get_srv_response_and_hosts(self, encapsulate_errors):
                 raise ConfigurationError("Invalid SRV host: %s" % (node[0],))
             if self.__plist != nlist:
                 raise ConfigurationError("Invalid SRV host: %s" % (node[0],))
-
+        if(self.__srv_max_hosts != 0):
+            ret = list(zip(results, nodes))
+            ret = random.sample(ret, min(self.__srv_max_hosts, len(ret)))
+            return [i[0] for i in ret], [i[1] for i in ret]
```
ShaneHarvey, line 119: You can "undo" the zip like this:
```
results, nodes = list(zip(*ret))
```

juliusgeo: Done.

pymongo/topology_description.py
```diff
@@ -446,16 +451,26 @@ def _updated_topology_description_srv_polling(topology_description, seedlist):
...
-            sds[address] = ServerDescription(address)
 
     # Remove SDs corresponding to servers no longer part of the SRV record.
     for address in list(sds.keys()):
         if address not in seedlist:
             sds.pop(address)
 
+    if topology_description.srv_max_hosts != 0:
+        available_new_hosts = set(seedlist) - set(sds.keys())
+        new_hosts_needed = topology_description.srv_max_hosts - len(sds)
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/topology_description.py
```diff
@@ -446,16 +451,26 @@ def _updated_topology_description_srv_polling(topology_description, seedlist):
...
+        if new_hosts_needed > 0:
+            seedlist = sample(available_new_hosts, min(new_hosts_needed,
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/topology_description.py
```diff
@@ -446,16 +451,26 @@ def _updated_topology_description_srv_polling(topology_description, seedlist):
...
+                                                              len(available_new_hosts)))
+        elif new_hosts_needed < 0:
+            while (len(sds) >
+                    topology_description.srv_max_hosts):
+                sds.pop()
```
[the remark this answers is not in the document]
juliusgeo: Done.

test/test_srv_polling.py
```diff
@@ -230,6 +230,47 @@ def erroring_seedlist():
...
+            client = MongoClient(self.CONNECTION_STRING, srvMaxHosts=2)
+            #self.assert_nodelist_change(response, client)
+            with SrvPollingKnobs(nodelist_callback=nodelist_callback):
+                sleep(2*common.MIN_SRV_RESCAN_INTERVAL)
+                expected_topology = {
+                    ("localhost.test.build.10gen.cc",  27019),
+                    ("localhost.test.build.10gen.cc",  27020)}
+                final_topology = set(
+                    client.topology_description.server_descriptions())
+                self.assertSetEqual(final_topology, final_topology | expected_topology)
```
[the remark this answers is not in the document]
juliusgeo: Because we don't want to assert that they are equal because the spec never says that they are equal, just that `expected_topology` is a subset of `final_topology`.

pymongo/srv_resolver.py
```diff
@@ -15,17 +15,19 @@
...
 import ipaddress
+import random
 
 try:
     from dns import resolver
     _HAVE_DNSPYTHON = True
 except ImportError:
     _HAVE_DNSPYTHON = False
 
+from random import sample
```
ShaneHarvey, line 26: This import is not needed since you use `random.sample` below.

test/test_srv_polling.py
```diff
@@ -230,6 +230,50 @@ def erroring_seedlist():
             raise ConfigurationError
         self._test_recover_from_initial(erroring_seedlist)
 
+    def test_10_all_dns_selected(self):
+        response = [("localhost.test.build.10gen.cc", 27017),
+                    ("localhost.test.build.10gen.cc", 27017),
```
ShaneHarvey, line 235: 27017 appears twice here. Is that a bug?

test/test_srv_polling.py
```diff
@@ -230,6 +230,50 @@ def erroring_seedlist():
...
+            with SrvPollingKnobs(nodelist_callback=nodelist_callback):
+                self.assert_nodelist_change(response, client)
+                sleep(2*common.MIN_SRV_RESCAN_INTERVAL)
+                expected_topology = {
+                    ("localhost.test.build.10gen.cc", 27017),
+                    ("localhost.test.build.10gen.cc",  27019),
+                    ("localhost.test.build.10gen.cc",  27020)}
+                final_topology = set(
+                    client.topology_description.server_descriptions())
+                self.assertSetEqual(final_topology, final_topology |
```
ShaneHarvey, line 253: please remove :
```python
                expected_topology = {
                    ("localhost.test.build.10gen.cc", 27017),
                    ("localhost.test.build.10gen.cc",  27019),
                    ("localhost.test.build.10gen.cc",  27020)}
                final_topology = set(
                    client.topology_description.server_descriptions())
                self.assertSetEqual(final_topology, final_topology |
```

These lines are all redundant because `self.assert_nodelist_change(response, client)` already does the same assertion.

test/test_srv_polling.py
```diff
@@ -230,6 +230,50 @@ def erroring_seedlist():
...
+            client = MongoClient(self.CONNECTION_STRING, srvMaxHosts=2)
+            with SrvPollingKnobs(nodelist_callback=nodelist_callback):
+                self.assert_nodelist_change(response, client)
+                sleep(2*common.MIN_SRV_RESCAN_INTERVAL)
+                expected_topology = {
+                    ("localhost.test.build.10gen.cc",  27019),
+                    ("localhost.test.build.10gen.cc",  27020)}
+                final_topology = set(
+                    client.topology_description.server_descriptions())
+                self.assertSetEqual(final_topology, final_topology |
```
ShaneHarvey, line 274: Same comment about removing the redundant lines.

test/test_srv_polling.py
```diff
@@ -230,6 +230,50 @@ def erroring_seedlist():
...
+                    ("localhost.test.build.10gen.cc", 27020)]
+
+        def nodelist_callback():
+            return response
+
+        with SrvPollingKnobs(
+                ttl_time=WAIT_TIME, min_srv_rescan_interval=WAIT_TIME):
+            client = MongoClient(self.CONNECTION_STRING, srvMaxHosts=2)
+            with SrvPollingKnobs(nodelist_callback=nodelist_callback):
+                self.assert_nodelist_change(response, client)
```
ShaneHarvey, line 267: This is the assertion that fails:
```
 [2021/10/21 19:40:48.759] FAIL [10.041s]: test_11_all_dns_selected (test_srv_polling.TestSrvPolling)
 [2021/10/21 19:40:48.759] ----------------------------------------------------------------------
 [2021/10/21 19:40:48.760] Traceback (most recent call last):
 [2021/10/21 19:40:48.760]   File "/data/mci/6c8659548e53884acc9814e9e91b6644/src/test/test_srv_polling.py", line 267, in test_11_all_dns_selected
 [2021/10/21 19:40:48.760]     self.assert_nodelist_change(response, client)
 [2021/10/21 19:40:48.760]   File "/data/mci/6c8659548e53884acc9814e9e91b6644/src/test/test_srv_polling.py", line 116, in assert_nodelist_change
 [2021/10/21 19:40:48.760]     wait_until(predicate, "see expected nodelist", timeout=100*WAIT_TIME)
 [2021/10/21 19:40:48.760]   File "/data/mci/6c8659548e53884acc9814e9e91b6644/src/test/utils.py", line 755, in wait_until
 [2021/10/21 19:40:48.760]     raise AssertionError("Didn't ever %s" % success_description)
 [2021/10/21 19:40:48.760] AssertionError: Didn't ever see expected nodelist
```
What is the current TopologyDescription when it fails? That would help determine what is not working as expected.

ShaneHarvey (changes requested):

pymongo/srv_resolver.py
juliusgeo: Done.

test/test_srv_polling.py
juliusgeo: https://github.com/mongodb/specifications/blob/master/source/polling-srv-records-for-mongos-discovery/tests/README.rst#configuring-the-test-environment So here it says that you should have two with 27017 but they both resolve to the same thing so I'm not sure why. I was just following exactly what the test said.

test/test_srv_polling.py
juliusgeo: I think that removing self.assert_nodelist_change would be better because the code I wrote actually checks for if it's a subset, which is what the spec says to do.

test/test_srv_polling.py
juliusgeo: <TopologyDescription id: 6171c8203509121876af7cf4, topology_type: Unknown, servers: [<ServerDescription ('localhost.test.build.10gen.cc', 27017) server_type: Unknown, rtt: None>, <ServerDescription ('localhost.test.build.10gen.cc', 27018) server_type: Unknown, rtt: None>]>

test/test_srv_polling.py
juliusgeo: Done.

test/test_srv_polling.py
juliusgeo: Done.

test/test_srv_polling.py
ShaneHarvey: The second 27017 address is for a different SRV record (test3.test.build.10gen.cc). There should only be one 27017 in this test.

test/test_srv_polling.py
ShaneHarvey: Hmm, that's unexpected. Try debugging _updated_topology_description_srv_polling. In this test it's supposed to be called with nodes [27019, 27020]. That should cause 27017 and 27018 to be removed and replaced with 27019 and 27020.

test/test_client.py
```diff
@@ -1640,7 +1640,23 @@ def test_service_name_from_kwargs(self):
...
+        client = MongoClient(
```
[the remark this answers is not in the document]
juliusgeo: I don't think that the spec or prose tests test the behavior when you specify srvMaxHosts in both the kwargs and in the URI, so unless I am missing something in the spec tests then I think we should keep it.

test/test_client.py
```diff
@@ -1640,7 +1640,23 @@ def test_service_name_from_kwargs(self):
...
+            len(client.topology_description.server_descriptions()), 1)
+        client = MongoClient(
+            'mongodb+srv://test1.test.build.10gen.cc/', srvmaxhosts=1)
+        self.assertEqual(
+            len(client.topology_description.server_descriptions()), 1)
+        client = MongoClient(
+            'mongodb+srv://test1.test.build.10gen.cc/?srvMaxHosts=1')
+        self.assertEqual(
+            len(client.topology_description.server_descriptions()), 1)
+        _updated_topology_description_srv_polling(
```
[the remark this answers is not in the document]
juliusgeo: Done.

test/test_srv_polling.py
```diff
@@ -230,6 +231,59 @@ def erroring_seedlist():
...
+    def test_10_all_dns_selected(self):
+        response = [("localhost.test.build.10gen.cc", 27017),
+                    ("localhost.test.build.10gen.cc", 27019),
+                    ("localhost.test.build.10gen.cc", 27020)]
+
+        def nodelist_callback():
+            return response
+        with SrvPollingKnobs(ttl_time=WAIT_TIME,
+                             min_srv_rescan_interval=WAIT_TIME):
+            client = MongoClient(self.CONNECTION_STRING, srvMaxHosts=0)
```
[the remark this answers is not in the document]
juliusgeo: Done.

test/test_srv_polling.py
```diff
@@ -230,6 +231,59 @@ def erroring_seedlist():
...
+
+        def nodelist_callback():
+            return response
+
+        with SrvPollingKnobs(
+                ttl_time=WAIT_TIME, min_srv_rescan_interval=WAIT_TIME):
+            client = MongoClient(self.CONNECTION_STRING, srvMaxHosts=2)
+            with SrvPollingKnobs(nodelist_callback=nodelist_callback):
+                sleep(2*common.MIN_SRV_RESCAN_INTERVAL)
+                final_topology = \
```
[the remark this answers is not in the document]
juliusgeo: Done.

test/test_srv_polling.py
```diff
@@ -230,6 +231,59 @@ def erroring_seedlist():
...
+                    set(client.topology_description.server_descriptions())
+
+                self.assertSetEqual(final_topology, final_topology | {(
```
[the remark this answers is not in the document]
juliusgeo: Done.

test/test_srv_polling.py
```diff
@@ -230,6 +231,59 @@ def erroring_seedlist():
...
+                    "localhost.test.build.10gen.cc", 27017)})
+                self.assertEqual(len(final_topology), 2)
+
+    def test_does_not_flipflop(self):
+        client = MongoClient(self.CONNECTION_STRING, srvMaxHosts=1)
+        old = set(client.topology_description.server_descriptions().keys())
```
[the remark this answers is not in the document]
juliusgeo: Done.

test/test_srv_polling.py
```diff
@@ -230,6 +231,59 @@ def erroring_seedlist():
...
+        sleep(15)
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/srv_resolver.py
```diff
@@ -111,7 +111,8 @@ def _get_srv_response_and_hosts(self, encapsulate_errors):
                 raise ConfigurationError("Invalid SRV host: %s" % (node[0],))
             if self.__plist != nlist:
                 raise ConfigurationError("Invalid SRV host: %s" % (node[0],))
-
+        if(self.__srv_max_hosts != 0):
```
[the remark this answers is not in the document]
juliusgeo: Done.

PR #767, opened 2021-10-28
PYTHON-2973 Revert back to using quote_plus/unquote_plus

juliusgeo:


test/test_uri_spec.py
```diff
@@ -143,6 +143,7 @@ def run_scenario(self):
                 options['database'] += "." + options['collection']
             for elm in auth:
                 if auth[elm] is not None:
+                    options[elm] = options[elm].replace(" ", "+")
```
ShaneHarvey, line 146: Please add a comment referring to the fact that we use unquote_plus so whereas the spec says that "+" -> "+", we actually do "+" -> " ".

ShaneHarvey (changes requested):
Two more suggestions:
- Could you update the PR title (and when merging, the final commit message) to say something like "Revert back to using quote_plus/unquote_plus"?
- We should add a test for a password that has a slash and other special chars. The test should do this:
```python
user = "user@ /9+:?~!$&'()*+,;="
pwd =  "pwd@ /9+:?~!$&'()*+,;="
uri  = 'mongodb://%s:%s@localhost' % (quote_plus(user), quote_plus(pwd))
res = parse_uri(uri)
self.assertEqual(user, res['user'])
self.assertEqual(pwd, res['password'])
```

test/test_uri_spec.py
juliusgeo: Done.

ShaneHarvey (approved):

PR #770, opened 2021-11-01
TEST: PyPy7.3.7 fixed unicode repr

mattip:
Only keep the old behaviour around for older PyPy versions. xref https://github.com/conda-forge/pymongo-feedstock/pull/50


ShaneHarvey (approved):
Thanks!

PR #773, opened 2021-11-02
PYTHON-2992 Implement unified test format loop operation

juliusgeo:


test/unified_format.py
```diff
@@ -1048,6 +1085,43 @@ def _testOperation_assertNumberConnectionsCheckedOut(self, spec):
...
+        if failure_key:
+            self.entity_map[failure_key] = []
+        if error_key:
+            self.entity_map[error_key] = []
+        if successes_key:
+            self.entity_map[successes_key] = []
+        if iteration_key:
+            self.entity_map[iteration_key] = 0
+        while True:
+            if IS_INTERRUPTED:
```
ShaneHarvey, line 1103: Do we still need to catch KeyboardInterrupt?

test/unified_format.py
```diff
@@ -1131,16 +1205,14 @@ def run_scenario(self, spec):
 
         # process operations
         self.run_operations(spec['operations'])
+        self.verify_outcome(spec.get('outcome', []))
```
ShaneHarvey, line 1208: Is this change important for some reason?

test/unified_format.py
```diff
@@ -68,6 +70,27 @@
...
+    # re-entrant (invokable multiple times without needing to wait for prior
+    # invocations to return/complete) which is made possible by this pattern.
+    IS_INTERRUPTED = True
+
+
+if WIN32:
+    # CTRL_BREAK_EVENT is mapped to SIGBREAK
+    signal.signal(signal.SIGBREAK, interrupt_handler)
+else:
+    signal.signal(signal.SIGINT, interrupt_handler)
```
ShaneHarvey, line 92: Hmm I was expecting this code to live in the drivers-atlas-testing workload runner. What do you think about replacing this with a function that the workload can use:
```python
def interrupt_loop():
    global IS_INTERRUPTED
    IS_INTERRUPTED = True
```

Then we can use this api from the workload executor.

ShaneHarvey (changes requested):
How will the workload executor actually retrieve the results from the entity map? Can we add a test that demos how this will happen?

test/unified_format.py
```diff
@@ -269,15 +304,17 @@ def _create_entity(self, entity_spec):
             kwargs = {}
             observe_events = spec.get('observeEvents', [])
             ignore_commands = spec.get('ignoreCommandMonitoringEvents', [])
+
             observe_sensitive_commands = spec.get(
                 'observeSensitiveCommands', False)
             # TODO: PYTHON-2511 support storeEventsAsEntities
```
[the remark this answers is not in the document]
juliusgeo: Done.

test/unified_format.py
juliusgeo: This is now for the interrupt_loop function that can be called by the workload executor.

test/unified_format.py
juliusgeo: No, I will revert it.

test/unified_format.py
ShaneHarvey: I'm not sure that answers my question. What happens when you crtl-C the tests during a loop op?

ShaneHarvey:
Did you forget to push your changes?

test/unified_format.py
```diff
@@ -1048,6 +1075,43 @@ def _testOperation_assertNumberConnectionsCheckedOut(self, spec):
...
+                for op in spec["operations"]:
+                    self.run_entity_operation(op)
+                    if successes_key:
+                        self.entity_map._entities[successes_key] += 1
+                if iteration_key:
+                    self.entity_map._entities[iteration_key] += 1
+            except AssertionError as exc:
+                if failure_key or error_key:
+                    self.entity_map[failure_key or error_key].append({
+                        "error": exc, "time": time.time()})
```
ShaneHarvey, line 1105: What is time used for? Should time be a datetime instead?

test/unified_format.py
```diff
@@ -1048,6 +1075,43 @@ def _testOperation_assertNumberConnectionsCheckedOut(self, spec):
...
+        failure_key = spec.get('storeFailuresAsEntity')
+        error_key = spec.get('storeErrorsAsEntity')
+        successes_key = spec.get('storeSuccessesAsEntity')
+        iteration_key = spec.get('storeIterationsAsEntity')
+        if failure_key:
+            self.entity_map[failure_key] = []
+        if error_key:
+            self.entity_map[error_key] = []
+        if successes_key:
+            self.entity_map[successes_key] = []
```
ShaneHarvey, line 1089: Is successes a list or a counter?

test/unified_format.py
```diff
@@ -1048,6 +1075,43 @@ def _testOperation_assertNumberConnectionsCheckedOut(self, spec):
...
+                break
+            try:
+                for op in spec["operations"]:
+                    self.run_entity_operation(op)
+                    if successes_key:
+                        self.entity_map._entities[successes_key] += 1
+                if iteration_key:
+                    self.entity_map._entities[iteration_key] += 1
+            except AssertionError as exc:
+                if failure_key or error_key:
```
ShaneHarvey, line 1103: Is the `failure_key or error_key` behavior mandated by the spec?

test/unified_format.py
```diff
@@ -197,6 +206,18 @@ def __init__(self, observe_events, ignore_commands,
...
+        self.entity_map = entity_map
+        if store_events:
+            for i in store_events:
+                id = i["id"]
+                events = (i.lower() for i in i["events"])
+                for i in events:
+                    if i in self._event_mapping:
+                        self._event_mapping[i].append(id)
+                    else:
+                        self._event_mapping[i] = [id]
```
ShaneHarvey, line 219: This is the perfect place to use the builtin defaultdict: `self._event_mapping = collections.defaultdict(list)`. Checkout the docs for it to see how we can simplify this code.

test/unified_format.py
```diff
@@ -23,7 +23,9 @@
 import os
 import re
 import sys
+import time
 import types
+import signal
```
ShaneHarvey, line 28: unneeded now.

test/unified_format.py
```diff
@@ -205,8 +226,14 @@ def get_events(self, event_type):
         return [e for e in self.events if 'Command' not in type(e).__name__]
 
     def add_event(self, event):
-        if type(event).__name__.lower() in self._event_types:
+        event_name = type(event).__name__.lower()
+        if event_name in self._event_types:
             super(EventListenerUtil, self).add_event(event)
+        for id in self._event_mapping.get(event_name,[]):
+            if id in self.entity_map._entities.keys():
```
ShaneHarvey, line 233: We don't need to call `.keys()` here.

test/test_create_entities.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class TestCreateEntities(unittest.TestCase):
+
+    def setUp(self):
+        self.scenario_runner = UnifiedSpecTestMixinV1()
+
+    def test_store_events_as_entities(self):
+        spec = {
+                  "description": "blank",
```
ShaneHarvey, line 27: I think this is overindented. Try running https://pypi.org/project/autopep8/ on this file.

test/test_create_entities.py
```diff
@@ -0,0 +1,56 @@
...
+                      "operations": []
+                    }
+                  ]
+                }
+        self.scenario_runner.TEST_SPEC = spec
+        self.scenario_runner.setUp()
+        self.scenario_runner.run_scenario(spec["tests"][0])
+        final_entity_map = self.scenario_runner.entity_map._entities
+        self.assertIn("events1", final_entity_map)
+        self.assertGreater(len(final_entity_map["events1"]), 0)
```
ShaneHarvey, line 56: Cool nice test!

ShaneHarvey (changes requested):

test/test_create_entities.py
```diff
@@ -0,0 +1,56 @@
...
+import unittest
+import time
+
+from test.unified_format import UnifiedSpecTestMixinV1
+
+
+class TestCreateEntities(unittest.TestCase):
+
+    def setUp(self):
+        self.scenario_runner = UnifiedSpecTestMixinV1()
```
ShaneHarvey, line 23: I think we can remove the setUp method here since there's only one test.

test/unified_format.py
juliusgeo: Nothing appears to happen which is why I'm confused we got rid of that code that used the sigint to stop the loop.

test/test_create_entities.py
juliusgeo: Done.

juliusgeo:
So I had to implement a `numIterations` field to allow the tests to be ran. I tried to implement it using processes (multiprocessing.Process) and that did not work well with the `interrupt_loop()` function, so I'm kind of unsure how to proceed without having to implement a new field in the spec.

test/test_create_entities.py
ShaneHarvey: Sure but each test can create its own UnifiedSpecTestMixinV1. Normally we'd reach for setUp() only once there's more boilerplate.

test/unified_format.py
```diff
@@ -205,8 +222,14 @@ def get_events(self, event_type):
         return [e for e in self.events if 'Command' not in type(e).__name__]
 
     def add_event(self, event):
-        if type(event).__name__.lower() in self._event_types:
+        event_name = type(event).__name__.lower()
+        if event_name in self._event_types:
             super(EventListenerUtil, self).add_event(event)
+        for id in self._event_mapping[event_name]:
+            if id in self.entity_map:
```
ShaneHarvey, line 229: I think this "if" is always try and the else is unreachable right? If an event name maps to an id then we know for a fact that the id already exists in entity_map.

test/unified_format.py
ShaneHarvey: Thanks for the explanation. SGTM.

test/unified_format.py
ShaneHarvey: > Nothing appears to happen

Really? I'd expect ctrl-C to cause a KeyboardInterrupt to be raised from _testOperation_loop (even without using signals).

ShaneHarvey (changes requested):

test/test_create_entities.py
```diff
@@ -0,0 +1,135 @@
...
+                    "operations": [
+                        {
+                            "name": "loop",
+                            "object": "testRunner",
+                            "arguments": {
+                                "storeIterationsAsEntity": "iterations",
+                                "storeSuccessesAsEntity": "successes",
+                                "storeFailuresAsEntity": "failures",
+                                "storeErrorsAsEntity": "errors",
+                                "numIterations": 100,
```
ShaneHarvey, line 98: I like numIterations idea! Let's reduce this to 5 to speed up the test.

test/test_create_entities.py
```diff
@@ -0,0 +1,135 @@
...
+                }
+            ]
+        }
+
+        self.scenario_runner.TEST_SPEC = spec
+        self.scenario_runner.setUp()
+        self.scenario_runner.run_scenario(spec["tests"][0])
+        final_entity_map = self.scenario_runner.entity_map
+        for entity in ["errors", "failures", "successes", "iterations"]:
+            self.assertIn(entity, final_entity_map)
```
ShaneHarvey, line 135: Please do some assertions on the saved objects to make sure they are the types we expect.

test/test_create_entities.py
ShaneHarvey: Please assert the type of the objects in "events1" is what we expect.

ShaneHarvey (changes requested):

test/test_create_entities.py
juliusgeo: That is true there isn't much boilerplate. Done.

test/test_create_entities.py
juliusgeo: Done.

test/test_create_entities.py
juliusgeo: Done.

test/test_create_entities.py
juliusgeo: Done.

test/unified_format.py
juliusgeo: What if it's the first event? Though removing these lines does not cause the tests to fail.

test/unified_format.py
juliusgeo: It appears it was something with pycharm causing an issue. If I send it a ctrl-C through the command line it instantly exits.

test/test_create_entities.py
```diff
@@ -0,0 +1,142 @@
...
+                            }
+                        }
+                    ]
+                }
+            ]
+        }
+
+        self.scenario_runner.TEST_SPEC = spec
+        self.scenario_runner.setUp()
+        self.scenario_runner.run_scenario(spec["tests"][0])
```
juliusgeo, line 134: This potentially could be written as:
```
   for test in spec["tests"]:
         self.scenario_runner.run_scenario(test)
```
This would be more general. However, if we know we aren't going to be adding on to the tests in this chunk of spec test, then I would prefer the first.

ShaneHarvey: Let's cross that bridge when we come to it. Glancing at  https://github.com/mongodb-labs/drivers-atlas-testing/tree/master/tests it appears that we only need to run a single test right now. 

Calling run_scenario twice will create a new entity_map so all the results from the previous test(s) will be lost. So I think the behavior of multiple tests is a little underspecified.

ShaneHarvey (approved):

PR #774, opened 2021-11-03
PYTHON-2554 Support $merge and $out executing on secondaries

ShaneHarvey:
Implementation of https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read-preferences-and-server-selection including the changes in DRIVERS-823 and DRIVERS-1969.

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -1161,7 +1161,7 @@ class MongoClient(common.BaseObject):
 
         with self._get_socket(server, session) as sock_info:
             secondary_ok = (single and not sock_info.is_mongos) or (
-                read_preference != ReadPreference.PRIMARY)
+                read_preference.mode != ReadPreference.PRIMARY.mode)
             yield sock_info, secondary_ok
 
     @contextlib.contextmanager
--- a/pymongo/read_preferences.py
+++ b/pymongo/read_preferences.py
@@ -442,26 +442,23 @@ class _AggWritePref:
 
     def __call__(self, selection):
         """Apply this read preference to a Selection."""
+        has_50 = any(s for s in selection.server_descriptions
+                     if s.max_wire_version >= 13)
+        if not has_50:
+            # Fallback to using the primary:
+            self.effective_pref = ReadPreference.PRIMARY
+            return selection.primary_selection
         # Apply the original selection:
-        pref_selection = self.pref(selection)
-        # Filter 5.0+ or writable servers:
-        suitable = [s for s in pref_selection.server_descriptions
-                    if s.max_wire_version >= 13 or s.is_writable]
-        if suitable:
-            self.effective_pref = self.pref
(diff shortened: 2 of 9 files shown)


pymongo/topology_description.py
```diff
@@ -263,21 +263,24 @@ def apply_selector(self, selector, address=None, custom_selector=None):
...
-        elif self.topology_type == TOPOLOGY_TYPE.Sharded:
-            # Ignore read preference.
-            selection = Selection.from_topology_description(self)
-        else:
-            selection = selector(Selection.from_topology_description(self))
+
+        selection = Selection.from_topology_description(self)
+        # Ignore read preference for sharded clusters.
+        if self.topology_type != TOPOLOGY_TYPE.Sharded:
+            selection = selector(selection)
```
ShaneHarvey, line 283: This bit is just a small cleanup with no functional change.

pymongo/topology_description.py
```diff
@@ -263,21 +263,24 @@ def apply_selector(self, selector, address=None, custom_selector=None):
                                              selector.min_wire_version,
                                              common_wv))
 
+        if isinstance(selector, _AggWritePref):
+            selector.selection_hook(self)
```
ShaneHarvey, line 267: Initially I was expecting to place the "selection_hook" logic inside `_AggWritePref.__call__` but it turns out that we skip calling the selector in many cases. In fact the only time we actually do call the selector is for replica sets. I attempted to refactor this code to actually call the selector in all cases but that doesn't work either because read preference selectors are written to only work with replica sets.

So "selection_hook" is what I came up with instead.

pymongo/mongo_client.py
```diff
@@ -1161,7 +1161,7 @@ def _secondaryok_for_server(self, read_preference, server, session):
 
         with self._get_socket(server, session) as sock_info:
             secondary_ok = (single and not sock_info.is_mongos) or (
-                read_preference != ReadPreference.PRIMARY)
+                read_preference.mode != ReadPreference.PRIMARY.mode)
```
ShaneHarvey, line 1164: This change is required because `read_preference` might be a _AggWritePref.

pymongo/read_preferences.py
```diff
@@ -424,6 +424,45 @@ def __call__(self, selection):
                 self.max_staleness, selection))
 
 
+class _AggWritePref:
+    """Agg $out/$merge write preference.
+
+    * If there are no readable servers, or there is any pre-5.0 server, use
```
blink1073, line 430: Aren't we using Primary() in the first case and `pref` in the second?

ShaneHarvey: Yeah good catch. Fixed.

blink1073 (approved):

ShaneHarvey:
I'll wait for @juliusgeo to review as well.

juliusgeo (approved):
LGTM!

PR #775, opened 2021-11-03
Reformat to comply with PEP 8 coding style

Mosallamy:


ShaneHarvey:
Thanks for the PR @Mosallamy, however I'm not going to accept it. The issue is that PyMongo has many pep8 violations besides the ones fixed in this PR. Instead of fixing them a few at a time, we are planning to add an auto code formatter.

PR #776, opened 2021-11-04
PYTHON-2998 Remove md5 checksums from gridfs and remove disable_md5

ShaneHarvey:


gridfs/grid_file.py
```diff
@@ -183,8 +183,6 @@ def __init__(
         coll = _clear_entity_type_registry(
             root_collection, read_preference=ReadPreference.PRIMARY)
 
-        if not disable_md5:
```
behackett, line 186: What about the update in __flush_data?

ShaneHarvey: Good catch!

ShaneHarvey: Fixed.

doc/changelog.rst
```diff
@@ -170,12 +170,15 @@ Breaking Changes in 4.0
...
-  when more than one URI is passed into the ``hosts`` argument.
+- :class:`~pymongo.mongo_client.MongoClient` now raises a
+  :exc:`~pymongo.errors.ConfigurationError` when more than one URI is passed
+  into the ``hosts`` argument.
 - :class:`~pymongo.mongo_client.MongoClient`` now raises an
   :exc:`~pymongo.errors.InvalidURI` exception
   when it encounters unescaped percent signs in username and password when
   parsing MongoDB URIs.
+- Removed the `disable_md5` parameter for :class:`~gridfs.GridFSBucket` and
+  :class:`~gridfs.GridFS`. MD5 checksums are now always disabled in GridFS.
```
behackett, line 181: Instead of saying "checksums are now always disabled" I think you should say something like "gridfs no longer generates a checksum" and link to the guidance in the upgrade docs.

doc/migrate-to-pymongo4.rst
```diff
@@ -819,6 +819,30 @@ Changed the default JSON encoding representation from legacy to relaxed.
...
 :const:`~bson.json_util.RELAXED_JSON_OPTIONS`.
 
+GridFS changes
+--------------
+
+disable_md5 parameter is removed
+................................
+
+Removed the `disable_md5` option for :class:`~gridfs.GridFSBucket` and
+:class:`~gridfs.GridFS`. MD5 checksums are now always disabled in GridFS.
```
behackett, line 829: Same comment here about "are now always disabled".

gridfs/__init__.py
```diff
@@ -48,14 +48,17 @@ def __init__(self, database, collection="fs", disable_md5=False):
         :Parameters:
           - `database`: database to use
           - `collection` (optional): root collection to use
-          - `disable_md5` (optional): When True, MD5 checksums will not be
-            computed for uploaded files. Useful in environments where MD5
-            cannot be used for regulatory or other reasons. Defaults to False.
+
+        .. versionchanged:: 4.0
+           Removed the `disable_md5` parameter.
```
behackett, line 53: Also mention that gridfs no longer generates an md5sum

gridfs/__init__.py
```diff
@@ -442,13 +442,16 @@ def __init__(self, db, bucket_name="fs",
             (the default) db.write_concern is used.
           - `read_preference` (optional): The read preference to use. If
             ``None`` (the default) db.read_preference is used.
-          - `disable_md5` (optional): When True, MD5 checksums will not be
-            computed for uploaded files. Useful in environments where MD5
-            cannot be used for regulatory or other reasons. Defaults to False.
+
+        .. versionchanged:: 4.0
+           Removed the `disable_md5` parameter.
```
behackett, line 447: Same here.

gridfs/grid_file.py
```diff
@@ -152,12 +150,14 @@ def __init__(
...
             :class:`~pymongo.client_session.ClientSession` to use for all
             commands
-          - `disable_md5` (optional): When True, an MD5 checksum will not be
-            computed for the uploaded file. Useful in environments where
-            MD5 cannot be used for regulatory or other reasons. Defaults to
-            False.
           - `**kwargs` (optional): file level options (see above)
 
+        .. versionchanged:: 4.0
+           Removed the `disable_md5` parameter.
```
behackett, line 156: And here.

behackett (changes requested):

doc/changelog.rst

doc/migrate-to-pymongo4.rst
ShaneHarvey: Done.

gridfs/__init__.py
ShaneHarvey: Updated to link to the migration guide which describes the change in more detail.

gridfs/__init__.py
ShaneHarvey: Updated to link to the migration guide which describes the change in more detail.

gridfs/grid_file.py
ShaneHarvey: Updated to link to the migration guide which describes the change in more detail.

juliusgeo (approved):
LGTM!

behackett (approved):

ShaneHarvey:
I made a small update to use delete_many+drop_indexes to clear the collection state rather than use drop_collection. This speeds up the gridfs test quite a bit. On a MongoDB 5.1.0-alpha-1170-g4c2c9d1 3 member replica set the following used to run in 219 seconds:
```
$ python3 -m unittest discover -s test -v -k 'grid'  
...
Ran 133 tests in 219.580s

OK
```

The same command now runs in 102.092s. So we've shaved off about 2 minutes in runtime! Still 100 seconds seems high for these tests. There's probably something else we could do but I'll save that for a future project.

behackett:
Still LGTM

PR #779, opened 2021-11-05
PYTHON-2721 Test failure - invalid flag in regex options on v5.0.0-alpha0-442-g0520668

juliusgeo:


ShaneHarvey (approved):
LGTM, please update the commit message to something like:
"PYTHON-2721 Reenable regex flags test on MongoDB 5.0+"

PR #780, opened 2021-11-05
PYTHON-2941 Add a CMAP test that verifies the background thread hands over connections to threads doing checkout

ShaneHarvey:
Python already has the correct behavior so we just need to add the new test.


ShaneHarvey:
The new test passes everywhere. The failures are due to known flakey tests.

PR #781, opened 2021-11-05
PYTHON-3011 Fix test_connections_are_only_returned_once

ShaneHarvey:


test/test_load_balancer.py
```diff
@@ -45,7 +45,7 @@ def test_connections_are_only_returned_once(self):
         nconns = len(pool.sockets)
         self.db.test.find_one({})
         self.assertEqual(len(pool.sockets), nconns)
-        self.db.test.aggregate([{'$limit': 1}])
+        list(self.db.test.aggregate([{'$limit': 1}]))
```
ShaneHarvey, line 48: I suspect the problem here was that we didn't explicitly cleanup or exhaust the cursor. So sometimes the connection would remain pinned to the cursor until it eventually gets garbage collected at some later point.

ShaneHarvey:
Weird this test still failed in the patch but with a different error:
```
 [2021/11/05 22:44:18.870] FAILURE: 13 != 10 (AssertionError)
 [2021/11/05 22:44:18.870] Traceback (most recent call last):
 [2021/11/05 22:44:18.870]   File "/data/mci/2ff0ab031c178ebccd3988f926b1f3cf/src/test/test_load_balancer.py", line 47, in test_connections_are_only_returned_once
 [2021/11/05 22:44:18.870]     self.assertEqual(len(pool.sockets), nconns)
 [2021/11/05 22:44:18.870] AssertionError: 13 != 10
```

This indicates that something is adding 3 sockets in the time that the find_one executes:
```python
        pool = get_pool(self.client)
        nconns = len(pool.sockets)
        self.db.test.find_one({})
        self.assertEqual(len(pool.sockets), nconns)
```

I think we can go ahead and merge this PR. Then I will leave the jira ticket open to investigate this new error.

juliusgeo (approved):
Sounds good to me!

PR #782, opened 2021-11-08
PYTHON-2460 Client can create more than minPoolSize background connections

juliusgeo:


test/test_pooling.py
```diff
@@ -520,6 +520,14 @@ def test_max_pool_size_with_connection_failure(self):
             self.assertNotIn('waiting for socket from pool',
                              str(context.exception))
 
+    def test_min_pool_size_always_positive(self):
```
ShaneHarvey, line 523: Rather than adding a new test I think we can re-use `test_return_socket_after_reset`: https://github.com/mongodb/mongo-python-driver/blob/b342990934b55370ebde40d20ba6423e035b1961/test/test_pooling.py#L291-L296

We just want to add some more assertions for active_sockets and probably operation_count too.

ShaneHarvey (changes requested):

test/test_pooling.py
juliusgeo: Done. I'm not sure why you mean by adding assertions for operation_count as that appears to stay at 0 during the entirety of the test.

test/test_pooling.py
```diff
@@ -295,6 +295,7 @@ def test_return_socket_after_reset(self):
 
         self.assertTrue(sock.closed)
         self.assertEqual(0, len(pool.sockets))
+        self.assertGreaterEqual(pool.active_sockets, 0)
```
ShaneHarvey, line 298: This should be assertEqual 0. Please also add the same assertion for operation_count. 

Finally, let's add the same two assertions inside the with-statement but replace 0 with 1.

ShaneHarvey (changes requested):

test/test_pooling.py
juliusgeo: Done.

ShaneHarvey (approved):
LGTM assuming the test passes.

PR #783, opened 2021-11-09
PYTHON-2512 Update Astrolabe's Workload Executor to use the unified test runner

juliusgeo:


test/unified_format.py
```diff
@@ -315,7 +315,10 @@ def _create_entity(self, entity_spec):
                 kwargs['server_api'] = ServerApi(
                     server_api['version'], strict=server_api.get('strict'),
                     deprecation_errors=server_api.get('deprecationErrors'))
-            client = rs_or_single_client(**kwargs)
+            if uri:
+                client = rs_or_single_client(h=uri, **kwargs)
+            else:
+                client = rs_or_single_client(**kwargs)
```
ShaneHarvey, line 321: We can simplify this to:
```python
            if uri:
                kwargs['h'] = uri
            client = rs_or_single_client(**kwargs)
```

test/unified_format.py
```diff
@@ -1103,15 +1102,21 @@ def _testOperation_loop(self, spec):
...
                 if failure_key or error_key:
                     self.entity_map[failure_key or error_key].append({
-                        "error": exc, "time": time.time()})
+                        "error": str(exc), "time": time.time(), "type": type(
+                            exc)})
                 else:
                     raise exc
             except Exception as exc:
+                if iteration_key:
+                    self.entity_map._entities[iteration_key] += 1
```
ShaneHarvey, line 1115: If we want to increment iteration_key every time it would be simpler to do it _before_  at the top of the loop rather than in every case.

ShaneHarvey (changes requested):
Please transition the PR out of the draft state before requesting a review.

test/unified_format.py
juliusgeo: Done.

test/unified_format.py
juliusgeo: Done.

test/unified_format.py
```diff
@@ -1080,38 +1080,39 @@ def _testOperation_loop(self, spec):
...
                     if successes_key:
                         self.entity_map._entities[successes_key] += 1
-                if iteration_key:
-                    self.entity_map._entities[iteration_key] += 1
             except AssertionError as exc:
                 if failure_key or error_key:
                     self.entity_map[failure_key or error_key].append({
-                        "error": exc, "time": time.time()})
+                        "error": str(exc), "time": time.time(), "type": type(
+                            exc)})
```
ShaneHarvey, line 1108: Are these keys defined somewhere? I'm wondering if it makes more sense to nest "type" under the "error" field instead of a new top level field.

ShaneHarvey (changes requested):

test/unified_format.py
juliusgeo: The spec says that the error should be the "textual description of the failure encountered.", which to me implies that it should be just a string, not a document. It also places no limits on additional top level fields.

test/unified_format.py
```diff
@@ -1080,38 +1080,42 @@ def _testOperation_loop(self, spec):
...
             try:
+                if iteration_key:
+                    self.entity_map._entities[iteration_key] += 1
                 for op in spec["operations"]:
                     self.run_entity_operation(op)
                     if successes_key:
                         self.entity_map._entities[successes_key] += 1
-                if iteration_key:
-                    self.entity_map._entities[iteration_key] += 1
             except AssertionError as exc:
```
ShaneHarvey, line 1104: Let's combine the AssertionError and Exception excepts into one:
```python
            except Exception as exc:
                if isinstance(exc, AssertionError):
                    key = failure_key or error_key
                else:
                    key = error_key or failure_key
                if not key:
                    raise
                    
                self.entity_map[key].append({
                        "error": str(exc),
                         "time": time.time(),
                         "type": type(exc)
                    })
```

Note that bare `raise` is preferred over `raise exc` since it will preserver the original traceback.

ShaneHarvey (changes requested):

test/unified_format.py
juliusgeo: Done. Noted.

ShaneHarvey (approved):
LGTM assuming EVG passes.

PR #789, opened 2021-11-16
PYTHON-3005 Make maxConnecting configurable

ShaneHarvey:


doc/changelog.rst
```diff
@@ -184,7 +184,8 @@ Notable improvements
 ....................
 
 - Enhanced connection pooling to create connections more efficiently and
-  avoid connection storms.
+  avoid connection storms. Added the ``maxConnecting`` URI and
```
behackett, line 187: I know these two things are related, but I would suggest making them separate bullet points.

behackett (approved):
Looking at the existing code for limiting the number of concurrent connection attempts, it's almost like you expected this to become a configurable. :-)

doc/changelog.rst

ShaneHarvey (commented):
Ready for another look.

IIRC when I implemented this we were still discussing whether to make it configurable or not.

behackett:
Still looks good to me.

PR #790, opened 2021-11-16
PYTHON-3017 Properly check for closed KMS connections

ShaneHarvey:


pymongo/encryption.py
```diff
@@ -129,6 +129,8 @@ def kms_request(self, kms_context):
             conn.sendall(message)
             while kms_context.bytes_needed > 0:
                 data = conn.recv(kms_context.bytes_needed)
+                if not data:
+                    raise OSError('KMS connection closed')
```
ShaneHarvey, line 133: What was happening here is that the connection was closed, recv returns `b''` and then we get stuck in an infinite loop.

behackett (approved):

PR #791, opened 2021-11-16
PYTHON-3015 Document cipher mismatch issues

behackett:


doc/examples/tls.rst
```diff
@@ -213,3 +213,21 @@ revocation checking failed::
   [('SSL routines', 'tls_process_initial_server_flight', 'invalid status response')]
 
 See :ref:`OCSP` for more details.
+
+Python 3.10+ incompatibilities with TLS/SSL on MongoDB <= 4.0
```
ShaneHarvey, line 217: One question, you mentioned that this error happens on the latest releases of 3.7+ on some platforms too. Is that still true? Should we mention that problem?

behackett: I suspect that is a Gentoo Linux specific problem. They seem to have backported the cipher changes to older versions of Python. I don't think it's worth getting into in this document.

ShaneHarvey (approved):
LGTM

PR #795, opened 2021-11-17
PYTHON-2981 Stop using MongoClient.address for hashing and equality

ShaneHarvey:


pymongo/topology.py
```diff
@@ -803,6 +803,20 @@ def __repr__(self):
             msg = 'CLOSED '
         return '<%s %s%r>' % (self.__class__.__name__, msg, self._description)
 
+    def eq_props(self):
+        """The properties to use for MongoClient/Topology equality checks."""
+        ts = self._settings
+        return (tuple(sorted(ts.seeds)), ts.replica_set_name, ts.fqdn,
+                ts.srv_service_name)
```
ShaneHarvey, line 810: I decided to use a few key client settings to determine equality. Comparing every single configurable setting is a bit overkill. Let me know if you have any opinions here.

behackett (commented):
This change should be called out in the changelog and possibly in the migration guide?

ShaneHarvey:
Updated to mention this change in the changelog. I don't think it's worth calling out in the migration guide. I also expanded the testing a bit.

behackett (approved):

PR #799, opened 2021-11-18
PYTHON-3021 Send primaryPreferred when connected to standalone servers

ShaneHarvey:


pymongo/mongo_client.py
```diff
@@ -1170,8 +1170,8 @@ def _socket_for_reads(self, read_preference, session):
         # for topology type Single."
         # Thread safe: if the type is single it cannot change.
         topology = self._get_topology()
-        single = topology.description.topology_type == TOPOLOGY_TYPE.Single
         server = self._select_server(read_preference, session)
+        single = topology.description.topology_type == TOPOLOGY_TYPE.Single
```
ShaneHarvey, line 1174: This fixes the bug. The problem is that before this _select_server call the topology_type could be Unknown in which case we would incorrectly set single to False.

juliusgeo: That makes sense.

juliusgeo (approved):
LGTM!

PR #800, opened 2021-11-19
PYTHON-2984 Fix retry behavior for bulk write writeConcernError

ShaneHarvey:
This PR resyncs the spec tests for DRIVERS-1385. It also fixes a bug where pymongo did not properly retry after writeConcernErrors encountered in bulk writes.


juliusgeo (approved):
LGTM!

PR #801, opened 2021-11-20
PYTHON-3018 Add docs for removed methods from Collection

juliusgeo:


doc/api/pymongo/collection.rst
```diff
@@ -58,9 +58,134 @@
...
+
+         Removed. MongoDB 4.2 removed the `parallelCollectionScan command`_.
+         There is no replacement.
+
+         .. _parallelCollectionScan command: https://docs.mongodb.com/manual/reference/command/parallelCollectionScan/
+      .. describe:: initialize_ordered_bulk_op || initialize_unordered_bulk_op
+
+         Removed. Use :meth:`pymongo.collection.Collection.bulk_write`
+         instead. Code like this::
+           batch = coll.initialize_(un)ordered_bulk_op()
```
juliusgeo, line 98: Not sure if it would be better to separate into two sections. Saying "(un)ordered" makes sense to me but might not be clear to the user.

ShaneHarvey (changes requested):
Instead of duplicating the docs from the migration guide, let's link to the guide itself. Something like this:
```
..versionchanged:: 4.0
  Removed the insert, save, count, .... methods. See the <link to migration guide>.
```
We can put this in the Collection docstring.

pymongo/collection.py
```diff
@@ -115,6 +115,13 @@ def __init__(self, database, name, create=False, codec_options=None,
           - `**kwargs` (optional): additional keyword arguments will
             be passed as options for the create collection command
 
+        .. versionchanged:: 4.0
+           Removed the reindex, map_reduce, inline_map_reduce,
+           parallel_scan, initialize_unordered_bulk_op,
+           initialize_ordered_bulk_op, group, count, insert, save,
+           update, remove, find_and_modify, and ensure_index methods. See the
+           :ref:`pymongo4-migration-guide`.
```
ShaneHarvey, line 123: Perfect, can you make a similar change for MongoClient and Database?

ShaneHarvey (changes requested):

pymongo/collection.py
juliusgeo: Done.

pymongo/mongo_client.py
```diff
@@ -503,6 +503,11 @@ def __init__(
 
         .. seealso:: The MongoDB documentation on `connections <https://dochub.mongodb.org/core/connections>`_.
 
+        .. versionchanged:: 4.0
```
ShaneHarvey, line 506: Please combine this with the existing versionchanged below.

ShaneHarvey (changes requested):

pymongo/mongo_client.py
juliusgeo: Done.

ShaneHarvey (approved):
LGTM

PR #804, opened 2021-11-24
PYTHON-2957 Support 'let' option for multiple CRUD commands

juliusgeo:


pymongo/collection.py
```diff
@@ -619,6 +619,9 @@ def _update(self, sock_info, criteria, document, upsert=False,
         command = SON([('update', self.name),
                        ('ordered', ordered),
                        ('updates', [update_doc])])
+        if let:
+            common.validate_is_document_type("let", let)
```
ShaneHarvey, line 623: Should this be the more generic `validate_is_mapping` instead?

pymongo/collection.py
```diff
@@ -795,6 +798,8 @@ def update_one(self, filter, update, upsert=False,
             MongoDB 4.2 and above.
           - `session` (optional): a
             :class:`~pymongo.client_session.ClientSession`.
+          - `let` (optional): Specifies a document with a list of variables
+            that can then be accessed using the form `$$<variable_name>`.
```
ShaneHarvey, line 802: Hmm my first impression is that this is very vague. Are there any docs that we can link to? Or any documentation guidance in the CRUD spec?

pymongo/collection.py
```diff
@@ -752,7 +755,7 @@ def replace_one(self, filter, replacement, upsert=False,
     def update_one(self, filter, update, upsert=False,
                    bypass_document_validation=False,
                    collation=None, array_filters=None, hint=None,
-                   session=None):
+                   session=None, let=None):
```
ShaneHarvey, line 758: For all the helpers that didn't support "let" before, can you add a `versionchanged:: 4.1` section?

pymongo/collection.py
```diff
@@ -2236,6 +2254,8 @@ def __find_and_modify(self, filter, projection, sort, upsert=None,
         cmd = SON([("findAndModify", self.__name),
                    ("query", filter),
                    ("new", return_document)])
+        if let:
+            cmd["let"] = let
```
ShaneHarvey, line 2258: Missing a validate_is_mapping?

ShaneHarvey (changes requested):
Great work! Just a couple comments

pymongo/collection.py
ShaneHarvey: Also can we add a test to make sure that validation works?

pymongo/collection.py
juliusgeo: Done.

test/test_collection.py
```diff
@@ -2178,6 +2178,24 @@ def test_bool(self):
         with self.assertRaises(NotImplementedError):
             bool(Collection(self.db, 'test'))
 
+    @client_context.require_version_min(5, 0, 0)
+    def test_helpers_with_let(self):
+        c = self.db.test
+        helpers = [("delete_many", ({}, {})), ("delete_one", ({}, {})),
```
ShaneHarvey, line 2184: Please refactor this to use the methods instead of strings, eg:
```
helpers = [(c.delete_many,...
```

It makes it easier to find usages of the method.

test/test_collection.py
```diff
@@ -2178,6 +2178,24 @@ def test_bool(self):
...
+                   ("find", ({})), ("update_many", ({}, {'$inc': {'x': 3}})),
+                   ("update_one", ({}, {'$inc': {'x': 3}})),
+                   ("find_one_and_delete", ({}, {})),
+                   ("find_one_and_replace", ({}, {})),
+                   ("aggregate", ([], {}))]
+        for let in [10, "str"]:
+            for helper, args in helpers:
+                with self.assertRaises(TypeError) as cm:
+                    getattr(c, helper)(*args, let=let)
+                self.assertIn("let must be an instance of dict",
```
ShaneHarvey, line 2194: Let's use self.assertRaisesRegex instead: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

pymongo/collection.py
ShaneHarvey: Can you update this with the docs from the CRUD spec?:
```
Map of parameter names and values. Values must be constant or closed
expressions that do not reference document fields. Parameters can then be
accessed as variables in an aggregate expression context (e.g. "$$var").
```
I think this version more clearly describes the api.

ShaneHarvey (changes requested):

test/test_collection.py
juliusgeo: Done.

test/test_collection.py
juliusgeo: Done.

pymongo/collection.py
juliusgeo: Done.

ShaneHarvey (commented):
Sorry, one last thing. Can you add this to the 4.1 changelog?

ShaneHarvey (approved):
LGTM!

PR #805, opened 2021-11-30
Preserve traceback when configuration errors occur

naomielst:
When resolving SRV records, the pymongo library disregards the previous error which lead to the configuration error.

This is making them harder to debug, especially when they occur only in CI (which is the case for us with [Celery](https://github.com/celery/celery/runs/4371284502?check_suite_focus=true#step:8:3264)).

### DIFF
--- a/pymongo/srv_resolver.py
+++ b/pymongo/srv_resolver.py
@@ -62,8 +62,8 @@ class _SrvResolver(object):
 
         try:
             self.__plist = self.__fqdn.split(".")[1:]
-        except Exception:
-            raise ConfigurationError(_INVALID_HOST_MSG % (fqdn,))
+        except Exception as exc:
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Can you provide an example of the error behavior before and after this change?

ShaneHarvey:
Any update here? Could you provide output before and after this change to show how this makes it easier to debug the `ConfigurationError: not a name` exception?

The real underlying issue may be that the Celery test is using mock to patching dnspython calls but pymongo calls different methods depending on the version:
```python
    if hasattr(resolver, 'resolve'):
        # dnspython >= 2
        return resolver.resolve(*args, **kwargs)
    # dnspython 1.X
    return resolver.query(*args, **kwargs)
```

This was originally changed in: https://jira.mongodb.org/browse/PYTHON-2364

ShaneHarvey:
Closing since I don't see how this change is an improvement.

PR #806, opened 2021-12-01
PYTHON-3027 Fix server selection when topology type is Unknown

ShaneHarvey:
This change fixes the bug and also implements DRIVERS-1999.


blink1073 (commented):
LGTM, thanks!

juliusgeo (approved):
LGTM!

PR #809, opened 2021-12-07
PYTHON-1864 PYTHON-2931 Spec complaint $readPreference 

ShaneHarvey:
This change fixes how pymongo sends $readPreference to the server via OP_MSG:
- Stop sending $readPreference to standalone servers. The spec says to omit it.
- Stop sending $readPreference with primary read preference to all servers. The spec says to omit "primary" since it's the server's default.
- The final rule is to always send $readPreference=primaryPreferred when directly connected to a single replica set member. This is to ensure that the app does not get NotPrimary errors for direct connections.

This change also removes the `secondary_ok` variable which is no longer needed. Instead of tracking the "secondary ok" boolean we replace the input read preference with the effective read preference (primary, primaryPreferred, or the original read preference). 

This change depends on https://github.com/mongodb/mongo-python-driver/pull/774 so please review that one first.

### DIFF
--- a/pymongo/aggregation.py
+++ b/pymongo/aggregation.py
@@ -92,11 +92,6 @@ class _AggregationCommand(object):
         """The database against which the aggregation command is run."""
         raise NotImplementedError
 
-    def _process_result(self, result, session, server, sock_info, secondary_ok):
-        if self._result_processor:
-            self._result_processor(
-                result, session, server, sock_info, secondary_ok)
-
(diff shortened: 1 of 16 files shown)


blink1073 (approved):
LGTM!

ShaneHarvey:
Update to fix the merge conflicts. @juliusgeo please take a look.

juliusgeo (approved):
LGTM!

PR #819, opened 2021-12-10
PYTHON-2888 Migrate from json.send to perf.send

ShaneHarvey:
Followed the migration guide here: https://github.com/10gen/performance-tooling-docs/blob/main/getting_started/performance_monitoring_setup.md#legacy-data-migration

Testing here: https://spruce.mongodb.com/version/61b2a9a5562343496412629b

### DIFF
--- a/test/performance/perf_test.py
+++ b/test/performance/perf_test.py
@@ -47,9 +47,7 @@ OUTPUT_FILE = os.environ.get('OUTPUT_FILE')
 result_data = []
 
 def tearDownModule():
-    output = json.dumps({
-        'results': result_data
-        }, indent=4)
+    output = json.dumps(result_data, indent=4)
     if OUTPUT_FILE:
         with open(OUTPUT_FILE, 'w') as opf:
             opf.write(output)
--- a/test/performance/perf_test.py
+++ b/test/performance/perf_test.py
@@ -77,7 +77,7 @@ class PerformanceTest(object):
     def tearDown(self):
         name = self.__class__.__name__
         median = self.percentile(50)
-        result = self.data_size / median
+        bytes_per_sec = self.data_size / median
         print('Running %s. MEDIAN=%s' % (self.__class__.__name__,
                                          self.percentile(50)))
         result_data.append({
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
Patch failed with:
```
[2021/12/10 02:07:17.431] Running command 'perf.send' in "send dashboard data" (step 4 of 4)
[2021/12/10 02:07:17.431] Task failed: command failed: problem reading tests from '/data/mci/5dd3bffd60584c7d823485c03b3eab22/src/results.json': problem unmarshaling report data: json: cannot unmarshal object into Go value of type []poplar.Test
[2021/12/10 02:07:17.431] Command failed: problem reading tests from '/data/mci/5dd3bffd60584c7d823485c03b3eab22/src/results.json': problem unmarshaling report data: json: cannot unmarshal object into Go value of type []poplar.Test
[2021/12/10 02:07:17.431] StatsCollector ticker stopping.
[2021/12/10 02:07:17.432] Task completed - FAILURE.
```

The results file must be malformed.

ShaneHarvey:
Fixed: https://spruce.mongodb.com/version/61b2c28332f41770ff5823e9/tasks

test/performance/perf_test.py
```diff
@@ -83,12 +81,18 @@ def tearDown(self):
...
-            }
+            'info': {
+                'test_name': name,
+                'args': {
+                   'threads': 1,
+                },
+            },
+            'metrics': [
+                {
+                    'name': 'ops_per_sec',
```
behackett, line 92: So, ops_per_sec has always been used because that was the only option, but IIRC that's not what our metrics actually are. Is it possible to fix that now?

ShaneHarvey: Oh I see, the current result is more like "bytes of data processed per second". We can actually add multiple stats now, like can can add the median latency in addition to a bytes/sec number. Any suggestions for what to call the new field? Maybe "bytes_per_sec"?

behackett: If that's what it is (it's been so long I don't remember) than LGTM.

blink1073 (approved):
LGTM!

ShaneHarvey:
New patch: https://spruce.mongodb.com/version/61b7bbd23627e00295e72080

behackett (approved):

blink1073 (approved):

PR #821, opened 2021-12-10
PYTHON-2013 Skip authentication step on arbiters

ShaneHarvey:


blink1073 (approved):
LGTM!

juliusgeo (approved):
LGTM!

ShaneHarvey:
Closing as won't do. Implementing this spec changes means it will no longer possible to authenticate as the "local.__system" (the internal cluster auth member) or X.509 users on direct connections to arbiters. I would prefer not to implement this at all in python to avoid removing a feature. The original motivation for DRIVERS-753/SPEC-1301 was to fix a regression which pymongo never implemented.

If a user does not want to authenticate on a direct connection to an arbiter (or any other node) they just need to omit auth credentials from the MongoClient URI and keyword options.

PR #829, opened 2021-12-23
PYTHON-3052 Add Typings to PyMongo Itself

blink1073:
- [x] Finish manual port of `monitoring.py`, `server_description`, `server_type`, `topology_description`, `url_parser` (all failed when using [`retype`](https://pythonrepo.com/repo/ambv-retype-python-code-refactoring))
- [x] Double check all public interfaces in the auto-ported files:
```
     modified:   pymongo/__init__.py
	modified:   pymongo/collation.py
	modified:   pymongo/common.py
	modified:   pymongo/event_loggers.py
	modified:   pymongo/operations.py
	modified:   pymongo/read_concern.py
	modified:   pymongo/read_preferences.py
	modified:   pymongo/results.py
	modified:   pymongo/write_concern.py
```
- [x] Also look at addressing https://github.com/mongodb-labs/pymongo-stubs/issues/15
  - Used `Union[MutableMapping[str, Any], RawBSONDocumentRef]`
  - Investigate if we can make `Collection` a generic object


blink1073:
Local experiment with generic document types:

```python
from typing import Any, Mapping, TypeVar, Generic, Optional, Type, Union, cast
from bson.raw_bson import RawBSONDocument
from bson import encode

T = TypeVar('T', bound=Mapping[str, Any])
U = TypeVar('U', bound=Mapping[str, Any])


class Client(Generic[T]):
    document_type: Type[T]

    def __init__(self, document_type: Type[T]) -> None:
        self.document_type = document_type


class CodecOptions(Generic[T]):
    document_type: Type[T]

    def __init__(self, document_type: Type[T]) -> None:
        self.document_type = document_type


class Database(Generic[T]):
    _document_type: Type[T]

    def __init__(self, client: Client[T], codec_options: Optional[CodecOptions[U]] = None) -> None:
        self._document_type = client.document_type

    def bar(self, codec_options: Optional[CodecOptions[U]] = None) -> T:
        return self._document_type()



f = Database(Client(dict))
f.bar()['a'] = 1

g = Database(Client(RawBSONDocument))
# Throws type error: "__setitem__" method not defined on type "RawBSONDocument"
#g.bar()['a'] = 1
```

pymongo/collection.py
```diff
@@ -356,8 +380,13 @@ def with_options(self, codec_options=None, read_preference=None,
                           write_concern or self.write_concern,
                           read_concern or self.read_concern)
 
-    def bulk_write(self, requests, ordered=True,
-                   bypass_document_validation=False, session=None):
+    def bulk_write(
+        self,
+        requests: List[WriteOp],
```
adavis444, line 385: Given a file
```py
collection = Collection()  # type: ignore
requests = [InsertOne({})]
collection.bulk_write(requests)
```

running mypy gives us
```
error: Argument 1 to "bulk_write" of "Collection" has incompatible type "List[InsertOne]"; expected "List[Union[InsertOne, DeleteOne, DeleteMany, ReplaceOne, UpdateOne, UpdateMany]]"
note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
note: Consider using "Sequence" instead, which is covariant
```

Changing this line to
```py
        requests: Sequence[WriteOp],
```
resolves this issue.

`common.validate_list("requests", requests)` on [line R455](https://github.com/mongodb/mongo-python-driver/pull/829/files#diff-00c0d0047a08351562fe8a665b8530f6fc909d6634389104a8982e01c190feb5R455) would likely best be replaced by sequence validation instead of list validation.

(This issue also appears in [mongodb-labs/pymongo-stubs](https://github.com/mongodb-labs/pymongo-stubs/blob/d9d1b0d00f54d179ee81174aeb371356776fbe3a/pymongo-stubs/collection.pyi#L67).)

blink1073: Updated to use `Sequence[WriteOp]` and allow tuples as inputs in several places in eb6bf52

bson/son.py
```diff
@@ -136,51 +143,51 @@ def update(self, other=None, **kwargs):
...
                     for k, v in value.items()])
             else:
                 return value
 
-        return transform_value(dict(self))
+        return cast(Dict[_Key, _Value], transform_value(dict(self)))
 
-    def __deepcopy__(self, memo):
-        out = SON()
+    def __deepcopy__(self, memo: Any) -> "SON[Any, Any]":
```
ShaneHarvey, line 186: I think we can clean this up and revmoe the `cast` like this:
```
    def __deepcopy__(self, memo: MutableMapping[int, "SON[_Key, _Value]"]) -> "SON[_Key, _Value]":

```

bson/timestamp.py
```diff
@@ -54,7 +55,7 @@ def __init__(self, time, inc):
         """
         if isinstance(time, datetime.datetime):
             if time.utcoffset() is not None:
-                time = time - time.utcoffset()
+                time = time - time.utcoffset()  # type: ignore
```
ShaneHarvey, line 58: Is this ignore a workaround for a mypy bug?
```
bson/timestamp.py:58: error: No overload variant of "__sub__" of "datetime" matches argument type "None"
bson/timestamp.py:58: note: Possible overload variants:
bson/timestamp.py:58: note:     def __sub__(self, timedelta) -> datetime
bson/timestamp.py:58: note:     def __sub__(self, datetime) -> timedelta
bson/timestamp.py:58: note: Right operand is of type "Optional[timedelta]"
```

bson/tz_util.py
```diff
@@ -14,10 +14,12 @@
 
 """Timezone related utilities for BSON."""
 
-from datetime import (timedelta,
+from datetime import (datetime, timedelta,
                       tzinfo)
+from time import time
```
ShaneHarvey, line 19: "time" is unused AFAICT.

gridfs/__init__.py
```diff
@@ -664,10 +681,10 @@ def upload_from_stream_with_id(self, file_id, filename, source,
         """
         with self.open_upload_stream_with_id(
                 file_id, filename, chunk_size_bytes, metadata,
-                session=session) as gin:
+                session=session) as gin:  # type: GridIn
```
ShaneHarvey, line 684: Why is `# type: GridIn` needed here?

gridfs/__init__.py
```diff
@@ -848,6 +865,8 @@ def open_download_stream_by_name(self, filename, revision=-1, session=None):
         query = {"filename": filename}
         _disallow_transactions(session)
         cursor = self._files.find(query, session=session)
+        if revision is None:
+          revision = -1
```
ShaneHarvey, line 869: Why this change? Looking at the API we do not support `revision=None`.

gridfs/__init__.py
```diff
@@ -861,8 +880,9 @@ def open_download_stream_by_name(self, filename, revision=-1, session=None):
             raise NoFile(
                 "no version %d for filename %r" % (revision, filename))
 
-    def download_to_stream_by_name(self, filename, destination, revision=-1,
-                                   session=None):
+    def download_to_stream_by_name(self, filename: str, destination: Any,
+                                   revision: Optional[int] = -1,
+                                   session: Optional[ClientSession] = None) -> None:
```
ShaneHarvey, line 885: I saw elsewhere you used NoReturn. Should we change all the `-> None:` methods to use it too?

gridfs/grid_file.py
```diff
@@ -572,12 +575,14 @@ def read(self, size=-1):
...
 
-    def readline(self, size=-1):
+    def readline(self, size: Optional[int] = -1) -> bytes:
         """Read one line or up to `size` bytes from the file.
 
         :Parameters:
          - `size` (optional): the maximum number of bytes to read
         """
+        if size is None:
+            size = -1
```
ShaneHarvey, line 585: Why this change? The api does not support size=None and I'd like to avoid adding misc features in this PR.

gridfs/grid_file.py
```diff
@@ -761,15 +769,15 @@ def _next_with_retry(self):
         """
         if self._cursor is None:
             self._create_cursor()
-
+        assert self._cursor is not None
```
ShaneHarvey, line 772: suggest identing this line to avoid duplicating the None check twice in the common case.

ShaneHarvey (changes requested):
Great work so far! I've reviewed the bson/ and gridfs/ files and think it's probably a good time to stop and address the various comments. Some of them are applicable to all the changes as a whole.

bson/binary.py
```diff
@@ -337,10 +339,10 @@ def __eq__(self, other):
...
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return super(Binary, self).__hash__() ^ hash(self.__subtype)
 
-    def __ne__(self, other):
+    def __ne__(self, other: Any) -> bool:
         return not self == other
 
     def __repr__(self):
```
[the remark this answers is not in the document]
blink1073: It is required if we eventually move to strict mode, but I purposely didn't include it for now since it wasn't in the stubs repo

bson/__init__.py
```diff
@@ -61,9 +61,10 @@
 import re
 import struct
 import sys
+from typing import TYPE_CHECKING, Any, BinaryIO, Callable, Dict,  Generator, Iterator, List, Mapping, MutableMapping, NoReturn, Sequence, Tuple, Type, Union, cast
```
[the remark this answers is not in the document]
blink1073: I ran isort on all the files: 0f12b61

bson/__init__.py
```diff
@@ -1027,13 +1036,13 @@ def _decode_all_selective(data, codec_options, fields):
 
     # Decode documents for internal use.
     from bson.raw_bson import RawBSONDocument
-    internal_codec_options = codec_options.with_options(
+    internal_codec_options: CodecOptions = codec_options.with_options(
```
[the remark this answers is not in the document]
blink1073: It had been needed in an earlier iteration, removing it now has no effect

bson/codec_options.py
```diff
@@ -92,6 +98,9 @@ class TypeCodec(TypeEncoder, TypeDecoder):
     pass
 
 
+Codec = Union[TypeEncoder, TypeDecoder, TypeCodec]
+Fallback = Callable[[Any], Any]
```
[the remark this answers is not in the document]
blink1073: Do we want to do that for all types, including the ones in `pymongo.typing`?

bson/json_util.py
```diff
@@ -781,8 +789,8 @@ def default(obj, json_options=DEFAULT_JSON_OPTIONS):
             if not obj.tzinfo:
                 obj = obj.replace(tzinfo=utc)
             if obj >= EPOCH_AWARE:
-                off = obj.tzinfo.utcoffset(obj)
-                if (off.days, off.seconds, off.microseconds) == (0, 0, 0):
+                off = obj.tzinfo.utcoffset(obj)  # type: ignore
```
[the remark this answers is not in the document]
blink1073: I added an `assert obj.tzinfo is not None` above and was able to remove the ignore.

bson/objectid.py
```diff
@@ -133,14 +134,14 @@ def from_datetime(cls, generation_time):
             as the generation time for the resulting ObjectId.
         """
         if generation_time.utcoffset() is not None:
-            generation_time = generation_time - generation_time.utcoffset()
+            generation_time = generation_time - generation_time.utcoffset()  # type: ignore
```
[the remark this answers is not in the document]
blink1073: Apparently we can't use the `is not None` trick for the result of a function, so I created an intermediate variable and asserted that it was not None.

.github/workflows/test-python.yml
```diff
@@ -26,3 +26,27 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip mypy
+          pip install -e ".[zstd, srv]"
+      - name: Run mypy
+        run: |
+          mypy --install-types --non-interactive bson
+          mypy --install-types --non-interactive gridfs
+          mypy --install-types --non-interactive pymongo
+          mypy --install-types --non-interactive test
```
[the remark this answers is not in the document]
blink1073: Done

bson/raw_bson.py
```diff
@@ -119,16 +123,16 @@ class from the standard library so it can be used like a read-only
...
-    def raw(self):
+    def raw(self) -> bytes:
         """The raw BSON bytes composing this document."""
         return self.__raw
 
-    def items(self):
+    def items(self) -> ItemsView[str, Any]:
         """Lazily decode and iterate elements in this document."""
-        return self.__inflated.items()
+        return cast(ItemsView[str, Any], self.__inflated.items())
```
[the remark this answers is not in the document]
blink1073: Yep, that worked!

bson/raw_bson.py
```diff
@@ -119,16 +123,16 @@ class from the standard library so it can be used like a read-only
...
 
     @property
-    def __inflated(self):
+    def __inflated(self) -> Any:
```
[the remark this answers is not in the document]
blink1073: Done

bson/__init__.py
```diff
@@ -410,12 +415,12 @@ def _element_to_dict(data, view, position, obj_end, opts):
         return element_name, value, position
 
 
-def _raw_to_dict(data, position, obj_end, opts, result):
+def _raw_to_dict(data: Any, position: int, obj_end: int, opts: Any, result: Any) -> Any:
```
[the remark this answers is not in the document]
blink1073: Cool, done

bson/__init__.py
```diff
@@ -88,39 +89,43 @@
...
-BSONCOD = b"\x0D" # Javascript code
-BSONSYM = b"\x0E" # Symbol
-BSONCWS = b"\x0F" # Javascript code with scope
-BSONINT = b"\x10" # 32bit int
-BSONTIM = b"\x11" # Timestamp
-BSONLON = b"\x12" # 64bit int
-BSONDEC = b"\x13" # Decimal128
-BSONMIN = b"\xFF" # Min key
-BSONMAX = b"\x7F" # Max key
+EPOCH_AWARE: datetime.datetime = datetime.datetime.fromtimestamp(0, utc)
```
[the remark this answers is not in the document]
blink1073: They are required for strict mode, so I had added them

bson/regex.py
```diff
@@ -49,7 +50,7 @@ class Regex(object):
     _type_marker = 11
 
     @classmethod
-    def from_native(cls, regex):
+    def from_native(cls: Type["Regex"], regex: Pattern[Any]) -> "Regex":
```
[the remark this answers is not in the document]
blink1073: https://jira.mongodb.org/browse/PYTHON-3062

bson/son.py
```diff
@@ -53,53 +60,53 @@ def __repr__(self):
...
+    def values(self) -> List[_Value]:  # type: ignore[override]
         return [v for _, v in self.items()]
 
-    def clear(self):
+    def clear(self) -> None:
         self.__keys = []
         super(SON, self).clear()
 
-    def setdefault(self, key, default=None):
+    def setdefault(self, key: _Key, default: Optional[Union[_Value, _T]] = None) -> Union[_Value, _T, None]:  # type: ignore[override]
```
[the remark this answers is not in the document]
blink1073: This was copied from https://github.com/mongodb-labs/pymongo-stubs/blob/d23bf60906959f33930fe67f3d9997745a28d6ee/bson-stubs/son.pyi#L29.  I agree it makes sense to remove the `_T` here and in `SON.get`

bson/son.py
```diff
@@ -136,51 +143,51 @@ def update(self, other=None, **kwargs):
...
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> bool:
         """Comparison to another SON is order-sensitive while comparison to a
         regular dictionary is order-insensitive.
         """
         if isinstance(other, SON):
             return len(self) == len(other) and list(self.items()) == \
                    list(other.items())
-        return self.to_dict() == other
+        return cast(bool, self.to_dict() == other)
```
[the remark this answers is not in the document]
blink1073: Removed

bson/son.py
```diff
@@ -136,51 +143,51 @@ def update(self, other=None, **kwargs):
...
-    def to_dict(self):
+    def to_dict(self) -> Dict[_Key, _Value]:
         """Convert a SON document to a normal Python dictionary instance.
 
         This is trickier than just *dict(...)* because it needs to be
         recursive.
         """
 
-        def transform_value(value):
+        def transform_value(value: Union[List[_Value], Dict[_Key, _Value]]) -> Union[List[_Value], Dict[_Key, _Value]]:
```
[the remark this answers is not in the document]
blink1073: Actually, it turns into a recursive mess of types when I do that.   I just punted and made this private method accept and return `Any`

bson/son.py
blink1073: Updated

bson/timestamp.py
blink1073: This is the same issue where the return type of a function can't be used as a "not None" guard

blink1073: I added the intermediate variable workaround

bson/tz_util.py
blink1073: Removed

gridfs/__init__.py
blink1073: Not needed anymore, must have been before full typing was added

gridfs/__init__.py
blink1073: the `Optional[int]` implies `None` is allowed.  I removed the `Optional`

gridfs/__init__.py
blink1073: Those are for functions that always raise an exception

gridfs/grid_file.py
blink1073: This is another case of `Optional[int]`.  In this case the superclass also uses it, we'd have to add an `# ignore: override`:

```
gridfs/grid_file.py:575: error: Argument 1 of "readline" is incompatible with supertype "IOBase"; supertype defines the argument type as "Optional[int]"
gridfs/grid_file.py:575: note: This violates the Liskov substitution principle
gridfs/grid_file.py:575: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

gridfs/grid_file.py
blink1073: None

bson/timestamp.py
```diff
@@ -54,7 +55,9 @@ def __init__(self, time, inc):
         """
         if isinstance(time, datetime.datetime):
             if time.utcoffset() is not None:
```
ShaneHarvey, line 57: Suggest rewriting to remove the duplicate utcoffset() call and redudant assert:
```python
offset = time.utcoffset()
if offset is not None:...
```

gridfs/__init__.py
```diff
@@ -618,14 +628,15 @@ def upload_from_stream(self, filename, source, chunk_size_bytes=None,
            Added ``session`` parameter.
         """
         with self.open_upload_stream(
-                filename, chunk_size_bytes, metadata, session=session) as gin:
+                filename, chunk_size_bytes, metadata, session=session) as gin:  # type: GridIn
```
ShaneHarvey, line 631: The type: GridIn comment shouldn't be needed anymore.

ShaneHarvey (changes requested):

bson/codec_options.py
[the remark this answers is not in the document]
blink1073: Sounds good, updated

bson/json_util.py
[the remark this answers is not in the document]
blink1073: I moved the assert below that line for clarity

gridfs/grid_file.py
blink1073: Agreed, but because `io.IOBase` uses `Optional[int]`, we have to add an ignore override comment.  I'll do that.

gridfs/grid_file.py
```diff
@@ -637,18 +639,19 @@ def seek(self, pos, whence=_SEEK_SET):
...
         if new_pos == self.__position:
-            return
+            return new_pos
 
         self.__position = new_pos
         self.__buffer = EMPTY
         if self.__chunk_iter:
             self.__chunk_iter.close()
             self.__chunk_iter = None
+        return new_pos
```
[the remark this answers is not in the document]
blink1073: Done

bson/objectid.py
```diff
@@ -133,14 +132,16 @@ def from_datetime(cls, generation_time):
             as the generation time for the resulting ObjectId.
         """
         if generation_time.utcoffset() is not None:
-            generation_time = generation_time - generation_time.utcoffset()
+            offset =  generation_time.utcoffset()
```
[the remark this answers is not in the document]
blink1073: Agreed, updated

bson/son.py
```diff
@@ -20,29 +20,36 @@
 
 import copy
 import re
-
 from collections.abc import Mapping as _Mapping
+from typing import (Any, Dict, Iterable, Iterator, List, Mapping,
+                    MutableMapping, Optional, Pattern, Tuple, Type, TypeVar,
+                    Union, cast)
```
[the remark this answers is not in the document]
blink1073: Another thing we can catch with `pre-commit`.  ;)

bson/son.py
```diff
@@ -176,14 +183,16 @@ def transform_value(value):
...
-    def __deepcopy__(self, memo):
-        out = SON()
+    def __deepcopy__(self, memo: MutableMapping[int, "SON[_Key, _Value]"]) -> "SON[_Key, _Value]":
+        out: SON[_Key, _Value] = SON()
         val_id = id(self)
         if val_id in memo:
-            return memo.get(val_id)
+            value = memo.get(val_id)
+            assert value is not None
+            return value
```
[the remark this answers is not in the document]
blink1073: Ideally `if val_id in memo` would guard against the `get()` being `None`, but it does not, which is why I added the intermediate variable and guard.

bson/son.py
```diff
@@ -176,14 +183,16 @@ def transform_value(value):
...
         memo[val_id] = out
         for k, v in self.items():
             if not isinstance(v, RE_TYPE):
-                v = copy.deepcopy(v, memo)
+                v = copy.deepcopy(v, dict(memo))
```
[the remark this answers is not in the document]
blink1073: Done

bson/son.py
```diff
@@ -136,35 +143,35 @@ def update(self, other=None, **kwargs):
...
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> bool:
         """Comparison to another SON is order-sensitive while comparison to a
         regular dictionary is order-insensitive.
         """
         if isinstance(other, SON):
             return len(self) == len(other) and list(self.items()) == \
                    list(other.items())
-        return self.to_dict() == other
+        return  self.to_dict() == other
```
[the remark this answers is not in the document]
blink1073: Fixed

bson/timestamp.py
blink1073: Done

gridfs/__init__.py
blink1073: Confirmed and removed

blink1073:
Note: I removed the changes to `pymongo` and `test` in order to make this PR easier to review.  I will follow up with those in other PRs.

gridfs/grid_file.py
ShaneHarvey: Oh I see. Yeah override is what typeshed does: https://github.com/python/typeshed/blob/df0fe1645681638359a4ab706732e79c62db6a71/stdlib/io.pyi#L128

test/mockupdb/operations.py
```diff
@@ -14,15 +14,14 @@
 
 from collections import namedtuple
 
-from mockupdb import *
-from mockupdb import OpMsgReply
+from mockupdb import OpMsgReply, OpMsg, OpReply  # type: ignore
```
ShaneHarvey, line 17: What if we use the mypy.ini file to ignore all mockupdb imports instead:
```
[mypy-mockupdb]
ignore_missing_imports = True
```

test/performance/perf_test.py
```diff
@@ -44,7 +45,7 @@
 
 OUTPUT_FILE = os.environ.get('OUTPUT_FILE')
 
-result_data = []
+result_data: List[Any] = []
```
ShaneHarvey, line 48: Isn't this annotation redundant?

bson/son.py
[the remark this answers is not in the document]
ShaneHarvey: Yeah I agree but I think it's similar to the other issue since mypy can't determine if the val_id might be removed by another thread in between the if-statement and the get() call. Regardless, I think we can replace these 3 lines with `return memo[val_id]` since we know val_id is always there. Then we don't need to add redundant code just to appease mypy.

ShaneHarvey (changes requested):
Thanks for pushing out the pymongo/+test/ changes!

test/performance/perf_test.py
blink1073: `test/performance/perf_test.py:48: error: Need type annotation for "result_data" (hint: "result_data: List[<type>] = ...")`

doc/changelog.rst
```diff
@@ -21,6 +21,8 @@ PyMongo 4.0 brings a number of improvements including:
   $merge and $out executing on secondaries on MongoDB >=5.0.
   aggregate() now always obeys the collection's :attr:`read_preference` on
   MongoDB >= 5.0.
+- :meth:~gridfs.GridOut.seek` now returns the new position in the file, to
```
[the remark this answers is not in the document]
blink1073: Fixed

test/mockupdb/operations.py
blink1073: Done

bson/son.py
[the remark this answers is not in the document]
blink1073: Used type ignore instead

test/performance/perf_test.py
ShaneHarvey: Oh that's surprising. I thought it would default to List. TIL.

bson/son.py
```diff
@@ -176,11 +182,11 @@ def transform_value(value):
...
         return transform_value(dict(self))
 
-    def __deepcopy__(self, memo):
-        out = SON()
+    def __deepcopy__(self, memo: Dict[int, "SON[_Key, _Value]"]) -> "SON[_Key, _Value]":
+        out: SON[_Key, _Value] = SON()
         val_id = id(self)
         if val_id in memo:
-            return memo.get(val_id)
+            return memo.get(val_id)  # type: ignore
```
ShaneHarvey, line 189: We can make this type-safe and remove `# type: ignore` with `return memo[val_id]` instead of using .get().

test/performance/perf_test.py
blink1073: It defaulted to `List[unknown]`, but adding `:  List` allowed it to reference as `List[Any]`, so the `[Any]` was redundant.

bson/son.py
blink1073: Oh, ha, duh, thanks

ShaneHarvey (approved):
LGTM!

PR #830, opened 2022-01-12
PYTHON-2863 Allow hint for unacknowledged writes using OP_MSG when supported by the server

blink1073:


pymongo/bulk.py
```diff
@@ -267,6 +267,14 @@ def _execute_command(self, generator, write_concern, session,
         # sock_info.write_command.
         sock_info.validate_session(client, session)
         while run:
+            # Check to see if this is the last run by peeking.
+            # On the last run we use the final write concern.
+            try:
+                next_run = next(generator)
+                generator = chain([next_run], generator)
+            except StopIteration:
+                write_concern = final_write_concern or write_concern
```
ShaneHarvey, line 276: A "run" is a sequence of the same operation type, eg `bulk_write([UpdateOne({}), InsertOne({}), InsertOne({})])` will have one run with one Update and then a second run with two inserts. I don't think this implementation handles the case where the last run is batched into multiple commands. In that case, all the commands must use w:1 writeConcern expect the last command.

I think the actual condition would be:
```
if last_run and last_command_in_run:
    use final_write_concern
else:
    use write_concern
```

However, the point at which we batch split happens _after_ encoding the writeConcern field to the wire protocol message. So this is somewhat of a chicken and egg problem.

pymongo/collection.py
```diff
@@ -608,9 +608,9 @@ def _update(self, sock_info, criteria, document, upsert=False,
             else:
                 update_doc['arrayFilters'] = array_filters
         if hint is not None:
-            if not acknowledged:
+            if sock_info.max_wire_version < 8:
```
ShaneHarvey, line 611: A strict reading of the spec (and spec tests) actually requires this to be: `if not acknowledged && sock_info.max_wire_version < 8`:
```
class UpdateOptions {
...
   * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
   * This option is only supported by servers >= 4.2. Older servers >= 3.4 will report an error for using this option.
   * For servers < 3.4, the driver MUST raise an error if the caller explicitly provides a value.
   * For unacknowledged writes using OP_UPDATE, the driver MUST raise an error if the caller explicitly provides a value.
   * For unacknowledged writes using OP_MSG and servers < 4.2, the driver MUST raise an error if the caller explicitly provides a value.
   *
   * @see https://docs.mongodb.com/manual/reference/command/update/
   */
  hint: Optional<(String | Document)>;
```
https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#insert-update-replace-delete-and-bulk-writes

I think this is why the test are failing.

pymongo/collection.py
```diff
@@ -2277,12 +2277,9 @@ def _find_and_modify(session, sock_info, retryable_write):
                         'writes.')
                 cmd["arrayFilters"] = array_filters
             if hint is not None:
-                if sock_info.max_wire_version < 8:
-                    raise ConfigurationError(
-                        'Must be connected to MongoDB 4.2+ to use hint.')
-                if not write_concern.acknowledged:
+                if sock_info.max_wire_version < 9:
```
ShaneHarvey, line 2280: A strict reading of the crud spec and tests means this needs to be:
```
if sock_info.max_wire_version < 8 or (not acknowledged && sock_info.max_wire_version < 9): raise <>
```
See:
```
class FindOneAndDeleteOptions {
...
  /**
   * The index to use. Specify either the index name as a string or the index key pattern.
   * If specified, then the query system will only consider plans using the hinted index.
   *
   * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
   * This option is only supported by servers >= 4.4. Older servers >= 4.2 will report an error for using this option.
   * For servers < 4.2, the driver MUST raise an error if the caller explicitly provides a value.
   * For unacknowledged writes and servers < 4.4, the driver MUST raise an error if the caller explicitly provides a value.
   *
   * @see https://docs.mongodb.com/manual/reference/command/findAndModify/
   */
  hint: Optional<(String | Document)>;
```
https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#find-and-modify

pymongo/collection.py
```diff
@@ -954,9 +954,9 @@ def _delete(
             else:
                 delete_doc['collation'] = collation
         if hint is not None:
-            if not acknowledged:
+            if sock_info.max_wire_version < 9:
```
ShaneHarvey, line 957: A strict reading of the spec (and spec tests) actually requires this to be: `if not acknowledged && sock_info.max_wire_version < 8`:

```
class DeleteOptions {
...
   * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
   * This option is only supported by servers >= 4.4. Older servers >= 3.4 will report an error for using this option.
   * For servers < 3.4, the driver MUST raise an error if the caller explicitly provides a value.
   * For unacknowledged writes using OP_DELETE, the driver MUST raise an error if the caller explicitly provides a value.
   * For unacknowledged writes using OP_MSG and servers < 4.4, the driver MUST raise an error if the caller explicitly provides a value.
   *
   * @see https://docs.mongodb.com/manual/reference/command/delete/
   */
  hint: Optional<(String | Document)>;
```

ShaneHarvey (changes requested):

pymongo/collection.py
blink1073: Fixed

pymongo/collection.py
blink1073: Fixed

pymongo/collection.py
blink1073: It should be `if not acknowledged && sock_info.max_wire_version < 9`  (driver 4.4), correct?

ShaneHarvey: Yes you're right, my mistake.

pymongo/bulk.py
blink1073: I think the latest commit addresses this

pymongo/bulk.py
```diff
@@ -254,25 +254,38 @@ def gen_unordered(self):
...
     def _execute_command(self, generator, write_concern, session,
-                         sock_info, op_id, retryable, full_result):
+                         sock_info, op_id, retryable, full_result, final_write_concern=None):
         db_name = self.collection.database.name
         client = self.collection.database.client
         listeners = client._event_listeners
 
         if not self.current_run:
             self.current_run = next(generator)
+            self.next_run = None
```
ShaneHarvey, line 264: next_run should be defined in `__init__`.

pymongo/bulk.py
```diff
@@ -254,25 +254,38 @@ def gen_unordered(self):
                 yield run
 
     def _execute_command(self, generator, write_concern, session,
-                         sock_info, op_id, retryable, full_result):
+                         sock_info, op_id, retryable, full_result, final_write_concern=None):
```
ShaneHarvey, line 257: Let's move final_write_concern on a new line to fit in 80chars.

pymongo/bulk.py
```diff
@@ -290,33 +303,50 @@ def _execute_command(self, generator, write_concern, session,
...
+                    _merge_command(run, full_result, run.idx_offset, result)
+
+                    # We're no longer in a retry once a command succeeds.
+                    self.retrying = False
+                    self.started_retryable_write = False
+
+                    if self.ordered and "writeErrors" in result:
+                        break
+                else:
+                    # Guard against unsupported unacknowledged writes.
```
ShaneHarvey, line 329: I think this guard should happen in the same place it was before (line 414) because we should raise ConfigurationError before executing any operation. To do that we might need two variables `self.hint_update` and `self.hint_delete`.

pymongo/bulk.py
```diff
@@ -393,16 +423,16 @@ def execute_command_no_results(self, sock_info, generator):
...
             self._execute_command(
-                generator, write_concern, None,
-                sock_info, op_id, False, full_result)
+                generator, initial_write_concern, None,
+                sock_info, op_id, False, full_result, write_concern)
         except OperationFailure:
             pass
 
-    def execute_no_results(self, sock_info, generator):
+    def execute_no_results(self, sock_info, generator, write_concern=None):
```
ShaneHarvey, line 435: Let's make write_concern a required argument since we always pass it.

pymongo/bulk.py
```diff
@@ -377,7 +407,7 @@ def execute_op_msg_no_results(self, sock_info, generator):
                 run.idx_offset += len(to_send)
             self.current_run = run = next(generator, None)
 
-    def execute_command_no_results(self, sock_info, generator):
+    def execute_command_no_results(self, sock_info, generator, write_concern=None):
```
ShaneHarvey, line 410: Let's make write_concern a required argument since we always pass it.

ShaneHarvey (changes requested):
Nice work! A few suggestions.

pymongo/bulk.py
blink1073: Done

pymongo/bulk.py
blink1073: Done

pymongo/bulk.py
blink1073: Done

pymongo/bulk.py
blink1073: Done

pymongo/bulk.py
blink1073: Done

ShaneHarvey (approved):
LGTM once all the tests pass.

PR #836, opened 2022-01-20
PYTHON-3046 Document support for backslashreplace and surrogateescape

ShaneHarvey:


bson/codec_options.py
```diff
@@ -233,7 +233,8 @@ class CodecOptions(_options_base):
       - `unicode_decode_error_handler`: The error handler to apply when
         a Unicode-related error occurs during BSON decoding that would
         otherwise raise :exc:`UnicodeDecodeError`. Valid options include
-        'strict', 'replace', and 'ignore'. Defaults to 'strict'.
+        'strict', 'replace', 'backslashreplace', 'surrogateescape', and
+        'ignore'. Defaults to 'strict'.
```
ShaneHarvey, line 237: I decided not to link to https://docs.python.org/3/library/codecs.html#error-handlers here because not all of Python's builtin handlers work with bson decoding. For example xmlcharrefreplace, namereplace, and surrogatepass do not work.

behackett (approved):

bson/codec_options.py
juliusgeo: Interesting. Makes sense.

juliusgeo (approved):
LGTM!

blink1073 (approved):

PR #837, opened 2022-01-21
PYTHON-3069 Require hello command + OP_MSG when 'loadBalanced=True'

juliusgeo:
This is a follow up to discussion in [DRIVERS-2033](https://github.com/mongodb/specifications/pull/1123) in which @ShaneHarvey mentioned that it should be implemented in Python first to ensure that the serverless tests still pass. 
Link to passing serverless tests Evergreen build: https://evergreen.mongodb.com/build/mongo_python_driver_serverless__platform~ubuntu_18.04_auth_ssl~auth_ssl_python_version~3.7_serverless~enabled_patch_9cb64775c97a800e2c3a7cf54cb732f95b37332b_61eb166e3627e067e3ccbb3b_22_01_21_20_25_11

### DIFF
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -550,6 +550,8 @@ class SocketInfo(object):
             self.close_socket(ConnectionClosedReason.STALE)
 
     def hello_cmd(self):
+        # Handshake spec requires us to use OP_MSG+hello command for the
+        # initial handshake in load balanced or versioned api mode
         if self.opts.server_api or self.hello_ok or self.opts.load_balanced:
             self.op_msg_enabled = True
             return SON([(HelloCompat.CMD, 1)])
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -551,6 +551,7 @@ class SocketInfo(object):
 
     def hello_cmd(self):
         if self.opts.server_api or self.hello_ok or self.opts.load_balanced:
+            self.op_msg_enabled = True
             return SON([(HelloCompat.CMD, 1)])
         else:
             return SON([(HelloCompat.LEGACY_CMD, 1), ('helloOk', True)])
@@ -592,7 +593,8 @@ class SocketInfo(object):
             auth_ctx = None
 
         doc = self.command('admin', cmd, publish_events=False,
-                           exhaust_allowed=awaitable)
+                           exhaust_allowed=awaitable, use,
+                           use_op_msg=self.op_msg_enabled)
         # PYTHON-2712 will remove this topologyVersion fallback logic.
         if self.opts.load_balanced and _MOCK_SERVICE_ID:
             process_id = doc.get('topologyVersion', {}).get('processId')
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -551,7 +551,7 @@ class SocketInfo(object):
 
     def hello_cmd(self):
         # Handshake spec requires us to use OP_MSG+hello command for the
-        # initial handshake in load balanced or versioned api mode
+        # initial handshake in load balanced or versioned api mode.
         if self.opts.server_api or self.hello_ok or self.opts.load_balanced:
             self.op_msg_enabled = True
             return SON([(HelloCompat.CMD, 1)])
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -550,7 +550,7 @@ class SocketInfo(object):
             self.close_socket(ConnectionClosedReason.STALE)
 
     def hello_cmd(self):
-        if self.opts.server_api or self.hello_ok:
+        if self.opts.server_api or self.hello_ok or self.opts.load_balanced:
             return SON([(HelloCompat.CMD, 1)])
         else:
             return SON([(HelloCompat.LEGACY_CMD, 1), ('helloOk', True)])
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -593,8 +593,7 @@ class SocketInfo(object):
             auth_ctx = None
 
         doc = self.command('admin', cmd, publish_events=False,
-                           exhaust_allowed=awaitable,
-                           use_op_msg=self.op_msg_enabled)
+                           exhaust_allowed=awaitable)
         # PYTHON-2712 will remove this topologyVersion fallback logic.
         if self.opts.load_balanced and _MOCK_SERVICE_ID:
             process_id = doc.get('topologyVersion', {}).get('processId')
@@ -683,6 +682,7 @@ class SocketInfo(object):
             using the TypeDecoders from codec_options, passed to
             bson._decode_all_selective.
         """
+
         self.validate_session(client, session)
         session = _validate_session_write_concern(session, write_concern)
 
@@ -697,7 +697,6 @@ class SocketInfo(object):
         if (write_concern and
                 not write_concern.is_server_default):
             spec['writeConcern'] = write_concern.document
-
         self.add_server_api(spec)
         if session:
             session._apply_to(spec, retryable_write, read_preference,
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -682,7 +682,6 @@ class SocketInfo(object):
             using the TypeDecoders from codec_options, passed to
             bson._decode_all_selective.
         """
-
         self.validate_session(client, session)
         session = _validate_session_write_concern(session, write_concern)
 
@@ -697,6 +696,7 @@ class SocketInfo(object):
         if (write_concern and
                 not write_concern.is_server_default):
             spec['writeConcern'] = write_concern.document
+
         self.add_server_api(spec)
         if session:
             session._apply_to(spec, retryable_write, read_preference,
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -593,7 +593,7 @@ class SocketInfo(object):
             auth_ctx = None
 
         doc = self.command('admin', cmd, publish_events=False,
-                           exhaust_allowed=awaitable, use,
+                           exhaust_allowed=awaitable,
                            use_op_msg=self.op_msg_enabled)
         # PYTHON-2712 will remove this topologyVersion fallback logic.
         if self.opts.load_balanced and _MOCK_SERVICE_ID:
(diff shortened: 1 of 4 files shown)


pymongo/pool.py
```diff
@@ -550,7 +550,7 @@ def unpin(self):
             self.close_socket(ConnectionClosedReason.STALE)
 
     def hello_cmd(self):
-        if self.opts.server_api or self.hello_ok:
+        if self.opts.server_api or self.hello_ok or self.opts.load_balanced:
```
ShaneHarvey, line 553: We still need to change it to use OP_MSG for the initial handshake.

ShaneHarvey (changes requested):

pymongo/pool.py
juliusgeo: I believe it does use it for the initial handshake. It is called in the same file
```
def _hello(self, cluster_time, topology_version, heartbeat_frequency):
        cmd = self.hello_cmd()
```
And then it modifies command but when it performs the initial handshake it uses that command which is instantiated from that function:
```
doc = self.command('admin', cmd, publish_events=False,
                           exhaust_allowed=awaitable)
        # PYTHON-2712 will remove this topologyVersion fallback logic.
        if self.opts.load_balanced and _MOCK_SERVICE_ID:
            process_id = doc.get('topologyVersion', {}).get('processId')
            doc.setdefault('serviceId', process_id)
        if not self.opts.load_balanced:
            doc.pop('serviceId', None)
        hello = Hello(doc, awaitable=awaitable)
        self.is_writable = hello.is_writable
        self.max_wire_version = hello.max_wire_version
        self.max_bson_size = hello.max_bson_size
        self.max_message_size = hello.max_message_size
        self.max_write_batch_size = hello.max_write_batch_size
```
I also took a look and it appears that the `hello` function is being called from the connect function:
```
def connect(self):
        sock_info = SocketInfo(sock, self, self.address, conn_id)
        try:
            if self.handshake:
                sock_info.hello()
```

ShaneHarvey: The hello command != OP_MSG. This change is sending the initial hello command with OP_QUERY. Take a look at where we decide how to use op_msg vs op_query.

test/test_client_context.py
```diff
@@ -21,6 +21,7 @@
 
 
 class TestClientContext(unittest.TestCase):
+    @client_context.require_connection
```
ShaneHarvey, line 24: These tests are testing client_context itself so you can't use require_connection here. If this test is failing then somethign else is broken.

test/test_client_context.py
```diff
@@ -30,6 +31,7 @@ def test_must_connect(self):
                         'PYMONGO_MUST_CONNECT is set. Failed attempts:\n%s' %
                         (client_context.connection_attempt_info(),))
 
+    @client_context.require_connection
```
ShaneHarvey, line 34: Same here.

ShaneHarvey (changes requested):

pymongo/pool.py
```diff
@@ -550,7 +550,8 @@ def unpin(self):
...
+            self.op_msg_enabled = True
```
ShaneHarvey, line 554: I think it's worth adding a comment here saying why. Something like "Handshake spec requires us to use OP_MSG+hello command for the initial handshake in load balanced or versioned api mode."

ShaneHarvey (changes requested):
Please run the load balancer, serverless, and versioned API tests on the latest commit. We should also add a mockupdb test to verify that the driver always uses OP_MSG when versioned api or load balanced mode is configured.

test/mockupdb/test_handshake.py
```diff
@@ -11,14 +11,56 @@
...
+from pymongo.server_api import ServerApi
 
 import unittest
-
+from copy import deepcopy
+import time
+
+def test_handshake_with_option(self, protocol, num_servers, **kwargs):
+    primary = MockupDB(verbose=False)
+    secondaries = [MockupDB(verbose=False) for _ in range(num_servers-1)]
```
ShaneHarvey, line 27: Is `verbose=False` needed? We never use it in any other tests.
Also since the handshake is the same for primaries and secondaries we can use a single MockupDB node instead of 10 servers.

test/mockupdb/test_handshake.py
```diff
@@ -11,14 +11,56 @@
...
+    for server in servers:
+        server.autoresponds(hangup)
+        server.run()
+        self.addCleanup(server.stop)
+    hosts = [server.address_string for server in servers if not kwargs.get(
+        "loadBalanced") or server == primary]
+    address_str = "mongodb://"+','.join(hosts)
+    client = MongoClient(address_str,
+                         appname='my app',
+                         heartbeatFrequencyMS=500, # Speed up the test.
```
ShaneHarvey, line 47: Calling hangup() is what's causing the test to be slow. We should autorespond with a normal hello response instead and heartbeatFrequencyMS wouldn't be needed. This way the test won't take >500ms.

test/mockupdb/test_handshake.py
```diff
@@ -11,14 +11,56 @@
...
+                         **kwargs)
+    if isinstance(kwargs.get("server_api"), ServerApi):
+        kwargs.pop("server_api", None)
+        kwargs["apiVersion"] = "1"
+
+    self.addCleanup(client.close)
+    future = go(client.db.command, "whatever")
+    time.sleep(.1)
```
ShaneHarvey, line 55: This sleep isn't needed once we actually run the command. I suggest using `client.db.command(hello)` to reuse the autoresponder.

test/mockupdb/test_handshake.py
```diff
@@ -11,14 +11,56 @@
...
+    primary = MockupDB(verbose=False)
+    secondaries = [MockupDB(verbose=False) for _ in range(num_servers-1)]
+    servers = [primary]+secondaries
+    hello = "ismaster" if isinstance(protocol(), OpQuery) else "hello"
+
+    # set up a custom handler to log and then immediately close all connections
+    # this makes sure that we only are checking the hello command, because
+    # after the first response it has been closed
+    reqs = []
+    def hangup(response):
```
ShaneHarvey, line 35: response -> request?

test/mockupdb/test_handshake.py
```diff
@@ -11,14 +11,56 @@
...
+        reqs.append(deepcopy([response.doc, response.__class__]))
```
ShaneHarvey, line 36: Why is the deepcopy needed? Can we do `reqs.append(request)`?

test/mockupdb/test_handshake.py
```diff
@@ -11,14 +11,56 @@
...
+        kwargs["apiVersion"] = "1"
+
+    self.addCleanup(client.close)
+    future = go(client.db.command, "whatever")
+    time.sleep(.1)
+    # check each server to see if it received messages, if it did then
+    # check them to make sure they match our protocol and kwargs
+    assert len(reqs) == len(hosts)
+    for i, t in reqs:
+        i = Request(i)
```
ShaneHarvey, line 60: Can we reuse the original request instead of creating a new one? Then we can reply on assert_matches to handle all the assertions, eg: `req.assert_matches(protocol(hello, **kwargs))

pymongo/pool.py
```diff
@@ -550,7 +550,10 @@ def unpin(self):
             self.close_socket(ConnectionClosedReason.STALE)
 
     def hello_cmd(self):
-        if self.opts.server_api or self.hello_ok:
+        # Handshake spec requires us to use OP_MSG+hello command for the
+        # initial handshake in load balanced or versioned api mode
```
ShaneHarvey, line 554: Missing period.

test/mockupdb/test_handshake.py
```diff
@@ -11,14 +11,56 @@
...
+
+def test_handshake_with_option(self, protocol, num_servers, **kwargs):
+    primary = MockupDB(verbose=False)
+    secondaries = [MockupDB(verbose=False) for _ in range(num_servers-1)]
+    servers = [primary]+secondaries
+    hello = "ismaster" if isinstance(protocol(), OpQuery) else "hello"
+
+    # set up a custom handler to log and then immediately close all connections
+    # this makes sure that we only are checking the hello command, because
+    # after the first response it has been closed
```
ShaneHarvey, line 33: Comments should be grammatically correct sentences (eg start with capital letter, end with punctuation).

ShaneHarvey (changes requested):
Great, now that you've confirmed the serverless tests are passing I've approved the DRIVERS-2033.

test/mockupdb/test_handshake.py
juliusgeo: Done.

test/mockupdb/test_handshake.py
juliusgeo: This was I think due to having the same autoresponder on every server. It is fixed now.

blink1073 (approved):
LGTM!

test/mockupdb/test_handshake.py
```diff
@@ -12,14 +12,58 @@
...
+                                        ServerApiVersion.V1))}
+    client = MongoClient("mongodb://"+primary.address_string,
+                         appname='my app', # For _check_handshake_data()
+                         **dict([k_map.get((k, v), (k, v)) for k, v
+                                 in kwargs.items()]))
+    
+    self.addCleanup(client.close)
+
+    # We have an autoresponder luckily, so no need for `go()`.
+    assert client.db.command(hello)["ok"], f"%s command failed" % (hello)
```
ShaneHarvey, line 57: FYI db.command() already checks the "ok" field and raises an error if the command failed so this line can be:
```
assert client.db.command(hello)
```

test/mockupdb/test_handshake.py
```diff
@@ -12,14 +12,58 @@
...
+
+    # We do this checking in here rather than hangup() because hangup runs
```
ShaneHarvey, line 59: hangup doesn't exist anymore.

ShaneHarvey (changes requested):

test/mockupdb/test_handshake.py
juliusgeo: Ok cool. Fixed.

test/mockupdb/test_handshake.py
juliusgeo: Done.

ShaneHarvey (approved):
LGTM

PR #840, opened 2022-01-31
PYTHON-3075 bulk_write does not apply CodecOptions to upserted_ids result

blink1073:


juliusgeo:
LGTM!

juliusgeo (approved):

pymongo/bulk.py
```diff
@@ -99,8 +101,17 @@ def _merge_command(run, full_result, offset, result):
...
         if upserted:
             n_upserted = len(upserted)
+            rep = codec_options.uuid_representation
             for doc in upserted:
                 doc["index"] = run.index(doc["index"] + offset)
+                # Normalize uuid representation.
+                _id = doc['_id']
+                if rep == UuidRepresentation.UNSPECIFIED:
+                    if isinstance(_id, uuid.UUID):
+                        _id = Binary.from_uuid(_id)
```
blink1073, line 111: I couldn't figure out how to trigger this condition

ShaneHarvey: This condition will only happen on pymongo 3.

pymongo/bulk.py
```diff
@@ -317,10 +329,10 @@ def _execute_command(self, generator, write_concern, session,
                         # Synthesize the full bulk result without modifying the
                         # current one because this write operation may be retried.
                         full = copy.deepcopy(full_result)
-                        _merge_command(run, full, run.idx_offset, result)
+                        _merge_command(run, full, run.idx_offset, result, codec_options)
```
ShaneHarvey, line 332: One problem with this approach is that it doesn't handle `_id` that are subdocuments, eg: `_id={'f': Binary(), 'f2': ...}`. Could you add a test for this as well?

Instead of converting the result after the fact, we should ensure that BulkWriteContext.execute() decodes the command response with the correct uuid_representation/CodecOptions to begin with. This will involve passing the Collection's `__write_response_codec_options` to the BulkWriteContext to later use for decoding.

test/test_bulk.py
```diff
@@ -376,6 +379,27 @@ def test_client_generated_upsert_id(self):
                           {'index': 2, '_id': 2}]},
             result.bulk_api_result)
 
+    def test_upsert_uuid_standard(self):
+        options = CodecOptions(uuid_representation=UuidRepresentation.STANDARD)
```
ShaneHarvey, line 383: Nice. Could you add a similar test to ensure that UuidRepresentation.UNSPECIFIED results in Binary instances?

ShaneHarvey (changes requested):

pymongo/bulk.py
blink1073: Cool, fixed the decoding in the latest commit.  I'll address the tests in the morning.

test/test_bulk.py
blink1073: As far as I can tell we can't do that because it raises a [`ValueError`](https://github.com/mongodb/mongo-python-driver/blob/308b4f4e08c9fe91210ee498adbbafe0c7ebebe4/bson/binary.py#L253).

pymongo/bulk.py
blink1073: Added the subdocument test

test/test_bulk.py
ShaneHarvey: Yes the test will need to convert the uuids to Binary manually before inserting them:
```
uuids = [Binary.from_uuid(uuid.uuid4()) for _ in range(3)]
```

pymongo/message.py
```diff
@@ -1273,9 +1273,9 @@ def unpack_response(self, cursor_id=None,
         return bson._decode_all_selective(
             self.payload_document, codec_options, user_fields)
 
-    def command_response(self):
+    def command_response(self, codec_options=_UNICODE_REPLACE_CODEC_OPTIONS):
```
ShaneHarvey, line 1276: Since we only call this method in one place, let's make codec_options a required argument to help prevent this kind of bug in the future.

pymongo/pool.py
```diff
@@ -775,7 +775,7 @@ def unack_write(self, msg, max_doc_size):
         self._raise_if_not_writable(True)
         self.send_message(msg, max_doc_size)
 
-    def write_command(self, request_id, msg):
+    def write_command(self, request_id, msg, codec_options=DEFAULT_CODEC_OPTIONS):
```
ShaneHarvey, line 778: Same for codec_options here.

test/test_bulk.py
```diff
@@ -376,6 +380,57 @@ def test_client_generated_upsert_id(self):
...
+             'upserted': [{'index': 0, '_id': uuids[0]},
+                          {'index': 1, '_id': uuids[1]},
+                          {'index': 2, '_id': uuids[2]}]},
+            result.bulk_api_result)
+
+    def test_upsert_uuid_standard_subdocuments(self):
+        options = CodecOptions(uuid_representation=UuidRepresentation.STANDARD)
+        coll = self.coll.with_options(codec_options=options)
+        ids = [
+            {'f': Binary(uuid.uuid4().bytes), 'f2': uuid.uuid4()}
```
ShaneHarvey, line 408: instead of `Binary(uuid.uuid4().bytes)` can we do `Binary(bytes(i))`? The use of uuid there is a little confusing.

ShaneHarvey (changes requested):

pymongo/message.py
blink1073: I also changed `_OpReply.command_response()` for uniformity.

pymongo/pool.py
blink1073: Done

test/test_bulk.py
blink1073: Done

test/test_bulk.py
blink1073: Done

ShaneHarvey (approved):
LGTM!

pymongo/message.py
```diff
@@ -798,7 +798,7 @@ def write_command(self, cmd, request_id, msg, docs):
             self._start(cmd, request_id, docs)
             start = datetime.datetime.now()
         try:
-            reply = self.sock_info.write_command(request_id, msg)
+            reply = self.sock_info.write_command(request_id, msg, self.codec)
```
ShaneHarvey, line 801: We need a similar change in _EncryptedBulkWriteContext.execute().

ShaneHarvey (changes requested):
Sorry I just realized we're missing the encrypted bulk case. We should add a similar test (just one is enough) to test_encryption.py.

pymongo/message.py
blink1073: I wasn't sure how to test that locally, waiting for evergreen

ShaneHarvey (approved):
LGTM!

PR #844, opened 2022-02-03
PYTHON-3064 Add typings to test package

blink1073:


test/test_cursor.py
```diff
@@ -22,6 +22,8 @@
 import time
 import threading
 
+from typing import no_type_check
```
ShaneHarvey, line 25: Is this needed? test_cursor seems to type check without errors locally for me.

mypy.ini
```diff
@@ -29,5 +35,11 @@ ignore_missing_imports = True
 [mypy-snappy.*]
 ignore_missing_imports = True
 
+[mypy-test.*]
+allow_redefinition = true
```
ShaneHarvey, line 36: Should we add allow_untyped_globals too?

test/utils.py
```diff
@@ -864,21 +864,14 @@ def frequent_thread_switches():
     """Make concurrency bugs more likely to manifest."""
     interval = None
     if not sys.platform.startswith('java'):
```
ShaneHarvey, line 866: You can remove the 'java' if statement altogether because we don't test Jython anymore.

test/utils.py
```diff
@@ -864,21 +864,14 @@ def frequent_thread_switches():
...
-        else:
-            interval = sys.getcheckinterval()
-            sys.setcheckinterval(1)
+        interval = sys.getswitchinterval()
+        sys.setswitchinterval(1e-6)
 
     try:
         yield
     finally:
         if not sys.platform.startswith('java'):
```
ShaneHarvey, line 873: Same here.

.github/workflows/test-python.yml
```diff
@@ -46,4 +46,4 @@ jobs:
           pip install -e ".[zstd, srv]"
       - name: Run mypy
         run: |
-          mypy --install-types --non-interactive bson gridfs tools
+          mypy --install-types --non-interactive bson gridfs tools test
```
ShaneHarvey, line 49: Does it make sense to separate the test/ type checking so that we can globally ignore certain error codes in the test/ package? Eg:
```
mypy --install-types --non-interactive bson gridfs tools
mypy --install-types --non-interactive --disable-error-code var-annotated --disable-error-code attr-defined --disable-error-code union-attr --disable-error-code assignment test
```

Disabling some checks could make the tests easier to maintain.

test/mockupdb/test_reset_and_request_check.py
```diff
@@ -45,7 +45,7 @@ def responder(request):
         kwargs = {'socketTimeoutMS': 100}
         # Disable retryable reads when pymongo supports it.
         kwargs['retryReads'] = False
-        self.client = MongoClient(self.server.uri, **kwargs)
+        self.client = MongoClient(self.server.uri, **kwargs)  # type: ignore
```
ShaneHarvey, line 48: What's up with the type ignore here? It seems perfectly reasonable to pass `**kwargs` to MongoClient because it accepts `**kwargs`. Is this something we should report to mypy?

test/test_change_stream.py
```diff
@@ -121,6 +122,7 @@ def kill_change_stream_cursor(self, change_stream):
 
 
 class APITestsMixin(object):
```
ShaneHarvey, line 124: I think this file is improperly using inheritance. Could you open a tech debt ticket so we can fix this at some point?

ShaneHarvey (changes requested):
A few points of feedback.

pymongo/socket_checker.py
```diff
@@ -43,7 +42,7 @@ def __init__(self) -> None:
         else:
             self._poller = None
 
-    def select(self, sock: Any, read: bool = False, write: bool = False, timeout: int = 0) -> bool:
+    def select(self, sock: Any, read: bool = False, write: bool = False, timeout: Optional[Union[float, int]] = 0) -> bool:
```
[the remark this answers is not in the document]
blink1073: We call it with `None` explicitly in `test_pooling.py`, and there is internal handling for `None`.  I'll change it to `Optional[float]`

test/test_son.py
```diff
@@ -152,28 +154,28 @@ def test_iteration(self):
...
-        self.assertTrue(2 in test_son, "in failed")
-        self.assertFalse(22 in test_son, "in succeeded when it shouldn't")
-        self.assertTrue(test_son.has_key(2), "has_key failed")
-        self.assertFalse(test_son.has_key(22), "has_key succeeded when it shouldn't")
+        test_son = SON([("1", 100), ("2", 200), ("3", 300)])
+        self.assertIn("1", test_son)
+        self.assertTrue("2" in test_son, "in failed")
+        self.assertFalse("22" in test_son, "in succeeded when it shouldn't")
+        self.assertTrue(test_son.has_key("2"), "has_key failed")
+        self.assertFalse(test_son.has_key("22"), "has_key succeeded when it shouldn't")
```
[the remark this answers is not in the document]
blink1073: Fixed

test/test_bulk.py
```diff
@@ -439,7 +444,7 @@ def test_upsert_uuid_standard_subdocuments(self):
 
         # The `Binary` values are returned as `bytes` objects.
         for _id in ids:
-            _id['f'] = bytes(_id['f'])
+            _id['f'] = bytes(_id['f'])  #
```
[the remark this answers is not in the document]
blink1073: Fixed

test/test_cursor.py
blink1073: Yeah it works fine with the relaxed rules

blink1073: (I removed it)

mypy.ini
blink1073: Done

test/utils.py
blink1073: Done

test/utils.py
blink1073: Done

.github/workflows/test-python.yml
blink1073: Done

test/mockupdb/test_reset_and_request_check.py
blink1073: Yeah it is a type limitation of mypy, I'll look for/file an issue

test/test_change_stream.py
blink1073: I addressed the code review comments, I'll handle the mypy/JIRA tickets tomorrow

test/mockupdb/test_reset_and_request_check.py
blink1073: Found this [comment](https://github.com/python/mypy/issues/11084#issuecomment-917426934): I'll try `TypedDict`

blink1073: Actually, it really is a bug, already reported: https://github.com/python/mypy/issues/8862

blink1073: Created https://jira.mongodb.org/browse/PYTHON-3105 to track this

test/test_change_stream.py
blink1073: https://jira.mongodb.org/browse/PYTHON-3106

test/test_bson.py
```diff
@@ -186,7 +186,7 @@ def test_encode_then_decode_any_mapping_legacy(self):
             decoder=lambda *args: BSON(args[0]).decode(*args[1:]))
 
     def test_encoding_defaultdict(self):
-        dct = collections.defaultdict(dict, [('foo', 'bar')])
+        dct = collections.defaultdict(dict, [('foo', 'bar')])  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
blink1073: Actually, thinking about it more, I don't think this is a bug.   We're declaring that new values in the dictionary are of dictionary type, but the dictionary itself is initialized with a value of string type.  If I use `foo = defaultdict(dict, {'a': dict(a=1)})
` in the [mypy playground](https://mypy-play.net) it works fine.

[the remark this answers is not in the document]
ShaneHarvey: This _is_ a mypy bug. The docs for defaultdict says:
> The first argument provides the initial value for the [default_factory](https://docs.python.org/3/library/collections.html#collections.defaultdict.default_factory) attribute; it defaults to None. All remaining arguments are treated the same as if they were passed to the [dict](https://docs.python.org/3/library/stdtypes.html#dict) constructor, including keyword arguments.

eg:
```
>>> dict([('foo', 'bar')])
{'foo': 'bar'}
>>> dct = collections.defaultdict(dict, [('foo', 'bar')])
>>> dct
defaultdict(<class 'dict'>, {'foo': 'bar'})
>>> dct['g']
{}
>>> dct
defaultdict(<class 'dict'>, {'foo': 'bar', 'g': {}})
```

test/test_examples.py
```diff
@@ -954,12 +956,13 @@ def _insert_employee_retry_commit(session):
...
 
         employee = employees.find_one({"employee": 4})
+        assert employee is not None
         self.assertIsNotNone(employee)
         self.assertEqual(employee['status'], 'Active')
 
         # Start Transactions Retry Example 3
 
-        def run_transaction_with_retry(txn_func, session):
+        def run_transaction_with_retry(txn_func, session):  # type: ignore[no-redef]
```
ShaneHarvey, line 965: Here and below: We can't have type ignore comments in any of these examples otherwise they will get published to the docs. For example: https://docs.mongodb.com/manual/core/transactions-in-applications/#example-1

Is there another way? What if we `del` the old method beforehand?

test/test_examples.py
```diff
@@ -1091,6 +1095,8 @@ def test_causal_consistency(self):
 
         # Start Causal Consistency Example 2
         with client.start_session(causal_consistency=True) as s2:
+            assert s1.cluster_time is not None
+            assert s1.operation_time is not None
```
ShaneHarvey, line 1099: For the same reason as above, let's move this asserts to outside the example (above "Start Causal Consistency Example 2"). Also I'm very confused how this assert helps at all since we're accessing `s1.cluster_time` again below. From earlier conversations I would have expected mypy to not narrow the type here since each property access can theoretically return a different instance.

test/test_grid_file.py
```diff
@@ -238,7 +241,7 @@ def test_grid_out_cursor_options(self):
         cursor_dict.pop('_Cursor__session')
         cursor_clone_dict = cursor_clone.__dict__.copy()
         cursor_clone_dict.pop('_Cursor__session')
-        self.assertDictEqual(cursor_dict, cursor_clone_dict)
+        self.assertEqual(cursor_dict, cursor_clone_dict)
```
ShaneHarvey, line 244: Explanation? Mypy should understand that `cursor.__dict__.copy()` is a dict, no?

test/__init__.py
```diff
@@ -642,7 +654,7 @@ def supports_secondary_read_pref(self):
         if self.has_secondaries:
             return True
         if self.is_mongos:
-            shard = self.client.config.shards.find_one()['host']
+            shard = self.client.config.shards.find_one()['host']  # type: ignore[index]
```
ShaneHarvey, line 657: This is a very common pattern in the test suite. Should we globally ignore the "index" code? Eg:
```python
coll.insert_one({})
coll.find_one()['_id']  # Test can safely assume find_one will never return None here
```

test/test_bson.py
```diff
@@ -530,7 +530,9 @@ def test_large_datetime_truncation(self):
     def test_aware_datetime(self):
         aware = datetime.datetime(1993, 4, 4, 2,
                                   tzinfo=FixedOffset(555, "SomeZone"))
-        as_utc = (aware - aware.utcoffset()).replace(tzinfo=utc)
+        offset = aware.utcoffset()
+        assert offset is not None
+        as_utc = (aware - offset).replace(tzinfo=utc)
```
ShaneHarvey, line 535: This is disappointing. IIUC a datetime with a tzinfo should always return a non-None utcoffset(). I think it could be related to https://github.com/python/mypy/issues/10067 but I think it's worth reporting as a separate issue.

test/test_bson.py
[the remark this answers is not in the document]
blink1073: I guess my point is, they'd have to change the signature of `defaultdict` in [`typeshed`](https://github.com/python/typeshed/blob/494481a0aed2ef0e00bbe190476ace0b8261bce6/stdlib/collections/__init__.pyi#L303) to not use generics, or perhaps add more additional overrides to provide non-generic versions.   I'll update the line instead.  :smile:

pymongo/socket_checker.py
```diff
@@ -16,9 +16,8 @@
 
 import errno
 import select
-import socket
 import sys
-from typing import Any, Optional
+from typing import Any, Optional, Union
```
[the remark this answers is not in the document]
blink1073: Fixed

test/test_bson.py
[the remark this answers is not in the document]
blink1073: Oh, I see what you meant.  Do you still think this is worth bringing up to `typeshed`?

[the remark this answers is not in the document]
ShaneHarvey: Yes we should report this bug to typeshed.

test/test_examples.py
blink1073: That didn't work, I added `--disable-error-code no-redef` to the options instead.

test/test_examples.py
blink1073: Fixed.  Yeah, that does seem like a limitation.

test/test_grid_file.py
blink1073: Ah, this is left over from earlier work, reverting.

test/__init__.py
blink1073: Makes sense, updated the rule and removed the now unused comments.

pymongo/socket_checker.py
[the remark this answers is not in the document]
blink1073: Fixed

test/test_examples.py
```diff
@@ -1089,8 +1093,12 @@ def test_causal_consistency(self):
                  'start': current_date}, session=s1)
         # End Causal Consistency Example 1
 
+        assert s1.cluster_time is not None
+        assert s1.operation_time is not None
+
         # Start Causal Consistency Example 2
         with client.start_session(causal_consistency=True) as s2:
+
```
ShaneHarvey, line 1101: Unneeded blank line.

ShaneHarvey (changes requested):
Just one minor change

test/test_bson.py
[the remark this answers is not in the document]
blink1073: https://github.com/python/typeshed/issues/7159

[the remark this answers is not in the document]
ShaneHarvey: Oh sorry, I just realized I'm wrong here. The problem is that we're passing a 'str' value to a dictionary of type `Dict['str', dict]`. This is not a typeshed bug:
```
>>> dct = collections.defaultdict(dict, [('foo', 'bar')]) # This is a type error
>>> dct = collections.defaultdict(dict, [('foo', {})])  # This works.

```

test/test_bson.py
```diff
@@ -302,7 +302,7 @@ def test_basic_decode(self):
 
     def test_decode_all_buffer_protocol(self):
         docs = [{'foo': 'bar'}, {}]
-        bs = b"".join(map(encode, docs))
+        bs = b"".join(map(encode, docs))  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
blink1073: I'm not sure about this one.  This is another example of a generic in the second argument affecting the value in the first, but I think this time it is warranted.  If we change it to `docs: list[Mapping[str, Any]]` then it matches and passes type check.

test/test_bson.py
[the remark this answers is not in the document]
blink1073: Haha, that is what I was poorly trying to convey.  I still think the raised issue stands, but they are likely to ignore/close it.

test/test_examples.py
blink1073: Fixed

test/test_bson.py
blink1073: I think the only way to do it is to do what I suggested in https://github.com/python/mypy/issues/10067#issuecomment-1032041938

ShaneHarvey (approved):
LGTM!

blink1073:
Merging since the failure is fixed by https://github.com/mongodb/mongo-python-driver/pull/848

PR #845, opened 2022-02-04
PYTHON-2970 SDAM should give priority to electionId over setVersion to detect a stale primary

ShaneHarvey:
Draft implementation of DRIVERS-1954: https://github.com/mongodb/specifications/pull/1122

TODO:
- [X] ServerDescription.election_tuple is now incorrect. Ideally we'd reverse it but doing so would be a breaking change. Perhaps we should just deprecate it? Edit: I opened https://jira.mongodb.org/browse/PYTHON-3162 to deprecation this.


blink1073:
> Perhaps we should just deprecate it?

Makes sense to me

blink1073 (approved):
LGTM!

PR #849, opened 2022-02-08
PYTHON-3110 [v4.0] Remove use of example.com in CSFLE tests

blink1073:


ShaneHarvey:
Please schedule the encryption tests for python 3 and 2.7 to make sure those pass. Also I think the link in your commit message is off. Please revert and use `git cherry-pick -x 561ee7cf77fcbdefb9e2f46691f2b2ba4c65198b`. The `-x` flag adds a message indicating it's a backport with the original commit's sha.

blink1073:
I fixed the commit and scheduled an encryption run for Python 3.9, but we don't support Python 2.7 on 4.0.x.  :smile:

PR #859, opened 2022-02-10
PYTHON-3073 Copy the unit tests from pymongo-stubs into pymongo

blink1073:


behackett (commented):
Isn't this going to require us to have mypy in evergreen and anywhere else we want to run tests?

juliusgeo (approved):
LGTM!

behackett (approved):

PR #860, opened 2022-02-10
PYTHON-3062 Make Regex generic

blink1073:


bson/regex.py
```diff
@@ -116,7 +119,7 @@ def __ne__(self, other: Any) -> bool:
     def __repr__(self):
         return "Regex(%r, %r)" % (self.pattern, self.flags)
 
-    def try_compile(self) -> Pattern[Any]:
+    def try_compile(self) -> "Pattern[_T]":
```
juliusgeo, line 122: This quoting could be prevented by importing:
``from __future__ import annotations``
Is this correct?

juliusgeo (approved):
LGTM!

bson/regex.py
blink1073: That is only supported in Python 3.7+

PR #861, opened 2022-02-10
PYTHON-2850 [v3.13] Deprecate mapReduce command

juliusgeo:
Deprecate mapReduce helpers

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -3121,8 +3121,11 @@ class Collection(common.BaseObject):
         .. seealso:: The MongoDB documentation on `mapreduce <https://dochub.mongodb.org/core/mapreduce>`_.
 
         """
-        warnings.warn('map_reduce is deprecated, use aggregate instead',
-                      DeprecationWarning, stacklevel=2)
+        warnings.warn(
+            'map_reduce is deprecated, use aggregate instead',
+            DeprecationWarning,
+            stacklevel=2
+        )
         if not isinstance(out, (string_type, abc.Mapping)):
             raise TypeError(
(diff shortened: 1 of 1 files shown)


pymongo/collection.py
```diff
@@ -3091,6 +3093,10 @@ def map_reduce(self, map, reduce, out, full_response=False, session=None, **kwar
            this collection is automatically applied to this operation (if the
            output is not inline) when using MongoDB >= 3.4.
 
+        .. versionchanged:: 3.13
+           Deprecated :meth:`map_reduce`. Support for this function is deprecated in
```
behackett, line 3097: We should also say that this is removed in PyMongo 4.0:
https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#collection-map-reduce-and-collection-inline-map-reduce-are-removed

You should say here what users should do instead. An abbreviated version of what the PyMongo 4 upgrade guide says.

pymongo/collection.py
```diff
@@ -3153,11 +3163,16 @@ def inline_map_reduce(self, map, reduce, full_response=False, session=None, **kw
 
             >>> db.test.inline_map_reduce(map, reduce, limit=2)
 
+        .. versionchanged:: 3.13
+           Deprecated :meth:`inline_map_reduce`. Support for this function is deprecated in
```
behackett, line 3167: Same comment here.

behackett (changes requested):

juliusgeo:
Managed to accidentally merge with master rather than the v3.13 branch so there are a tonnn of messed up changes. I think I'm just gonna create this from scratch in a new PR given how small it is.

blink1073 (approved):
LGTM, thanks!

PR #864, opened 2022-02-11
PYTHON-3078 Remove Use of Unsupported NoReturn Type Class

blink1073:


behackett (commented):
I think we should add RHEL 7.0 to our test matrix using the system python for testing, since it has Python 3.6.0 and will catch problems like this.

blink1073:
> I think we should add RHEL 7.0 to our test matrix using the system python for testing, since it has Python 3.6.0 and will catch problems like this.

Sounds good, will do

juliusgeo (approved):
LGTM!

PR #865, opened 2022-02-11
PYTHON-2786 [v3.13] Drop support for Python 3.4

blink1073:


behackett:
You also need to remove it from the test matrix. See https://github.com/mongodb/mongo-python-driver/blob/v3.13/.evergreen/config.yml and possibly other files in that directory.

blink1073:
> You also need to remove it from the test matrix. See [https://github.com/mongodb/mongo-python-driver/blob/v3.13/.evergreen/config.yml](https://github.com/mongodb/mongo-python-driver/blob/v3.13/.evergreen/config.yml?rgh-link-date=2022-02-11T15%3A19%3A07Z) and possibly other files in that directory.

Done, thanks

.evergreen/build-manylinux.sh
```diff
@@ -39,7 +36,6 @@ ls dist
 # Check for any unexpected files.
 unexpected=$(find dist \! \( -iname dist -or \
                              -iname '*cp27*' -or \
-                             -iname '*cp34*' -or \
```
behackett, line 42: You'll also need to remove 3.4 from build-mac.sh and build-windows.sh. That last one lists the version as "34". You'll probably want to search around for that more broadly.

blink1073: Done

PR #869, opened 2022-02-14
PYTHON-3088 [v4.0] Update load balancer tests to support dedicated load balancer port

juliusgeo:
PYTHON-3088 Update load balancer tests to support dedicated load balancer port (#866)
    
(cherry picked from commit 341d489f38ad51620fab50bfc7c3f8c1227fefee)


blink1073 (approved):
LGTM!

PR #870, opened 2022-02-15
PYTHON-3088 [v3.13] Update load balancer tests to support dedicated load balancer port

juliusgeo:
PYTHON-3088 Update load balancer tests to support dedicated load balancer port (#866)
    
(cherry picked from commit 341d489f38ad51620fab50bfc7c3f8c1227fefee)


blink1073 (approved):
LGTM!

.evergreen/config.yml
```diff
@@ -1770,8 +1772,12 @@ tasks:
       commands:
         - func: "bootstrap mongo-orchestration"
           vars:
+            VERSION: "latest"
```
ShaneHarvey, line 1775: VERSION should not be here. The VERSION gets set in the matrix here:
```
- matrix_name: "load-balancer"
  matrix_spec:
    platform: ubuntu-18.04
    mongodb-version: ["5.0", "latest"]   <---------- VERSION set here
    auth-ssl: "*"
 ...
```

We'll need to fix this in master as well.

pymongo/pool.py
```diff
@@ -584,12 +581,7 @@ def _hello(self, cluster_time, topology_version,
...
 
-        doc = self.command('admin', cmd, publish_events=False,
-                           exhaust_allowed=awaitable)
-        # PYTHON-2712 will remove this topologyVersion fallback logic.
-        if self.opts.load_balanced and _MOCK_SERVICE_ID:
-            process_id = doc.get('topologyVersion', {}).get('processId')
-            doc.setdefault('serviceId', process_id)
+        doc = self.command("admin", cmd, publish_events=False, exhaust_allowed=awaitable)
         if not self.opts.load_balanced:
             doc.pop('serviceId', None)
```
ShaneHarvey, line 586: Should this be removed now too?

ShaneHarvey (changes requested):

.evergreen/config.yml
ShaneHarvey: We also need to change "5.0" to "rapid". We can no longer test load-balancer support against 5.0. The only reason the test pass is because the above bug makes it so that the "5.0" test are actually using "latest".

juliusgeo: Done.

pymongo/pool.py
juliusgeo: Done. And removed another occurrence.

ShaneHarvey (changes requested):
Is this PR supposed to be merging into the v3.12 branch or v3.13? The title says v3.13 but the branch is v3.12

.evergreen/config.yml
```diff
@@ -1771,7 +1773,10 @@ tasks:
         - func: "bootstrap mongo-orchestration"
           vars:
             TOPOLOGY: "sharded_cluster"
+            LOAD_BALANCER: true
         - func: "run load-balancer"
+          vars:
+            LOAD_BALANCER: true
```
ShaneHarvey, line 1806: I don't think vars/LOAD_BALANCER is needed for "run load-balancer".

ShaneHarvey (changes requested):
Oh I see, we still need to add "rapid" as a choice for the "mongodb-version" axes.

.evergreen/config.yml
juliusgeo: Done.

ShaneHarvey: LOAD_BALANCER is not needed for run tests either is it?

ShaneHarvey (changes requested):

.evergreen/config.yml
ShaneHarvey: LOAD_BALANCER is only needed for run-orchestration.sh. Let's remove it here. Note we can also remove the `export LOAD_BALANCER=1` line above.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #874, opened 2022-02-16
PYTHON-3068 Support 'let' option in BulkWriteOptions

juliusgeo:


pymongo/bulk.py
```diff
@@ -315,6 +316,8 @@ def _execute_command(
                     cmd["writeConcern"] = write_concern.document
                 if self.bypass_doc_val:
                     cmd["bypassDocumentValidation"] = True
+                if self.let and cmd_name in ("update", "delete"):
```
blink1073, line 319: Do we need to enforce a minwireversion here?

juliusgeo: That would, I believe, fall under the philosophy of letting the server deliver the error message if possible. For the spec tests for regular 'let' (non-BulkWrite), we rely on server error messages rather than the validation from the driver itself to catch version incompatibility.

.evergreen/resync-specs.sh
```diff
@@ -56,10 +56,14 @@ cpjson () {
...
-    printf "\nIgnored files for ${PWD}"
-    printf "\n%s\n" "$(diff <(find . -name '*.json' | sort) \
+    printf "\nIgnored files for ${PWD}\n"
+    IGNORED_FILES="$(printf "\n%s\n" "$(diff <(find . -name '*.json' | sort) \
     <(find . -name '*.json' | grep -Ev "${BLOCKLIST}" | sort))" | \
-    sed -e '/^[0-9]/d' | sed -e 's|< ./||g'
+    sed -e '/^[0-9]/d' | sed -e 's|< ./||g' )"
+    printf "%s\n" $IGNORED_FILES
+    cd "$PYMONGO"/test/$2
+    printf "%s\n" $IGNORED_FILES | xargs git checkout master
```
juliusgeo, line 65: This is mainly tweaks to the functionality of `resync_specs.sh` that I discovered while implementing this ticket. The main issue is just that when it "ignores" a file, that means it does not copy it over. However, because we delete all files in the directory prior, that means that now the file has been deleted. The solution is to checkout those ignored files after copying to ensure they will not show in the PR rather than showing up as a deletion.

blink1073 (approved):
LGTM!

pymongo/bulk.py
ShaneHarvey: `cmd_name in ("update", "delete"):` -> `run.op_type in (_DELETE, _UPDATE)` 

We already use run.op_type elsewhere for these kind of checks and it doesn't rely on string comparison.

ShaneHarvey (changes requested):
One minor suggestion

pymongo/bulk.py
```diff
@@ -315,6 +316,8 @@ def _execute_command(
                     cmd["writeConcern"] = write_concern.document
                 if self.bypass_doc_val:
                     cmd["bypassDocumentValidation"] = True
+                if self.let and run.op_type in (_DELETE, _UPDATE):
```
ShaneHarvey, line 319: Should be `self.let is not None`?

ShaneHarvey (approved):

pymongo/bulk.py
```diff
@@ -138,13 +139,16 @@ def _raise_bulk_write_error(full_result):
...
-    def __init__(self, collection, ordered, bypass_document_validation, comment=None):
+    def __init__(self, collection, ordered, bypass_document_validation, comment=None, let=None):
         """Initialize a _Bulk instance."""
         self.collection = collection.with_options(
             codec_options=collection.codec_options._replace(
                 unicode_decode_error_handler="replace", document_class=dict
             )
         )
+        self.let = let
+        if self.let:
```
ShaneHarvey, line 150: `if let is not None`

ShaneHarvey (changes requested):

pymongo/bulk.py
juliusgeo: Done.

ShaneHarvey (approved):

PR #876, opened 2022-02-16
PYTHON-2956 Drivers should check out an implicit session only after checking out a connection

juliusgeo:


pymongo/client_session.py
```diff
@@ -947,9 +947,16 @@ def _txn_read_preference(self):
             return self._transaction.opts.read_preference
         return None
 
+    def _start_serv_sesh(self):
+        while isinstance(self._server_session, Empty):
+            self._server_session, old = self._client._get_server_session(), self._server_session
+            self._server_session.generation = old.generation
+            self._server_session.last_use = old.last_use
+            self._server_session._transaction_id = old._transaction_id
```
ShaneHarvey, line 955: You cannot overwrite a session's _transaction_id otherwise the session will have undefined behavior. You'll have to figure our another way to handle the `session._start_retryable_write()` behavior with an implicit session.

pymongo/client_session.py
```diff
@@ -947,9 +947,18 @@ def _txn_read_preference(self):
...
+                self._client._topology.get_server_session(),
+                self._server_session,
+            )
+            for _ in range(old.started):
+                self._server_session.inc_transaction_id()
+
     def _apply_to(self, command, is_retryable, read_preference, sock_info):
         self._check_ended()
-
+        self._start_serv_sesh()
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/client_session.py
```diff
@@ -947,9 +947,18 @@ def _txn_read_preference(self):
             return self._transaction.opts.read_preference
         return None
 
+    def _start_serv_sess(self):
```
blink1073, line 950: Actually, I think `ensure_server_session` is a better name.  It is more explicit, and I don't think we should abbreviate method name.

juliusgeo: Done.

ShaneHarvey: How about `_materialize`? That's what ruby named their method and it conveys the meaning well.

pymongo/client_session.py
```diff
@@ -947,9 +947,18 @@ def _txn_read_preference(self):
...
+        if isinstance(self._server_session, _EmptyServerSession):
+            self._server_session, old = (
+                self._client._topology.get_server_session(),
+                self._server_session,
```
ShaneHarvey, line 954: I think this would be more clear without using tuple assignment:
```
old = self._server_session
self._server_session = self._client._topology.get_server_session()
```

pymongo/client_session.py
```diff
@@ -1000,6 +1009,24 @@ def __copy__(self):
...
+        self.started = 0
+
+    def timed_out(self, session_timeout_minutes):
+        return True
+
+    def mark_dirty(self):
+        self.dirty = True
+
+    def inc_transaction_id(self):
+        self.started += 1
```
ShaneHarvey, line 1027: `old.started` should just be a boolean since inc_transaction_id() will only be called once. All that matters is wether we should forward the inc_transaction_id() call once when the session is materialized.

pymongo/client_session.py
```diff
@@ -947,9 +947,18 @@ def _txn_read_preference(self):
             return self._transaction.opts.read_preference
         return None
 
+    def _start_serv_sess(self):
+        if isinstance(self._server_session, _EmptyServerSession):
+            self._server_session, old = (
+                self._client._topology.get_server_session(),
+                self._server_session,
+            )
+            for _ in range(old.started):
```
ShaneHarvey, line 956: This should be an if-statement instead of a loop (see by other comment below.

pymongo/cursor.py
```diff
@@ -207,6 +207,7 @@ def __init__(
         if session:
             self.__session = session
             self.__explicit_session = True
+            self.__session._start_serv_sess()
```
ShaneHarvey: Shouldn't this call be deferred to when the session is actually used (in apply_to)?

pymongo/mongo_client.py
```diff
@@ -1602,7 +1603,12 @@ def _process_periodic_tasks(self):
 
     def __start_session(self, implicit, **kwargs):
         # Raises ConfigurationError if sessions are not supported.
-        server_session = self._get_server_session()
+        if implicit:
+            with self._topology._lock:
+                self._topology._check_session_support()
```
ShaneHarvey, line 1608: We should never take the topology lock outside of methods in the Topology class. Let's make this a new topology method or maybe just a flag to the existing get_server_session() method. EG:
```
server_session = self._get_server_session(implicit)
```

pymongo/mongo_client.py
```diff
@@ -1602,7 +1603,12 @@ def _process_periodic_tasks(self):
...
+            server_session = _EmptyServerSession(self._topology._session_pool.generation)
```
ShaneHarvey, line 1609: Does the generation matter here? If not let's remove it.

test/test_session.py
```diff
@@ -173,6 +176,36 @@ def _test_ops(self, client, *ops):
...
+            (client.db.test.bulk_write, [[UpdateOne({}, {"$set": {"x": 2}})]]),
+            (client.db.test.find_one_and_delete, [{}]),
+            (client.db.test.find_one_and_update, [{}, {"$set": {"x": 1}}]),
+            (client.db.test.find_one_and_replace, [{}, {}]),
+            (client.db.test.find, [{}]),
+        ]
+        threads = []
+        listener.results.clear()
+        for op, args in ops:
+            threads.append(threading.Thread(target=op, args=typing.cast(typing.Iterable, args)))
```
ShaneHarvey, line 198: What's the `typing.cast(typing.Iterable` here? Should we type hint `ops` instead?

pymongo/client_session.py
```diff
@@ -1000,6 +1009,24 @@ def __copy__(self):
         raise TypeError("A ClientSession cannot be copied, create a new session instead")
 
 
+class _EmptyServerSession(object):
+    def __init__(self, generation):
+        self.generation = generation
+        self.last_use = time.monotonic()
```
ShaneHarvey, line 1015: last_use is not needed is it?

pymongo/client_session.py
```diff
@@ -1000,6 +1009,24 @@ def __copy__(self):
...
+        self.lsid = None
+        self.dirty = False
+        self.started = 0
+
+    def timed_out(self, session_timeout_minutes):
```
ShaneHarvey, line 1020: timed_out is not needed is it?

ShaneHarvey (changes requested):

pymongo/client_session.py
juliusgeo: Done.

pymongo/client_session.py
juliusgeo: Done.

pymongo/client_session.py
juliusgeo: Done.

pymongo/client_session.py
juliusgeo: It is needed because of `session_pool.return_server_session()`.

pymongo/client_session.py
juliusgeo: Done.

pymongo/cursor.py
juliusgeo: Done.

pymongo/mongo_client.py
juliusgeo: I do not believe it does, removed.

pymongo/client_session.py
ShaneHarvey: I think it would be cleaner for the session_pool to have no knowledge of the _EmptyServerSession. Let's avoid calling session_pool.return_server_session or session_pool.return_server_session_no_lock with a _EmptyServerSession.

ShaneHarvey (changes requested):
A few more comments

test/test_session.py
```diff
@@ -173,6 +175,36 @@ def _test_ops(self, client, *ops):
...
+
+        ops: List[Tuple[Callable, List[Any]]] = [
+            (client.db.test.find_one, [{"_id": 1}]),
+            (client.db.test.delete_one, [{}]),
+            (client.db.test.update_one, [{}, {"$set": {"x": 2}}]),
+            (client.db.test.bulk_write, [[UpdateOne({}, {"$set": {"x": 2}})]]),
+            (client.db.test.find_one_and_delete, [{}]),
+            (client.db.test.find_one_and_update, [{}, {"$set": {"x": 1}}]),
+            (client.db.test.find_one_and_replace, [{}, {}]),
+            (client.db.test.find, [{}]),
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/client_session.py
```diff
@@ -1000,6 +1007,19 @@ def __copy__(self):
         raise TypeError("A ClientSession cannot be copied, create a new session instead")
 
 
+class _EmptyServerSession(object):
+    def __init__(self):
+        self.lsid = None
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/client_session.py
```diff
@@ -1000,6 +1007,19 @@ def __copy__(self):
         raise TypeError("A ClientSession cannot be copied, create a new session instead")
 
 
+class _EmptyServerSession(object):
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/client_session.py
```diff
@@ -1000,6 +1007,19 @@ def __copy__(self):
...
+    def __init__(self):
+        self.lsid = None
+        self.dirty = False
+        self.started = False
```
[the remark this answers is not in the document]
juliusgeo: Done.

test/test_session.py
```diff
@@ -173,6 +175,40 @@ def _test_ops(self, client, *ops):
...
+            threads.append(threading.Thread(target=op, args=args))
+            threads[-1].start()
+        for thread in threads:
+            thread.join()
+
+        lsid_set = set()
+        for i in listener.results["started"]:
+            if i.command.get("lsid"):
+                lsid_set.add(i.command.get("lsid")["id"])
+        self.assertEqual(len(lsid_set), 1)
```
ShaneHarvey, line 210: I worry a bit that this test will be flaky. The issue is that the session is checked back into the pool _after_ the connection is checked back in. So there's always the risk that this test uses >1 server session. Let's try a "wait and see" approach where we merge the test as is while keeping in mind that it might fail.

ShaneHarvey (approved):
LGTM once the latest tests pass.

test/test_session.py
juliusgeo: I actually did notice this test failing in CI one time. So I believe you are correct. However, I tried to replicate it locally by running the test 1000s of times and could not reproduce.

juliusgeo: Thanks for opening that drivers ticket. Right now it appears that on replica sets it still fails even with 5 iterations.

ShaneHarvey: How was the test passing on replica sets before?

test/test_session.py
```diff
@@ -173,6 +175,50 @@ def _test_ops(self, client, *ops):
...
+                    thread.join()
+
+                lsid_set = set()
+                for i in listener.results["started"]:
+                    if i.command.get("lsid"):
+                        lsid_set.add(i.command.get("lsid")["id"])
+                self.assertEqual(len(lsid_set), 1)
+                succeeded = True
+                break
+            except:
```
ShaneHarvey, line 218: Let's remove the try/except and change `self.assertEqual` to `succeeded = len(lsid_set) == 1`. `except:`  is probably masking a bug.

test/test_session.py
```diff
@@ -173,6 +175,50 @@ def _test_ops(self, client, *ops):
...
                     )
 
+    def test_implicit_sessions_checkout(self):
+        # "To confirm that implicit sessions only allocate their server session after a
+        # successful connection checkout" test from Driver Sessions Spec.
+        succeeded = False
+        for _ in range(5):
+            try:
+                listener = EventListener()
+                client = rs_or_single_client(
```
ShaneHarvey, line 185: Let's explicitly close these clients.

ShaneHarvey (changes requested):

test/test_session.py
```diff
@@ -173,6 +175,46 @@ def _test_ops(self, client, *ops):
...
+                (client.db.list_collections, []),
+                (cursor.distinct, ["_id"]),
+            ]
+            threads = []
+            listener.results.clear()
+            for op, args in ops:
+                threads.append(threading.Thread(target=op, args=args))
+                threads[-1].start()
+            for thread in threads:
+                thread.join()
```
ShaneHarvey, line 209: There's one more hidden problem with this test. The test does not assert that the threads actually succeeded in running their operation. Change this to use ExceptionCatchingThread and then assert that `exc` is None for each thread.

ShaneHarvey (changes requested):

test/test_session.py
juliusgeo: I changed it to use exception catching thread and it is still failing on Github Actions.

juliusgeo: Nevermind, I see what you mean. Will fix.

ShaneHarvey:
Looks like this test is failing on pypy 3.8:
```
FAIL: test_implicit_sessions_checkout (test_session.TestSession)
  File "/home/runner/work/mongo-python-driver/mongo-python-driver/test/test_session.py", line 220, in test_implicit_sessions_checkout
    self.assertTrue(succeeded)
```

Could be related to delayed garbage collection in pypy? Does it also fail locally for you?

test/test_session.py
```diff
@@ -171,6 +177,54 @@ def _test_ops(self, client, *ops):
...
+            client.close()
+            lsid_set = set()
+            for i in listener.results["started"]:
+                if i.command.get("lsid"):
+                    lsid_set.add(i.command.get("lsid")["id"])
+            if len(lsid_set) == 1:
+                succeeded = True
+            else:
+                failures += 1
+        print(failures)
```
blink1073, line 225: Do we not need to assert that there were no failures?

juliusgeo: Nope, because we expect some failures, we just want at least one success.

blink1073: Gotcha

blink1073 (approved):
LGTM!

ShaneHarvey (changes requested):
Still waiting for an explanation for why changing from $listLocalSessions to `client.admin.aggregate, [[{"$currentOp": {}}]]` fixes the test failure on pypy. Is there a bug in our CommandCursor implementation? Does $listLocalSessions require multiple batches whereas $currentOp is a single batch? A multi-batch cursor would cause this test to almost always create one extra session since the session needs to be kept alive until all batches complete. In this case I believe we should update the test to impose a limit=1 for all find/aggregate cursors.

test/test_session.py
```diff
@@ -171,6 +177,54 @@ def _test_ops(self, client, *ops):
...
+            ops: List[Tuple[Callable, List[Any]]] = [
+                (client.db.test.find_one, [{"_id": 1}]),
+                (client.db.test.delete_one, [{}]),
+                (client.db.test.update_one, [{}, {"$set": {"x": 2}}]),
+                (client.db.test.bulk_write, [[UpdateOne({}, {"$set": {"x": 2}})]]),
+                (client.db.test.find_one_and_delete, [{}]),
+                (client.db.test.find_one_and_update, [{}, {"$set": {"x": 1}}]),
+                (client.db.test.find_one_and_replace, [{}, {}]),
+                (client.db.test.aggregate, [[{"$limit": 1}]]),
+                (client.db.test.find, [{"$limit": 1}]),
```
ShaneHarvey: One problem here is that the thread is only calling find() but is never iterating the cursor. The test needs to fully iterate all the cursors operations in this list.

ShaneHarvey (changes requested):

test/test_session.py
ShaneHarvey: Resolve by you latest commit.

ShaneHarvey (approved):
LGTM

PR #877, opened 2022-02-16
PYTHON-2147 Use verified peer cert chain in OCSP when available

ShaneHarvey:
I opened https://jira.mongodb.org/browse/PYTHON-3132 to update our dependency to pyopenssl>=20.0 in which case we can remove the fallback. I'm preserving the pyopenssl<20.0 behavior to ease backports.


blink1073 (approved):
LGTM!

PR #878, opened 2022-02-17
PYTHON-3136 [DevOps] Resync-specs.sh removes ignored files from working tree 

juliusgeo:


ShaneHarvey:
Can you describe how you use this feature? What's the benefit of this change?

juliusgeo:
> Can you describe how you use this feature? What's the benefit of this change?

Sorry, perhaps this was confusing because it has the DevOps tag. This is a bug fix. The current behavior is that `resync-specs.sh` completely removes ignored files from the working tree, rather than *actually* ignoring them and leaving them as they stood prior to that spec resync.

ShaneHarvey (approved):

PR #879, opened 2022-02-17
PYTHON-3090 Clean up Database Command Typing

blink1073:


pymongo/mongo_client.py
```diff
@@ -1700,7 +1706,7 @@ def _process_response(self, reply, session):
         if session is not None:
             session._process_response(reply)
 
-    def server_info(self, session: Optional[client_session.ClientSession] = None) -> Dict[str, Any]:
+    def server_info(self, session: Optional[client_session.ClientSession] = None) -> _DocumentOut:
```
ShaneHarvey, line 1709: This method always returns a dict so isn't the previous return type more correct?

pymongo/database.py
```diff
@@ -926,7 +926,7 @@ def validate_collection(
         session: Optional["ClientSession"] = None,
         background: Optional[bool] = None,
         comment: Optional[Any] = None,
-    ) -> Dict[str, Any]:
+    ) -> _DocumentOut:
```
ShaneHarvey, line 929: This method always returns a dict so isn't the previous return type more correct?

pymongo/database.py
```diff
@@ -620,7 +620,7 @@ def command(
         session: Optional["ClientSession"] = None,
         comment: Optional[Any] = None,
         **kwargs: Any,
-    ) -> Dict[str, Any]:
+    ) -> _DocumentOut:
```
ShaneHarvey, line 623: Is there a simple way to make this method generic such that it returns the type of codec_options.document_class?

bson/codec_options.py
```diff
@@ -293,7 +289,7 @@ class CodecOptions(_options_base):
 
     def __new__(
         cls: Type["CodecOptions"],
-        document_class: Union[Type[MutableMapping], Type["RawBSONDocument"]] = dict,
+        document_class: Union[Type[MutableMapping], Type[Mapping], Type[Dict]] = dict,
```
ShaneHarvey, line 292: It's not clear to me how this change will benefit users. Can you explain?

ShaneHarvey (changes requested):
Let's add tests for this change (mypy_fails + test_mypy)

pymongo/mongo_client.py
blink1073: Is it not the result of a command call, which is decoded with codec options?

pymongo/database.py
blink1073: Same comment as above, it is the result of a `command()` call

pymongo/database.py
blink1073: We had to pick one source of truth, which was the document type of the client given to the constructor.  Technically the `codec_options` given to the constructor any of the methods can override it, but it was too difficult to express as a type.  We had agreed that the most common case is to specify it once in the client, and the user can cast if they are using explicit `codec_options` in other places.

bson/codec_options.py
blink1073: It was to match the types used in `_DocumentOut` for symmetry.

blink1073:
> Let's add tests for this change (mypy_fails + test_mypy) if possible.

Working on this now

pymongo/mongo_client.py
ShaneHarvey: It's the result of a command() call _using the default codec_options_ whose document_class is always dict hence this method always returns dict.

pymongo/database.py
ShaneHarvey: db.command() doesn't use the client's or the database's document_class. It only uses the document_class from the codec_options argument.

bson/codec_options.py
ShaneHarvey: Okay then I have the same question for _DocumentOut.

pymongo/mongo_client.py
blink1073: Okay, I see the distinction now

pymongo/database.py
blink1073: Understood now, thanks

blink1073: Drat, this is a problem: https://bugs.python.org/issue43923.  We can't have a generic namedtuple.

blink1073: We could explicitly add the properties or add the appropriate `__getattr__`.  The downside would be that `isinstance` would fail if anyone were relying on that.

ShaneHarvey: In that case I'm okay with having db.command (somewhat incorrectly) return _DocumentOut. We can open a new issue to support generics and have it depend on that cpython bug.

pymongo/mongo_client.py
blink1073: Used a cast to force this to be a dict, since we're limited by https://jira.mongodb.org/browse/PYTHON-3143

pymongo/database.py
blink1073: Used a cast here as well

pymongo/database.py
blink1073: Done and done: https://jira.mongodb.org/browse/PYTHON-3143

bson/codec_options.py
blink1073: I simplified all of the document types to be either `MutableMapping[str, Any]` or `RawBSONDocument`.

bson/codec_options.py
```diff
@@ -37,7 +37,6 @@
     UuidRepresentation,
 )
 
-# Import RawBSONDocument for type-checking only to avoid circular dependency.
```
juliusgeo, line 40: This seems like a helpful comment. Why remove it and now the code below?

blink1073: Ah, I had removed the import itself at one point, I'll put the comment back

pymongo/database.py
```diff
@@ -783,7 +784,7 @@ def list_collections(
         filter: Optional[Mapping[str, Any]] = None,
         comment: Optional[Any] = None,
         **kwargs: Any,
-    ) -> CommandCursor[Dict[str, Any]]:
+    ) -> CommandCursor[MutableMapping[str, Any]]:
```
ShaneHarvey, line 787: Why change from Dict to MutableMapping?

pymongo/mongo_client.py
```diff
@@ -1710,16 +1715,19 @@ def server_info(self, session: Optional[client_session.ClientSession] = None) ->
...
+            ),
         )
 
     def list_databases(
         self,
         session: Optional[client_session.ClientSession] = None,
         comment: Optional[Any] = None,
         **kwargs: Any,
-    ) -> CommandCursor[Dict[str, Any]]:
+    ) -> CommandCursor[MutableMapping[str, Any]]:
```
ShaneHarvey, line 1730: Same question here.

ShaneHarvey (commented):
Are you still planning to add tests?

pymongo/database.py
blink1073: I removed `Dict` from `_DocumentType` to align with `_DocumentOut`, which meant this method and `list_databases` had to be changed as well.

blink1073:
> Are you still planning to add tests?

Yep

pymongo/database.py
ShaneHarvey: But dict is a MutableMapping so why can't we return it? Is there a mypy error preventing it?

blink1073: `Dict` is more specific than `MutableMapping`.  We can either put `Dict` back in `_DocumentType` or use `MutableMapping` here.

ShaneHarvey:
```
pymongo/database.py: note: In member "list_collections" of class "Database":
pymongo/database.py:787: error: Value of type variable "_DocumentType" of "CommandCursor" cannot be "dict"  [type-var]
        ) -> CommandCursor[Dict[str, Any]]:
             ^
Found 1 error in 1 file (checked 55 source files)
```

I think this is a bug. It has to be possible to use Dict/dict as a generic for cursor/collection/db/client otherwise the generic feature is not useful:
```
$ cat generic.py 
from typing import Dict, Any

client: MongoClient[Dict[str, Any]] = MongoClient()
doc = client.test.test.find_one()
if doc is not None:
    doc['a'] = 1
$ python generic.py 
$ mypy generic.py 
generic.py:4: error: Value of type variable "_DocumentType" of "MongoClient" cannot be "dict"  [type-var]
    client: MongoClient[Dict[str, Any]] = MongoClient()
            ^
generic.py:4: error: Incompatible types in assignment (expression has type "MongoClient[MutableMapping[str, Any]]", variable has type
"MongoClient[Dict[str, Any]]")  [assignment]
    client: MongoClient[Dict[str, Any]] = MongoClient()
                                          ^
Found 2 errors in 1 file (checked 1 source file)
```

blink1073: Yeah, I think this is related to https://github.com/python/mypy/issues/3737.  I say we do not try and make it generic on the input value

pymongo/mongo_client.py
blink1073: Switched to `Dict`

ShaneHarvey (changes requested):

pymongo/encryption.py
```diff
@@ -290,7 +290,7 @@ def _get_internal_client(encrypter, mongo_client):
         db, coll = opts._key_vault_namespace.split(".", 1)
         key_vault_coll = key_vault_client[db][coll]
 
-        mongocryptd_client = MongoClient(
+        mongocryptd_client: MongoClient = MongoClient(
```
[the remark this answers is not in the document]
blink1073: No, it was required

pymongo/collection.py
```diff
@@ -2127,7 +2128,7 @@ def list_indexes(
         self,
         session: Optional["ClientSession"] = None,
         comment: Optional[Any] = None,
-    ) -> CommandCursor[MutableMapping[str, Any]]:
+    ) -> CommandCursor[Dict[str, Any]]:
```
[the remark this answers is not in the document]
blink1073: Done and done: https://jira.mongodb.org/browse/PYTHON-3148

pymongo/mongo_client.py
```diff
@@ -119,7 +119,7 @@ def __init__(
         self,
         host: Optional[Union[str, Sequence[str]]] = None,
         port: Optional[int] = None,
-        document_class: Type[_DocumentType] = dict,
+        document_class: Type[Mapping[str, Any]] = dict,
```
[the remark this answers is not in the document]
blink1073: Done

pymongo/typings.py
```diff
@@ -36,6 +35,5 @@
 _CollationIn = Union[Mapping[str, Any], "Collation"]
 _DocumentIn = Union[MutableMapping[str, Any], "RawBSONDocument"]
 _Pipeline = Sequence[Mapping[str, Any]]
-_DocumentType = TypeVar(
-    "_DocumentType", Mapping[str, Any], MutableMapping[str, Any], Dict[str, Any]
-)
+_DocumentOut = _DocumentIn
+_DocumentType = TypeVar("_DocumentType", bound=Mapping[str, Any])
```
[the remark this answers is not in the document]
blink1073: I cleaned this up to be `_DocumentType = TypeVar("_DocumentType", MutableMapping[str, Any], "RawBSONDocument")` now that we're not confined by the default value limitation.

test/test_mypy.py
```diff
@@ -87,6 +90,38 @@ def test_bulk_write(self) -> None:
         result = self.coll.bulk_write(requests)
         self.assertTrue(result.acknowledged)
 
+    def test_command(self) -> None:
+        result = self.client.admin.command("ping")
+        items = result.items()
```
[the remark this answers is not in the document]
blink1073: Easy fix, done

blink1073:
Wow, this is quicksand indeed.   There is no way to declare compatibility with `TypedDict` in general:

```python
bson/codec_options.py:119: error: Variable "typing.TypedDict" is not valid as a type  [valid-type]
    _DocumentClass = Union[Type[MutableMapping], Type[TypedDict], Type["RawBSONDocument"]]
                                                      ^
bson/codec_options.py:119: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

If we don't include `TypedDict`, we get:

```python
test/test_mypy.py: note: In member "test_default_typedict" of class "TestPymongo":
test/test_mypy.py:140: error: Invalid type argument value for "MongoClient"  [type-var]
            client: MongoClient[Movie] = MongoClient(document_class=Movie)
                    ^
test/test_mypy.py:140: error: Incompatible types in assignment (expression has type
"MongoClient[Dict[str, Any]]", variable has type "MongoClient[Movie]")  [assignment]
            client: MongoClient[Movie] = MongoClient(document_class=Movie)
                                         ^
test/test_mypy.py:140: error: Argument "document_class" to "MongoClient" has incompatible type
"Type[Movie]"; expected "Union[Type[MutableMapping[Any, Any]], Type[RawBSONDocument]]"  [arg-type]
            client: MongoClient[Movie] = MongoClient(document_class=M
```

The only way for us to gain compatibility with `TypedDict` is for us to use `Any` for both `_DocumentClass` and `_DocumentType` and rely on explicit type declaration.

pymongo/encryption.py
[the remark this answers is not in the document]
ShaneHarvey: I don't understand. Why is it needed?

pymongo/typings.py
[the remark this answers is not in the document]
ShaneHarvey: Wait I thought `TypeVar("_DocumentType", bound=Mapping[str, Any])` was the better option? In my testing it supports TypedDict and everything else. The only quirk is that user will need to add type annotations for their MongoClients:
```
generic.py:7: error: Need type annotation for "client"  [var-annotated]
        client = MongoClient()
                 ^
```
Thats not so bad since list/dict/set/etc all have the same requirement: https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#explicit-types-for-collections

And in the future when mypy gets their act together we can add `dict` as the default. 

One other quirk is that it leads to this internal pymongo mypy error:
```
mypy pymongo                                     
pymongo/mongo_client.py: note: In member "__init__" of class "MongoClient":
pymongo/mongo_client.py:127: error: Incompatible default for argument "document_class" (default has type "Type[Dict[Any, Any]]", argument has type
"Type[_DocumentType]")  [assignment]
            document_class: Type[_DocumentType] = dict,
                                                  ^
```

We could fix this by changing it to `document_class: Optional[Type[_DocumentType]] = None` instead of dict.

pymongo/encryption.py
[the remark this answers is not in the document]
blink1073: As for why, I can't say why that would be different that all of the other places we use the class.

pymongo/typings.py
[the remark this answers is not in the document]
blink1073: Updated

pymongo/encryption.py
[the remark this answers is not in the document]
blink1073: No longer needed with recent updates to generics handling, removed.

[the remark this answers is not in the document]
blink1073: Had to add it back in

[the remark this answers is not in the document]
ShaneHarvey: Ack.

pymongo/typings.py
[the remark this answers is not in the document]
ShaneHarvey: Ack.

test/test_mypy.py
```diff
@@ -87,6 +104,66 @@ def test_bulk_write(self) -> None:
...
+    def test_default_document_type(self) -> None:
+        client = rs_or_single_client()
+        coll = client.test.test
+        doc = {"my": "doc"}
+        coll.insert_one(doc)
+        retreived = coll.find_one({"_id": doc["_id"]})
+        assert retreived is not None
+        retreived["a"] = 1
+
+    def test_explicit_document_type(self) -> None:
```
ShaneHarvey, line 130: This test looks the same as test_default_document_type

blink1073: Updated

ShaneHarvey (approved):

PR #880, opened 2022-02-17
PYTHON-3088 [4.0] Test rapid releases with load balancers

juliusgeo:
PYTHON-3088 [v3.13] Update load balancer tests to support dedicated load balancer port (#870)
(cherry picked from commit 1443d7687edb9f6f62a182c0d3b5e9386b4a568c)

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -2639,7 +2639,7 @@ buildvariants:
 
 - matrix_name: "load-balancer"
   matrix_spec:
-    platform: awslinux
+    platform: ubuntu-18.04
     mongodb-version: ["rapid", "latest"]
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -1718,8 +1718,6 @@ tasks:
             TOPOLOGY: "sharded_cluster"
```
ShaneHarvey, line 1718: Still need to remove the "VERSION" above.

ShaneHarvey (changes requested):
Still need to remove the `doc.pop('serviceId...` logic from the pool.

ShaneHarvey (approved):

PR #882, opened 2022-02-19
PYTHON-2504 Apply pyupgrade --py36-plus v2.31.0

ShaneHarvey:
Applied `pyupgrade --py36-plus` to all files except setup.py with v2.31.0. I omitted setup.py so that the file can still produce a good error message on python2, otherwise the user gets this error due to format strings:
```
$ python2 setup.py                                                      
  File "setup.py", line 8, in <module>
    raise RuntimeError("Python version >= 3.6 required.")
RuntimeError: Python version >= 3.6 required.
$ pyupgrade --py36-plus setup.py                                        
Rewriting setup.py
$ python2 setup.py              
  File "setup.py", line 168
    raise RuntimeError(f"documentation step '{mode}' failed")
                                                           ^
SyntaxError: invalid syntax
```

I tested out adding this to pre-commit but pyupgrade's formatting conflicts with black so it's more of a nuisance.


behackett:
I don't have a strong opinion on this being a blocker, but won't this make backports to 3.x more difficult?

ShaneHarvey:
Great point. This will definitely make backports more tedious. Let's hold off until pymongo 3 is EOL.

PR #883, opened 2022-02-19
Add InvalidId as valid error in bson exports

nadavge:
It is "impossible" (or unconventional) to catch exceptions for an invalid ID, even though there's no good reason not to IMHO.

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -98,7 +98,7 @@ from bson.code import Code
 from bson.codec_options import DEFAULT_CODEC_OPTIONS, CodecOptions, _raw_document_class
 from bson.dbref import DBRef
 from bson.decimal128 import Decimal128
-from bson.errors import InvalidBSON, InvalidDocument, InvalidStringData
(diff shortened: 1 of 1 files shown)


juliusgeo:
@nadavge Hi, I appreciate you opening this PR. That error is already importable--the following code for me runs fine on PyMongo 4.1 with Python 3.8.12 (in a new venv). 
```
from bson.errors import InvalidId
try:
        pass
except InvalidId:
        # Handle exception
        pass
```
The only thing that your change allows is importing from the package namespace like so:
```
From bson import InvalidId
```
All of the other errors that are imported in `bson/__init__.py` are also used in that same file--hence the import. I believe that users are intended to import from `bson.errors` based on the [PyMongo BSON documentation for errors](https://pymongo.readthedocs.io/en/stable/api/bson/errors.html). 
Could I ask for the specific reason why you wish it to be in the packages namespace?

agolin95:
We're closing this out since we haven't heard anything further in this thread, but please let us know if you'd like to continue the conversation!

PR #886, opened 2022-02-23
PYTHON-3142 Stop using $where in test_maxConnecting

ShaneHarvey:
Removing $where fixes the test failures.

### DIFF
--- a/test/test_pooling.py
+++ b/test/test_pooling.py
@@ -378,12 +378,14 @@ class TestPooling(_TestPoolingBase):
     def test_maxConnecting(self):
         client = rs_or_single_client()
         self.addCleanup(client.close)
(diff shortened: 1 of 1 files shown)


juliusgeo:
Is there a way to add a unittest to ensure that this test does not fail within `n` iterations? I think that would be a good addition to this PR.

ShaneHarvey:
> ensure that this test does not fail within n iterations

I don't want to do that because it would make this test slower and more flakey.

We could "ensure the test passes at least once in n iterations" but I also don't want to do that here since that could hide legitimate server or driver bugs (such a change would have masked this server regression).

blink1073 (approved):
LGTM

juliusgeo (approved):
LGTM!

PR #890, opened 2022-02-28
PYTHON-3141 Add slotscheck to pre-commit checks

blink1073:


.pre-commit-config.yaml
```diff
@@ -34,6 +34,7 @@ repos:
   rev: 3.9.2
   hooks:
   - id: flake8
+    files: \.py$
```
ShaneHarvey, line 37: What does this do?

ShaneHarvey (approved):
Let's update the PR title to "Add slotscheck" now that we're done considering :)

ShaneHarvey:
The CI job looks like it's _only_ running slotscheck now:
```
/opt/hostedtoolcache/Python/3.10.2/x64/bin/pre-commit run --show-diff-on-failure --color=always --all-files --hook-stage=manual
[INFO] Initializing environment for https://github.com/ariebovenberg/slotscheck.
[INFO] Installing environment for https://github.com/ariebovenberg/slotscheck.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
slotscheck...............................................................Passed
```

Here's the base commit:
```
Cache Size: ~34 MB (36171498 B)
/usr/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/68aea[56](https://github.com/mongodb/mongo-python-driver/runs/5337839218?check_suite_focus=true#step:4:56)d-7da8-46ea-8aae-ff15f24e4ad8/cache.tzst -P -C /home/runner/work/mongo-python-driver/mongo-python-driver
/opt/hostedtoolcache/Python/3.10.2/x[64](https://github.com/mongodb/mongo-python-driver/runs/5337839218?check_suite_focus=true#step:4:64)/bin/pre-commit run --show-diff-on-failure --color=always --all-files --hook-stage=manual
check for added large files..............................................Passed
check for case conflicts.................................................Passed
check toml...........................................(no files to check)Skipped
check yaml...............................................................Passed
debug statements (python)................................................Passed
fix end of files.........................................................Passed
forbid new submodules................................(no files to check)Skipped
trim trailing whitespace.................................................Passed
black....................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
shellcheck...............................................................Passed
Check GitHub Workflows...................................................Passed
```

Is this expected?

.pre-commit-config.yaml
blink1073: It limits it to run on only python files.  We have the same on black and isort.

blink1073:
> Is this expected?

No, I had an extra `repos:` heading that was excluding the previous checks, fixed

ShaneHarvey:
> No, I had an extra repos: heading that was excluding the previous checks, fixed

Funny, that's one of the things the yaml check would have complained about if it ran :)

ShaneHarvey (approved):

PR #893, opened 2022-03-04
PYTHON-3085 Audit consistent and correct types for _DocumentOut

blink1073:
- [x] Make `CodecOptions` and `decode` generic
- [x] Add tests for `decode` that include dict, `RAWBSONDocument`, and custom dict types
- [x] Make `database.command` generic
- [x] Add tests for `database.command`


ShaneHarvey (commented):
Before adding tests for the namedtuple api, what do you think about only making this "inherit from tuple" change when TYPE_CHECKING? Eg something like:
```
if TYPE_CHECKING:
    class _CodecOptionsBase(Tuple, Generic[_DocumentType]):
        ....
else:
    _CodecOptionsBase = namedtuple("CodecOptions" ...)
```

blink1073:
> Before adding tests for the namedtuple api, what do you think about only making this "inherit from tuple" change when TYPE_CHECKING? 

Done

test/test_binary.py
```diff
@@ -158,53 +158,53 @@ def test_uuid_subtype_4(self):
     def test_legacy_java_uuid(self):
         # Test decoding
         data = self.java_data
-        docs = bson.decode_all(data, CodecOptions(SON, False, PYTHON_LEGACY))
+        docs = bson.decode_all(data, CodecOptions(SON, False, PYTHON_LEGACY))  # type: ignore[type-var]
```
ShaneHarvey, line 161: what about:
```
SON = SON[str, Any]
...
```
or 
```
son = SON[str, Any]
...CodecOptions(son,...
...
```
Would that remove the need for `# type:` comments?

test/test_mypy.py
```diff
@@ -127,65 +138,212 @@ def test_default_document_type(self) -> None:
...
+        assert next(rt_documents2).foo() == "bar"
+
+        codec_options3 = CodecOptions(RawBSONDocument)
+        bsonbytes3 = encode(doc, codec_options=codec_options3)
+        bsonbytes3 += encode(doc, codec_options=codec_options3)
+        rt_documents3 = decode_iter(bsonbytes3, codec_options3)
+        assert next(rt_documents3).raw
+
+    def make_tempfile(self, content: bytes) -> Any:
+        fileobj = tempfile.TemporaryFile()
```
ShaneHarvey, line 236: I suggest adding `self.addCleanup(fileobj.close)` to close the file.

ShaneHarvey (changes requested):
Cool I think this is the most promising approach so far! A few minor comments.

bson/codec_options.py
```diff
@@ -202,20 +198,16 @@ def __eq__(self, other: Any) -> Any:
...
-        "type_registry",
-    ),
-)
+class _BaseCodecOptions(NamedTuple):
+    document_class: Type[Mapping[str, Any]]
+    tz_aware: bool
+    uuid_representation: int
+    unicode_decode_error_handler: Optional[str]
+    tzinfo: Optional[datetime.tzinfo]
+    type_registry: TypeRegistry
```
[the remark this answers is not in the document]
ShaneHarvey: Although I'm a little wary of this change in 3.6.1: https://docs.python.org/3/library/typing.html#typing.NamedTuple

> Changed in version 3.6.1: Added support for default values, methods, and docstrings.

Does this new typing.NamedTuple approach work on python 3.6.0?

bson/__init__.py
```diff
@@ -967,15 +970,16 @@ def decode(
...
         raise _CODEC_OPTIONS_TYPE_ERROR
 
     return _bson_to_dict(data, codec_options)
 
 
 def decode_all(
-    data: _ReadableBuffer, codec_options: CodecOptions = DEFAULT_CODEC_OPTIONS
-) -> List[Dict[str, Any]]:
+    data: _ReadableBuffer, codec_options: "Optional[CodecOptions[_DocumentType]]" = None
+) -> List[_DocumentType]:
```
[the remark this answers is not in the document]
blink1073: Fixed, and added test

bson/__init__.py
```diff
@@ -967,15 +970,16 @@ def decode(
 
     .. versionadded:: 3.9
     """
+    codec_options = codec_options or DEFAULT_CODEC_OPTIONS  # type: ignore[assignment]
```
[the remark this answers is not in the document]
blink1073: Done here and in `mongo_client` and `database.command`

bson/codec_options.py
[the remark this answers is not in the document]
blink1073: Spun up a RHEL 7.0 FIPS host to test tomorrow

pymongo/encryption.py
```diff
@@ -572,7 +572,7 @@ def encrypt(
             encrypted_doc = self._encryption.encrypt(
                 doc, algorithm, key_id=key_id, key_alt_name=key_alt_name
             )
-            return decode(encrypted_doc)["v"]
+            return decode(encrypted_doc)["v"]  # type: ignore[index]
```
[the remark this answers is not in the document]
blink1073: The revealed type of the returned value is `Any`, and the error is `Value of type <nothing> is not indexable`.  Once again we're getting hit by the lack of default generic handling.

test/test_binary.py
blink1073: I had thought it wouldn't work on Python 3.6, but it looks like it does, updating.

test/test_mypy.py
blink1073: Done

bson/codec_options.py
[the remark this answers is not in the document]
blink1073: Wow, I can't even get `mypy` to run on a file that has `print('hello')`:

```python
  File "/root/.local/bin/mypy", line 5, in <module>
    from mypy.__main__ import console_entry
  File "/root/.local/lib/python3.6/site-packages/mypy/__main__.py", line 6, in <module>
    from mypy.main import main, process_options
  File "mypy/main.py", line 13, in <module>
  File "mypy/build.py", line 32, in <module>
  File "mypy/semanal.py", line 85, in <module>
  File "mypy/errors.py", line 14, in <module>
  File "mypy/message_registry.py", line 29, in <module>
TypeError: __new__() missing 1 required positional argument: 'code'
```

I think we should stipulate that we support typing on Python 3.7+ only.

test/test_binary.py
blink1073: Okay, the actual issue was that you can't use `issubclass` on a generic alias in Python 3.10 and PyPy, I added a workaround.

bson/codec_options.py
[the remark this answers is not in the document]
ShaneHarvey: Oof. Can you report that error as a bug to mypy?

So mypy doesn't work, but does the driver function correctly? And does it work with my suggestion above? If not we should add a comment that _BaseCodecOptions is required for 3.6.0 support.

bson/codec_options.py
```diff
@@ -294,26 +286,34 @@ class CodecOptions(_options_base):
...
+        document_class: Optional[Type[Mapping[str, Any]]] = None,
         tz_aware: bool = False,
         uuid_representation: Optional[int] = UuidRepresentation.UNSPECIFIED,
         unicode_decode_error_handler: Optional[str] = "strict",
         tzinfo: Optional[datetime.tzinfo] = None,
         type_registry: Optional[TypeRegistry] = None,
     ) -> "CodecOptions":
-        if not (issubclass(document_class, _MutableMapping) or _raw_document_class(document_class)):
+        doc_class = document_class or dict
+        # issubclass can raise TypeError for generic aliases like SON[str, Any].
```
ShaneHarvey, line 297: Wow. Can we create a bugs.python.org feature request for this?

bson/codec_options.py
```diff
@@ -294,26 +286,34 @@ class CodecOptions(_options_base):
...
+                "subclass of collections.abc.MutableMapping"
             )
         if not isinstance(tz_aware, bool):
             raise TypeError("tz_aware must be True or False")
         if uuid_representation not in ALL_UUID_REPRESENTATIONS:
             raise ValueError(
                 "uuid_representation must be a value from bson.binary.UuidRepresentation"
             )
-        if not isinstance(unicode_decode_error_handler, (str, None)):  # type: ignore
+        if not isinstance(unicode_decode_error_handler, (str, None)):  # type: ignore[arg-type]
```
ShaneHarvey, line 316: I think this is hiding a real bug:
```
>>> isinstance(None, (str, None))
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

Let's remove None here and make unicode_decode_error_handler non-Optional since None was never supported.

pymongo/encryption.py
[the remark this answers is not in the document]
ShaneHarvey: Woah why is the return type `<nothing>`? Shouldn't it be Any? Subscripting Any shouldn't generate an error.

test/test_bson.py
```diff
@@ -993,6 +993,18 @@ def test_decode_all_defaults(self):
...
+        # Test decode_all()'s default document_class is dict and tz_aware is
+        # False.
+        doc = {"sub_document": {}, "dt": datetime.datetime.utcnow()}
+
+        decoded = bson.decode_all(bson.encode(doc), None)[0]
+        self.assertIsInstance(decoded["sub_document"], dict)
+        self.assertIsNone(decoded["dt"].tzinfo)
+        # The default uuid_representation is UNSPECIFIED
+        with self.assertRaisesRegex(ValueError, "cannot encode native uuid"):
+            bson.decode_all(bson.encode({"uuid": uuid.uuid4()}), None)
```
ShaneHarvey, line 1006: This test never reaches the `bson.decode_all` because the error is raised by `bson.encode`. Use `Binary.from_uuid(uuid.uuid4())`.

ShaneHarvey (changes requested):

bson/codec_options.py
blink1073: Notes inline here: the issue comes from the fact that [`PyType_Check`](https://docs.python.org/3/c-api/type.html#c.PyType_Check) is failing [here](https://github.com/python/cpython/blob/54ab9ad312ea53db40e31712454272e1d4c0315f/Modules/_abc.c#L672) despite the fact that `isinstance(SON[str, Any], type)` is `True`.  This is the check used in the Python [fallback](https://github.com/python/cpython/blob/54ab9ad312ea53db40e31712454272e1d4c0315f/Lib/_py_abc.py#L110).   The type of `SON[str, Any]` in Python 3.10 is `types.GenericAlias`, and in Python 3.8 it is `typing._GenericAlias`.  `issubclass(a, MutableMapping)` also fails in Python 3.8.

Minimal repro:

```python
from typing import Generic, Any, TypeVar
from collections.abc import MutableMapping

T = TypeVar("T")

class Test(MutableMapping, Generic[T]):
   pass

type(Test)  #  abc.ABCMeta
issubclass(Test, MutableMapping)   # True
type(Test[int]))  # types.GenericAlias
isinstance(Test[int], type)  # True
issubclass(Test[int], MutableMapping)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 issubclass(Test[int], MutableMapping)

File ~/miniconda/envs/mongo-python-driver_env/lib/python3.10/abc.py:123, in ABCMeta.__subclasscheck__(cls, subclass)
    121 def __subclasscheck__(cls, subclass):
    122     """Override for issubclass(subclass, cls)."""
--> 123     return _abc_subclasscheck(cls, subclass)

TypeError: issubclass() arg 1 must be a class
```

bson/codec_options.py
blink1073: Done

test/test_bson.py
blink1073: Fixed

pymongo/encryption.py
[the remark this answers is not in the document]
blink1073: It looks like `<nothing>` is what is says when you don't define a value for a [generic](https://github.com/python/mypy/blob/4037b6a082556dab352a182c7671939073f25b26/test-data/unit/check-inference-context.test#L19).  I think the fact that it reveals as `Any` is the bug here.  There probably should be an `Unknown` type shown for both (that's what `TypeScript` calls it).

bson/codec_options.py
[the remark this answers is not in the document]
blink1073: It runs fine on Python 3.6.15, so I don't think there's anything to report, since Python 3.6 is EOL anyway.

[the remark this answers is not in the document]
blink1073: It looks like with Python 3.6.0 we can't use `CodecOptions(SON[str, Any])`.  It passes in CI but not in FIPS.  

```python
[2022/03/11 18:54:31.415] ----------------------------------------------------------------------
[2022/03/11 18:54:31.415] Traceback (most recent call last):
[2022/03/11 18:54:31.415]   File "/data/mci/5eb993a02c41b93b14bde7085d451fe9/src/test/__init__.py", line 561, in wrap
[2022/03/11 18:54:31.415]     return f(*args, **kwargs)
[2022/03/11 18:54:31.415]   File "/data/mci/5eb993a02c41b93b14bde7085d451fe9/src/test/test_binary.py", line 261, in test_legacy_csharp_uuid_roundtrip
[2022/03/11 18:54:31.415]     docs = bson.decode_all(data, CodecOptions(SON[str, Any], False, CSHARP_LEGACY))
[2022/03/11 18:54:31.415]   File "/data/mci/5eb993a02c41b93b14bde7085d451fe9/src/bson/son.py", line 69, in __new__
[2022/03/11 18:54:31.415]     instance = super(SON, cls).__new__(cls, *args, **kwargs)
[2022/03/11 18:54:31.415] TypeError: super(type, obj): obj must be an instance or subtype of type
[2022/03/11 18:54:31.415] ======================================================================
[2022/03/11 18:54:31.415] ERROR: test_legacy_java_uuid (test_binary.TestBinary)
[2022/03/11 18:54:31.415] ----------------------------------------------------------------------
```

We should be able work around this in the tests by having a version guard.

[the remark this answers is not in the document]
blink1073: Alternately, we could set our minimum supported Python to 3.6.2  (enforceable in `python_requires`) and bring back support for `NoReturn` and several other things that were added to typing in 3.6.1 and 3.6.2.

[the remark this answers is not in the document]
ShaneHarvey: > It runs fine on Python 3.6.15, so I don't think there's anything to report, since Python 3.6 is EOL anyway.

Mypy claims to support >=3.6.0 so this would be a real bug to report.

bson/codec_options.py
blink1073: Based on my reading of https://bugs.python.org/issue45665, it was never intended for `GenericAlias` to be considered a `type`.

ShaneHarvey (changes requested):
Related to the issubclass workaround, I found this in the original PEP:

> If a library is faced with a parameterized generic and would like to perform an isinstance() check using the base type, that type can be retrieved using the __origin__ attribute on the parameterized generic.

https://peps.python.org/pep-0585/#making-isinstance-obj-list-str-perform-a-check-ignoring-generics


Let's use this approach instead of mro().
```python
>>> s = SON[str, Any]
>>> s.__origin__
<class 'bson.son.SON'>
```

bson/codec_options.py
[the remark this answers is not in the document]
blink1073: > Mypy claims to support >=3.6.0 so this would be a real bug to report.

It feels like a waste of our time to find a repro and report this in my opinion.

bson/codec_options.py
```diff
@@ -294,27 +286,35 @@ class CodecOptions(_options_base):
...
     ) -> "CodecOptions":
-        if not (issubclass(document_class, _MutableMapping) or _raw_document_class(document_class)):
+        doc_class = document_class or dict
+        # issubclass can raise TypeError for generic aliases like SON[str, Any].
+        # In that case we can use the base class for the comparison.
+        is_mapping = False
+        try:
+            is_mapping = issubclass(doc_class, _MutableMapping)
+        except TypeError:
+            is_mapping = issubclass(doc_class.__origin__, _MutableMapping)
```
ShaneHarvey, line 303: This should  protect against the case where doc_class.__origin__ does not exist:
```
getattr(doc_class, "__origin__", type)
```

blink1073: Already fixed.  ;)

blink1073:
I bumped the minimum version and scheduled a FIPS build.  If that build passes we can consider [PYTHON-3168](https://jira.mongodb.org/browse/PYTHON-3168) to be fixed.

blink1073:
FIPS passed, but we're now getting this error, I'll look into it tomorrow.

```
[2022/03/14 22:21:39.972] FAILURE: OperationFailure not raised (AssertionError)
[2022/03/14 22:21:39.972] Traceback (most recent call last):
[2022/03/14 22:21:39.972]   File "/data/mci/4cea1b66713276ffb9d51b01c40e8ad5/src/test/__init__.py", line 561, in wrap
[2022/03/14 22:21:39.972]     return f(*args, **kwargs)
[2022/03/14 22:21:39.972]   File "/data/mci/4cea1b66713276ffb9d51b01c40e8ad5/src/test/test_examples.py", line 1285, in test_versioned_api_migration
[2022/03/14 22:21:39.972]     client.db.command("count", "sales", query={})
[2022/03/14 22:21:39.972] AssertionError: OperationFailure not raised
```

doc/python3.rst
```diff
@@ -20,7 +20,7 @@ with subtype 0.
 For example, let's insert a :class:`bytes` instance using Python 3 then
 read it back. Notice the byte string is decoded back to :class:`bytes`::
 
-  Python 3.6.1 (v3.6.1:69c0db5050, Mar 21 2017, 01:21:04)
+  Python 3.6.2 (v3.6.1:69c0db5050, Mar 21 2017, 01:21:04)
```
ShaneHarvey, line 23: Here and below, 3.6.1 is still present. Let's use some real output:
```
Python 3.6.8 (v3.6.8:3c6b436a57, Dec 24 2018, 02:04:31)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
```

ShaneHarvey (changes requested):
Let's also update the changelog to warn that we've dropped 3.6.0 and 3.6.1.

doc/python3.rst
blink1073: Done

blink1073:
> Let's also update the changelog to warn that we've dropped 3.6.0 and 3.6.1.

Done

ShaneHarvey (approved):

PR #894, opened 2022-03-05
PYTHON-3159 Fix typo in zlib compression support

ShaneHarvey:


pymongo/compression_support.py
```diff
@@ -123,16 +122,10 @@ class ZlibContext(object):
...
-        self.compress: Callable[[bytes], bytes]
-
-        # Jython zlib.compress doesn't support -1
-        if level == -1:
-            self.compress = zlib.compress
-        # Jython zlib.compress also doesn't support 0
-        elif level == 0:
-            self.compress = _zlib_no_compress
-        else:
-            self.compresss = lambda data, _: zlib.compress(data, level)
```
ShaneHarvey, line 135: `compresss` -> `compress` was the bug fix. I also removed the Jython workarounds.

pymongo/compression_support.py
```diff
@@ -123,16 +122,10 @@ class ZlibContext(object):
     compressor_id = 2
 
     def __init__(self, level):
-        self.compress: Callable[[bytes], bytes]
```
juliusgeo, line 126: Why as this code removed as opposed to moving it into `def compress` below? Is this logic no longer necessary for Jython compatibility?

ShaneHarvey: We don't support Jython.

juliusgeo: Ah, I see your comment below addresses part of this. I'm still curious why the Jython workarounds are no longer necessary, did we drop support for Jython?

juliusgeo (approved):
LGTM once tests pass.

PR #896, opened 2022-03-15
PYTHON-3079 Investigate Adding Async APIs to PyMongo

blink1073:


michaeloliverx:
Would a pure `asyncio` implementation not be faster than using threading to avoid blocking on network operations?

See https://github.com/sakal/aiomongo:

> Aiomongo is a rewrite of [Pymongo](https://github.com/mongodb/mongo-python-driver) driver that uses [asyncio](https://docs.python.org/3/library/asyncio.html) for networking. The reason why this client was written is that an alternative [Motor](https://github.com/mongodb/motor) implementation uses thread pool to avoid blocking on network operations. But it appears (see benchmarks below) to work slower than pure asyncio implementation.

ShaneHarvey:
Yes, I would expect a pure asyncio approach to be an improvement (both latency and throughput) over the current version of motor. However, we want to make sure we don't degrade the performance of pymongo. So we're also interested in comparing the the current version of pymongo with a version of pymongo built on top of asyncio at the core. This second comparison is not as cut and dry.

michaeloliverx:
> Yes, I would expect a pure asyncio approach to be an improvement (both latency and throughput) over the current version of motor. However, we want to make sure we don't degrade the performance of pymongo. So we're also interested in comparing the the current version of pymongo with a version of pymongo built on top of asyncio at the core. This second comparison might not be as cut and dry but the initial perf benchmarks show only a <0.5% throughput decrease which would be acceptable.
> 
> We have not yet benchmarked this branch vs motor but I imagine the improvement would be similar to aiomongo.

Interesting, I was initially surprised to find three threads were spawned when using `motor` in one of my projects. It would be interesting to see how a pure asyncio version would perform! Is there anywhere I can subscribe to to track the progress of that?

ShaneHarvey:
You can follow https://jira.mongodb.org/browse/PYTHON-3079 (you may need to create a mongodb jira account).

blink1073:
Closing, as the investigation is complete

PR #899, opened 2022-03-15
PYTHON-3174 Don't reinit client_context.client

ShaneHarvey:
This code was mistakenly added in PYTHON-2463 (#737).


blink1073 (approved):
LGTM!

PR #909, opened 2022-03-28
PYTHON-3084 MongoClient/Database/Collection should not implement Iterable

ShaneHarvey:
As noted in PYTHON-3084, this was the only way to get the typing.Iterable instance check to fail. We could also remove `__next__`+`next` (from client/Database only) but I kept them around so that users get better error messages `TypeError: 'MongoClient' object is not iterable` vs `TypeError: name must be an instance of str`.

### DIFF
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -211,8 +211,12 @@ class ClientUnitTest(unittest.TestCase):
 
     def test_iteration(self):
         client = self.client
+        if "PyPy" in sys.version:
+            msg = "'NoneType' object is not callable"
+        else:
+            msg = "'MongoClient' object is not iterable"
         # Iteration fails
-        with self.assertRaisesRegex(TypeError, "'MongoClient' object is not iterable"):
+        with self.assertRaisesRegex(TypeError, msg):
             for _ in client:  # type: ignore[misc] # error: "None" not callable  [misc]
                 break
         # Index fails
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -26,7 +26,7 @@ import struct
 import sys
 import threading
 import time
-from typing import Type, no_type_check
+from typing import Iterable, Type, no_type_check
 
 sys.path[0:0] = [""]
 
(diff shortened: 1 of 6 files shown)


test/test_collection.py
```diff
@@ -124,7 +124,22 @@ def test_getattr(self):
         self.assertEqual(coll2.write_concern, coll4.write_concern)
 
     def test_iteration(self):
-        self.assertRaises(TypeError, next, self.db)
```
ShaneHarvey, line 127: Notice the old code was accidentally testing Database, not collection.

test/test_collection.py
```diff
@@ -124,7 +124,22 @@ def test_getattr(self):
...
+        coll = self.db.coll
+        # Iteration fails
+        with self.assertRaisesRegex(TypeError, "'Collection' object is not iterable"):
+            for _ in coll:  # type: ignore[misc] # error: "None" not callable  [misc]
+                break
+        # Non-string indices will start failing in PyMongo 5.
+        self.assertEqual(coll[0].name, "coll.0")
+        self.assertEqual(coll[{}].name, "coll.{}")
```
ShaneHarvey, line 134: I opened https://jira.mongodb.org/browse/PYTHON-3184 for this.

ShaneHarvey:
Ah it turns out this approach does not work on PyPy. I'm not surprised since it feels like an undefined behavior hack:
```
FAIL: test_iteration (test_client.ClientUnitTest)
  File "/home/runner/work/mongo-python-driver/mongo-python-driver/test/test_client.py", line 216, in test_iteration
    for _ in client:  # type: ignore[misc] # error: "None" not callable  [misc]
TypeError: 'NoneType' object is not callable

  File "/home/runner/work/mongo-python-driver/mongo-python-driver/test/test_client.py", line 217, in test_iteration
    break
AssertionError: "'MongoClient' object is not iterable" does not match "'NoneType' object is not callable"
```

ShaneHarvey:
With further research it appears that setting `__iter__ = None` is actually well defined behavior:
> Setting a special method to None indicates that the corresponding operation is not available. For example, if a class sets [__iter__()](https://docs.python.org/3/reference/datamodel.html#object.__iter__) to None, the class is not iterable, so calling [iter()](https://docs.python.org/3/library/functions.html#iter) on its instances will raise a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) (without falling back to [__getitem__()](https://docs.python.org/3/reference/datamodel.html#object.__getitem__)). [2](https://docs.python.org/3/reference/datamodel.html#id9)

https://docs.python.org/3/reference/datamodel.html#special-method-names

I'll adjust the assertions so they pass on PyPy.

test/test_client.py
```diff
@@ -210,10 +210,26 @@ def test_getattr(self):
         self.assertIn("has no attribute '_does_not_exist'", str(context.exception))
 
     def test_iteration(self):
-        def iterate():
-            [a for a in self.client]
-
-        self.assertRaises(TypeError, iterate)
+        client = self.client
+        if "PyPy" in sys.version:
+            msg = "'NoneType' object is not callable"
```
ShaneHarvey, line 215: I opened https://foss.heptapod.net/pypy/pypy/-/issues/3716 for PyPy to consider improving this error message.

blink1073 (approved):
LGTM!

PR #914, opened 2022-03-30
PYTHON-3160 Fix MMAPv1 tests

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3160

### DIFF
--- a/test/sessions/driver-sessions-dirty-session-errors.json
+++ b/test/sessions/driver-sessions-dirty-session-errors.json
@@ -448,7 +448,6 @@
           "name": "insertOne",
           "object": "collection0",
           "arguments": {
-            "session": "session0",
             "document": {
               "_id": 2
             }
--- a/test/test_session.py
+++ b/test/test_session.py
@@ -183,12 +183,11 @@ class TestSession(IntegrationTest):
         # "To confirm that implicit sessions only allocate their server session after a
         # successful connection checkout" test from Driver Sessions Spec.
         succeeded = False
+        lsid_set = set()
         failures = 0
         for _ in range(5):
             listener = EventListener()
-            client = rs_or_single_client(
-                event_listeners=[listener], maxPoolSize=1, retryWrites=True
-            )
+            client = rs_or_single_client(event_listeners=[listener], maxPoolSize=1)
             cursor = client.db.test.find({})
             ops: List[Tuple[Callable, List[Any]]] = [
                 (client.db.test.find_one, [{"_id": 1}]),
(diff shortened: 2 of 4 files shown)


test/unified_format.py
```diff
@@ -766,11 +766,30 @@ def setUp(self):
...
+            if name == "listIndexNames":
+                self.skipTest("PyMongo does not support list_index_names()")
+            if client_context.storage_engine == "mmapv1":
+                if name == "createChangeStream":
+                    self.skipTest("MMAPv1 does not support change streams")
+                if name == "withTransaction" or name == "startTransaction":
+                    self.skipTest("MMAPv1 does not support document-level locking")
+            if not client_context.test_commands_enabled:
+                if name == "failPoint" or name == "targetedFailPoint":
+                    self.skipTest("Test commands must be enabled to use fail points")
```
ShaneHarvey, line 791: The explanation here is that we're trying to skip createChangeStream tests on mmapv1 at run time which would cause problems when the test used both createChangeStream and expectError. The new code checks if we should skip the test before executing it. This also speeds up the test suite a bit since we don't have to run half of a bunch of tests just to find out we can skip it.

test/test_session.py
```diff
@@ -225,16 +224,15 @@ def thread_target(op, *args):
...
             for i in listener.results["started"]:
                 if i.command.get("lsid"):
                     lsid_set.add(i.command.get("lsid")["id"])
             if len(lsid_set) == 1:
                 succeeded = True
             else:
                 failures += 1
-        print(failures)
-        self.assertTrue(succeeded)
+        self.assertTrue(succeeded, lsid_set)
```
ShaneHarvey, line 235: This is a small improvement to include the last lsid_set in the error output when the test fails.

test/test_session.py
```diff
@@ -183,12 +183,11 @@ def test_implicit_sessions_checkout(self):
...
         # successful connection checkout" test from Driver Sessions Spec.
         succeeded = False
+        lsid_set = set()
         failures = 0
         for _ in range(5):
             listener = EventListener()
-            client = rs_or_single_client(
-                event_listeners=[listener], maxPoolSize=1, retryWrites=True
-            )
+            client = rs_or_single_client(event_listeners=[listener], maxPoolSize=1)
```
ShaneHarvey, line 190: This test doesn't need retryWrite=True (which doesn't work on MMAPv1).

test/sessions/driver-sessions-dirty-session-errors.json
```diff
@@ -448,7 +448,6 @@
           "name": "insertOne",
           "object": "collection0",
           "arguments": {
-            "session": "session0",
```
ShaneHarvey, line 451: This is a bug fix in the spec test that I will upstream.

blink1073 (approved):
LGTM!

PR #921, opened 2022-04-06
PYTHON-3214 Fix typing markers not being included in the distribution

terencehonles:
The typing markers should be declared in the `package_data` to get picked up and packaged into the distributed wheels.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -12,6 +12,3 @@ include tools/README.rst
 recursive-include test *.pem
 recursive-include test *.py
 recursive-include bson *.h
-include bson/py.typed
-include gridfs/py.typed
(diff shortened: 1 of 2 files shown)


ShaneHarvey (approved):
Thanks for fixing this! LGTM according to https://peps.python.org/pep-0561/#packaging-type-information and double checked the wheel:
```
 unzip -l dist/pymongo-4.2.0.dev0-cp37-cp37m-macosx_10_9_x86_64.whl| grep py.typed
      170  04-05-2022 20:40   bson/py.typed
      170  04-05-2022 20:40   gridfs/py.typed
      170  04-05-2022 20:40   pymongo/py.typed
```

PR #924, opened 2022-04-08
PYTHON-3189 Change Stream event document missing to field for rename events

blink1073:


pymongo/collection.py
```diff
@@ -2584,7 +2584,8 @@ def watch(
 
     def rename(
         self,
-        new_name: str,
+        new_name: Optional[str] = None,
+        to: Optional[str] = None,
```
ShaneHarvey, line 2588: I assume you only added "to" so that the new test would pass. Since we already have "new_name" we don't need a "to" parameter. Let's remove it and instead change the unified test runner to translate to "to" parameter to "new_name". We already do this kind of parameter translation here: https://github.com/mongodb/mongo-python-driver/blob/5578999a90e439fbca06fc0ffc98f4d04e96f7b4/test/utils.py#L1021

ShaneHarvey (changes requested):

pymongo/collection.py
blink1073: Oh good, will do

juliusgeo (approved):
LGTM!

ShaneHarvey (approved):

blink1073:
Both failures are transient (centos URL and `test_dns.TestDNSRepl.test_loadBalanced-false-txt`).

PR #925, opened 2022-04-08
PYTHON-3219 Use $$unsetOrMatches for upsert and multi in update command expectations

juliusgeo:


ShaneHarvey:
This was done in https://github.com/mongodb/mongo-python-driver/pull/934

PR #928, opened 2022-04-13
PYTHON-3225 Stop testing delete on capped collections

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3225

I thought it simpler to stop testing this case altogether rather than try to parse which version supports it or not. The driver doesn't do anything different for a capped collection anyway.


blink1073 (approved):
👍🏼

PR #930, opened 2022-04-18
PYTHON-3228 _tmp_session should validate session input

blink1073:


test/test_client.py
```diff
@@ -505,6 +505,15 @@ def test_client_options(self):
         self.assertIsInstance(c.options.retry_writes, bool)
         self.assertIsInstance(c.options.retry_reads, bool)
 
+    def test_session_validation(self):
+        c = MongoClient(connect=False)
+
+        def try_improper_session():
+            with c._tmp_session({}):  # type:ignore
```
ShaneHarvey, line 512: Instead of testing _tmp_session we should call a public method, like insert_one.

pymongo/mongo_client.py
```diff
@@ -40,6 +40,7 @@
     Any,
     Dict,
     FrozenSet,
+    Generator,
```
ShaneHarvey, line 43: The docs say: 
> Deprecated since version 3.9: [collections.abc.Generator](https://docs.python.org/3/library/collections.abc.html#collections.abc.Generator) now supports []. See [PEP 585](https://www.python.org/dev/peps/pep-0585) and [Generic Alias Type](https://docs.python.org/3/library/stdtypes.html#types-genericalias).

https://docs.python.org/3/library/typing.html#typing.Generator 

Do we need to use collections.abc.Generator on >=3.9?

ShaneHarvey (changes requested):
Could you also update the changelog to mention this in a "Unavoidable breaking changes:" section?

test/test_client.py
blink1073: Done

pymongo/mongo_client.py
blink1073: Done

ShaneHarvey (changes requested):

test/test_collection.py
```diff
@@ -1538,6 +1538,13 @@ def test_aggregation_cursor_alive(self):
 
             self.assertTrue(cursor.alive)
 
+    def test_invalid_session_parameter(self):
+        def try_invalid_session():
+            with self.db.test.aggregate({}, {}):  # type:ignore
```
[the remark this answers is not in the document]
blink1073: Fixed

ShaneHarvey (approved):

PR #932, opened 2022-04-28
PYTHON-3243 Pin version of mypy used

blink1073:


juliusgeo (approved):
These test failures look expected. LGTM!

PR #933, opened 2022-04-28
PYTHON-3038 The doc should clarify that the resulting documents that are produced with upserts are constructed from both the filter and the update params

juliusgeo:


blink1073 (approved):
LGTM, looks like you need to run `pre-commit` locally and pull from `master` to pick up the `mypy` version pins.

pymongo/collection.py
```diff
@@ -951,7 +951,22 @@ def update_one(
           {'x': 4, '_id': 0}
           {'x': 1, '_id': 1}
           {'x': 1, '_id': 2}
-
+          
+        If ``upsert=True`` and no documents match the filter, create a
+        new document based on the filter criteria and update modifications.
+        
+          >>> result = db.test.update_one({'x': -1*2**32}, {'$inc': {'x': 3}}, upsert=True)
```
ShaneHarvey, line 958: What do you think about using `update_one({'x': -10}, ...` instead (for clarity)?

ShaneHarvey (changes requested):

pymongo/collection.py
juliusgeo: That makes sense to me. Fixed.

pymongo/collection.py
```diff
@@ -952,6 +952,21 @@ def update_one(
...
+
+          >>> result = db.test.update_one({'x': -10}, {'$inc': {'x': 3}}, upsert=True)
+          >>> result.matched_count
+          0
+          >>> result.modified_count
+          0
+          >>> result.upserted_id
+          ObjectId('626a678eeaa80587d4bb3fb7')
+          >>> for doc in db.test.find():
+          ...     print(doc)
```
ShaneHarvey, line 966: I just noticed this would make more sense as a find_one:
```
>>> db.test.find_one(result.upserted_id)
{'_id': ObjectId('626a678eeaa80587d4bb3fb7'), 'x': -7}
```

ShaneHarvey (changes requested):

PR #934, opened 2022-04-28
PYTHON-3167 Revert to using the count command for estimatedDocumentCount

ShaneHarvey:
This change implements https://jira.mongodb.org/browse/PYTHON-3167 and fixes https://jira.mongodb.org/browse/PYTHON-2885. By resyncing the spec tests for crud, stable-api, retryable-reads, and atlas-data-lake-testing this change also resolves:
- https://jira.mongodb.org/browse/PYTHON-3166
- https://jira.mongodb.org/browse/PYTHON-3224
- https://jira.mongodb.org/browse/PYTHON-3219


.evergreen/resync-specs.sh
```diff
@@ -69,6 +69,9 @@ cpjson () {
 for spec in "$@"
 do
   case "$spec" in
+    atlas-data-lake-testing|data_lake)
```
juliusgeo, line 72: Is there a specific reason to mix underscores and dashes? It would be more consistent with the rest of the script to use `*`

doc/changelog.rst
```diff
@@ -1,6 +1,37 @@
...
+
+- Fixed a bug where :meth:`~pymongo.collection.Collection.estimated_document_count`
+  would fail with a "CommandNotSupportedOnView" error on views (`PYTHON-2885`_).
+
+Unavoidable breaking changes
+............................
+
+- :meth:`~pymongo.collection.Collection.estimated_document_count` now always uses
+  the `count`_ command. Due to an oversight in versions 5.0.0-5.0.8 of MongoDB,
+  the count command was not included in v1 of the :ref:`versioned-api-ref`.
```
juliusgeo, line 18: This should be "V1" to match the name in `pymongo.server_api.ServerApiVersion`

pymongo/collection.py
```diff
@@ -1694,34 +1694,25 @@ def estimated_document_count(self, comment: Optional[Any] = None, **kwargs: Any)
             command.
           - `**kwargs` (optional): See list of options above.
 
+        .. versionchanged:: 4.2
+           This method now always uses the `count`_ command. Due to an oversight in versions
+           5.0.0-5.0.8 of MongoDB, the count command was not included in v1 of the
```
juliusgeo, line 1699: Same comment as above.

.evergreen/resync-specs.sh
ShaneHarvey: My reason was that the * makes it harder to copy paste, eg:
```
.evergreen/resync-specs.sh data*lake      
zsh: no matches found: data*lake
```

doc/changelog.rst
ShaneHarvey: Fixed.

pymongo/collection.py
ShaneHarvey: Fixed.

.evergreen/resync-specs.sh
ShaneHarvey: If the script is like this:
```
  case "$spec" in
    atlas*data*lake)
```
I don't want to type out "atlas-data-lake" manually, instead I want to copy/paste the match condition "atlas*data*lake" which ends up erroring:
```
.evergreen/resync-specs.sh atlas*data*lake      
zsh: no matches found: atlas*data*lake
```

juliusgeo: ```pbpaste | python -c "import sys; s = sys.stdin.read(); print(s.replace('*', '_'));"``` works on my machine. I think it makes more sense to adhere to the style of the rest of the script in the interest of readability and consistency. Conversely, I think that if you would like to make the change to explicit underscores, then perhaps we should open a new ticket to update the rest of the cases as well. I'm also confused what the purpose of having shortened names is if the intended use case is copying from the switch statement. Is it not faster to simply type "data-lake" as opposed to opening the script, finding the right switch statement, copying it, pasting it, doing that for every resync? Another solution is we could add comments above each switch statement with the string ready for copy pasting.

ShaneHarvey: The intention here is to be user friendly (which ` pbpaste | python -c "import sys; s = sys.stdin.read(); print(s.replace('*', '_'));"` is mostly certainly not) and match both the spec dir name and the python test dir name (and match common abbreviations like CMAP/SDAM/etc..). I used `*` initially because I was lazy and didn't want to include both `_` and `-` names like `bson_corpus|bson-corpus)` but I now realized this was a mistake. I've updated the script to not use `*`.

juliusgeo: SGTM

juliusgeo:
LGTM!

test/unified_format.py
```diff
@@ -883,6 +887,17 @@ def _databaseOperation_listCollections(self, target, *args, **kwargs):
         cursor = target.list_collections(*args, **kwargs)
         return list(cursor)
 
+    def _databaseOperation_createCollection(self, target, *args, **kwargs):
+        # PYTHON-1936 Ignore the listCollections event from create_collection.
```
ShaneHarvey, line 891: I had to add this workaround to fix the listCollections/createCollection problem described here: https://github.com/mongodb/specifications/pull/1202#discussion_r863158569

PR #935, opened 2022-05-02
PYTHON-3247 Mitigate user issues caused by change in directConnection defaults in 4.x

juliusgeo:


pymongo/mongo_client.py
```diff
@@ -137,6 +137,10 @@ def __init__(
     ) -> None:
         """Client for a MongoDB instance, a replica set, or a set of mongoses.
 
+        .. warning:: Starting in PyMongo 4.0, ``directConnection`` now has a default value of
+          False instead of None (which is the same as specifying ``directionConnection=True``).
```
ShaneHarvey, line 141: "(which is the same as specifying ``directionConnection=True``)" is not quite correct. Let's remove that part.

doc/changelog.rst
```diff
@@ -89,6 +89,14 @@ Changes in Version 4.0
 
 .. warning:: PyMongo 4.0 drops support for MongoDB 2.6, 3.0, 3.2, and 3.4.
 
+.. warning:: PyMongo 4.0 changes the default value of the ``directConnection`` URI option and
+  keyword argument to :class:`~pymongo.mongo_client.MongoClient`
+  to ``False`` instead of ``None``, allowing for the automatic
+  discovery of replica sets. This means that if you
+  want a direct connection to a single server you must pass
+  ``directConnection=True`` as a URI option or keyword argument.
+  See the :doc:`migrate-to-pymongo4` for more details.
```
ShaneHarvey, line 98: Can we link to the directConnection section in the migration guide?

doc/migrate-to-pymongo4.rst
```diff
@@ -74,6 +74,30 @@ allowing for the automatic discovery of replica sets. This means that if you
...
+        Topology Description: ...
+
+
+Additionally, the "isWritablePrimary" attribute of a hello command sent back by the server will
+always be True if ``directConnection=False``::
+
+    doc = client.admin.command('hello')
+    result = doc['isWritablePrimary']
+    print(result)
+    >> True
```
ShaneHarvey, line 98: The `>> True` seems a little odd. How about?:
```
    >>> client.admin.command('hello')['isWritablePrimary']
    True
```

doc/migrate-to-pymongo4.rst
```diff
@@ -74,6 +74,30 @@ allowing for the automatic discovery of replica sets. This means that if you
 want a direct connection to a single server you must pass
 ``directConnection=True`` as a URI option or keyword argument.
 
+Here are some example errors that you might see
```
ShaneHarvey, line 77: What do you think about adding something like "If you see any ServerSelectionTimeoutError after upgrading from 3 to 4, you likely need to add directConnection=True when creating the client. Here are some example errors:"

doc/changelog.rst
```diff
@@ -217,11 +225,6 @@ Breaking Changes in 4.0
 - ``tz_aware``, an argument for :class:`~bson.json_util.JSONOptions`,
   now defaults to ``False`` instead of ``True``. ``json_util.loads`` now
   decodes datetime as naive by default.
-- ``directConnection`` URI option and keyword argument to :class:`~pymongo.mongo_client.MongoClient`
-  defaults to ``False`` instead of ``None``, allowing for the automatic
-  discovery of replica sets. This means that if you
-  want a direct connection to a single server you must pass
-  ``directConnection=True`` as a URI option or keyword argument.
```
ShaneHarvey, line 224: I think we should keep this in the list here too just for more visibility.

ShaneHarvey (changes requested):

doc/changelog.rst

doc/changelog.rst

doc/migrate-to-pymongo4.rst
juliusgeo: SGTM, done.

doc/migrate-to-pymongo4.rst
juliusgeo: Done.

pymongo/mongo_client.py
juliusgeo: Bernie mentioned that fact explicitly in the ticket, so I thought it would be good to include so that user understands that None does not equal False (as I would assume). What about making it more clear by saying that is "functionally" the same? Or is that still not the case.

pymongo/mongo_client.py
```diff
@@ -138,8 +138,9 @@ def __init__(
         """Client for a MongoDB instance, a replica set, or a set of mongoses.
 
         .. warning:: Starting in PyMongo 4.0, ``directConnection`` now has a default value of
-          False instead of None (which is the same as specifying ``directionConnection=True``).
-          Please see the :ref:`pymongo4-migration-guide` for more details.
+          False instead of None.
+         For more details, see the relevant section of the PyMongo 4.x migration guide:
+         :ref:`pymongo4-migration-direct-connection`.
```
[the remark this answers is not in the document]
juliusgeo: Oops, my bad. Copy and pasted it wrong. Fixed.

doc/migrate-to-pymongo4.rst
```diff
@@ -74,6 +76,30 @@ allowing for the automatic discovery of replica sets. This means that if you
...
 ``directConnection=True`` as a URI option or keyword argument.
 
+If you see any :exc:`~pymongo.errors.ServerSelectionTimeoutError`'s after upgrading from PyMongo 3 to 4.x, you likely
+need to add ``directConnection=True`` when creating the client.
+Here are some example errors:
+
+.. code-block::
+
+        _select_servers_loop
+            raise ServerSelectionTimeoutError(
```
ShaneHarvey, line 86: Let's remove these two lines:
```
        _select_servers_loop
            raise ServerSelectionTimeoutError(
```

ShaneHarvey (approved):
LGTM with one small change

pymongo/mongo_client.py
juliusgeo: Oh, ok that makes sense. Will leave it removed then.

PR #939, opened 2022-05-05
PYTHON-3235 Drop support for Python 3.6

ShaneHarvey:


doc/changelog.rst
```diff
@@ -4,6 +4,8 @@ Changelog
 Changes in Version 4.2
 ----------------------
 
+.. warning:: PyMongo 4.2 drops support for Python 3.6, Python 3.7+ is now required.
```
blink1073, line 7: I think this would read clearer with a semicolon after 3.6.  I had to read it twice.

blink1073 (approved):

doc/changelog.rst

ShaneHarvey:
Running a bigger patch build to make sure all the tasks that use system "python3" are using 3.7+.

PR #947, opened 2022-05-10
PYTHON-3276 [pymongo] FLE 1.0 shared library

juliusgeo:


pymongo/encryption_options.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         mongocryptd_spawn_path: str = "mongocryptd",
         mongocryptd_spawn_args: Optional[List[str]] = None,
         kms_tls_options: Optional[Mapping[str, Any]] = None,
+        extra_options: Optional[Mapping[str, Any]] = None,
```
ShaneHarvey, line 48: AutoEncryptionOpts accepts the spec's "extra_options" as named parameters. So this should be:
```
csfle_path: ...,
csfle_required: ...,
```

pymongo/encryption_options.py
```diff
@@ -140,6 +142,12 @@ def __init__(
             Or to supply a client certificate::
 
               kms_tls_options={'kmip': {'tlsCertificateKeyFile': 'client.pem'}}
+          - `csfle_path` (optional): Override the path to load the CSFLE library.
+          - `csfle_required` (optional): If 'true', refuse to continue encryption without a CSFLE
+          library
```
blink1073, line 147: I think this needs to align with the backtick on the line above, no the hyphen

pymongo/encryption.py
```diff
@@ -296,7 +296,11 @@ def _get_internal_client(encrypter, mongo_client):
 
         io_callbacks = _EncryptionIO(metadata_client, key_vault_coll, mongocryptd_client, opts)
         self._auto_encrypter = AutoEncrypter(
-            io_callbacks, MongoCryptOptions(opts._kms_providers, schema_map)
+            io_callbacks,
+            MongoCryptOptions(opts._kms_providers, schema_map),
+            opts._csfle_path,
+            opts._csfle_required,
+            opts._bypass_auto_encryption,
```
ShaneHarvey, line 303: These are arguments for MongoCryptOptions, not AutoEncrypter.

pymongo/encryption_options.py
```diff
@@ -45,6 +45,8 @@ def __init__(
         mongocryptd_spawn_path: str = "mongocryptd",
         mongocryptd_spawn_args: Optional[List[str]] = None,
         kms_tls_options: Optional[Mapping[str, Any]] = None,
+        csfle_path: Optional[str] = None,
+        csfle_required: Optional[bool] = False,
```
ShaneHarvey, line 49: Remove Optional, just "bool" is fine. Can you do the same for bypass_auto_encryption?

ShaneHarvey (changes requested):
We need to test with the csfle library and retain the existing mongocryptd tests. There are a few options:
# create new EVG tasks that runs test_encryption with auto-loaded csfle.
# duplicate the tests programmatically, run one set without csfle and one set with csfle_path.
# do what we did in pymongocrypt's testing and run setup.py test twice.

I like option 2 the best since it makes it makes the EVG test logs easier to reason about without bloating the test matrix.

pymongo/encryption.py
juliusgeo: Done.

pymongo/encryption_options.py
juliusgeo: Done.

pymongo/encryption_options.py
juliusgeo: Fixed.

test/test_encryption.py
```diff
@@ -1753,6 +1764,10 @@ def test_case_8(self):
 
 # https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#bypass-spawning-mongocryptd
 class TestBypassSpawningMongocryptdProse(EncryptionIntegrationTest):
+    @unittest.skipIf(
+        os.environ.get("TEST_CSFLE"),
+        "this prose test does not work when CSFLE is " "on a system dynamic library search path.",
```
juliusgeo, line 1769: https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#8-bypass-spawning-mongocryptd
The note at the top says that you cannot run this test unless you disable CSFLE. The only way to disable CSFLE is to both not specify `csfle_path` and make sure CSFLE is not on any search paths. When `TEST_CSFLE` is specified, CSFLE is definitely on the search path, so we have to skip.

.evergreen/run-tests.sh
```diff
@@ -153,7 +155,17 @@ if [ -z "$DATA_LAKE" ]; then
 else
     TEST_ARGS="-s test.test_data_lake"
 fi
-
+if [ -z $TEST_CSFLE ]; then
+    echo "CSFLE not being tested"
+else
+    git clone git@github.com:mongodb-labs/drivers-evergreen-tools.git
```
ShaneHarvey, line 161: drivers-evergreen-tools is already cloned into `$DRIVERS_TOOLS`, eg:
```
$PYTHON $DRIVERS_TOOLS/.evergreen/mongodl.py ...
```

test/test_encryption.py
```diff
@@ -82,6 +82,17 @@ def get_client_opts(client):
 
 
 class TestAutoEncryptionOpts(PyMongoTestCase):
+    @unittest.skipUnless(_HAVE_PYMONGOCRYPT, "pymongocrypt is not installed")
+    @unittest.skipUnless(os.environ.get("TEST_CSFLE"), "csfle is not installed")
+    def test_csfle(self):
+        # Test that we can pick up csfle automatically
+        MongoClient(
```
ShaneHarvey, line 89: We still need to close this client to clean up pymongocrypt resources.

.evergreen/config.yml
```diff
@@ -2235,8 +2231,13 @@ buildvariants:
     auth: "auth"
     ssl: "nossl"
     encryption: "*"
-  display_name: "Encryption ${platform} ${auth} ${ssl}"
-  tasks: *encryption-server-versions
+  display_name: "Encryption ${platform} ${auth} ${ssl} ${encryption}"
+  tasks: &encryption-server-versions
+    - ".latest"
```
ShaneHarvey, line 2236: Add 6.0 and rapid.

.evergreen/config.yml
```diff
@@ -2235,8 +2231,13 @@ buildvariants:
...
+    - ".5.0"
+    - ".4.4"
+    - ".4.2"
+    - ".4.0"
```
ShaneHarvey, line 2240: Since we don't want to test CSFLE with <6.0 we should use the "rules" field to ensure we're only testing classic fle with 4.0+ and shared lib fle with 6.0+: https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#the-rules-field

.evergreen/config.yml
```diff
@@ -2235,8 +2231,13 @@ buildvariants:
     auth: "auth"
     ssl: "nossl"
     encryption: "*"
-  display_name: "Encryption ${platform} ${auth} ${ssl}"
-  tasks: *encryption-server-versions
+  display_name: "Encryption ${platform} ${auth} ${ssl} ${encryption}"
```
ShaneHarvey, line 2234: For all these display_name's we can do this: `${encryption} ${platform} ${auth} ${ssl}`. It will use the `display_name` defined in the axis values above (`Encryption` and `Encryption with CSFLE`)

ShaneHarvey (changes requested):
Nice work so far!

.evergreen/config.yml
juliusgeo: I added a rules field but it does not seem to be removing the tasks.

ShaneHarvey: Due to a longstanding bug in EVG, the `if` condition needs to have all matrix axis in it:
```
  rules:
    - if:
        platform: "*"
        auth: "*"
        ssl: "*"
        encryption: [ "encryption_with_csfle" ]
      then:...
```

.evergreen/config.yml
```diff
@@ -1232,6 +1235,32 @@ tasks:
...
+        - func: "run tests"
+
+    - name: "test-6.0-sharded_cluster"
+      tags: [ "6.0", "sharded_cluster" ]
+      commands:
+        - func: "bootstrap mongo-orchestration"
+          vars:
+            VERSION: "6.0"
+            TOPOLOGY: "sharded_cluster"
+        - func: "run tests"
```
ShaneHarvey, line 1263: We already added 6.0 testing a while back. Why is this showing up here? Can we git merge master and fix the merge conflicts?

.evergreen/config.yml
juliusgeo: It is now working for mac os, but not for the other platforms.

ShaneHarvey: The other ones have different axis.

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM, waiting for this patch to succeed testing the shared library on all server versions on all platforms: https://spruce.mongodb.com/version/628ff72b56234342524f04d3

PR #950, opened 2022-05-17
PYTHON-2683 Convert change stream spec tests to unified test format

blink1073:


blink1073:
RedHat wiki is back up, I kicked the link check job.

test/unified_format.py
```diff
@@ -508,6 +520,7 @@ def _operation_unsetOrMatches(self, spec, actual, key_to_compare):
             # we add a dummy value for the compared key to pass map size check
             actual[key_to_compare] = "dummyValue"
             return
+
```
ShaneHarvey, line 523: Unintentional?

test/unified_format.py
```diff
@@ -1181,19 +1194,30 @@ def check_events(self, spec):
...
+                self.assertGreaterEqual(len(actual_events), len(events), actual_events)
+            else:
+                self.assertEqual(len(actual_events), len(events), actual_events)
 
             for idx, expected_event in enumerate(events):
                 self.match_evaluator.match_event(event_type, expected_event, actual_events[idx])
 
+            if server_connection_id is not None:
+                assert isinstance(server_connection_id, int)
+                assert server_connection_id < 0
```
ShaneHarvey, line 1219: server_connection_id >= 0?

test/unified_format.py
```diff
@@ -1181,19 +1194,30 @@ def check_events(self, spec):
...
+            server_connection_id = event_spec.get("serverConnectionId")
+
             assert event_type in ("command", "cmap")
 
             listener = self.entity_map.get_listener_for_client(client_name)
             actual_events = listener.get_events(event_type)
             if len(events) == 0:
-                self.assertEqual(actual_events, [])
+                if not ignore_extra_events:
+                    self.assertEqual(actual_events, [])
```
ShaneHarvey, line 1206: What do you think of clamping actual_events when ignoreExtraEvents is set?:
```
if ignore_extra_events:
    actual_events = actual_events[:len(events)]
```

That way we only need to check ignore_extra_events in one place.

test/unified_format.py
```diff
@@ -1011,7 +1024,7 @@ def run_entity_operation(self, spec):
                 return self.process_error(exc, expect_error)
             raise
         else:
-            if expect_error:
+            if expect_error and opname != "iterateUntilDocumentOrError":
```
ShaneHarvey, line 1027: What is the iterateUntilDocumentOrError check for?

test/unified_format.py
```diff
@@ -470,9 +470,21 @@ def __init__(self, test_class):
 
     def _operation_exists(self, spec, actual, key_to_compare):
         if spec is True:
-            self.test.assertIn(key_to_compare, actual)
+            if key_to_compare is None:
```
ShaneHarvey, line 473: Why would key_to_compare be None?

ShaneHarvey (changes requested):

test/unified_format.py
blink1073: Fixed

test/unified_format.py
blink1073: For the case where this an "expectError", but the error condition isn't raised, because we got a document instead, which as I understand it is what `iterateUntilDocumentOrError` implies.

test/unified_format.py
blink1073: Done

test/unified_format.py
blink1073: The [spec](https://github.com/mongodb/specifications/blob/223df1f58beb0d3aed1e6ef7f1ee74a707a01a43/source/unified-test-format/unified-test-format.rst#hasserverconnectionid) says it is a "nonpositive Int32"

test/unified_format.py
blink1073: It gets set to `None` [here](https://github.com/mongodb/mongo-python-driver/blob/e02eb287e896ff0e301e67648ba9398f6fe5d799/test/unified_format.py#L544) when the outer operation is special, as is the case with:

```
"$$unsetOrMatches": {
   "$$exists": true
}
```

blink1073:
Moving to draft to make sure we handle [PYTHON-3273](https://jira.mongodb.org/browse/PYTHON-3273) and [PYTHON-3237](https://jira.mongodb.org/browse/PYTHON-3237).

test/unified_format.py
ShaneHarvey: It looks like iterateUntilDocumentOrError is usually (always?) accompanied by "expectError" or "expectResult". When expectError is there but the iterateUntilDocumentOrError op does not error, shouldn't we still fail the test?

test/unified_format.py
```diff
@@ -1181,19 +1193,32 @@ def check_events(self, spec):
...
             actual_events = listener.get_events(event_type)
+            if ignore_extra_events:
+                actual_events = actual_events[: len(events)]
+
             if len(events) == 0:
                 self.assertEqual(actual_events, [])
                 continue
 
-            self.assertGreaterEqual(len(actual_events), len(events), actual_events)
+            self.assertEqual(len(actual_events), len(events), actual_events)
```
ShaneHarvey, line 1211: Oh I see. So we always had the ignoreExtraEvents=True behavior before this.

test/unified_format.py
ShaneHarvey: Hmm, the existing code here is pretty confusing to me. I would have expected key_to_compare to be the key $$unsetOrMatches is comparing against. Is this a bug in _operation_unsetOrMatches?

test/unified_format.py
blink1073: I removed the modification based on feedback about how "expectError" should be handled in the spec: "If [operation.expectError](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#operation-expecterror) is specified, the test runner MUST assert that the operation yielded an error; otherwise, the test runner MUST assert that the operation did not yield an error.

I suspect this will make "test_change_stream.TestUnifiedChangeStreamsErrors.test_change_stream_errors_on_ElectionInProgress" fail again on Server 4.2, let's find out.

test/unified_format.py
blink1073: Okay, by the time we get to this point we are already comparing the value of the key.  For example, if the key was `resumeAfter`, we're checking the value of the `resumeAfter`, which is `{'_data': '82628D384B000000022B0229296E04'}`, against the condition `{{ $exists: true }}`.  

This is for the `test.test_change_stream.TestUnifiedChangeStreamsResumeAllowlist.test_change_stream_resumes_after_CursorNotFound` test case.  The triggering spec is:

```json
"resumeAfter": {
  "$$unsetOrMatches": {
    "$$exists": true
  }
}
```

test/unified_format.py
blink1073: Sure enough:

```python
AssertionError: Excepted error {'errorCode': 216} but "iterateUntilDocumentOrError" succeeded: 
{'_id':
 {'_data': '82628D50B2000000042B022C0100296E5A10044BBFDAAD009A4B0A8CF9E843E82603C946645F69640064628D50B26500615C7FE84A840004'},
 'operationType': 'insert', 'clusterTime': Timestamp(1653428402, 4), 'fullDocument':
 {'_id': ObjectId('628d50b26500615c7fe84a84'), 'z': 3}, 'ns': {'db': 'database0', 'coll': 'collection0'},
 'documentKey': {'_id': ObjectId('628d50b26500615c7fe84a84')}}
```

test/unified_format.py
blink1073: Done and done

test/unified_format.py
blink1073: I don't think it is a spec test failure, given that several other drivers have implemented the unified spec change.

ShaneHarvey: How should we proceed?

blink1073: I'll do some testing with server 4.2 to see if I can spot a bug in our driver.

PR #952, opened 2022-05-24
PYTHON-3283 Remove Generic Typing from the ClientSession Class

blink1073:


test/test_mypy.py
```diff
@@ -313,6 +313,16 @@ def test_son_document_type(self) -> None:
     def test_son_document_type_runtime(self) -> None:
         client = MongoClient(document_class=SON[str, Any], connect=False)
 
+    @only_type_check
+    def test_create_index(self) -> None:
```
ShaneHarvey, line 317: Should this test run create_index instead of a transaction?

ShaneHarvey (changes requested):

test/test_mypy.py
blink1073: Done

ShaneHarvey (approved):

PR #957, opened 2022-06-02
PYTHON-3241 Add Queryable Encryption API to AutoEncryptionOpts

juliusgeo:


pymongo/database.py
```diff
@@ -911,7 +916,34 @@ def drop_collection(
...
+            ):
+                fields = self.client.options.auto_encryption_opts._encrypted_fields_map.get(
+                    full_name
+                )
+                try:
+                    fields = next(self.list_collections(session=session, filter={"name": name}))[
+                        "encryptedFields"
+                    ]
+                except (KeyError, StopIteration):
+                    pass
```
ShaneHarvey, line 936: What do you think about removing the try/expect to simplify the code like this:
```python
opts = list(self.list_collections(filter={"name": name}, session=session))
if opts:
    fields = opts[0].get("encryptedFields")
```

pymongo/database.py
```diff
@@ -911,7 +916,34 @@ def drop_collection(
...
+        full_name = "%s.%s" % (self.name, name)
+        fields = None
+        if not kwargs.get("top_level"):
+            if encrypted_fields:
+                fields = encrypted_fields.get(full_name)
+            if (
+                self.client.options.auto_encryption_opts
+                and self.client.options.auto_encryption_opts._encrypted_fields_map
+            ):
+                fields = self.client.options.auto_encryption_opts._encrypted_fields_map.get(
```
ShaneHarvey, line 928: When fields exists in the `auto_encryption_opts._encrypted_fields_map` why do we still go on to run list_collections? Assuming this is supposed to work similar to schemaMap, then shouldn't we only run list_collections when `encrypted_fields` is not given and the collection doesn't exist in auto_encryption_opts._encrypted_fields_map?

pymongo/encryption_options.py
```diff
@@ -145,9 +147,15 @@ def __init__(
           - `csfle_path` (optional): Override the path to load the CSFLE library.
           - `csfle_required` (optional): If 'true', refuse to continue encryption without a CSFLE
             library
+          - `encrypted_fields_map`: Optional map encoded to BSON `bytes`.
```
ShaneHarvey, line 150: This doc string description needs some work. Maybe:
```
- `encrypted_fields_map`: Map of collection namespace ("db.coll") to documents that described the encrypted fields for Queryable Encryption.
  For example::

    {
      "db.coll": {
          "escCollection": "escCollectionName",
          "eccCollection": "eccCollectionName",
          "ecocCollection": "ecocCollectionName",
          "fields": [
              {
                  "path": "firstName",
                  "keyId": Binary.from_uuid(UUID('00000000-0000-0000-0000-000000000000')),
                  "bsonType": "string",
                  "queries": {"queryType": "equality"}
              },
              {
                  "path": "ssn",
                  "keyId": Binary.from_uuid(UUID('04104104-1041-0410-4104-104104104104')),
                  "bsonType": "string"
              }
          ]
      }
    }
```

test/test_encryption.py
```diff
@@ -605,14 +606,19 @@ def parse_auto_encrypt_opts(self, opts):
...
-    def parse_client_options(self, opts):
+    def parse_client_options(self, opts, encrypted_fields=None):
         """Override clientOptions parsing to support autoEncryptOpts."""
         encrypt_opts = opts.pop("autoEncryptOpts")
-        if encrypt_opts:
-            opts["auto_encryption_opts"] = self.parse_auto_encrypt_opts(encrypt_opts)
+        if encrypt_opts or encrypted_fields:
+            opts["auto_encryption_opts"] = self.parse_auto_encrypt_opts(
+                encrypt_opts, encrypted_fields=encrypted_fields
+            )
```
ShaneHarvey, line 621: From my reading of the spec test, a top-level `encrypted_fields` only means that the test setup code should pass encrypted_fields to create_collection. I don't think it means we pass it to auto_encryption_opts. I think the expected behavior should be similar to how the json_shema field is handled in setup_scenario().

pymongo/collection.py
```diff
@@ -115,6 +116,7 @@ def __init__(
         write_concern: Optional[WriteConcern] = None,
         read_concern: Optional["ReadConcern"] = None,
         session: Optional["ClientSession"] = None,
+        encrypted_fields: Optional[Mapping[str, Any]] = {},
```
ShaneHarvey, line 119: We should use a mutable default parameter like `{}`, let's use None instead.

ShaneHarvey (changes requested):
Nice work so far! A few pointers and questions.

pymongo/database.py
```diff
@@ -874,11 +877,34 @@ def list_collection_names(
 
         return [result["name"] for result in self.list_collections(session=session, **kwargs)]
 
+    def _drop_helper(
+        self,
+        name,
+        session=None,
+        comment=None,
+        **kwargs,
+    ) -> Dict[str, Any]:
```
ShaneHarvey, line 886: Looks like this can fit on one line.

pymongo/database.py
```diff
@@ -911,20 +937,24 @@ def drop_collection(
...
+        full_name = "%s.%s" % (self.name, name)
+        if (
+            not encrypted_fields
+            and self.client.options.auto_encryption_opts
+            and self.client.options.auto_encryption_opts._encrypted_fields_map
+        ):
+            encrypted_fields = self.client.options.auto_encryption_opts._encrypted_fields_map.get(
+                full_name
             )
+            if not encrypted_fields or not encrypted_fields.get(full_name):
```
ShaneHarvey, line 949: `or not encrypted_fields.get(full_name)` is incorrect. encrypted_fields will never have a collection name in it, only encrypted_fields_map does.

ShaneHarvey (changes requested):

pymongo/database.py
```diff
@@ -911,20 +985,38 @@ def drop_collection(
...
+        full_name = "%s.%s" % (self.name, name)
+        if (
+            not encrypted_fields
+            and self.client.options.auto_encryption_opts
+            and self.client.options.auto_encryption_opts._encrypted_fields_map
+        ):
+            encrypted_fields = self.client.options.auto_encryption_opts._encrypted_fields_map.get(
+                full_name
+            )
+        if not encrypted_fields and self.client.options.auto_encryption_opts:
```
blink1073, line 997: Were we going to factor this logic out to a helper method?

pymongo/encryption_options.py
```diff
@@ -45,9 +45,10 @@ def __init__(
         mongocryptd_spawn_path: str = "mongocryptd",
         mongocryptd_spawn_args: Optional[List[str]] = None,
         kms_tls_options: Optional[Mapping[str, Any]] = None,
+        encrypted_fields_map: Optional[Mapping] = None,
+        bypass_query_analysis: bool = False,
         crypt_shared_lib_path: Optional[str] = None,
         crypt_shared_lib_required: bool = False,
-        bypass_query_analysis: bool = False,
```
ShaneHarvey, line 50: The order should be:
```
        crypt_shared_lib_required: bool = False,
        bypass_query_analysis: bool = False,
        encrypted_fields_map: Optional[Mapping] = None,
```

When adding new args, we always append to the end to maintain backwards compat for positional args.

pymongo/encryption_options.py
```diff
@@ -166,6 +190,8 @@ def __init__(
                 "install a compatible version with: "
                 "python -m pip install 'pymongo[encryption]'"
             )
+        self._encrypted_fields_map = encrypted_fields_map
```
ShaneHarvey, line 193: It would be good to call validate_is_mapping here.

ShaneHarvey (changes requested):

pymongo/encryption_options.py
juliusgeo: Done.

pymongo/database.py
juliusgeo: The logic is slightly different for create_collection and drop_collection, so I'm not sure it makes sense to factor it out

pymongo/database.py
```diff
@@ -889,6 +940,29 @@ def drop_collection(
             :class:`~pymongo.client_session.ClientSession`.
           - `comment` (optional): A user-provided comment to attach to this
             command.
+          - `encrypted_fields_map`: Map of collection namespace ("db.coll") to documents that described the encrypted fields for Queryable Encryption.
+            For example::
+
+                {
+                  "db.coll": {
```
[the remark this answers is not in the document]
ShaneHarvey: This is still not fixed. You've copy/pasted the docs for encrypted_fields_map here. It needs to be a description for encrypted_fields.

pymongo/encryption_options.py
```diff
@@ -150,10 +152,33 @@ def __init__(
             outgoing commands. Set `bypass_query_analysis` to use explicit
             encryption on indexed fields without the MongoDB Enterprise Advanced
             licensed crypt_shared library.
+          - `encrypted_fields_map`: Map of collection namespace ("db.coll") to documents that
+            described the encrypted fields for Queryable Encryption. For example::
+
+                {
+                  "db.coll": {
```
ShaneHarvey, line 159: Update this to "db.encryptedCollection" to match the names you use in escCollection and friends below.

pymongo/database.py
```diff
@@ -911,20 +985,38 @@ def drop_collection(
...
+            self._drop_helper(
+                encrypted_fields.get("eccCollection", f"enxcol_.{name}.ecc"),
+                session=session,
+                comment=comment,
+            )
+            self._drop_helper(
+                encrypted_fields.get("ecocCollection", f"enxcol_.{name}.ecoc"),
                 session=session,
+                comment=comment,
             )
```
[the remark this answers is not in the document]
ShaneHarvey: Readability is important. Please update this to
```python
            self._drop_helper(_esc_coll_name(encrypted_fields), session, comment)
            self._drop_helper(_ecc_coll_name(encrypted_fields), session, comment)
            self._drop_helper(_ecoc_coll_name(encrypted_fields), session, comment)
```

test/utils_spec_runner.py
```diff
@@ -229,7 +229,19 @@ def check_result(self, expected_result, result):
 
             return True
         else:
-            self.assertEqual(result, expected_result)
+
+            def _helper(expected_result, result):
+                if isinstance(expected_result, abc.Mapping):
+                    for i in expected_result.keys():
+                        _helper(expected_result[i], result[i])
```
ShaneHarvey, line 236: I believe this needs to be `self.assertEqual` since only the top level documents should ignore missing keys.

pymongo/database.py
ShaneHarvey: The large majority of the logic is identical and we can refactor it but we can address that in a future change.

ShaneHarvey (changes requested):

pymongo/database.py
```diff
@@ -369,14 +393,24 @@ def create_collection(
...
-            # supported.
-            if (not s or not s.in_transaction) and name in self.list_collection_names(
-                filter={"name": name}, session=s
-            ):
-                raise CollectionInvalid("collection %s already exists" % name)
 
+            if (
+                (not s or not s.in_transaction)
+                and not encrypted_fields
+                and not self.client.options.auto_encryption_opts
```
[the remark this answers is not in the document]
ShaneHarvey: SGTM, I opened a new ticket to discuss adding a toggle to create_collection (we don't need it for the beta): https://jira.mongodb.org/browse/PYTHON-3298

ShaneHarvey (approved):
LGTM!

PR #960, opened 2022-06-03
PYTHON-3245 Fix docs for index_key_id

ShaneHarvey:
Better docs for encrypt() which you can see rendered here: https://pymongo--960.org.readthedocs.build/en/960/api/pymongo/encryption.html#pymongo.encryption.ClientEncryption.encrypt


blink1073 (approved):
LGTM!

PR #974, opened 2022-06-14
PYTHON-1552 Adds test that an exception thrown while writing a GridFS file ensures chunks are saved, but not associated with a file.

bcwarner:
The [spec for GridFS](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#id21) says not to rollback any changes that occur as a result of errors while uploading a file, so this test was added to check that the state doesn't get rolled back.

### DIFF
--- a/gridfs/__init__.py
+++ b/gridfs/__init__.py
@@ -109,11 +109,8 @@ class GridFS(object):
 
         Equivalent to doing::
 
-          try:
-              f = new_file(**kwargs)
+          with GridIn(self.database[collection], **kwargs) as f:
               f.write(data)
-          finally:
-              f.close()
 
         `data` can be either an instance of :class:`bytes` or a file-like
         object providing a :meth:`read` method. If an `encoding` keyword
@@ -134,11 +131,10 @@ class GridFS(object):
         .. versionchanged:: 3.0
            w=0 writes to GridFS are now prohibited.
         """
-        grid_file = GridIn(self.__collection, **kwargs)
-        grid_file.write(data)
-        grid_file.close()
 
-        return grid_file._id
+        with GridIn(self.__collection, **kwargs) as grid_file:
+            grid_file.write(data)
+            return grid_file._id
 
     def get(self, file_id: Any, session: Optional[ClientSession] = None) -> GridOut:
         """Get a file from GridFS by ``"_id"``.
--- a/gridfs/__init__.py
+++ b/gridfs/__init__.py
@@ -522,11 +522,11 @@ class GridFSBucket(object):
 
           my_db = MongoClient().test
           fs = GridFSBucket(my_db)
-          grid_in = fs.open_upload_stream(
+          with fs.open_upload_stream(
                 "test_file", chunk_size_bytes=4,
-                metadata={"contentType": "text/plain"})
-          grid_in.write("data I want to store!")
-          grid_in.close()  # uploaded on close
+                metadata={"contentType": "text/plain"}) as grid_in:
+              grid_in.write("data I want to store!")
+          # uploaded on close
 
         Returns an instance of :class:`~gridfs.grid_file.GridIn`.
 
@@ -578,13 +578,13 @@ class GridFSBucket(object):
 
           my_db = MongoClient().test
           fs = GridFSBucket(my_db)
-          grid_in = fs.open_upload_stream_with_id(
+          with fs.open_upload_stream_with_id(
                 ObjectId(),
                 "test_file",
                 chunk_size_bytes=4,
-                metadata={"contentType": "text/plain"})
-          grid_in.write("data I want to store!")
-          grid_in.close()  # uploaded on close
+                metadata={"contentType": "text/plain"}) as grid_in:
+              grid_in.write("data I want to store!")
+          # uploaded on close
 
         Returns an instance of :class:`~gridfs.grid_file.GridIn`.
 
--- a/gridfs/__init__.py
+++ b/gridfs/__init__.py
@@ -109,7 +109,7 @@ class GridFS(object):
 
         Equivalent to doing::
 
-          with GridIn(self.database[collection], **kwargs) as f:
+          with new_file(**kwargs) as f:
               f.write(data)
 
         `data` can be either an instance of :class:`bytes` or a file-like
--- a/gridfs/__init__.py
+++ b/gridfs/__init__.py
@@ -135,10 +135,8 @@ class GridFS(object):
            w=0 writes to GridFS are now prohibited.
         """
         grid_file = GridIn(self.__collection, **kwargs)
-        try:
-            grid_file.write(data)
-        finally:
-            grid_file.close()
+        grid_file.write(data)
+        grid_file.close()
 
         return grid_file._id
 
--- a/gridfs/__init__.py
+++ b/gridfs/__init__.py
@@ -109,7 +109,7 @@ class GridFS(object):
 
         Equivalent to doing::
 
-          with new_file(**kwargs) as f:
+          with fs.new_file(**kwargs) as f:
               f.write(data)
 
         `data` can be either an instance of :class:`bytes` or a file-like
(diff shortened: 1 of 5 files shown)


test/test_grid_file.py
```diff
@@ -675,6 +675,23 @@ def test_context_manager(self):
         with GridOut(self.db.fs, infile._id) as outfile:
             self.assertEqual(contents, outfile.read())
 
+    def test_exception_file_non_existence(self):
+        contents = b"Imagine this is some important data..."
+
+        try:
```
ShaneHarvey, line 681: Instead of try/except it would be cleaner to use `self.assertRaises(CorruptGridFile):` https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

test/test_grid_file.py
```diff
@@ -675,6 +675,23 @@ def test_context_manager(self):
...
+            with GridIn(self.db.fs, filename="important") as infile:
+                infile.write(contents)
+                raise CorruptGridFile("Test exception")
```
ShaneHarvey, line 684: CorruptGridFile would never actually be raised by GridIn.write so it probably makes more sense to use pymongo.errors.ConnectionFailure or something.

test/test_grid_file.py
```diff
@@ -675,6 +675,23 @@ def test_context_manager(self):
...
+                # Never calls infile.flush.
```
ShaneHarvey, line 685: Is it important to call (or not call) flush here?

test/test_grid_file.py
```diff
@@ -675,6 +675,23 @@ def test_context_manager(self):
...
+        except CorruptGridFile as e:
+            pass
+
+        # Expectation: File chunks are written, entry in files doesn't appear.
+        with GridOut(self.db.fs, infile._id) as outfile:
+            self.assertEqual(contents, outfile.read())
+
+        self.assertIsNone(self.db.files.find_one({"_id": infile._id}))
```
ShaneHarvey, line 693: This should be `self.db.fs.files.find_one`

ShaneHarvey (changes requested):
Nice work so far. The test is essentially what we want but now we need to actually fix the bug and update the test to prove the bug is fixed.

gridfs/__init__.py
```diff
@@ -135,10 +135,8 @@ def put(self, data: Any, **kwargs: Any) -> Any:
            w=0 writes to GridFS are now prohibited.
         """
         grid_file = GridIn(self.__collection, **kwargs)
-        try:
-            grid_file.write(data)
-        finally:
-            grid_file.close()
+        grid_file.write(data)
+        grid_file.close()
```
ShaneHarvey, line 139: Let's use a with-statement here instead of calling close explicitly.

test/test_grid_file.py
bcwarner: Turns out flush is internally handled

gridfs/grid_file.py
```diff
@@ -396,9 +396,10 @@ def __enter__(self) -> "GridIn":
     def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> Any:
         """Support for the context manager protocol.
 
-        Close the file and allow exceptions to propagate.
+        Close the file if no exceptions occur and allow exceptions to propagate.
         """
-        self.close()
+        if exc_type is None:
+            self.close()
```
ShaneHarvey, line 402: I think we should still mark the GridIn object as closed. Otherwise calling close explicitly after the with-statement would end up saving the file incorrectly:
```python
with GridIn(...) as file:
    file.write(...)
    raise Exception()
file.close() # This should be a no-op because the file is already marked as "closed"
```

Can you update the test to cover this case as well?

test/test_gridfs.py
```diff
@@ -540,7 +540,7 @@ def test_gridfs_secondary_lazy(self):
 
         # Connects, doesn't create index.
         self.assertRaises(NoFile, fs.get_last_version)
-        self.assertRaises(NotPrimaryError, fs.put, "data")
+        self.assertRaises(NotPrimaryError, fs.put, "data", encoding="utf-8")
```
ShaneHarvey, line 543: Why was this needed?

bcwarner:
Changed `__exit__` so that it only calls `close()` when no exceptions are thrown. Before this it would always call close() and create the associated doc in `files`. I also fixed a unit test where a call to `GridFS.put` didn't provide an encoding and failed.

ShaneHarvey (changes requested):
There are a few more places in GridFS+the docs that still use try/finally. Can you update those to use with-statements too?

test/test_gridfs.py
bcwarner: I think the try-finally statement originally in `GridFS.put` suppressed an error that should have shown up without the `encoding` kwarg there. `GridFS.put` relies on `GridIn.write`, the latter of which requires that an encoding is set for `str` objects, otherwise it will raise an exception.

gridfs/__init__.py
```diff
@@ -109,11 +109,8 @@ def put(self, data: Any, **kwargs: Any) -> Any:
 
         Equivalent to doing::
 
-          try:
-              f = new_file(**kwargs)
+          with GridIn(self.database[collection], **kwargs) as f:
```
ShaneHarvey, line 112: Th9is example should still use new_file, actually `fs.new_file`.

test/test_gridfs.py
ShaneHarvey: Can you try removing this change and share the traceback of the test failure (if it fails)?

ShaneHarvey (changes requested):

test/test_gridfs.py
bcwarner: Yeah, this is the one test that fails:

```
======================================================================
ERROR: test_gridfs_secondary_lazy (test.test_gridfs.TestGridfsReplicaSet)
  File "/Users/benwarner/Documents/pymongo/gridfs/grid_file.py", line 350, in write
    read = data.read
AttributeError: 'str' object has no attribute 'read'

  File "/Users/benwarner/Documents/pymongo/gridfs/grid_file.py", line 357, in write
    data = data.encode(self.encoding)
  File "/Users/benwarner/Documents/pymongo/gridfs/grid_file.py", line 252, in __getattr__
    raise AttributeError("GridIn object has no attribute '%s'" % name)
AttributeError: GridIn object has no attribute 'encoding'

  File "/Users/benwarner/Documents/pymongo/test/test_gridfs.py", line 543, in test_gridfs_secondary_lazy
    self.assertRaises(NotPrimaryError, fs.put, "data")
  File "/Users/benwarner/opt/anaconda3/lib/python3.9/unittest/case.py", line 739, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/Users/benwarner/opt/anaconda3/lib/python3.9/unittest/case.py", line 201, in handle
    callable_obj(*args, **kwargs)
  File "/Users/benwarner/Documents/pymongo/gridfs/__init__.py", line 136, in put
    grid_file.write(data)
  File "/Users/benwarner/Documents/pymongo/gridfs/grid_file.py", line 359, in write
    raise TypeError("must specify an encoding for file in order to write str")
TypeError: must specify an encoding for file in order to write str
```

ShaneHarvey: Oooh I get it now.

ShaneHarvey (changes requested):
Great stuff! I just did a final check for places that should be using a with statement and found these examples that should be updated:
```
doc/migrate-to-pymongo4.rst:885:  grid_in.write(file_data)
gridfs/__init__.py:528:          grid_in.write("data I want to store!")
gridfs/__init__.py:586:          grid_in.write("data I want to store!")
```

ShaneHarvey (approved):
When merging please update the commit message to something a little more succinct like:
`PYTHON-1552 Avoid uploading partial/corrupt GridFS file after an error occurs`

If you want to add more info you can add it to the body of the commit message:
```
PYTHON-1552 Avoid uploading partial/corrupt GridFS file after an error occurs

If an exception is thrown while...
```

PR #975, opened 2022-06-15
PYTHON-3300 Fix Explicit Queryable Encryption Example

juliusgeo:


juliusgeo:
I merged the previous PR without noticing @blink1073 's comment about the key_vault_client. This is to address that.

ShaneHarvey (approved):

PR #977, opened 2022-06-16
Add string equality and non-equality to ObjectId

fuegoio:
Hello,

As we are using MongoDB and Pymongo in a HTTP API, we got a lot of strings coming in as query parameters, json body, ...
We were quite surprised that a comparison between a `string` and its `ObjectId` equivalent would not return `True`.

We don't really know what this change would imply, but we would like to propose it if it can help someone else. If it is something that you have thought of and is not desired upstream, feel free to close ;)

Thanks a lot !

### DIFF
--- a/bson/objectid.py
+++ b/bson/objectid.py
@@ -255,11 +255,15 @@ class ObjectId(object):
     def __eq__(self, other: Any) -> bool:
         if isinstance(other, ObjectId):
             return self.__id == other.binary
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
Let us know if you have any questions about using [Extended JSON](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/)!

PR #978, opened 2022-06-18
PYTHON-3310 Test Failure - query_type must be str or None, not: <enum 'QueryType'>

juliusgeo:


ShaneHarvey (changes requested):
Just need to update the type annotation and docs for the query_type arg of encrypt() as well.

ShaneHarvey (approved):

PR #981, opened 2022-06-22
PYTHON-1824 Allow encoding/decoding out-of-range datetimes via DatetimeMS and datetime_conversion

bcwarner:
This adds an immutable `UTCDatetimeRaw` object that contains the number of milliseconds since the Unix epoch in the UTC timezone. The intent is only for passing back and forth between client and server. I added some basic dunder methods since some of them were going to default to other ones anyways (e.g. `__eq__` as `__id__`).

I also added a new codec option `datetime_conversion`, which has 3 options: `datetime` (current behavior, except on out-of-range), `datetime_clamp` (clamp to min and max datetimes), and `raw` (return the milliseconds as `UTCDatetimeRaw`).

There's no C extension equivalent right now, but it'll crash with C extensions turned on since the extra codec option is an unexpected argument when passed to the equivalent C extensions. I'll add that part later if the Python part looks good.


bson/__init__.py
```diff
@@ -887,26 +897,41 @@ def _dict_to_bson(doc: Any, check_keys: bool, opts: CodecOptions, top_level: boo
 if _USE_C:
     _dict_to_bson = _cbson._dict_to_bson  # noqa: F811
 
+# Inclusive, not inclusive.
```
juliusgeo, line 900: One of these should be "exclusive" I think

test/test_bson.py
```diff
@@ -551,6 +552,71 @@ def test_datetime_encode_decode(self):
         dt2 = decode(encode({"date": dt1}))["date"]
         self.assertEqual(dt1, dt2)
 
+        # Other timestamp formats.
```
juliusgeo, line 555: I think this comment is somehow stuck in the previous indentation block. If you remove the space below this line it should be formatted correctly by pre-commit.

test/test_bson.py
```diff
@@ -551,6 +552,71 @@ def test_datetime_encode_decode(self):
...
+        pairs = [
+            (UTCDatetimeRaw(-1), UTCDatetimeRaw(1)),
+            (UTCDatetimeRaw(0), UTCDatetimeRaw(0)),
+            (UTCDatetimeRaw(1), UTCDatetimeRaw(-1)),
+        ]
+
+        comp_ops = ["__lt__", "__le__", "__eq__", "__ne__", "__gt__", "__ge__"]
+        for lh, rh in pairs:
+            for op in comp_ops:
+                self.assertEqual(getattr(lh, op)(rh), getattr(lh._value, op)(rh._value))
```
juliusgeo, line 569: This is my favorite pattern for writing unit tests :)

bson/__init__.py
```diff
@@ -114,6 +114,7 @@
 from bson.son import RE_TYPE, SON
 from bson.timestamp import Timestamp
 from bson.tz_util import utc
+from bson.utc_datetime import UTCDatetimeRaw
```
ShaneHarvey, line 117: UTCDatetimeRaw is good name but I think we should discuss other options too:
- UTCDatetime
- BsonDatetime
- DatetimeMS
- TimestampMS (potentially confusing with the existing Timestamp class) 
- ?

Overall I'm leaning to DatetimeMS since that's what it represents but do you have any other ideas? We can discuss in slack too.

bson/utc_datetime.py
```diff
@@ -0,0 +1,65 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Representation of a UTC datetime in milliseconds as an int64.
+"""
+import numbers
+from datetime import datetime, timezone
+
+
+class UTCDatetimeRaw:
```
ShaneHarvey, line 21: We should add some docstrings to the class and to the to_datetime method.

ShaneHarvey (changes requested):
Nice! This is a great foundation to build off.

ShaneHarvey:
> There's no C extension equivalent right now, but it'll crash with C extensions turned on since the extra codec option is an unexpected argument when passed to the equivalent C extensions. I'll add that part later if the Python part looks good.

Let's at least update the codec options parsing logic some the C extensions don't crash. That way we can make sure the tests pass.

bson/utc_datetime.py
```diff
@@ -0,0 +1,65 @@
...
+    def __ge__(self, other):
+        return self._value.__ge__(other._value)
+
+    _type_marker = 9
+
+    def __init__(self, value):
+        if isinstance(value, numbers.Number):
+            self._value = int(value)
+        elif isinstance(value, datetime):
+            self._value = value.timestamp() * 1000
```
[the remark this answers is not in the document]
bcwarner: This results in a circular import, I'm copying and pasting since it's 3 lines I'm assuming it won't change.

bson/__init__.py
```diff
@@ -887,26 +897,41 @@ def _dict_to_bson(doc: Any, check_keys: bool, opts: CodecOptions, top_level: boo
 if _USE_C:
     _dict_to_bson = _cbson._dict_to_bson  # noqa: F811
 
+# Inclusive, not inclusive.
+MIN_PYDATETIME = datetime.datetime.min.replace(tzinfo=datetime.timezone.utc).timestamp() * 1000
+MAX_PYDATETIME = datetime.datetime.max.replace(tzinfo=datetime.timezone.utc).timestamp() * 1000 - 1
```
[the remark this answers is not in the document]
ShaneHarvey: Thanks, this all makes sense but it would be better to use _datetime_to_millis here since it already handles with these issues:
```
>>> bson._datetime_to_millis(datetime.datetime.max)
253402300799999
>>> bson._millis_to_datetime(253402300799999, DEFAULT_CODEC_OPTIONS)
datetime.datetime(9999, 12, 31, 23, 59, 59, 999000)
```

bson/__init__.py
ShaneHarvey: Let's go with DatetimeMS since we might want to attach the CodecOptions' timezone to the DatetimeMS in the future.

bson/__init__.py
```diff
@@ -778,7 +788,6 @@ def _encode_maxkey(name: bytes, dummy0: Any, dummy1: Any, dummy2: Any) -> bytes:
     _abc.Mapping: _encode_mapping,
 }
 
-
```
ShaneHarvey, line 781: Unneeded?

bson/utc_datetime.py
```diff
@@ -0,0 +1,96 @@
...
+        seconds = (self._value - diff) // 1000
+        micros = diff * 1000
+
+        if tz_aware:
+            dt = EPOCH_AWARE + datetime.timedelta(seconds=seconds, microseconds=micros)
+            if tzinfo:
+                dt = dt.astimezone(tzinfo)
+            return dt
+        else:
+            return EPOCH_NAIVE + datetime.timedelta(seconds=seconds, microseconds=micros)
```
ShaneHarvey, line 93: This should reuse _millis_to_datetime rather than duplicating the logic.

test/test_bson.py
```diff
@@ -1142,5 +1144,86 @@ def test_bson_encode_decode(self) -> None:
...
+        below = encode(
+            {"x": UTCDatetimeRaw(_datetime_to_millis(datetime.datetime.min, opts1) - 24 * 60 * 60)}
+        )
+        # Could be any timezone.
+        dec_below = decode(below, opts1)
+        self.assertEqual(
+            dec_below["x"], datetime.datetime.min.replace(tzinfo=datetime.timezone.utc)
+        )
+
+        # Aware clamping.
```
ShaneHarvey, line 1225: These tests are missing.

bson/__init__.py
```diff
@@ -887,22 +896,47 @@ def _dict_to_bson(doc: Any, check_keys: bool, opts: CodecOptions, top_level: boo
...
+                dt = dt.astimezone(opts.tzinfo)
+            return dt
+        else:
+            return EPOCH_NAIVE + datetime.timedelta(seconds=seconds, microseconds=micros)
+    elif opts.datetime_conversion == "raw":
+        return UTCDatetimeRaw(millis)
 
 
-def _datetime_to_millis(dtm: datetime.datetime) -> int:
+def _datetime_to_millis(dtm: Union[datetime.datetime]) -> int:
```
ShaneHarvey, line 939: no need for Union anymore.

bson/__init__.py
```diff
@@ -887,22 +896,47 @@ def _dict_to_bson(doc: Any, check_keys: bool, opts: CodecOptions, top_level: boo
 if _USE_C:
     _dict_to_bson = _cbson._dict_to_bson  # noqa: F811
 
+# Inclusive and exclusive min and max for timezones.
+# Timezones are hashed by their offset, which is a timedelta
+# and therefore there are more than 24 possible timezones.
+@functools.cache
+def _min_datetime_ms(tz=datetime.timezone.utc):
+    return datetime.datetime.min.replace(tzinfo=tz).timestamp() * 1000
```
ShaneHarvey, line 904: This should use _datetime_to_millis

bson/codec_options.py
```diff
@@ -198,13 +199,20 @@ def __eq__(self, other: Any) -> Any:
         )
 
 
+class DatetimeConversionOpts(str, enum.Enum):
+    DATETIME = "datetime"
+    DATETIME_CLAMP = "datetime_clamp"
+    RAW = "raw"
```
ShaneHarvey, line 205: How do you feel about renaming RAW to DATETIME_MS?

ShaneHarvey (changes requested):

bson/__init__.py
```diff
@@ -524,6 +523,70 @@ def _bson_to_dict(data: Any, opts: CodecOptions) -> Any:
...
+        return hash(self._value)
+
+    def __repr__(self) -> str:
+        return type(self).__name__ + "(" + str(self._value) + ")"
+
+    __str__ = __repr__
+
+    # Avoids using functools.total_ordering for speed.
+
+    def __lt__(self, other):
```
blink1073, line 557: For consistency, these type signatures should be `other: Any) -> bool`

bson/__init__.py
```diff
@@ -524,6 +523,70 @@ def _bson_to_dict(data: Any, opts: CodecOptions) -> Any:
...
+    Represents a BSON UTC datetime, which is defined as an int64 of
+    milliseconds since the Unix epoch. Principal use is to represent
+    datetimes outside the range of the Python builtin
+    :class:`~datetime.datetime` class when encoding/decoding BSON.
+    To decode UTC datetimes as a ``DatetimeMS``,
+    `datetime_conversion` in :class:`~bson.CodecOptions` must be set
+    to 'datetime_ms'.
+    """
+
+    def __init__(self, value: [int, datetime.datetime]):
```
blink1073, line 537: We've been explicitly adding `-> None` for init methods

bson/json_util.py
```diff
@@ -810,6 +818,12 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
         if json_options.datetime_representation == DatetimeRepresentation.LEGACY:
             return {"$date": millis}
         return {"$date": {"$numberLong": str(millis)}}
+    if isinstance(obj, DatetimeMS):
+        if json_options.datetime_representation == DatetimeRepresentation.ISO8601:
+            raise NotImplementedError()
+        elif json_options.datetime_representation == DatetimeRepresentation.LEGACY:
+            return {"$date": int(obj)}
+        return {"$date": {"$numberLong": int(obj)}}
```
bcwarner, line 826: Converting from `DatetimeMS` to an ISO8601 date would require calculating the date to get the entire range of valid ISO8601 dates, does this mean this be a separate PR like you initially mentioned @ShaneHarvey? 

Going from `DatetimeMS -> datetime.datetime ->` ISO8601 would essentially require clamping.

ShaneHarvey: According to the Extended JSON spec, datetimes before the epoch (1970) or after year 9999 must always be encoded as `{"$date": {"$numberLong": <64-bit signed integer giving millisecs relative to the epoch, as a string>}}`: https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#conversion-table

bson/__init__.py
```diff
@@ -351,9 +353,93 @@ def _get_boolean(
...
+
+    def __repr__(self) -> str:
+        return type(self).__name__ + "(" + str(self._value) + ")"
+
+    __str__ = __repr__
+
+    # Avoids using functools.total_ordering for speed.
+
+    def __lt__(self, other: "DatetimeMS") -> bool:
+        return self._value.__lt__(other._value)
```
ShaneHarvey, line 388: It's a little odd to call `__lt__` (and other "dunder" methods) directly. Let's use <,>,==, etc... instead. EG:
```
        return self._value < other
```

bson/__init__.py
```diff
@@ -351,9 +353,93 @@ def _get_boolean(
...
+    def __gt__(self, other: "DatetimeMS") -> bool:
+        return self._value.__gt__(other._value)
+
+    def __ge__(self, other: "DatetimeMS") -> bool:
+        return self._value.__ge__(other._value)
+
+    _type_marker = 9
+
+    def to_datetime(
+        self, tz_aware: bool = True, tzinfo: datetime.timezone = datetime.timezone.utc
```
ShaneHarvey, line 412: we should accept a codec_options instead of `tz_aware/tzinfo`. This has multiple benefits:
- we don't need to create a new CodecOptions instead for each to_datetime() call which can be slow.
- future datetime options would be supported automatically.

We then need to assert that the datetime_conversion is "datetime" or "datetime_clamp" to avoid returning a DatetimeMS.

bson/codec_options.py
```diff
@@ -198,13 +199,25 @@ def __eq__(self, other: Any) -> Any:
...
 
+class DatetimeConversionOpts(str, enum.Enum):
+    DATETIME = "datetime"
+    DATETIME_CLAMP = "datetime_clamp"
+    DATETIME_MS = "datetime_ms"
+
+    def __str__(self):
+        return f"'{self.value}'"
+
+    __repr__ = __str__
```
ShaneHarvey, line 210: I don't think `__repr__/__str__` is needed for enums is it?

bson/codec_options.py
```diff
@@ -327,6 +346,11 @@ def __new__(
         if not isinstance(type_registry, TypeRegistry):
             raise TypeError("type_registry must be an instance of TypeRegistry")
 
+        if not isinstance(datetime_conversion, str):
+            raise TypeError("datetime_conversion must be a string")
+        if datetime_conversion not in DatetimeConversionOpts.__members__.values():
+            raise ValueError(f"{datetime_conversion} is not a valid option for datetime_conversion")
```
ShaneHarvey, line 352: We can defer to the enum class for all the validation. For example:
```
datetime_conversion = DatetimeConversionOpts(datetime_conversion)
```

On a valid input:
```
>>> DatetimeConversionOpts("datetime")
<DatetimeConversionOpts.DATETIME: 'datetime'>
>>> DatetimeConversionOpts(_)
<DatetimeConversionOpts.DATETIME: 'datetime'>
```
On invalid input:
```
DatetimeConversionOpts("asd")
ValueError: 'asd' is not a valid DatetimeConversionOpts

  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 315, in __call__
    return cls.__new__(cls, value)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 569, in __new__
    raise exc
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 553, in __new__
    result = cls._missing_(value)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 582, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 'asd' is not a valid DatetimeConversionOpts
```
On a non-string:
```
>>> DatetimeConversionOpts(2)
ValueError: 2 is not a valid DatetimeConversionOpts

  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 315, in __call__
    return cls.__new__(cls, value)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 569, in __new__
    raise exc
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 553, in __new__
    result = cls._missing_(value)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 582, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 2 is not a valid DatetimeConversionOpts
```

bson/json_util.py
```diff
@@ -647,8 +655,12 @@ def _parse_canonical_datetime(doc: Any, json_options: JSONOptions) -> datetime.d
         if json_options.tz_aware:
             if json_options.tzinfo:
                 aware = aware.astimezone(json_options.tzinfo)
+            if json_options.datetime_conversion == DatetimeConversionOpts.DATETIME_MS:
+                return DatetimeMS(aware)
             return aware
         else:
+            if json_options.datetime_conversion == DatetimeConversionOpts.DATETIME_MS:
+                return DatetimeMS(aware.replace(tzinfo=None))
```
ShaneHarvey, line 663: Suggest refactoring to avoid duplicating the `aware.replace(tzinfo=None)` call.

bson/json_util.py
```diff
@@ -810,6 +822,15 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
...
         return {"$date": {"$numberLong": str(millis)}}
+    if isinstance(obj, DatetimeMS):
+        if (
+            json_options.datetime_representation == DatetimeRepresentation.ISO8601
+            and 0 <= int(obj) <= bson._max_datetime_ms()
+        ):
+            return default(obj.to_datetime(), json_options)
+        elif json_options.datetime_representation == DatetimeRepresentation.LEGACY:
+            return {"$date": int(obj)}
+        return {"$date": {"$numberLong": int(obj)}}
```
ShaneHarvey, line 833: `int(obj)` -> `str(int(obj))`

test/test_json_util.py
```diff
@@ -241,6 +243,69 @@ def test_datetime(self):
...
+        self.assertEqual(
+            dat_max["x"].to_datetime(tz_aware=False),
+            json_util.loads(json_util.dumps(dat_max))["x"],
+        )
+
+        # Test ISO8601 out-of-range
+        dat_min = {"x": DatetimeMS(-1)}
+        dat_max = {"x": DatetimeMS(_max_datetime_ms() + 1)}
+
+        self.assertEqual("""{"x": {"$date": {"$numberLong": -1}}}""", json_util.dumps(dat_min))
```
ShaneHarvey, line 265: Here and below you should use single quotes instead of `"""`, for example `'{"x": {"$date": {"$numberLong": -1}}}'`

ShaneHarvey (changes requested):
Great work cleaning up the BSON encoding and hooking in the JSON support!

bson/codec_options.py
bcwarner: I included it for `__repr__` in `CodecOptions`, since the behavior of enums is different and requires referencing `.value` otherwise. If we don't include it, `__repr__` and `__str__` look like this:

```
In [127]: DatetimeConversionOpts.DATETIME
Out[127]: <DatetimeConversionOpts.DATETIME: 'datetime'>

In [128]: str(Out[127])
Out[128]: 'DatetimeConversionOpts.DATETIME'
```

ShaneHarvey: I see. I think it's preferable to have `DatetimeConversionOpts.DATETIME` in the repr/str output for CodecOptions.

bson/__init__.py
```diff
@@ -351,9 +353,82 @@ def _get_boolean(
...
+        return self._value > other._value
+
+    def __ge__(self, other: "DatetimeMS") -> bool:
+        return self._value >= other._value
+
+    _type_marker = 9
+
+    def to_datetime(
+        self,
+        codec_options=CodecOptions(
```
ShaneHarvey, line 411: We should change this to be `codec_options=DEFAULT_CODEC_OPTIONS` and update the docstring.

bson/__init__.py
```diff
@@ -113,13 +120,10 @@
 from bson.regex import Regex
 from bson.son import RE_TYPE, SON
 from bson.timestamp import Timestamp
-from bson.tz_util import utc
```
ShaneHarvey, line 116: We need to keep this import to maintain the `from bson import utc` api. To avoid a "imported but not used" error I think we can do:
```
from bson.tz_util import utc as utc
```

bson/_cbsonmodule.c
```diff
@@ -1854,8 +1923,50 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
+                value = datetime_ms_from_millis(self, millis);
+                break;
+            }
+
+            int dt_clamp = options->datetime_conversion == DATETIME_CONVERSION_OPTS_DATETIME_CLAMP;
+            int dt_auto = options->datetime_conversion == DATETIME_CONVERSION_OPTS_DATETIME_AUTO;
+
+
+            if (dt_clamp || dt_auto){
+                PyObject *min_millis_fn = _get_object(state->_min_datetime_ms, "bson.datetime_ms", "_min_datetime_ms");
```
ShaneHarvey, line 1938: All `_get_object` calls need the return value to be NULL checked for error handling.

ShaneHarvey (changes requested):

bson/_cbsonmodule.c
```diff
@@ -1049,6 +1115,10 @@ static int _write_element_to_buffer(PyObject* self, buffer_t buffer,
         }
         *(pymongo_buffer_get_buffer(buffer) + type_byte) = 0x09;
         return buffer_write_int64(buffer, (int64_t)millis);
+    } else if (PyObject_TypeCheck(value, (PyTypeObject *) state->DatetimeMSType)) {
+        long long millis = millis_from_datetime_ms(value);
+        *(pymongo_buffer_get_buffer(buffer) + type_byte) = 0x09;
+        return buffer_write_int64(buffer, (int64_t)millis);
```
ShaneHarvey, line 1121: What should happen when millis_from_datetime_ms returns an error? Hint: let's add a test case for it.

bson/_cbsonmodule.c
```diff
@@ -179,6 +188,58 @@ static long long millis_from_datetime(PyObject* datetime) {
...
+            Py_DECREF(ll_millis);
+            return -1;
+        }
+    }
+
+    if ((millis = PyLong_AsLongLong(ll_millis)) == -1){
+        if (PyErr_Occurred()) {
+            PyObject* OverflowError = _error("OverflowError");
+            if (OverflowError) {
+                PyErr_SetString(OverflowError,
```
ShaneHarvey, line 232: We should use PyExc_OverflowError directly without loading "OverflowError".

bson/_cbsonmodule.c
```diff
@@ -179,6 +188,58 @@ static long long millis_from_datetime(PyObject* datetime) {
...
+                                "Unable to represent this DatetimeMS internally as a long long.");
```
ShaneHarvey, line 233: "MongoDB datetimes can only handle up to 8-byte ints"

bson/_cbsonmodule.c
```diff
@@ -369,6 +433,7 @@ static int _load_python_objects(PyObject* module) {
     state->REType = Py_TYPE(compiled);
     Py_DECREF(empty_string);
     Py_DECREF(compiled);
+
```
ShaneHarvey, line 436: unneeded whitespace

bson/_cbsonmodule.c
```diff
@@ -1854,8 +1923,50 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
+                int64_t max_millis;
+
+                if (options->tz_aware){
+                    PyObject* tzinfo = options->tzinfo;
+                    if (tzinfo == Py_None) {
+                        // Default to UTC.
+                        utc_type = _get_object(state->UTC, "bson.tz_util", "utc");
+                        tzinfo = utc_type;
+                    }
+                    min_millis = PyLong_AsLongLong(PyObject_CallFunctionObjArgs(min_millis_fn, tzinfo, NULL));
```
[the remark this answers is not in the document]
ShaneHarvey: These two PyObject_CallFunctionObjArgs calls still need to be NULL checked. Also the return value from PyObject_CallFunctionObjArgs needs to be decrefed after converting to a `long long`.

bson/_cbsonmodule.c
```diff
@@ -1854,8 +1924,57 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
+                    max_millis = PyLong_AsLongLong(PyObject_CallFunctionObjArgs(max_millis_fn, tzinfo, NULL));
+                } else {
+                    min_millis = PyLong_AsLongLong(PyObject_CallObject(min_millis_fn, NULL));
+                    max_millis = PyLong_AsLongLong(PyObject_CallObject(max_millis_fn, NULL));
+                }
```
ShaneHarvey, line 1960: need to decref min_millis_fn + max_millis_fn to avoid leaking them on every call.

ShaneHarvey (changes requested):

bson/_cbsonmodule.c
bcwarner: Added test cases are in `test_c_millis_from_datetime_ms` in `test_bson`.

ShaneHarvey: This is still not resolved. The problem is that even when `millis_from_datetime_ms` fails, we still call buffer_write_int64 with millis==-1 as if the call succeeded. I suggest refactoring `millis_from_datetime_ms` to accept an "out" parameter so that an error can be returned more easily, for example:
```c
long long millis;
if (!millis_from_datetime_ms(value, &millis)) {
    return 0;
}
*(pymongo_buffer_get_buffer(buffer) + type_byte) = 0x09;
return buffer_write_int64(buffer, (int64_t)millis);
```

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
+import calendar
```
ShaneHarvey, line 1: Missing the standard apache header comment.

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
...
+import datetime
```
ShaneHarvey, line 2: Missing a top-level docstring:
```
"""Tools for representing the BSON datetime type."""

import calendar
``` 

We also need to:
- add a doc/api/bson/datetime_ms.rst which you can model after doc/api/bson/ecimal128.rst. 
- add datetime_ms to doc/api/bson/index.rst

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
...
+)
+from bson.tz_util import utc
+
+EPOCH_AWARE = datetime.datetime.fromtimestamp(0, utc)
+EPOCH_NAIVE = datetime.datetime.utcfromtimestamp(0)
+
+
+class DatetimeMS:
+    """
+    Represents a BSON UTC datetime.
```
ShaneHarvey, line 19: This should be on the same line:
```
"""Represents a BSON UTC datetime
```

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
...
+    DatetimeConversionOpts,
+)
+from bson.tz_util import utc
+
+EPOCH_AWARE = datetime.datetime.fromtimestamp(0, utc)
+EPOCH_NAIVE = datetime.datetime.utcfromtimestamp(0)
+
+
+class DatetimeMS:
+    """
```
ShaneHarvey, line 18: Let's move this docstring to `__init__` so that we can also document the `value` parameter. See Decimal128 for an example format to use, eg: `:Parameters:` and `.. versionadded:: 4.3`.

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
...
+    BSON UTC datetimes are defined as an int64 of milliseconds since the Unix
+    epoch. The principal use of DatetimeMS is to represent datetimes outside
+    the range of the Python builtin :class:`~datetime.datetime` class when
+    encoding/decoding BSON.
+
+    To decode UTC datetimes as a ``DatetimeMS``,`datetime_conversion` in
+    :class:`~bson.CodecOptions` must be set to 'datetime_ms' or
+    'datetime_auto'.
+    """
+
```
ShaneHarvey, line 30: Let's add `__slots__ = ("_value",)` for better performance.

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
...
+    def as_datetime(self, codec_options: CodecOptions = DEFAULT_CODEC_OPTIONS) -> datetime.datetime:
+        """Create a Python :class:`~datetime.datetime` from this DatetimeMS object.
+
+        :Parameters:
+          - `codec_options`: A CodecOptions instance for specifying how the
+            resulting DatetimeMS object will be formatted using ``tz_aware``
+            and ``tz_info``. Defaults to
+            :const:`~bson.codec_options.DEFAULT_CODEC_OPTIONS`.
+
+        .. versionadded:: 4.3
```
ShaneHarvey, line 78: we can remove versionadded here once we add it to the `__init__` docstring.

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        return cast(datetime.datetime, _millis_to_datetime(self._value, codec_options))
```
ShaneHarvey, line 80: Should we have a guard here to only allow DATETIME and DATETIME_CLAMP? Otherwise this could return a DatetimeMS, not a datetime. If you agree, let's open a new ticket in the epic for it rather than changing this PR.

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __int__(self) -> int:
+        return int(self._value)
```
ShaneHarvey, line 83: `_value` is already an int here so can we just return it directly?

bson/datetime_ms.py
```diff
@@ -0,0 +1,134 @@
...
+    Represents a BSON UTC datetime.
+
+    BSON UTC datetimes are defined as an int64 of milliseconds since the Unix
+    epoch. The principal use of DatetimeMS is to represent datetimes outside
+    the range of the Python builtin :class:`~datetime.datetime` class when
+    encoding/decoding BSON.
+
+    To decode UTC datetimes as a ``DatetimeMS``,`datetime_conversion` in
+    :class:`~bson.CodecOptions` must be set to 'datetime_ms' or
+    'datetime_auto'.
```
ShaneHarvey, line 28: We should add a link to:
```
See  :ref:`handling-out-of-range-datetimes` for details.
```

bson/json_util.py
```diff
@@ -228,6 +233,13 @@ class JSONOptions(CodecOptions):
       - `tzinfo`: A :class:`datetime.tzinfo` subclass that specifies the
         timezone from which :class:`~datetime.datetime` objects should be
         decoded. Defaults to :const:`~bson.tz_util.utc`.
+      - `datetime_conversion`: Specifies how UTC datetimes should be decoded
+        within BSON. Valid options include 'datetime_ms' to return as a
+        DatetimeMS, 'datetime' to return as a datetime.datetime and
+        raising a ValueError for out-of-range values, 'datetime_auto' to
+        return DatetimeMS objects when the underlying datetime is
+        out-of-range and 'datetime_clamp' to clamp to the minimum and
+        maximum possible datetimes. Defaults to 'datetime'.
```
ShaneHarvey, line 242: We should add a link to:
```
See  :ref:`handling-out-of-range-datetimes` for details.
```

pymongo/common.py
```diff
@@ -684,6 +699,7 @@ def validate_auto_encryption_opts_or_none(option: Any, value: Any) -> Optional[A
     "uuidrepresentation": validate_uuid_representation,
     "waitqueuemultiple": validate_non_negative_integer_or_none,
     "waitqueuetimeoutms": validate_timeout_or_none,
+    "datetime_conversion": validate_datetime_conversion,
```
ShaneHarvey, line 702: Let's document the datetime_conversion parameter in `MongoClient.__init__`. We should add a link to  :ref:`handling-out-of-range-datetimes` there too.

bson/_cbsonmodule.c
```diff
@@ -179,6 +188,46 @@ static long long millis_from_datetime(PyObject* datetime) {
...
+
+/* Extended-range datetime, takes a DatetimeMS object and extracts the long long value. */
+static long long millis_from_datetime_ms(PyObject* dt){
+    PyObject* ll_millis;
+    long long millis;
+
+    if (!(ll_millis = PyNumber_Long(dt))){
+        if (PyErr_Occurred()) {
+            PyErr_SetString(PyExc_TypeError,
+                            "MongoDB datetimes must be int types.");
```
ShaneHarvey, line 215: We can remove this PyErr_SetString call completely so that the PyNumber_Long error propagates directly to the user.

bson/_cbsonmodule.c
```diff
@@ -179,6 +188,46 @@ static long long millis_from_datetime(PyObject* datetime) {
...
+            return -1;
+        }
+    }
+
+    if ((millis = PyLong_AsLongLong(ll_millis)) == -1){
+        if (PyErr_Occurred()) { /* Overflow */
+            PyErr_SetString(PyExc_OverflowError,
+                            "MongoDB datetimes can only handle up to 8-byte ints.");
```
ShaneHarvey, line 223: "ints." -> "ints", remove the trailing `.` . Errors messages in python generally don't end in a period.


This if statement is also missing a `return -1;`.

bson/_cbsonmodule.c
```diff
@@ -1854,8 +1911,75 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
+                } else {
+                    min_millis_fn_res = PyObject_CallObject(min_millis_fn, NULL);
+                    max_millis_fn_res = PyObject_CallObject(max_millis_fn, NULL);
+                }
+
+                if (!min_millis_fn_res || !max_millis_fn_res){
+                    Py_DECREF(min_millis_fn);
+                    Py_DECREF(max_millis_fn);
+                    Py_DECREF(min_millis_fn_res);
+                    Py_DECREF(max_millis_fn_res);
```
ShaneHarvey, line 1955: Since either or both min_millis_fn_res and max_millis_fn_res can be NULL this needs to be Py_XDECREF:
```
                    Py_XDECREF(min_millis_fn_res);
                    Py_XDECREF(max_millis_fn_res);
```

bson/_cbsonmodule.c
```diff
@@ -1854,8 +1911,75 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
+                        utc_type = _get_object(state->UTC, "bson.tz_util", "utc");
+                        tzinfo = utc_type;
+                    }
+                    min_millis_fn_res = PyObject_CallFunctionObjArgs(min_millis_fn, tzinfo, NULL);
+                    max_millis_fn_res = PyObject_CallFunctionObjArgs(max_millis_fn, tzinfo, NULL);
+                } else {
+                    min_millis_fn_res = PyObject_CallObject(min_millis_fn, NULL);
+                    max_millis_fn_res = PyObject_CallObject(max_millis_fn, NULL);
+                }
+
```
ShaneHarvey, line 1950: It would be simpler to decref min_millis_fn+max_millis_fn here. That way we don't need to duplicate it below.

bson/_cbsonmodule.c
```diff
@@ -1854,8 +1911,75 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
+                if (!min_millis_fn_res || !max_millis_fn_res){
+                    Py_DECREF(min_millis_fn);
+                    Py_DECREF(max_millis_fn);
+                    Py_DECREF(min_millis_fn_res);
+                    Py_DECREF(max_millis_fn_res);
+                    goto invalid;
+                }
+
+                min_millis = PyLong_AsLongLong(min_millis_fn_res);
+                max_millis = PyLong_AsLongLong(max_millis_fn_res);
```
ShaneHarvey, line 1960: Missing error checks for PyLong_AsLongLong.

ShaneHarvey (changes requested):

bson/datetime_ms.py
bcwarner: Yeah I think that makes sense, made an issue at [PYTHON-3365](https://jira.mongodb.org/browse/PYTHON-3365).

bcwarner:
Changed it so instances of DatetimeMS cannot go above a 64-bit int upon instantiation. `mypy` looks like it is failing on something SSL related.

bson/_cbsonmodule.c
```diff
@@ -1948,21 +1947,21 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
                 }
 
                 min_millis = PyLong_AsLongLong(min_millis_fn_res);
                 max_millis = PyLong_AsLongLong(max_millis_fn_res);
 
-                Py_DECREF(min_millis_fn);
-                Py_DECREF(max_millis_fn);
-                Py_DECREF(min_millis_fn_res);
-                Py_DECREF(max_millis_fn_res);
+                if (PyErr_Occurred()){ // min/max_millis check
```
ShaneHarvey, line 1962: Should be:
```
if ((min_millis == -1 || max_millis == -1) && PyErr_Occurred()) {
```

bson/datetime_ms.py
```diff
@@ -15,27 +31,36 @@
...
         if isinstance(value, int):
+            if not (-(2**63) <= value <= 2**63 - 1):
+                raise OverflowError("Must be a 64-bit integer of milliseconds")
             self._value = value
         elif isinstance(value, datetime.datetime):
             self._value = _datetime_to_millis(value)
         else:
             raise TypeError(f"{type(value)} is not a valid type for DatetimeMS")
 
+    __slots__ = ("_value",)
```
ShaneHarvey, line 62: `__slots__` is usually placed first, before __init__.

bson/_cbsonmodule.c
```diff
@@ -1854,8 +1910,75 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
+
+            if (dt_clamp || dt_auto){
+                PyObject *min_millis_fn = _get_object(state->_min_datetime_ms, "bson.datetime_ms", "_min_datetime_ms");
+                PyObject *max_millis_fn = _get_object(state->_max_datetime_ms, "bson.datetime_ms", "_max_datetime_ms");
+                PyObject *min_millis_fn_res;
+                PyObject *max_millis_fn_res;
+                int64_t min_millis;
+                int64_t max_millis;
+
+                if (min_millis_fn == NULL || max_millis_fn == NULL) {
```
ShaneHarvey, line 1932: This leaks when one is NULL and another is not, should be:
```
                if (min_millis_fn == NULL || max_millis_fn == NULL) {
                    Py_XDECREF(min_millis_fn);
                    Py_XDECREF(max_millis_fn);
```

bson/codec_options.py
```diff
@@ -406,9 +428,15 @@ def _parse_codec_options(options: Any) -> CodecOptions:
...
         "type_registry",
+        "datetime_conversion",
     }:
         if k == "uuidrepresentation":
             kwargs["uuid_representation"] = options[k]
+        elif k == "datetime_conversion":
+            if isinstance(options[k], str):
+                kwargs["datetime_conversion"] = DatetimeConversionOpts[options[k]]
+            else:
+                kwargs["datetime_conversion"] = DatetimeConversionOpts(options[k])
```
ShaneHarvey, line 439: These two branches are the same. Can we remove the isinstance check?

ShaneHarvey (changes requested):
Still need to resolve this issue as well: https://github.com/mongodb/mongo-python-driver/pull/981#discussion_r927164808

bson/_cbsonmodule.c
```diff
@@ -223,8 +223,8 @@ static long long millis_from_datetime_ms(PyObject* dt){
         }
     }
     Py_DECREF(ll_millis);
-
-    return millis;
+    *out = millis;
+    return 0;
```
ShaneHarvey, line 227: Nice! One followup nit: Can you refactor this method to return 0 on error and 1 on success? That's the pattern we use for all the other methods.

ShaneHarvey (approved):
LGTM!

PR #982, opened 2022-06-24
PYTHON-3313 Cache AWS Credentials Where Possible

blink1073:


test/auth_aws/test_auth_aws.py
```diff
@@ -53,6 +56,15 @@ def test_connect_uri(self):
         with MongoClient(self.uri) as client:
             client.get_database().test.find_one()
 
+    def test_cache_credentials(self):
+        with MongoClient(self.uri) as client:
+            client.get_database().test.find_one()
+            pool = get_pool(client)
+        pool.reset()
```
ShaneHarvey, line 63: Can we remove the get_pool/reset logic?

test/auth_aws/test_auth_aws.py
```diff
@@ -53,6 +56,15 @@ def test_connect_uri(self):
...
+        with mock.patch.dict(os.environ, {}, clear=True):
+            with MongoClient(self.uri) as client:
+                client.get_database().test.find_one()
```
ShaneHarvey, line 66: Is this test supposed to fail without caching? If so it would be good to make that explicit by asserting that this fails initially:
```
        # Auth fails without env variables.
        with mock.patch.dict(os.environ, {}, clear=True):
            with MongoClient(self.uri) as client:
                    client.get_database().test.find_one()

        # Auth succeeds and caches credential.
        with MongoClient(self.uri) as client:
            client.get_database().test.find_one()
        # Auth succeeds with cached credential.
        with mock.patch.dict(os.environ, {}, clear=True):
            with MongoClient(self.uri) as client:
                client.get_database().test.find_one()
```

ShaneHarvey (changes requested):

test/auth_aws/test_auth_aws.py
blink1073: This may not be possible to test.  We don't cache when the creds environment variables are given directly, which is why those tests are failing.   We have no way of knowing if cached creds were used on EC2 without using `requests-mock` which seems like too much of invasion into implementation details.

ShaneHarvey: We at least need to test what we can. For example, the cache reset on auth failure logic. We can use failCommand to trigger a server side auth failure.

Would it be possible to  use the connectionStatus command to determine which credential was authed?
```
>>> client.admin.command('connectionStatus')
{'authInfo': {'authenticatedUsers': [...]
```
https://www.mongodb.com/docs/manual/reference/command/connectionStatus/#mongodb-data-connectionStatus.authinfo.authenticatedUsers

Probably not I suppose, since I imagine different temp creds will map to the same database level user.

ShaneHarvey: Or a test that "poisons" the cache with invalid creds. The first auth attempt should fail and clear the cache, the next attempt should generate a new cred and succeed.

pymongo/auth_aws.py
```diff
@@ -88,7 +96,13 @@ def _authenticate_aws(credentials, sock_info):
                 # SASL complete.
                 break
     except PyMongoAuthAwsError as exc:
+        # Clear the cached credentials if we hit a failure in auth.
+        _set_cached_credentials(None)
```
blink1073, line 100: I think perhaps we should add an API to `AwsSaslContext` to handle exceptions, which would internally clear the cache.

ShaneHarvey: That would be a bit odd since a new AwsSaslContext is created for each auth whereas the creds are cached globally.

blink1073: It's either that or make the concept of a cache public I think.

ShaneHarvey: What's wrong with the current approach?

blink1073: It feels odd to use a private API from another library outside of testing, even if it is a sibling package.

ShaneHarvey: I think it's fine to make it "public". pymongo-auth-aws isn't used by anyone except pymongo anyway.

.evergreen/run-mongodb-aws-ecs-test.sh
```diff
@@ -29,8 +29,10 @@ authtest () {
     $PYTHON --version
     $PYTHON -m pip install --upgrade wheel setuptools pip
     cd src
+    apt-get update || true
```
ShaneHarvey, line 32: We already run apt-get update above so we can remove this line.

.evergreen/run-mongodb-aws-test.sh
```diff
@@ -60,8 +60,10 @@ authtest () {
     else
       . venvaws/bin/activate
     fi
+    sudo apt-get update || true
+    sudo apt-get install -y git || true
```
ShaneHarvey, line 64: We shouldn't run this because it will affect other tasks run on this host. Can we devise an alternative way to test against the "latest" version of pymongo-auth-aws vs the stable version? Perhaps we should switch all the current tests to use the latest version (using `pip install 'https://...'` so we don't need git) and then add a few tests for the stable version (installed with `pip install '.[aws]'`). We could also double the matrix. What do you think?

test/auth_aws/test_auth_aws.py
```diff
@@ -53,6 +56,57 @@ def test_connect_uri(self):
         with MongoClient(self.uri) as client:
             client.get_database().test.find_one()
 
+    def setup_cache(self):
```
ShaneHarvey, line 59: Should we skip these tests on pymongo_auth_aws <=1.0.2?

test/auth_aws/test_auth_aws.py
```diff
@@ -53,6 +56,57 @@ def test_connect_uri(self):
...
+        self.assertEqual(auth.get_cached_credentials(), None)
+
+        client = MongoClient(self.uri)
+        client.get_database().test.find_one()
+        client.close()
+        return auth.get_cached_credentials()
+
+    def test_cache_credentials(self):
+        creds = self.setup_cache()
+        assert creds is not None
```
ShaneHarvey, line 73: `self.assertIsNotNone`

ShaneHarvey (changes requested):
Could you open a PYTHON ticket to track this work?

blink1073:
> Could you open a PYTHON ticket to track this work?

There's already a https://jira.mongodb.org/browse/PYTHON-3313

.evergreen/run-mongodb-aws-ecs-test.sh
blink1073: Done

.evergreen/run-mongodb-aws-test.sh
blink1073: I think we should test against released unless we are specifically testing a new feature, updated to use the zip download links.

test/auth_aws/test_auth_aws.py
blink1073: Added a guard to check for the attribute

test/auth_aws/test_auth_aws.py
blink1073: Done

.evergreen/config.yml
```diff
@@ -642,6 +642,12 @@ functions:
         working_dir: "src"
         script: |
           ${PREPARE_SHELL}
+          if [ "${skip_EC2_auth_test}" = "true" ]; then
+             echo "This platform does not support the EC2 auth test, skipping..."
+             exit 0
+          fi
+          # Remove any prepare_mongodb_aws.sh files created by other tasks .
+          rm "${PROJECT_DIRECTORY}/prepare_mongodb_aws.sh"
```
blink1073, line 650: I found this oversight when working on the OIDC testing.  We weren't actually testing EC2 credential fetch.

.evergreen/run-mongodb-aws-ecs-test.sh
```diff
@@ -29,8 +29,9 @@ authtest () {
     $PYTHON --version
     $PYTHON -m pip install --upgrade wheel setuptools pip
     cd src
-    $PYTHON -m pip install '.[aws]'
-    $PYTHON test/auth_aws/test_auth_aws.py
+    $PYTHON -m pip install '.'
+    $PYTHON -m pip install https://github.com/blink1073/pymongo-auth-aws/archive/refs/heads/DRIVERS-2333-2.zip
```
ShaneHarvey, line 33: Can we merge https://github.com/mongodb/pymongo-auth-aws/pull/10 and update this to be pymongo-auth-aws@master?

setup.py
```diff
@@ -280,7 +280,7 @@ def build_extension(self, ext):
     "ocsp": pyopenssl_reqs,
     "snappy": ["python-snappy"],
     "zstd": ["zstandard"],
-    "aws": ["pymongo-auth-aws<2.0.0"],
+    "aws": ["pymongo-auth-aws"],
```
ShaneHarvey, line 283: We still need to keep the `<2`.

pymongo/auth_aws.py
```diff
@@ -88,7 +96,13 @@ def _authenticate_aws(credentials, sock_info):
                 # SASL complete.
                 break
     except PyMongoAuthAwsError as exc:
+        # Clear the cached credentials if we hit a failure in auth.
+        set_cached_credentials(None)
         # Convert to OperationFailure and include pymongo-auth-aws version.
         raise OperationFailure(
             "%s (pymongo-auth-aws version %s)" % (exc, pymongo_auth_aws.__version__)
         )
+    except BaseException:
```
ShaneHarvey, line 105: BaseException or Exception? Do we care about things like KeyboardInterrupt here?

ShaneHarvey (changes requested):

.evergreen/run-mongodb-aws-ecs-test.sh
blink1073: Done

setup.py
blink1073: Fixed

pymongo/auth_aws.py
blink1073: Exception seems reasonable, updated

.evergreen/config.yml
blink1073: Updated

.evergreen/run-mongodb-aws-ecs-test.sh
ShaneHarvey: I still see `/refs/heads/DRIVERS-2333-2.zip`.

blink1073: Fixed

.evergreen/run-mongodb-aws-test.sh
```diff
@@ -60,8 +60,9 @@ authtest () {
     else
       . venvaws/bin/activate
     fi
-    python -m pip install '.[aws]'
-    python test/auth_aws/test_auth_aws.py
+    python -m pip install '.'
+    python -m pip install https://github.com/blink1073/pymongo-auth-aws/archive/refs/heads/DRIVERS-2333-2.zip
```
ShaneHarvey, line 64: One more DRIVERS-2333-2

blink1073: Whoops, hadn't save the file, fixed

ShaneHarvey (approved):

pymongo/auth_aws.py
```diff
@@ -88,7 +96,13 @@ def _authenticate_aws(credentials, sock_info):
...
+        # Clear the cached credentials if we hit a failure in auth.
+        set_cached_credentials(None)
         # Convert to OperationFailure and include pymongo-auth-aws version.
         raise OperationFailure(
             "%s (pymongo-auth-aws version %s)" % (exc, pymongo_auth_aws.__version__)
         )
+    except Exception:
+        # Clear the cached credentials if we hit a failure in auth.
+        set_cached_credentials(None)
+        raise
```
ShaneHarvey, line 108: Can you confirm that these `set_cached_credentials(None)` calls are not strictly needed? Otherwise we would be breaking PyMongo<4.3

ShaneHarvey: Should we consider adding `pymongo_auth_aws.enable_caching(True)` to make it an opt in?

blink1073: Done: https://github.com/mongodb/pymongo-auth-aws/pull/11

pymongo/pool.py
```diff
@@ -1375,7 +1375,8 @@ def connect(self, handler=None):
             if handler:
                 handler.contribute_socket(sock_info, completed_handshake=False)
 
-            sock_info.authenticate()
+            with self.lock:
+                sock_info.authenticate()
```
ShaneHarvey, line 1379: Taking the pool lock here is a bit too coarse. The problem is that no other threads will be able to access the pool at all if any other thread is authenticating.

Instead we should introduce a new lock and only acquire it for the first step of MONGODB-AWS auth. Even that is a little more coarse than needed since it will serialize MONGODB-AWS auth with explicit creds too but it's a good start.

pymongo/auth_aws.py
ShaneHarvey: One case this ignores is when an app uses both explicit and implicit aws creds for two different clients. For example, let's say an app:

1. creates a client with implicit creds and the temp creds are cached
1. then creates a new client with invalid explicit creds.
1. The invalid creds will cause an auth failure and will clear the cached temp creds (from 1) which are actually still valid.

Do you think this case is worth handling? One fix might be to pass the AwsCredential to a new function `clear_cache(creds)`. Then the lib can determine whether the fail auth attempt is caused by the cached creds or not.

ShaneHarvey (changes requested):

pymongo/auth_aws.py
blink1073: I intentionally wanted to keep the cache as simple as possible, since rate limiting is hard to hit in practice.

ShaneHarvey: SGTM

pymongo/pool.py
blink1073: I think it makes sense to put the lock in `auth_aws.py`, so we can use it for On-Demand KMS credentials as well, and it can utilize the `_create_lock` convention already present in this library.  Technically this will be too coarse as the lock will be held even for explicit credentials, but the lock will not be held for long in that case, since it is a simple environment lookup.

blink1073: We decided against locking, since it would interfere with the `maxConnecting` setting.

ShaneHarvey (commented):
This can be updated to use the new set_use_cached_credentials api now: https://github.com/mongodb/pymongo-auth-aws/pull/11

blink1073:
> This can be updated to use the new set_use_cached_credentials api 

Done

pymongo/auth_aws.py
```diff
@@ -60,6 +68,9 @@ def _authenticate_aws(credentials, sock_info):
         raise ConfigurationError("MONGODB-AWS authentication requires MongoDB version 4.4 or later")
 
     try:
+        # Enable credential caching.
+        set_use_cached_credentials(True)
```
ShaneHarvey, line 72: Should this be called once at the module level instead?

blink1073: Done

blink1073:
This is ready for review.  The remaining failure is tracked by https://jira.mongodb.org/browse/PYTHON-3468

ShaneHarvey (commented):
Code LGTM but could you add a changelog entry too?

blink1073:
> could you add a changelog entry too?

Done

ShaneHarvey (approved):

PR #983, opened 2022-06-25
Delete .evergreen directory

VaraPrasad08:


behackett:
That directory holds all the scripts for our internal CI system called Evergreen. It can not be deleted.

PR #985, opened 2022-06-28
PYTHON-2484 Added lock sanitization for MongoClient and ObjectId

bcwarner:
Adds some `_after_fork()` calls to reset locks for `ObjectId` and `MongoClient`, and some unit tests for both classes.


pymongo/__init__.py
```diff
@@ -155,3 +155,19 @@ def timeout(seconds: Optional[float]) -> ContextManager:
     if seconds is not None:
         seconds = float(seconds)
     return _csot._TimeoutContext(seconds)
+
+
+import os
+
+from bson.objectid import ObjectId
```
blink1073, line 162: We should `import ObjectId as _ObjectId` so it isn't implicitly in the `pymongo` namespace.

bson/objectid.py
```diff
@@ -285,3 +285,7 @@ def __ge__(self, other: Any) -> bool:
     def __hash__(self) -> int:
         """Get a hash value for this :class:`ObjectId`."""
         return hash(self.__id)
+
+    @classmethod
+    def _after_fork(cls):
```
blink1073, line 290: Needs a docstring

pymongo/__init__.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 """Python driver for MongoDB."""
-
```
juliusgeo, line 16: This change should not be here.

test/test_fork.py
```diff
@@ -0,0 +1,133 @@
...
+        Child => All locks should be reset.
+        """
+        lock_pid: int = -1
+
+        def _fork():
+            nonlocal lock_pid
+            lock_pid = os.fork()
+
+        with patch.object(
+            self.db.client, "_MongoClient__lock", TestFork.LockWrapper(_after_enter=_fork)
```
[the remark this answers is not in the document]
bcwarner: `patch.object` requires a string, so I changed it to get the mangled name indirectly.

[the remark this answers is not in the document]
ShaneHarvey: Let's change it back. Using "_MongoClient__lock" is way more readable than `lock_name: str = next(filter(lambda y: "__lock" in y, dir(self.db.client)))`

pymongo/mongo_client.py
```diff
@@ -839,6 +846,12 @@ def target():
 
             self._encrypter = _Encrypter(self, self.__options.auto_encryption_opts)
 
+    def _after_fork(self):
+        """
+        Resets topology in a child after successfully forking.
+        """
+        self._topology = Topology(self._topology_settings)
```
ShaneHarvey, line 853: This needs to run more of the client `__init__` logic including:
```python
        self._topology = Topology(self._topology_settings)

        def target():
            client = self_ref()
            if client is None:
                return False  # Stop the executor.
            MongoClient._process_periodic_tasks(client)
            return True

        executor = periodic_executor.PeriodicExecutor(
            interval=common.KILL_CURSOR_FREQUENCY,
            min_interval=common.MIN_HEARTBEAT_INTERVAL,
            target=target,
            name="pymongo_kill_cursors_thread",
        )

        # We strongly reference the executor and it weakly references us via
        # this closure. When the client is freed, stop the executor soon.
        self_ref: Any = weakref.ref(self, executor.close)

        if connect:
            self._get_topology()
```

pymongo/mongo_client.py
```diff
@@ -830,6 +833,10 @@ def target():
         self_ref: Any = weakref.ref(self, executor.close)
         self._kill_cursors_executor = executor
 
+        # Add this client to the list of weakly referenced items.
+        # This will be used later if we fork.
+        MongoClient._clients.add(self)
```
ShaneHarvey, line 838: Let's move this to be the last thing in `__init__`. Also is WeakSet.add thread safe?

pymongo/mongo_client.py
```diff
@@ -2125,3 +2138,23 @@ def __enter__(self):
...
+    """
+    # Perform cleanup in clients (i.e. get rid of topology)
+    for client in MongoClient._clients:
+        client._after_fork()
+
+    # Reinitialize locks
+    MongoClientLock._reset_locks()
+
+    # These can't use MongoClientLock due to circular imports.
+    ObjectId._reset_lock()
```
ShaneHarvey, line 2156: If an app _only_ uses bson and never imports pymongo, we still need ObjectId to be fork() safe. Let's move the ObjectId logic to the bson module.

ShaneHarvey (changes requested):

pymongo/lock.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __getattr__(self, item):
+        return getattr(self._lock, item)
+
+    def __enter__(self):
+        self._lock.__enter__()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self._lock.locked():
+            self._lock.__exit__(exc_type, exc_val, exc_tb)
```
ShaneHarvey, line 25: Shouldn't this should always call `__exit__` regardless of the lock's state?

pymongo/mongo_client.py
bcwarner: I don't think it's thread-safe for `add`s. For iteration, `WeakSet` uses `_IterationGuard` to protect items in the set, so I think only `add` might need a lock.

bcwarner:
The locking mechanism gets invoked a lot so I timed it once for each set of tests. There's no performance impact as far as I can tell from one set of runs comparing `mongodb:master` and commit [`2d1c42d`](https://github.com/mongodb/mongo-python-driver/pull/985/commits/2d1c42d3e16dcaafcafb6debed6baf056e6a941d) I ran all tests, the new tests, and tests from `TestClientLazyConnect` since I debugged deadlocking behavior there and thought I saw a performance decline. Here's a table of what I ran on my M1 MBP:

 | Version | Tests Ran | Test Count | Run Time (s) | 
 | --- | --- | --- | --- |
 | Fork Safe | All + New | 2670 | 241.346 | 
 | Upstream | All | 2667 | 243.946 | 
 | Fork Safe | `TestClientLazyConnect` | 4 | 8.714 | 
 | Upstream | `TestClientLazyConnect` | 4 | 8.888 | 
 | Fork Safe | New Only | 3 | 0.024 |

pymongo/lock.py
```diff
@@ -0,0 +1,126 @@
+import platform
```
ShaneHarvey, line 1: Wee need to add the standard license comment at the top.

pymongo/lock.py
```diff
@@ -0,0 +1,126 @@
...
+                _ForkLock._global_locked_cv.notify_all()
+                # Notify all waiting threads in the parent.
+                # In the child, there should be no waiting threads as only the
+                # calling thread gets forked.
+
+        @classmethod
+        def _acquire_locks(cls):
+            _ForkLock._global_lock.acquire()  # Indicate we're about to fork.
+            while _ForkLock._atomic_counter > 0:
+                pass  # Spin while the other threads finish up.
```
ShaneHarvey, line 101: This spin is problematic since it's possible that with many threads  `_atomic_counter > 0` is always true.

pymongo/lock.py
```diff
@@ -0,0 +1,126 @@
...
+            for lock in cls._locks:  # Acquire them all.
```
ShaneHarvey, line 102: I don't think we should try to acquire _all_ the locks since that is vulnerable to deadlock. Let's only acquire the locks protecting the WeakSets.

pymongo/lock.py
```diff
@@ -0,0 +1,126 @@
...
+    # - Is Condition not lowered and are locks active => Continue (not
+    # forking)
+    # - Atomic increment lock counter. (done first and last to ensure
+    # correct state)
+    # - Acquire lock.
+    # - Critical region (and any nested critical regions)
+    # - Release lock.
+    # - Decrement lock counter.
+
+    if platform.python_implementation() == "CPython":
```
ShaneHarvey, line 45: We should be able to remove this CPython specific code here after fixing the deadlock issues.

ShaneHarvey (changes requested):

bcwarner:
I moved the more complicated locking mechanism to `PYTHON-2484-auto-lock` in case it's needed again in the future, the latest commit onward only locks the lock protecting the `WeakSet` for tracking other locks before forking.

bcwarner:
`test_client.TestClient.test_list_databases` was the only failure in the last commit, which I don't believe is related

pymongo/topology.py
```diff
@@ -175,8 +175,8 @@ def open(self):
         elif pid != self._pid:
             self._pid = pid
             warnings.warn(
```
ShaneHarvey, line 177: Should we only raise this warning when register_at_fork is unavailable?

test/test_fork.py
```diff
@@ -0,0 +1,147 @@
...
+        return getattr(self.__lock, item)
+
+
+# Not available for versions of Python without "register_at_fork"
+@skipIf(
+    not hasattr(os, "register_at_fork"), "register_at_fork not available in this version of Python"
+)
+class TestFork(IntegrationTest):
+    def setUp(self):
+        self.db = self.client.pymongo_test
```
ShaneHarvey, line 71: self.db is already defined by IntegrationTest.

test/test_fork.py
```diff
@@ -0,0 +1,147 @@
...
+        self.fork_thread = ForkThread()
```
ShaneHarvey, line 72: `self.fork_thread` should be defined in each test so exit_cond can be passed to the constructor instead of patching exit_cond.

test/test_fork.py
```diff
@@ -0,0 +1,147 @@
...
+
+class LockWrapper:
+    def __init__(self, fork_thread: ForkThread, lock_type: Any = _ForkLock):
+        self.__lock = lock_type()
+        self.fork_thread = fork_thread
+
+    def __enter__(self):
+        self.__lock.__enter__()
+        if not self.fork_thread.is_alive():
+            self.fork_thread.start()
```
ShaneHarvey, line 56: start() is non-blocking. The thread might not actually start (and finally call fork()) until much later. Is that expected or did you want to block here until fork() was called?

test/test_fork.py
```diff
@@ -0,0 +1,147 @@
...
+        Parent => _inc_lock should remain locked.
+        Child => _inc_lock should be unlocked.
+        Must use threading.Lock as ObjectId uses this.
+        """
+        self.fork_thread.exit_cond = lambda: 0 if not ObjectId._inc_lock.locked() else 1
+        with patch.object(
+            ObjectId,
+            "_inc_lock",
+            LockWrapper(fork_thread=self.fork_thread, lock_type=threading.Lock),
+        ):
```
ShaneHarvey, line 117: Same here:
```python
with ObjectId._inc_lock:
    # call fork() assert child process can create an ObjectId().
```

test/test_fork.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def test_topology_reset(self):
+        """
+        Tests that topologies are different from each other.
+        Cannot use ID because virtual memory addresses may be the same.
+        Cannot reinstantiate ObjectId in the topology settings.
+        Relies on difference in PID when opened again.
+        """
+        parent_conn, child_conn = Pipe()
+        cl_test = MongoClient()
```
ShaneHarvey, line 136: Let's use self.client here instead of creating a new one.

test/test_fork.py
```diff
@@ -0,0 +1,147 @@
...
+        init_id = cl_test._topology._pid
+        lock_pid: int = os.fork()
+
+        if lock_pid == 0:  # Child
+            cl_test._topology.open()
```
ShaneHarvey, line 141: Instead of `cl_test._topology.open()` let's run a command with the client, like:
```
cl_test.admin.command('ping')
```

ShaneHarvey (changes requested):

test/test_fork.py
```diff
@@ -0,0 +1,147 @@
...
+            # probabilistic as more locking behavior may ensue. Instead we
+            # check that we can acquire _insertion_lock, meaning we have
+            # successfully completed.
+            with _ForkLock._insertion_lock:
+                return 0  # success
+
+        self.fork_thread.exit_cond = exit_cond
+        with patch.object(
+            self.db.client, "_MongoClient__lock", LockWrapper(fork_thread=self.fork_thread)
+        ):
```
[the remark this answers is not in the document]
bcwarner: I had it doing that a few commits ago, and I realized it's probably more realistic to fork from another thread during a critical region. I think wrapping it also allows us to test lock behavior in functions where they are used.

test/test_fork.py
bcwarner: Yeah I think I assumed it would have run before leaving the critical region, I added a join in the `__exit__` wrapper to make sure it runs all the way in the critical region.

test/test_fork.py
```diff
@@ -0,0 +1,147 @@
...
+        Cannot use ID because virtual memory addresses may be the same.
+        Cannot reinstantiate ObjectId in the topology settings.
+        Relies on difference in PID when opened again.
+        """
+        parent_conn, child_conn = Pipe()
+        cl_test = MongoClient()
+        init_id = cl_test._topology._pid
+        lock_pid: int = os.fork()
+
+        if lock_pid == 0:  # Child
```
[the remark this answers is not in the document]
bcwarner: For the Topology, I think the `_pid` will always change, and the `_kill_cursors_executor._thread.ident` will change as well.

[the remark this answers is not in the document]
bcwarner: I don't think the `thread.ident` one will work. Thread IDs wrt processes can be recycled, and the only way to get a thread's ID wrt the kernel is to be calling `threading.get_native_id()` on a desired thread on Python 3.8. 

Is there another way it might be possible to verify `_kill_cursors_executor` has been reinitialized? I don't see anything in `PeriodicExecutor` whose state would be distinct upon restarting in a child process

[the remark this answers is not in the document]
ShaneHarvey: What about:
```
parent_cursor_exc = client._kill_cursors_executor
...
child_conn.send((parent_cursor_exc != client._kill_cursors_executor, "client._kill_cursors_executor was not reinitialized"))

...

passed, msg = parent_conn.recv()
self.assertTrue(passed, msg)
```

pymongo/mongo_client.py
```diff
@@ -848,6 +851,39 @@ def target():
...
 
+        # Add this client to the list of weakly referenced items.
+        # This will be used later if we fork.
+        with MongoClient._clients_lock:
+            MongoClient._clients[self._topology._topology_id] = self
+
+    def _after_fork(self):
+        """
+        Resets topology in a child after successfully forking.
+        """
```
ShaneHarvey, line 862: Make a single line:
```
        """Resets topology in a child after successfully forking."""
```

pymongo/mongo_client.py
```diff
@@ -848,6 +851,39 @@ def target():
             self._encrypter = _Encrypter(self, self.__options.auto_encryption_opts)
         self._timeout = options.timeout
 
+        # Add this client to the list of weakly referenced items.
+        # This will be used later if we fork.
+        with MongoClient._clients_lock:
+            MongoClient._clients[self._topology._topology_id] = self
+
+    def _after_fork(self):
```
ShaneHarvey, line 859: We should rename this method so we can call it in `__init__`, that way we aren't duplicating the Topology and PeriodicExecutor creation code.

pymongo/mongo_client.py
```diff
@@ -848,6 +851,39 @@ def target():
...
+            interval=common.KILL_CURSOR_FREQUENCY,
+            min_interval=common.MIN_HEARTBEAT_INTERVAL,
+            target=target,
+            name="pymongo_kill_cursors_thread",
+        )
+
+        # We strongly reference the executor and it weakly references us via
+        # this closure. When the client is freed, stop the executor soon.
+        self_ref: Any = weakref.ref(self, executor.close)
+        self._kill_cursors_executor = executor
```
ShaneHarvey, line 882: Do we have to do anything special with _Encrypter here?

pymongo/mongo_client.py
```diff
@@ -848,6 +851,39 @@ def target():
...
+
+        # Connects
+        self._get_topology()
```
ShaneHarvey, line 885: Let's not call _get_topology and instead rely on lazy connection.

pymongo/mongo_client.py
```diff
@@ -2150,3 +2186,35 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         return self.handle(exc_type, exc_val)
+
+
+def _before_fork():
+    # Ensure that we aren't in any critical region.
+    _acquire_locks()
```
ShaneHarvey, line 2193: Don't we have to acquire the _clients_lock here and release it after-fork (like we do for _insertion_lock)?

test/test_fork.py
```diff
@@ -0,0 +1,153 @@
...
+            ObjectId,
+            "_inc_lock",
+            LockWrapper(fork_thread=fork_thread, lock_type=threading.Lock),
+        ):
+            # Generate the ObjectId, will generate a thread to fork
+            # upon __enter__ing the with region.
+
+            ObjectId()
+            fork_thread.join()
+            lock_pid: int = ObjectId._inc_lock.fork_thread.pid
```
ShaneHarvey, line 120: `ObjectId._inc_lock.fork_thread` -> `fork_thread`

test/test_fork.py
```diff
@@ -0,0 +1,153 @@
...
+        def exit_cond():
+            # Checking all the locks after forking this way is highly
+            # probabilistic as more locking behavior may ensue. Instead we
+            # check that we can acquire _insertion_lock, meaning we have
+            # successfully completed.
+            with _insertion_lock:
+                return 0  # success
+
+        fork_thread = ForkThread()
+        fork_thread.exit_cond = exit_cond
```
ShaneHarvey, line 85: use this instead:
```
        fork_thread = ForkThread(exit_cond)
```

test/test_fork.py
```diff
@@ -0,0 +1,153 @@
...
+    def test_lock_object_id(self):
+        """
+        Forks the client with ObjectId's _inc_lock locked.
+        Will fork upon __enter__, waits for child to return.
+        Parent => _inc_lock should remain locked.
+        Child => _inc_lock should be unlocked.
+        Must use threading.Lock as ObjectId uses this.
+        """
+        fork_thread = ForkThread()
+        fork_thread.exit_cond = lambda: 0 if not ObjectId._inc_lock.locked() else 1
```
ShaneHarvey, line 109: use this instead:
```
        fork_thread = ForkThread(lambda: 0 if not ObjectId._inc_lock.locked() else 1)
```

test/test_fork.py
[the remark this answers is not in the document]
ShaneHarvey: I still think it would be preferable to remove LockWrapper or at least simplify it. For example LockWrapper should not be creating a new lock. Instead it should wrap the existing lock we're trying to test. IE:
```
        with patch.object(
            self.db.client, "_MongoClient__lock", LockWrapper(fork_thread=fork_thread, lock=self.client._MongoClient__lock)
        ):

```

Second, there is still a race where the ForkThread might not even run until _after_ the lock has been released.


Manually acquiring the lock is a much simpler and provides the same test coverage so I vote we remove ForkThread and LockWrapper altogether in favor of the approach described above:
```python
with self.client._MongoClient__lock:
    # call fork() assert child process can run an operation on the client.

with ObjectId._inc_lock:
    # call fork() assert child process can still create an ObjectId.
```

ShaneHarvey (changes requested):

bcwarner:
Currently fails on an pymongocrypt error that doesn't appear to be related.

pymongo/mongo_client.py
ShaneHarvey: Ping.

pymongo/mongo_client.py
ShaneHarvey: Ping.

bson/__init__.py
```diff
@@ -1336,3 +1337,24 @@ def decode(self, codec_options: "CodecOptions[_DocumentType]" = DEFAULT_CODEC_OP
 def has_c() -> bool:
     """Is the C extension installed?"""
     return _USE_C
+
+
+def _before_fork():
+    """
+    Acquires the ObjectId lock.
+    """
```
ShaneHarvey, line 1345: This does nothing? Can we remove it?

bson/__init__.py
```diff
@@ -1336,3 +1337,24 @@ def decode(self, codec_options: "CodecOptions[_DocumentType]" = DEFAULT_CODEC_OP
...
+
+
+def _after_fork():
+    """
+    Releases the ObjectID lock in parent and child.
+    """
```
ShaneHarvey, line 1351: We can't release the lock in the parent, only the child.

Also can we fix the docstring format to be a single line?

bson/__init__.py
```diff
@@ -1336,3 +1337,16 @@ def decode(self, codec_options: "CodecOptions[_DocumentType]" = DEFAULT_CODEC_OP
 def has_c() -> bool:
     """Is the C extension installed?"""
     return _USE_C
+
+
+def _after_fork():
+    """Releases the ObjectID lock in parent and child."""
```
ShaneHarvey, line 1343: "in parent and child" -> "in child"

pymongo/mongo_client.py
```diff
@@ -846,7 +857,11 @@ def target():
             from pymongo.encryption import _Encrypter
 
             self._encrypter = _Encrypter(self, self.__options.auto_encryption_opts)
-        self._timeout = options.timeout
+        self._timeout = self.__options.timeout
```
ShaneHarvey, line 860: I suggest moving this section out of _init_deadlockables since it only needs to happen once in `__init__`, not after a fork:
```
        if connect:
            self._get_topology()
        self._encrypter = None
        if self.__options.auto_encryption_opts:
            from pymongo.encryption import _Encrypter

            self._encrypter = _Encrypter(self, self.__options.auto_encryption_opts)
        self._timeout = options.timeout
```

pymongo/mongo_client.py
```diff
@@ -817,6 +821,14 @@ def __init__(
             srv_max_hosts=srv_max_hosts,
         )
 
+        self._init_deadlockables(connect)
```
ShaneHarvey, line 824: Suggest renaming to `self._init_sdam` or `self._init_background`

pymongo/mongo_client.py
```diff
@@ -832,7 +844,6 @@ def target():
             target=target,
             name="pymongo_kill_cursors_thread",
         )
-
```
ShaneHarvey, line 835: extra change here.

pymongo/lock.py
```diff
@@ -0,0 +1,50 @@
...
+import os
+import threading
+import weakref
+
+_HAS_REGISTER_AT_FORK = hasattr(os, "register_at_fork")
+
+# References to instances of _create_lock
+_forkable_locks: weakref.WeakSet = weakref.WeakSet()
+
+_insertion_lock = threading.Lock()
```
ShaneHarvey, line 24: WeakSet.add is implementing using set.add so it is threadsafe and we can remove _insertion_lock and _clients_lock:
https://github.com/python/cpython/blob/3.7/Lib/_weakrefset.py#L81-L84

ShaneHarvey (changes requested):

test/test_fork.py
```diff
@@ -0,0 +1,101 @@
...
+    def test_lock_client(self):
+        """
+        Forks the client with some items locked.
+        Parent => All locks should be as before the fork.
+        Child => All locks should be reset.
+        """
+
+        def exit_cond():
+            with _insertion_lock:
+                return 0  # success
```
ShaneHarvey, line 45: It might be better to run an operation on the client in the child process here, like:
```
self.client.admin.command('ping')
```

pymongo/lock.py
ShaneHarvey: Even asyncio uses WeakSet.add from multiple threads without a lock:
https://github.com/python/cpython/blob/3.7/Lib/asyncio/tasks.py#L860

pymongo/lock.py
```diff
@@ -0,0 +1,40 @@
...
+    """Represents a lock that is tracked upon instantiation using a WeakSet and
+    reset by pymongo upon forking.
+    """
+    lock = threading.Lock()
+    if _HAS_REGISTER_AT_FORK:
+        _forkable_locks.add(lock)
+    return lock
+
+
+def _release_locks(child: bool) -> None:
```
ShaneHarvey, line 35: `child` is no longer needed.

pymongo/mongo_client.py
```diff
@@ -2150,3 +2164,27 @@ def __enter__(self):
...
+def _after_fork_parent():
+    # Only unlock locks in child.
+    _release_locks(False)
+
+
+if hasattr(os, "register_at_fork"):
+    # This will run in the same thread as the fork was called.
+    # If we fork in a critical region on the same thread, it should break.
+    # This is fine since we would never call fork directly from a critical region.
+    os.register_at_fork(after_in_child=_after_fork_child, after_in_parent=_after_fork_parent)
```
ShaneHarvey, line 2190: after_in_parent is no longer needed.

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM!

PR #989, opened 2022-06-29
PYTHON-3315 Remove index_key_id option from ClientEncryption encrypt method

juliusgeo:


ShaneHarvey:
> juliusgeo and others added 30 commits 10 months ago

Something seems off with this branch.

ShaneHarvey:
I think you lost some changes when you force pushed. We still need to remove index_key_id in the tests too.

ShaneHarvey (approved):

PR #990, opened 2022-07-01
PYTHON-3337 Fix capped collection test on MMAPv1

ShaneHarvey:
Explanation:
> If the size field is less than or equal to 4096, then the collection will have a cap of 4096 bytes.

https://www.mongodb.com/docs/manual/core/capped-collections/#create-a-capped-collection


blink1073 (approved):
LGTM!

PR #991, opened 2022-07-02
PYTHON-3333 Fix bug where non-cursor read operations fail in a transaction with directConnection=True on primary

ShaneHarvey:


blink1073 (approved):
LGTM!

PR #996, opened 2022-07-06
PYTHON-3292 Remove ElectionInProgress (216) from ResumableChangeStreamError

blink1073:


ShaneHarvey (commented):
What's the explanation here?

blink1073:
I the explanation is in the ticket, we have never synced with SPEC-1653

ShaneHarvey:
Oh, in that case can you update the title to be "PYTHON-2210 Remove ElectionInProgress (216) from ResumableChangeStreamError"

ShaneHarvey (approved):

PR #998, opened 2022-07-07
PYTHON-3345 CSOT use connection handshake RTT for load balanced mode

ShaneHarvey:
Explanation: 

1.  In load balanced mode the driver skips SDAM so we need to fallback to using the connection's rtt calculated from the connection handshake.
1.  In load balanced mode the driver also skips server selection so we cannot rely on a ServerSelectionTimeoutError to abort the retry loop. Instead we need to explicitly check the CSOT timeout has not expired.

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -1336,6 +1336,10 @@ class MongoClient(common.BaseObject, Generic[_DocumentType]):
                 bulk.started_retryable_write = True
 
         while True:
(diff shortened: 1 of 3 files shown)


ShaneHarvey:
This is ready for review. The remaining load balancer test failures are being addressed by Steve in PYTHON-3336.

blink1073 (approved):
LGTM!

PR #1002, opened 2022-07-11
PYTHON-3153 Update initial DNS seedlist discovery tests to support dedicated load balancer port

blink1073:


blink1073:
Failing tests are fixed in #1001, will re-run after that is merged to double-check, but the expected tests are passing:

```
[2022/07/11 14:20:39.552]   test_loadBalanced-directConnection (test_dns.TestDNSLoadBalanced) ... ok (0.056s)
[2022/07/11 14:20:39.589]   test_loadBalanced-no-results (test_dns.TestDNSLoadBalanced) ... ok (0.033s)
[2022/07/11 14:20:39.651]   test_loadBalanced-replicaSet-errors (test_dns.TestDNSLoadBalanced) ... ok (0.065s)
[2022/07/11 14:20:39.661]   test_loadBalanced-true-multiple-hosts (test_dns.TestDNSLoadBalanced) ... ok (0.010s)
[2022/07/11 14:20:39.704]   test_loadBalanced-true-txt (test_dns.TestDNSLoadBalanced) ... ok (0.039s)
[2022/07/11 14:20:39.761]   test_srvMaxHosts-conflicts_with_loadBalanced-true (test_dns.TestDNSLoadBalanced) ... ok (0.061s)
[2022/07/11 14:20:39.793]   test_srvMaxHosts-conflicts_with_loadBalanced-true-txt (test_dns.TestDNSLoadBalanced) ... ok (0.032s)
[2022/07/11 14:20:39.829]   test_srvMaxHosts-zero (test_dns.TestDNSLoadBalanced) ... ok (0.032s)
[2022/07/11 14:20:39.864]   test_srvMaxHosts-zero-txt (test_dns.TestDNSLoadBalanced) ... ok (0.035s)
```

ShaneHarvey (approved):

pymongo/topology.py
```diff
@@ -692,6 +685,12 @@ def _handle_error(self, address, err_ctx):
                 # Clear the pool.
                 server.reset(service_id)
         elif issubclass(exc_type, ConnectionFailure):
+            # Ignore a handshake error if the server is behind a load balancer but
+            # the service ID is unknown. This indicates that the error happened
+            # when dialing the connection or during the MongoDB handshake, so we
+            # don't know the service ID to use for clearing the pool.
+            if self._settings.load_balanced and not service_id and not err_ctx.completed_handshake:
+                return
```
ShaneHarvey, line 693: Why limit to only connection errors?

blink1073: That was the wrong solution, I now see that we needed a way to distinguish between a handshake error and an auth error.

ShaneHarvey: Does the spec define how we’re supposed to handle an auth error?

pymongo/pool.py
```diff
@@ -1344,7 +1344,7 @@ def remove_stale_sockets(self, reference_generation):
                     self.requests -= 1
                     self.size_cond.notify()
 
-    def connect(self):
+    def connect(self, error_handler=None):
```
ShaneHarvey, line 1347: Suggest error_handler->handler since that's the name used in get_socket already.

pymongo/mongo_client.py
```diff
@@ -2086,6 +2089,7 @@ class _MongoClientErrorHandler(object):
         "completed_handshake",
         "service_id",
         "handled",
+        "sock_info",
```
ShaneHarvey, line 2092: I'm a little wary of storing the sock_info in more places since we've had odd garbage collection issues in the past (https://jira.mongodb.org/browse/PYTHON-2751 for one). 

What if we keep the same contribute_socket behavior as before but add a completed_handshake=True arg then in pool.connect:
```python
            if self.handshake:
                sock_info.hello()
                self.is_writable = sock_info.is_writable
            if handler:
                 handler.contribute_socket(sock_info, complete_handshake=False)
```

pymongo/pool.py
```diff
@@ -1408,7 +1410,10 @@ def get_socket(self, handler=None):
         if self.enabled_for_cmap:
             listeners.publish_connection_check_out_started(self.address)
 
-        sock_info = self._get_socket()
+        sock_info = self._get_socket(error_handler=handler)
+        if handler:
+            handler.contribute_socket(sock_info)
```
ShaneHarvey, line 1415: We don't need this call since contribute_socket is already called directly after the yield (in the calling code).

pymongo/pool.py
blink1073: Done

pymongo/pool.py
```diff
@@ -1374,6 +1374,8 @@ def connect(self):
             raise
 
         sock_info = SocketInfo(sock, self, self.address, conn_id)
+        if error_handler:
+            error_handler.contribute_socket(sock_info)
```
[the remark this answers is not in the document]
blink1073: Done

pymongo/mongo_client.py
blink1073: Done

pymongo/mongo_client.py
```diff
@@ -1184,6 +1184,7 @@ def _get_socket(self, server, session):
             # Reuse the pinned connection, if it exists.
             if in_txn and session._pinned_connection:
                 err_handler.contribute_socket(session._pinned_connection)
+                err_handler.completed_handshake = True
```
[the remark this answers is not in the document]
blink1073: Done

ShaneHarvey (approved):

PR #1014, opened 2022-07-19
PYTHON-3363 Allow change stream to be resumed after a timeout

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3363

### DIFF
--- a/test/unified_format.py
+++ b/test/unified_format.py
@@ -1076,10 +1076,6 @@ class UnifiedSpecTestMixinV1(IntegrationTest):
         self.__raise_if_unsupported("startTransaction", target, ClientSession)
         return target.start_transaction(*args, **kwargs)
 
-    def _cursor_iterateOnce(self, target, *args, **kwargs):
-        self.__raise_if_unsupported("iterateOnce", target, NonLazyCursor, ChangeStream)
-        return target.try_next()
-
     def _changeStreamOperation_iterateUntilDocumentOrError(self, target, *args, **kwargs):
         self.__raise_if_unsupported("iterateUntilDocumentOrError", target, ChangeStream)
         return next(target)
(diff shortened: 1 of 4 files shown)


test/unified_format.py
```diff
@@ -1076,10 +1076,6 @@ def _sessionOperation_startTransaction(self, target, *args, **kwargs):
         self.__raise_if_unsupported("startTransaction", target, ClientSession)
         return target.start_transaction(*args, **kwargs)
 
-    def _cursor_iterateOnce(self, target, *args, **kwargs):
-        self.__raise_if_unsupported("iterateOnce", target, NonLazyCursor, ChangeStream)
```
juliusgeo, line 1080: I'm confused. Isn't this change and the change below combined a no-op?

ShaneHarvey: Not a no-op. The old code only worked with find() cursors while the new code implements iterateOnce for all objects that have try_next().

juliusgeo: Couldn't that also be accomplished by just removing `ChangeStream` from this line? It feels strange to bring this logic into the body of the task execution logic, rather than keeping it in a separate function.

ShaneHarvey: iterate_once is just another name for "try_next" so I think it's fine. This would be the first time we apply a name translation to the op name but we already do similar name transformations when parsing spec args.

ShaneHarvey: The alternative would be:
```python
def _cursor_iterateOnce(self, target, *args, **kwargs):
    self.__raise_if_unsupported("iterateOnce", target, NonLazyCursor, ChangeStream)
    return target.try_next()

def _changeStreamOperation_iterateOnce(self, target, *args, **kwargs):
    self.__raise_if_unsupported("iterateOnce", target, NonLazyCursor, ChangeStream)
    return target.try_next()
```
Which seems subpar to me.

juliusgeo (approved):
LGTM!

PR #1023, opened 2022-07-28
PYTHON-3371 Remove DatetimeConversionOpts.__repr__

bcwarner:
Removed `__repr__` so that the default behavior is used, and changed test to match.

### DIFF
--- a/test/test_bson.py
+++ b/test/test_bson.py
@@ -980,7 +980,8 @@ class TestCodecOptions(unittest.TestCase):
             "uuid_representation=UuidRepresentation.UNSPECIFIED, "
             "unicode_decode_error_handler='strict', "
             "tzinfo=None, type_registry=TypeRegistry(type_codecs=[], "
-            "fallback_encoder=None), datetime_conversion=1)"
+            "fallback_encoder=None), "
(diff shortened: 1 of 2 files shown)


test/test_bson.py
```diff
@@ -980,7 +980,8 @@ def test_codec_options_repr(self):
             "uuid_representation=UuidRepresentation.UNSPECIFIED, "
             "unicode_decode_error_handler='strict', "
             "tzinfo=None, type_registry=TypeRegistry(type_codecs=[], "
-            "fallback_encoder=None), datetime_conversion=1)"
+            "fallback_encoder=None), "
+            "datetime_conversion=<DatetimeConversionOpts.DATETIME: 1>)"
```
ShaneHarvey, line 984: CodecOptions's repr tries to be able to be in a format that can be eval'd so in this case CodecOptions's repr should use `%s` for datetime_conversion. We would expect this line to be:
```

            "datetime_conversion=DatetimeConversionOpts.DATETIME)"
```

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #1024, opened 2022-07-28
PYTHON-3375 Added docstrings to DatetimeConversionOpts

bcwarner:
Added short docstrings to each of the `DatetimeConversionOpts`.

### DIFF
--- a/bson/codec_options.py
+++ b/bson/codec_options.py
@@ -200,12 +200,15 @@ class TypeRegistry(object):
 
 
 class DatetimeConversionOpts(enum.IntEnum):
+    """Options for decoding BSON datetimes."""
+
     DATETIME = 1
     """Decode a BSON UTC datetime as a :class:`datetime.datetime`.
 
     BSON UTC datetimes that cannot be represented as a
-    :class:`~datetime.datetime` will raise an :class:`~builtins.OverflowError`
-    or a :class:`~builtins.ValueError`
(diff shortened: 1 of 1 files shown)


bson/codec_options.py
```diff
@@ -201,9 +201,33 @@ def __eq__(self, other: Any) -> Any:
 
 class DatetimeConversionOpts(enum.IntEnum):
     DATETIME = 1
+    """Decode a BSON UTC datetime as a :class:`datetime.datetime`.
+
+    BSON UTC datetimes that cannot be represented as a
+    :class:`~datetime.datetime` will raise an :class:`~builtins.OverflowError`
```
ShaneHarvey, line 207: Is it always an OverflowError? I recall sometimes it's a ValueError.

bcwarner: Yeah, I think it can looking at the source for `datetime.datetime`

bcwarner:
Fails on an unrelated CSOT test

bson/codec_options.py
```diff
@@ -201,9 +201,34 @@ def __eq__(self, other: Any) -> Any:
...
+    or a :class:`~builtins.ValueError`
```
ShaneHarvey, line 208: missing punctuation.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #1025, opened 2022-07-28
PYTHON-3376/PYTHON-3378 Update FAQ about OverflowError when decoding out of range datetimes

bcwarner:
Added to the FAQ, and tweaked another doc. Will have to be tweaked again after renaming `DatetimeConversionOpts`.


doc/faq.rst
```diff
@@ -489,9 +489,35 @@ limited to years between :data:`datetime.MINYEAR` (usually 1) and
...
+
+Two other options involve decoding as a :class:`~bson.datetime_ms.DatetimeMS`
+object, which simply contains the milliseconds of the decoded UTC datetime.
+:attr:`~bson.datetime_ms.DatetimeConversionOpts.DATETIME_MS` will decode all
+BSON datetimes as this object.
+:attr:`~bson.datetime_ms.DatetimeConversionOpts.DATETIME_AUTO` is a blend of
+the default behavior and
+:attr:`~bson.datetime_ms.DatetimeConversionOpts.DATETIME_MS`, returning
+:class:`~datetime.datetime`s when possible, and
+:class:`~bson.datetime_ms.DatetimeMS` when representations are out-of-range.
```
ShaneHarvey, line 515: Rather than duplicating the behavior of each DatetimeConversionOpts, what if we just suggest using `DATETIME_AUTO` and then link to the DatetimeConversionOpts example page for the rest?

bcwarner:
There's a broken link to `DatetimeMS` which I think is contingent on [PYTHON-3377](https://github.com/mongodb/mongo-python-driver/pull/1026) being fixed.

doc/faq.rst
```diff
@@ -489,9 +489,25 @@ limited to years between :data:`datetime.MINYEAR` (usually 1) and
...
+
+The default option is
+:attr:`~bson.codec_options.DatetimeConversionOpts.DATETIME`, which will
+attempt to decode as a :class:`datetime.datetime`, allowing
+:class:`~builtin.OverflowError` to occur upon out-of-range dates.
+:attr:`~bson.codec_options.DatetimeConversionOpts.DATETIME_AUTO` alters
+this behavior to instead return :class:`~bson.datetime_ms.DatetimeMS` when
+representations are out-of-range, while returning :class:`~datetime.datetime`
+objects as before. For other options, please refer to
+:class:`~bson.codec_options.DatetimeConversionOpts`.
```
ShaneHarvey, line 505: We should add a code example here for setting DATETIME_AUTO at the client level and issuing a find().

bcwarner:
Started making edits to `datetime.rst` before I read [PYTHON-3378](https://jira.mongodb.org/browse/PYTHON-3378) was also a ticket, so I'm going to keep both here.

doc/faq.rst
```diff
@@ -489,9 +489,43 @@ limited to years between :data:`datetime.MINYEAR` (usually 1) and
...
+representations are out-of-range, while returning :class:`~datetime.datetime`
+objects as before:
+
+.. doctest::
+
+    >>> from datetime import datetime
+    >>> from bson.datetime_ms import DatetimeMS
+    >>> from bson.codec_options import DatetimeConversionOpts
+    >>> from pymongo import MongoClient
+    >>> client = MongoClient('localhost', 27017, datetime_conversion=DatetimeConversionOpts.DATETIME_AUTO)
```
ShaneHarvey, line 512: You can remove `'localhost', 27017` since that is the default.

doc/faq.rst
```diff
@@ -489,9 +489,43 @@ limited to years between :data:`datetime.MINYEAR` (usually 1) and
...
+    <pymongo.results.InsertOneResult object at 0x1032cf880>
+    >>> for x in client.db.collection.find():
+    ...     print(x)
+    {'_id': ObjectId('62e45f36e7a1bacf393dbf60'), 'x': datetime.datetime(1970, 1, 1, 0, 0)}
+    {'_id': ObjectId('62e45f36e7a1bacf393dbf61'), 'x': DatetimeMS(4611686018427387904)}
+
+For other options, please refer to
+:class:`~bson.codec_options.DatetimeConversionOpts`.
+
+If we need to use the default decoding behavior with datetimes not supported by
```
ShaneHarvey, line 525: I'd suggest changing from "If we need to use the default decoding behavior ... filter out documents ..." -> "Another option that does not involve setting `datetime_conversion` is to to filter out documents ..."

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM assuming the doctests pass.

ShaneHarvey:
FYI you can run the doctests locally by running `python setup.py doc -t` with single mongod server.

PR #1035, opened 2022-08-10
PYTHON-3256 Obtain AWS credentials for CSFLE in the same way as for MONGODB-AWS

blink1073:
Depends on https://github.com/mongodb/libmongocrypt/pull/440

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -139,7 +139,8 @@ if [ -n "$TEST_ENCRYPTION" ]; then
     export PYMONGOCRYPT_LIB
 
     # TODO: Test with 'pip install pymongocrypt'
-    git clone https://github.com/mongodb/libmongocrypt.git libmongocrypt_git
+    git clone -b PYTHON-3256
+ https://github.com/blink1073/libmongocrypt.git libmongocrypt_git
     python -m pip install --prefer-binary -r .evergreen/test-encryption-requirements.txt
     python -m pip install ./libmongocrypt_git/bindings/python
     python -c "import pymongocrypt; print('pymongocrypt version: '+pymongocrypt.__version__)"
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -140,11 +140,7 @@ if [ -n "$TEST_ENCRYPTION" ]; then
     export PYMONGOCRYPT_LIB
 
     # TODO: Test with 'pip install pymongocrypt'
-    git clone https://github.com/blink1073/libmongocrypt.git libmongocrypt_git
-    pushd libmongocrypt_git
-    git fetch origin PYTHON-3256
-    git checkout PYTHON-3256
-    popd
+    git clone https://github.com/mongodb/libmongocrypt.git libmongocrypt_git
     python -m pip install --prefer-binary -r .evergreen/test-encryption-requirements.txt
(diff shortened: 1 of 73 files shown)


.evergreen/run-tests.sh
```diff
@@ -139,13 +139,20 @@ if [ -n "$TEST_ENCRYPTION" ]; then
     export PYMONGOCRYPT_LIB
 
     # TODO: Test with 'pip install pymongocrypt'
-    git clone https://github.com/mongodb/libmongocrypt.git libmongocrypt_git
+    git clone https://github.com/blink1073/libmongocrypt.git libmongocrypt_git
+    pushd libmongocrypt_git
```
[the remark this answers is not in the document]
blink1073: It just makes it easier to go back to where you came from, I almost always use `pushd` instead of `cd`.

test/test_encryption.py
```diff
@@ -2202,6 +2202,37 @@ def test_05_roundtrip_encrypted_unindexed(self):
...
+
+    @unittest.skipIf(any(AWS_CREDS.values()), "AWS environment credentials are set")
+    def test_01_failure(self):
+        self.client_encryption = ClientEncryption(
+            kms_providers={"aws": {}},
+            key_vault_namespace="keyvault.datakeys",
+            key_vault_client=client_context.client,
+            codec_options=OPTS,
+        )
+        with self.assertRaises(EncryptionError) as ctx:
```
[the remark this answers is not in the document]
blink1073: Removed

setup.py
```diff
@@ -278,12 +278,14 @@ def build_extension(self, ext):
...
+aws_reqs = ["pymongo-auth-aws<2.0.0"]
+
 extras_require = {
-    "encryption": ["pymongocrypt>=1.3.0,<2.0.0"],
+    "encryption": ["pymongocrypt>=1.3.0,<2.0.0"] + aws_reqs,
     "ocsp": pyopenssl_reqs,
     "snappy": ["python-snappy"],
     "zstd": ["zstandard"],
-    "aws": ["pymongo-auth-aws<2.0.0"],
+    "aws": aws_reqs,
```
ShaneHarvey, line 288: Let's update the docs to say that pymongo-auth-aws is also required for encryption in these 3 places:
```
README.rst:135:  $ python -m pip install "pymongo[encryption]"
doc/examples/encryption.rst:30:  $ python -m pip install 'pymongo[encryption]'
doc/installation.rst:75:  $ python3 -m pip install "pymongo[encryption]"
```

ShaneHarvey (changes requested):

setup.py
blink1073: Isn't the main point of the extra that the user doesn't need to concern themselves with the libraries used?

ShaneHarvey: Sure but we should still document what libraries will be installed.

ShaneHarvey: Note I'm not saying to change those 3 lines in particular. Just the surrounding test. For example the readme says:
```
Client-Side Field Level Encryption requires `pymongocrypt
<https://pypi.org/project/pymongocrypt/>`_::

  $ python -m pip install "pymongo[encryption]"

```
But this is now incomplete because it doesn't mention pymongo-auth-aws.

blink1073: Fixed

ShaneHarvey (approved):

PR #1036, opened 2022-08-12
PYTHON-3393 Added fork-safety stress test.

bcwarner:


test/test_fork.py
```diff
@@ -98,3 +100,42 @@ def test_topology_reset(self):
...
+                super().__init__()
+
+            def run(self) -> None:
+                clients = []
+                for _ in range(10):
+                    clients.append(MongoClient())
+
+                # The sequence of actions should be somewhat reproducible.
+                # If truly random, there is a chance we never actually fork.
+                # The scheduling is somewhat random, so rely upon that.
```
ShaneHarvey, line 118: I don't understand this comment. The loop always calls fork() 3 times.

test/test_fork.py
```diff
@@ -98,3 +100,42 @@ def test_topology_reset(self):
...
+
+                for i in range(200):
+                    # Pick a random client.
+                    rc = clients[i % len(clients)]
+                    if i % 50 == 0:
+                        # Fork
+                        pid = os.fork()
+                        if pid == 0:  # Child => Can we use it?
+                            code = action(rc)
+                            os._exit(code)
```
ShaneHarvey, line 131: If the action() fails then `os._exit(code)` will never run. I think we need to call os._exit in a finally block (for all the tests that use fork).

test/test_fork.py
```diff
@@ -98,3 +100,42 @@ def test_topology_reset(self):
...
+                        else:  # Parent => Child work?
+                            self.runner.assertEqual(0, os.waitpid(pid, 0)[1])
+                    action(rc)
+
+        threads = [ForkThread(self) for _ in range(10)]
+        for t in threads:
+            t.start()
+
+        for t in threads:
+            t.join()
```
ShaneHarvey, line 141: This test will not fail even when one of the threads raises an exception because join() only says that the thread has stopped, not that it succeeded. Consider using ExceptionCatchingThread instead.

ShaneHarvey (changes requested):

test/test_fork.py
bcwarner: Yeah, the ticket mentioned randomly forking, and I was initially thinking like this:
```
if random.random() <= 0.01:
   os.fork()
```
But there would be a small probability a test would exhibit behavior where it never forks, and would be less clear if it fails/passes differently than in the majority of cases. So the number and relative frequency is fixed, but the ordering of the forks is somewhat random.

test/test_fork.py
```diff
@@ -98,3 +100,45 @@ def test_topology_reset(self):
...
+        # Fork randomly while doing operations.
+        class ForkThread(ExceptionCatchingThread):
+            def __init__(self, runner):
+                self.runner = runner
+                super().__init__()
+
+            def run(self) -> None:
+                clients = []
+                for _ in range(10):
+                    clients.append(MongoClient())
```
ShaneHarvey, line 114: MongoClient  -> rs_or_single_client()

test/test_fork.py
```diff
@@ -98,3 +107,53 @@ def test_topology_reset(self):
...
+                    if i % 50 == 0 and self.fork:
+                        # Fork
+                        pid = os.fork()
+                        if pid == 0:  # Child => Can we use it?
+                            code = -1
+                            try:
+                                code = action(rc)
+                            finally:
+                                os._exit(code)
+                        else:  # Parent => Child work?
```
ShaneHarvey, line 146: Can you clear up or remove `=> Child work?`

test/test_fork.py
```diff
@@ -98,3 +107,53 @@ def test_topology_reset(self):
...
+                    client.admin.command("ping")
+                    return 0
+
+                for i in range(200):
+                    # Pick a random client.
+                    rc = self.clients[i % len(self.clients)]
+                    if i % 50 == 0 and self.fork:
+                        # Fork
+                        pid = os.fork()
+                        if pid == 0:  # Child => Can we use it?
```
ShaneHarvey, line 140: Can you clear up or remove `=> Can we use it?`

test/test_fork.py
```diff
@@ -98,3 +107,53 @@ def test_topology_reset(self):
...
+                            code = -1
+                            try:
+                                code = action(rc)
```
ShaneHarvey, line 143: Let's have the child run a command on _all_ the clients, not just one.

test/test_fork.py
```diff
@@ -98,3 +107,53 @@ def test_topology_reset(self):
...
+        threads = [ForkThread(self, clients) for _ in range(10)]
+        threads[-1].fork = True
+        for t in threads:
+            t.start()
+
+        for t in threads:
+            t.join()
+
+        for c in clients:
+            c.close()
```
ShaneHarvey, line 159: Better to use self.addCleanup to close the clients directly after creating them to make sure they get closed even if the test fails.

ShaneHarvey (changes requested):

bcwarner:
Failed on `test_discovery_and_monitoring.TestUnifiedRediscoverQuicklyAfterStepDown.test_Rediscover_quickly_after_replSetStepDown` in `test-latest-replica_set`

ShaneHarvey (approved):

PR #1037, opened 2022-08-12
PYTHON-3390 Test for encrypted client post-fork

bcwarner:


test/test_encryption.py
```diff
@@ -329,6 +329,20 @@ def test_use_after_close(self):
         with self.assertRaisesRegex(InvalidOperation, "Cannot use MongoClient after close"):
             client.admin.command("ping")
 
+    def test_fork(self):
```
ShaneHarvey, line 332: Add:
```
# Not available for versions of Python without "register_at_fork"
@skipIf(
    not hasattr(os, "register_at_fork"), "register_at_fork not available in this version of Python"
)
```

test/test_encryption.py
```diff
@@ -329,6 +329,20 @@ def test_use_after_close(self):
...
+        opts = AutoEncryptionOpts(KMS_PROVIDERS, "keyvault.datakeys")
+        client = rs_or_single_client(auto_encryption_opts=opts)
+
+        lock_pid = os.fork()
+        if lock_pid == 0:
+            client.admin.command("ping")
+            client.close()
+            os._exit(0)
+        else:
+            self.assertEqual(0, os.waitpid(lock_pid, 0)[1] >> 8)
```
ShaneHarvey, line 342: What's the purpose of ` >> 8`?

bcwarner: I think if I remember correctly, early iterations of the fork feature were failing tests on Windows because `waitpid` shifts exit statuses left 8 bits, but I don't think it's needed now since we skip testing on Windows anyways. I'll remove the shifting from `test_fork` as well.

blink1073 (approved):
LGTM!

ShaneHarvey (approved):

PR #1039, opened 2022-08-17
PYTHON-3403 Skips unit test if eventlent or gevent is imported

bcwarner:


test/test_fork.py
```diff
@@ -32,6 +33,10 @@ def setUpModule():
 @skipIf(
     not hasattr(os, "register_at_fork"), "register_at_fork not available in this version of Python"
 )
+@skipIf(
+    "gevent" in sys.modules.keys() or "eventlet" in sys.modules.keys(),
```
blink1073, line 37: You can use `in sys.modules`, `.keys()` is not needed.

ShaneHarvey: Let's use is_greenthread_patched from test/utils.py instead.

ShaneHarvey (changes requested):

test/test_fork.py
```diff
@@ -32,6 +36,10 @@ def setUpModule():
 @skipIf(
     not hasattr(os, "register_at_fork"), "register_at_fork not available in this version of Python"
 )
+@skipIf(
+    is_greenthread_patched(),
+    "gevent and eventlet do not support POSIX-style forking.",
+)
```
ShaneHarvey, line 42: Should we add this to the encryption fork test too?

PR #1041, opened 2022-08-18
PYTHON-3405/PYTHON-2531 Fix tests for primary step down

ShaneHarvey:
This PR fixes https://jira.mongodb.org/browse/PYTHON-2531 and https://jira.mongodb.org/browse/PYTHON-3405.

PYTHON-2531 was a test bug where we had an infinite loop in ensure_all_connected. 
PYTHON-3405 was a test bug where we were incorrect comparing a ServerDescription with a _Address which always compare unequal.


ShaneHarvey:
Opened https://jira.mongodb.org/browse/PYTHON-3406 for the PyPy 3.8 test failure.

PR #1043, opened 2022-08-19
PYTHON-3408 Changing test_many_threaded to use multiprocessing

bcwarner:


test/test_fork.py
```diff
@@ -106,10 +107,27 @@ def action(client):
...
+                        proc.join(60)
+                        if proc.exitcode is None:
+                            # If it failed, SIGINT to get traceback and wait
+                            # 10s.
+                            os.kill(proc.pid, signal.SIGINT)
+                            proc.join(10)
+                            if proc.exitcode is None:
+                                # If that also failed, SIGKILL and resume after.
+                                os.kill(proc.pid, signal.SIGKILL)
+                            self.runner.fail("deadlock")
```
ShaneHarvey, line 130: Need one last `proc.join(10)` before failing here to avoid zombie process or a resource warning.

test/test_fork.py
```diff
@@ -106,10 +107,27 @@ def action(client):
                     rc = self.clients[i % len(self.clients)]
                     if i % 50 == 0 and self.fork:
                         # Fork
-                        with self.runner.fork() as pid:
-                            if pid == 0:  # Child
-                                for c in self.clients:
-                                    action(c)
+                        def target():
```
ShaneHarvey, line 110: Let's refactor the  `self.runner.fork()` method to use multiprocessing internally. The body of the with statement would then _only_ be run in the parent process:
```python
def target():
    for c in self.clients:
        action(c)
        c.close()  # Client no longer needed.

with self.runner.fork(target=target) as proc:
    self.assertTrue(proc.pid)
```

ShaneHarvey (changes requested):

ShaneHarvey:
Superseded by https://github.com/mongodb/mongo-python-driver/pull/1045

PR #1047, opened 2022-08-24
PYTHON-2722 Improve performance of find/aggregate_raw_batches

blink1073:


pymongo/collection.py
```diff
@@ -2472,11 +2472,12 @@ def aggregate_raw_batches(
             raise InvalidOperation("aggregate_raw_batches does not support auto encryption")
         if comment is not None:
             kwargs["comment"] = comment
+        cursor_class = kwargs.pop("cursor_class", RawBatchCommandCursor)
```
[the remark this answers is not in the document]
blink1073: That won't work, because we need to control the unpacking of the raw response as well before it is consumed in `__send_message`.   Instead we could pass in a custom unpacker for both `aggregate_raw_batches()` and `find_raw_batches()` that handles setting `user_fields` and unpacking the `raw_response` using the bson stream parser.  So all cases `_unpack_response` would check for a custom unpacker and use it instead of its own logic if present.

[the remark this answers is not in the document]
ShaneHarvey: I want to avoid making those warty apis public in any way. Can we instead make a `aggregate_raw_batches(raw_command=True)` cursor more lazily decode the command response without inflating the documents? Perhaps a new lazier RawBSONDocument that only decodes the one field being accessed instead of the entire level of bson?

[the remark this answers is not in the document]
blink1073: Not without having a bson stream parser in the bson package, unless there's a way I couldn't find while attempting to do just that.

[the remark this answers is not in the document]
ShaneHarvey: RawBSONDocument could be a quasi stream parser. What if we changed it to lazily decode an embedded BSON array field. For example, raw_bson._inflate_bson could wrap embeded arrays in a new, and more importantly, private   _RawBSONArray class. Then RawBSONDocument would only inflate the array once a user actually requests access to that specific field.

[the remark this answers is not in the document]
blink1073: We are still accessing `cursor["firstBatch"]` in `__send_message`.  I think if `raw_response` was true, we'd leave "firstBatch"/"nextBack" as raw bytes, otherwise we'd inflate it as we currently do.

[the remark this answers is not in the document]
ShaneHarvey: Yes that's another hurdle the new cursor type needs to jump over. We can't publicly exposing the internal cursor_class as a public api. It's too unstable and will prevent us from refactoring/improving the codebase in the future. The api needs to be something as simple as:
```
for raw_command in aggregate_raw_batches(raw_command=True):
    ....
```

Where PyMongo does all its normal bookkeeping (error checking (ok: 0), $clusterTime parsing, cursor id parsing) and then yields the raw command response so that pymongoarrow can parse the actual documents.

bson/__init__.py
```diff
@@ -1121,12 +1162,30 @@ def _decode_selective(rawdoc: Any, fields: Any, codec_options: Any) -> Mapping[A
...
-                cursor[key] = [stream]
+    if not cursor:
+        return
+    for key in ("firstBatch", "nextBatch"):
+        batch = cursor.get(key)
+        if not batch:
+            continue
+
+        # Extract the raw bytes of each document.
+        data, _ = get_data_and_view(batch.raw)
```
ShaneHarvey, line 1173: Should this be:
```
        _, data = get_data_and_view(batch.raw)
```
That way we use a memory view to avoid making many small byte copies.

blink1073: Done

blink1073: It turns out we needed the bytes so we can use the `index` method.

blink1073: Okay, back to using memoryview directly.

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2816,85 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+
+    if (!_get_buffer(arr, &view)) {
+        goto fail;
+    }
+
+    string = (char*)view.buf;
+    memcpy(&size, string, 4);
+    size = BSON_UINT32_FROM_LE(size);
+
+    if (view.len < size || view.len > BSON_MAX_SIZE) {
```
ShaneHarvey, line 2844: `... || view.len > BSON_MAX_SIZE` -> `... || size > BSON_MAX_SIZE`?

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2816,85 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+        return NULL;
+    }
+
+    buffer = pymongo_buffer_new();
+
+    if (!_get_buffer(arr, &view)) {
+        goto fail;
+    }
+
+    string = (char*)view.buf;
```
ShaneHarvey, line 2840: Is .buf a pointer to the start view's memory or a pointer to the start of the original bytes() the view points to (ie. memoryview.obj in pure python)?

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2816,85 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+        while (string[position] != '\0') {
+            position++;
+        }
+        position++;
+
+        memcpy(&value_length, string + position, 4);
+        value_length = BSON_UINT32_FROM_LE(value_length);
+
+        if (pymongo_buffer_write(buffer, string + position, value_length) == 1) {
+            PyObject* InvalidBSON = _error("InvalidBSON");
```
ShaneHarvey, line 2865: An exception should already be set in this case (it would be a MemoryError because the buffer could not be resized) so we don't need this InvalidBSON logic.

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2816,85 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+    position += 4;
+    while (position < size - 1) {
+        // Just skip the keys.
+        while (string[position] != '\0') {
+            position++;
+        }
+        position++;
+
+        memcpy(&value_length, string + position, 4);
+        value_length = BSON_UINT32_FROM_LE(value_length);
```
ShaneHarvey, line 2862: Do we need to validate value_length? Perhaps we should add a _pymongo_get_object_size helper method?

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2816,85 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+    const char* string;
+    PyObject* arr;
+    PyObject* result = NULL;
+    Py_buffer view = {0};
+
+    if (!PyArg_ParseTuple(args, "O", &arr)) {
+        return NULL;
+    }
+
+    buffer = pymongo_buffer_new();
```
ShaneHarvey, line 2834: We should use pymongo_buffer_save_space(size) to preallocate the max amount of space we need.

bson/raw_bson.py
```diff
@@ -111,7 +126,7 @@ class from the standard library so it can be used like a read-only
         # it refers to this class RawBSONDocument.
         if codec_options is None:
             codec_options = DEFAULT_RAW_BSON_OPTIONS
-        elif codec_options.document_class is not RawBSONDocument:
+        elif codec_options.document_class not in [RawBSONDocument, LazyRawBSONDocument]:
```
ShaneHarvey, line 129: `not issubclass(codec_options.document_class, RawBSONDocument)` would be more future proof.

bson/raw_bson.py
```diff
@@ -153,23 +172,23 @@ def __eq__(self, other: Any) -> bool:
...
-    """
-    # Use SON to preserve ordering of elements.
-    return _raw_to_dict(bson_bytes, 4, len(bson_bytes) - 1, codec_options, SON())
+    @staticmethod
+    def _inflate_bson(bson_bytes: bytes, codec_options: CodecOptions) -> Mapping[Any, Any]:
+        return _inflate_bson(bson_bytes, codec_options, lazy=True)
 
 
 DEFAULT_RAW_BSON_OPTIONS: CodecOptions = DEFAULT.with_options(document_class=RawBSONDocument)
+LAZY_RAW_BSON_OPTIONS: CodecOptions = DEFAULT.with_options(document_class=LazyRawBSONDocument)
```
ShaneHarvey, line 191: Let's make this private as well.

ShaneHarvey (changes requested):

bson/__init__.py
```diff
@@ -517,22 +516,56 @@ def _get_decimal128(
...
+    def raw(self):
+        return memoryview(self.__view)
+
+    @property
+    def value(self):
+        if not self.__inflated:
+            self.__inflated = decode(self.__view, self.__codec_options)
+            if self.__is_array:
+                assert self.__inflated is not None
+                self.__inflated = [a for a in self.__inflated.values()]
```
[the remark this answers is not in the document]
blink1073: Actually, we pass a flag through to treat an array as a stream of documents and do the transformation on the spot.  Then we don't need a LazyValue at all.  Does that sound reasonable?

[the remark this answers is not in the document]
ShaneHarvey: How would we be able to tell the difference between an array and a bytes?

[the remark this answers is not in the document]
blink1073: This would only be used by `OpMsg.raw_response`, to explicitly tell `_get_array` (and the C ext equivalent) to decode differently for a given field name.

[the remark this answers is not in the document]
ShaneHarvey: Oh right. Yeah I think that would work nicely. Then we wouldn't need a LazyValue or _cbson_array_of_documents_to_buffer.

blink1073:
@ShaneHarvey what do you think of the approach taken in [b1f9a41](https://github.com/mongodb/mongo-python-driver/pull/1047/commits/b1f9a4104ac0c8b0ea96bdcfa5b74ab5f111caa9) (not yet implemented in C).  It uses the existing `user_fields` in a way that we were not able to before (because we had to unpack arrays immediately).

bson/raw_bson.py
```diff
@@ -153,23 +172,23 @@ def __eq__(self, other: Any) -> bool:
...
+        return "%s(%r, codec_options=%r)" % (
+            self.__class__.__name__,
+            self.raw,
+            self.__codec_options,
+        )
 
 
-def _inflate_bson(bson_bytes: bytes, codec_options: CodecOptions) -> Mapping[Any, Any]:
-    """Inflates the top level fields of a BSON document.
+class LazyRawBSONDocument(RawBSONDocument):
```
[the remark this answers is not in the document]
blink1073: Done

bson/__init__.py
```diff
@@ -517,22 +516,56 @@ def _get_decimal128(
 }
 
 
+class LazyValue:
```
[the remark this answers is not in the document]
blink1073: Removed entirely

bson/_cbsonmodule.c
blink1073: Done

bson/_cbsonmodule.c
blink1073: This is the same approach we used in `_cbson_bson_to_dict`, as I understand it `.buf` in this case points to the underlying bytes in the `Py_Buffer` structure.

bson/_cbsonmodule.c
blink1073: Removed

bson/_cbsonmodule.c
blink1073: Added validation

bson/_cbsonmodule.c
blink1073: Done

bson/raw_bson.py
blink1073: Done

bson/raw_bson.py
blink1073: Done

bson/_cbsonmodule.c
```diff
@@ -2504,7 +2520,7 @@ static int _element_to_dict(PyObject* self, const char* string,
     }
     position += (unsigned)name_length + 1;
     *value = get_value(self, *name, string, &position, type,
-                       max - position, options);
+                           max - position, options, lazy);
```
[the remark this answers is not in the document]
blink1073: Fixed

bson/__init__.py
```diff
@@ -1119,14 +1160,44 @@ def _decode_selective(rawdoc: Any, fields: Any, codec_options: Any) -> Mapping[A
     return doc
 
 
+def _array_of_documents_to_buffer(view: memoryview) -> bytes:
+    # Extract the raw bytes of each document.
+    position = 0
+    _, end = _get_object_size(view, position, len(view))
+    position += 4
+    buffers: List[bytes] = []
```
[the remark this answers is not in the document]
blink1073: Fixed

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,86 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+    const char* string;
+    PyObject* arr;
+    PyObject* result = NULL;
+    Py_buffer view = {0};
+
+    if (!PyArg_ParseTuple(args, "O", &arr)) {
+        return NULL;
+    }
+
+    buffer = pymongo_buffer_new();
```
ShaneHarvey, line 2826: Need error checking for pymongo_buffer_new.

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,86 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+
+
+    if (!_get_buffer(arr, &view)) {
+        goto fail;
+    }
+
+    string = (char*)view.buf;
+    memcpy(&size, string, 4);
```
ShaneHarvey, line 2834: Missing a length check to guard against a zero length memoryview being passed in.

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,86 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+    size = BSON_UINT32_FROM_LE(size);
+    /* save space for length */
+    if (pymongo_buffer_save_space(buffer, size) == -1) {
+        goto fail;
+    }
+    initial_position = pymongo_buffer_get_position(buffer);
```
ShaneHarvey, line 2840: Can we remove initial_position? Isn't it always 0?

bson/_cbsonmodule.c
ShaneHarvey: This validation is different from the other places we parse obj/arr sizes. Looks like it was accidentally copied from bson string decoding. See decode_all for the correct validation (including the necessary InvalidBSON errors): https://github.com/mongodb/mongo-python-driver/blob/e07081599301fe9e53387f98ccc00d7d2c700eb0/bson/_cbsonmodule.c#L2743-L2773

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,86 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+        }
+        goto done;
+    }
+
+    position += 4;
+    while (position < size - 1) {
+        // Just skip the keys.
+        while (string[position] != '\0') {
+            position++;
+        }
```
ShaneHarvey, line 2856: Let's use strlen here.

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,86 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+
+    if (!PyArg_ParseTuple(args, "O", &arr)) {
+        return NULL;
+    }
+
+    buffer = pymongo_buffer_new();
+
+
+    if (!_get_buffer(arr, &view)) {
+        goto fail;
```
ShaneHarvey, line 2830: There's a bug here. When this line fails we'll mistakenly call PyBuffer_Release(&view).

ShaneHarvey (changes requested):

bson/raw_bson.py
```diff
@@ -111,7 +128,9 @@ class from the standard library so it can be used like a read-only
         # it refers to this class RawBSONDocument.
         if codec_options is None:
             codec_options = DEFAULT_RAW_BSON_OPTIONS
-        elif codec_options.document_class is not RawBSONDocument:
+        elif codec_options.document_class is not RawBSONDocument and not issubclass(
+            codec_options.document_class, RawBSONDocument
+        ):
```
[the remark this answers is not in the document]
blink1073: Done

pymongo/message.py
```diff
@@ -1379,7 +1384,7 @@ def raw_response(self, cursor_id=None, user_fields={}):  # noqa: B006
         user_fields is used to determine which fields must not be decoded
         """
         inflated_response = _decode_selective(
-            RawBSONDocument(self.payload_document), user_fields, DEFAULT_RAW_BSON_OPTIONS
+            RawBSONDocument(self.payload_document), user_fields, _RAW_ARRAY_BSON_OPTIONS
```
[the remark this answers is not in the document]
blink1073: It doesn't seem so:

```python
event.reply
{'cursor': {'nextBatch': [b'$\x00\x00\x00\x07_id\x00c$\xd2\xbf\x83g1s\xb4\xe3\xd1\xbd\x10x\x00e\x00\x00\x00\x10y\x00e\x00\x00\x00\x00$\x00\x00\x00\x07_id
```

bson/_cbsonmodule.c
```diff
@@ -1712,6 +1712,14 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
...
                 goto invalid;
             }
+
+            if (raw_array != 0) {
+                // Treat it as a binary buffer.
+                value = PyBytes_FromStringAndSize(buffer + *position, size);
+                *position += size;
+                break;
+            }
+
```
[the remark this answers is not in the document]
blink1073: Fixed

bson/_cbsonmodule.c
blink1073: Fixed

bson/_cbsonmodule.c
blink1073: isn't that what `total_size < BSON_MIN_SIZE` is checking?

bson/_cbsonmodule.c
blink1073: No, it is not always 0

bson/_cbsonmodule.c
blink1073: Done

bson/_cbsonmodule.c
blink1073: Fixed

bson/_cbsonmodule.c
blink1073: Done

bson/_cbsonmodule.c
blink1073: Removed

pymongo/message.py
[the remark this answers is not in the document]
ShaneHarvey: Great!

bson/_cbsonmodule.c
ShaneHarvey: The same bug now exists when pymongo_buffer_new fails.

bson/_cbsonmodule.c
ShaneHarvey: I don't see a `total_size < BSON_MIN_SIZE` check. We can't call memcpy without first validating that there's at least 4 bytes in the top-level buffer.

bson/_cbsonmodule.c
ShaneHarvey: The buffer was just created with pymongo_buffer_new() so when would it not be 0?

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,103 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+        }
+        goto done;
+    }
+
+    position += 4;
+    while (position < size - 1) {
+        // Just skip the keys.
+        position = position + strlen(string + position) + 1;
+
+        memcpy(&value_length, string + position, 4);
```
ShaneHarvey, line 2859: Need to bounds check position before calling memcpy here or else we could read out of bounds memory.

ShaneHarvey (changes requested):

bson/_cbsonmodule.c
blink1073: Fixed

bson/_cbsonmodule.c
blink1073: Done

bson/_cbsonmodule.c
blink1073: It was down further, I moved it up

bson/_cbsonmodule.c
blink1073: I added a print statement and it was 4036

ShaneHarvey: Oh I get it, pymongo_buffer_save_space resizes the buffer and advances the buffer's position. We have to reset the position to 0 directly after the pymongo_buffer_save_space call.

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,112 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+        return NULL;
+    }
+
+    if (!_get_buffer(arr, &view)) {
+        return NULL;
+    }
+
+    buffer = pymongo_buffer_new();
+    if (!buffer) {
+        return NULL;
```
ShaneHarvey, line 2831: need to call PyBuffer_Release(&view); here.

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,112 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+    string = (char*)view.buf;
+
+    if (view.len < BSON_MIN_SIZE) {
+        PyObject* InvalidBSON = _error("InvalidBSON");
+        if (InvalidBSON) {
+            PyErr_SetString(InvalidBSON,
+                            "not enough data for a BSON document");
+            Py_DECREF(InvalidBSON);
+        }
+        goto done;;
```
ShaneHarvey, line 2843: extra ;

bson/_cbsonmodule.c
blink1073: Done

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,111 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+    size = BSON_UINT32_FROM_LE(size);
+    /* save space for length */
+    if (pymongo_buffer_save_space(buffer, size) == -1) {
+        goto fail;
+    }
+    pymongo_buffer_update_position(buffer, 0);
+
+    position += 4;
+    while (position < size - 1) {
+        // Just skip the keys.
```
ShaneHarvey, line 2856: I think we also need to validate that each key's type is a BSON document.

bson/_cbsonmodule.c
```diff
@@ -2797,6 +2807,111 @@ static PyObject* _cbson_decode_all(PyObject* self, PyObject* args) {
...
+        position = position + strlen(string + position) + 1;
+        if (position >= size + 1) {
```
ShaneHarvey, line 2858: Should this be `if (position >= size  || (size - position) < BSON_MIN_SIZE)`? The BSON_MIN_SIZE is needed to make sure there's enough room for the memcpy.

bson/_cbsonmodule.c
blink1073: Done

bson/_cbsonmodule.c
blink1073: Fixed

bson/_cbsonmodule.c
blink1073: Done

ShaneHarvey (approved):
LGTM!

PR #1049, opened 2022-08-29
PYTHON-3411 Stop testing MongoDB 6.0 on Amazon1 2018

blink1073:


blink1073:
Failures are related to https://jira.mongodb.org/browse/PYTHON-3406

.evergreen/config.yml
```diff
@@ -2505,7 +2498,7 @@ buildvariants:
 
 - matrix_name: "tests-python-version-supports-openssl-102-test-ssl"
   matrix_spec:
-    platform: awslinux
+    platform: ubuntu-20.04
```
ShaneHarvey, line 2501: I think we want to keep this so that we still test OpenSSL 1.0.2
```
/opt/python/3.7/bin/python3 -c 'import ssl;print(ssl.OPENSSL_VERSION)'
OpenSSL 1.0.2k-fips  26 Jan 2017
```

ShaneHarvey (changes requested):

.evergreen/config.yml
blink1073: Done

ShaneHarvey (approved):

PR #1050, opened 2022-08-30
PYTHON-3406 Reinstate warning and docs that PyMongo is not fork safe

ShaneHarvey:


doc/changelog.rst
```diff
@@ -13,8 +13,10 @@ PyMongo 4.3 brings a number of improvements including:
   :class:`bson.codec_options.DatetimeConversion`, and
   :class:`bson.codec_options.CodecOptions`'s ``datetime_conversion``
   parameter for more details (`PYTHON-1824`_).
-- Added support for using a :class:`~pymongo.mongo_client.MongoClient` after
-  an :py:func:`os.fork` (`PYTHON-2484`_).
+- PyMongo now resets its locks and other shared state in the child process
```
behackett, line 16: I think you should say that libraries that PyMongo depends on (e.g. OpenSSL) are not fork safe, which makes PyMongo not fork safe. This text implies that a future version of PyMongo can be made fork safe with some amount of further effort.

ShaneHarvey: Done.

blink1073 (approved):
LGTM, thanks!

PR #1051, opened 2022-09-08
PYTHON-3424 PyMongo Universal Wheels Are Improperly Compiled

blink1073:


.evergreen/build-mac.sh
```diff
@@ -12,6 +12,14 @@ for VERSION in 3.7 3.8 3.9 3.10; do
     PYTHON=/Library/Frameworks/Python.framework/Versions/$VERSION/bin/python3
     rm -rf build
 
+    # Set the arch flags appropriately for the target platform.
+    PLATFORM=$($PYTHON -c "import sysconfig;print(sysconfig.get_platform())")
+    if [[ $PLATFORM =~ "universal2" ]]; then
+        ARCHFLAGS="-arch arm64 -arch x86_64"
+    else
+        ARCHFLAGS="-arch x86_64"
+    fi
```
ShaneHarvey, line 21: Is there any point when we can remove this hack? It seems like something that should happen automatically.

blink1073: We don't actually need this hack, I just haven't pushed yet because I'm using direct patch builds.

blink1073:
The produced macos wheels are:

```bash
-rw-r--r--@  1 steve.silvester  staff  407947 Sep  8 10:22 pymongo-4.2.1.dev0-cp37-cp37m-macosx_10_6_intel.whl
-rw-r--r--@  1 steve.silvester  staff  378895 Sep  8 10:23 pymongo-4.2.1.dev0-cp38-cp38-macosx_10_9_x86_64.whl
-rw-r--r--@  1 steve.silvester  staff  410370 Sep  8 11:28 pymongo-4.2.1.dev0-cp39-cp39-macosx_10_9_universal2.whl
-rw-r--r--@  1 steve.silvester  staff  410390 Sep  8 11:28 pymongo-4.2.1.dev0-cp310-cp310-macosx_10_9_universal2.whl
```

I verified that `pymongo-4.2.1.dev0-cp310-cp310-macosx_10_9_universal2.whl`  as installed with an arm64 version of Python has working c extensions.

tools/fail_if_no_c.py
```diff
@@ -26,3 +29,13 @@
...
     sys.exit("could not load C extensions")
+
+if sys.platform == "darwin":
+    wheel_file = sys.argv[-1]
+    if ".whl" in wheel_file and "universal2" in wheel_file:
+        parent_dir = os.path.dirname(pymongo.__path__[0])
+        for so_file in glob.glob(f"{parent_dir}/**/*.so"):
+            print(f"Checking {so_file}...")
+            output = subprocess.check_output(["file", so_file])
+            if "arm64" not in output.decode("utf-8"):
```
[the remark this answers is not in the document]
blink1073: Done

tools/fail_if_no_c.py
```diff
@@ -26,3 +29,13 @@
 
 if not pymongo.has_c() or not bson.has_c():
     sys.exit("could not load C extensions")
+
+if sys.platform == "darwin":
+    wheel_file = sys.argv[-1]
+    if ".whl" in wheel_file and "universal2" in wheel_file:
+        parent_dir = os.path.dirname(pymongo.__path__[0])
+        for so_file in glob.glob(f"{parent_dir}/**/*.so"):
```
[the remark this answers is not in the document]
blink1073: I was trying to save unpacking the wheel since it was already done by installing.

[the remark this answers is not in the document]
ShaneHarvey: Then can we pass a flag to fail_if_no_c.py instead of the wheel file name? Maybe `--ensure-universal2`.

tools/fail_if_no_c.py
```diff
@@ -26,3 +29,15 @@
...
+            print(f"Checking {so_file}...")
+            output = subprocess.check_output(["file", so_file])
+            if "arm64" not in output.decode("utf-8"):
+                sys.exit("Universal wheel was not compiled with arm64 support")
+            if "x86_64" not in output.decode("utf-8"):
+                sys.exit("Universal wheel was not compiled with arm64 support")
```
ShaneHarvey, line 43: arm64 -> x86_64

blink1073: D'oh, fixed

tools/fail_if_no_c.py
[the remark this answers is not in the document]
ShaneHarvey: I'd also be fine with a comment that explains the assumptions here.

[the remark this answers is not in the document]
blink1073: Fixed

.evergreen/build-mac.sh
```diff
@@ -8,22 +8,25 @@ rm -rf validdist
...
 
-    # Test that each wheel is installable.
-    for release in dist/*; do
-        testinstall $PYTHON $release
-        mv $release validdist/
-    done
+# Test that each wheel is installable.
+for release in dist/*; do
+    if [[ "$release" = "*universal2*'" ]]; then
+        testinstall $PYTHON --ensure-universal2
```
ShaneHarvey, line 25: testinstall doesn't accept a `--ensure-universal2` param. I wonder if it would make more sense to set an ENV var in the release-mac-1100 tasks rather then try to derive if we should be creating a universal2 wheel after the fact. Then we could use that ENV var in testinstall.

ShaneHarvey (changes requested):

.evergreen/build-mac.sh
blink1073: Done

blink1073: I verified that the wheels are checked in the build.

.evergreen/build-mac.sh
```diff
@@ -8,22 +8,24 @@ rm -rf validdist
...
 
-    # Test that each wheel is installable.
-    for release in dist/*; do
-        testinstall $PYTHON $release
-        mv $release validdist/
-    done
+# Test that each wheel is installable.
+for release in dist/*; do
+    if [[ "$release" =~ .*universal2.* ]]; then
+        export ENSURE_UNIVERSAL2="1"
```
ShaneHarvey, line 25: Per my last comment I think we should move ENSURE_UNIVERSAL2=1 to release-mac-1100, that way if the job stops producing universal2 wheels for some reason, we'll know:
```
    - name: "release-mac-1100"
      tags: ["release_tag"]
      run_on: macos-1100
      commands:
        - func: "build release"
          vars:
              VERSION: "3.10"
              ENSURE_UNIVERSAL2: "1"
        - func: "build release"
          vars:
              VERSION: "3.9"
              ENSURE_UNIVERSAL2: "1"
        - func: "upload release"
 ```

blink1073: Done

ShaneHarvey (approved):

PR #1055, opened 2022-09-14
PYTHON-3423 Make dnspython a required dependency

juliusgeo:


pymongo/uri_parser.py
```diff
@@ -469,8 +469,8 @@ def parse_uri(
             raise ConfigurationError(
                 'The "dnspython" module must be '
                 "installed to use mongodb+srv:// URIs. "
-                "To fix this error install pymongo with the srv extra:\n "
-                '%s -m pip install "pymongo[srv]"' % (python_path)
+                "To fix this error install pymongo again:\n "
+                '%s -m pip install "pymongo"' % (python_path)
```
blink1073, line 473: Maybe say `pymongo>=4.3`?

setup.py
```diff
@@ -281,7 +281,6 @@ def build_extension(self, ext):
     "snappy": ["python-snappy"],
     "zstd": ["zstandard"],
     "aws": ["pymongo-auth-aws<2.0.0"],
-    "srv": ["dnspython>=1.16.0,<3.0.0"],
```
blink1073, line 284: nb: if the user tries to use `pymongo[srv]`, they should see a warning that `srv` is not available but it should still work, from my experience with setuptools extras.

blink1073: e.g.:

```bash
Requirement already satisfied: pymongo[nonsense] in ./miniconda/lib/python3.10/site-packages (4.2.0)
WARNING: pymongo 4.2.0 does not provide the extra 'nonsense'
```

behackett: Can we add this back in as "srv": [""] to suppress any warnings?

pymongo/uri_parser.py
```diff
@@ -469,8 +469,8 @@ def parse_uri(
             raise ConfigurationError(
                 'The "dnspython" module must be '
                 "installed to use mongodb+srv:// URIs. "
-                "To fix this error install pymongo with the srv extra:\n "
-                '%s -m pip install "pymongo[srv]"' % (python_path)
```
ShaneHarvey, line 473: I guess it doesn't hurt to keep this error just in case but is this even possible now that dnspython is a required dependency?

ShaneHarvey (changes requested):
Nice work, can you also add this to the changelog?

blink1073 (approved):
LGTM!

setup.py
ShaneHarvey: According to https://setuptools.pypa.io/en/latest/userguide/dependency_management.html?highlight=extras#optional-dependencies this should be an empty list: `"srv": [],`

> Best practice: if a project ends up no longer needing any other packages to support a feature, it should keep an empty requirements list for that feature in its extras_require argument, so that packages depending on that feature don’t break (due to an invalid feature name).

doc/changelog.rst
```diff
@@ -4,6 +4,8 @@ Changelog
 Changes in Version 4.3
 ----------------------
 
+PyMongo 4.3 introduces the first required dependency: `dnspython <https://pypi.python.org/pypi/dnspython>`_.
```
ShaneHarvey, line 7: What about:
```
`dnspython <https://pypi.python.org/pypi/dnspython>`_ is now a required
dependency. This change makes PyMongo easier to install for use with "mongodb+srv://"
connection strings and `MongoDB Atlas <https://www.mongodb.com/cloud>`_.
```

ShaneHarvey (changes requested):

ShaneHarvey (approved):

ShaneHarvey (approved):

PR #1057, opened 2022-09-15
PYTHON-3420 Update ChangeStreamEvent type definition to include clusterTime

blink1073:


ShaneHarvey (changes requested):
We only want to pull in the change-streams-clusterTime.json test. The other files are for PYTHON-3232.

ShaneHarvey (approved):

PR #1058, opened 2022-09-16
PYTHON-3419 Use memoryview to avoid byte copies when decoding RawBSONDocument

blink1073:


blink1073:
Current status:  `6.3 MiB` increment vs. `4.3 MiB` increment in `master` in the memory check script.

bson/_cbsonmodule.c
```diff
@@ -2461,10 +2472,12 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
  *
  * Returns the position of the next element in the document, or -1 on error.
  */
-static int _element_to_dict(PyObject* self, const char* string,
+static int _element_to_dict(PyObject* self, PyObject* bson,
                             unsigned position, unsigned max,
                             const codec_options_t* options,
                             PyObject** name, PyObject** value) {
+    const Py_buffer* buffer = PyMemoryView_GET_BUFFER(bson);
```
blink1073, line 2479: According to the docs this should only be a pointer to the memory

blink1073:
Interesting, testing in Ubuntu gives me more debug information with memray:


image

blink1073:
There is 3MB allocated in that call.

blink1073:
Yeah, the memoryview itself is consuming memory, even though I created it with a slice.  If I throw out the memoryview in the constructor we don't see the memory growth.

blink1073:
Closing for now, we could not find a way to avoid the memory duplication.

PR #1062, opened 2022-09-26
PYTHON-3232 Improved change stream event visibility for C2C Replication

juliusgeo:


pymongo/collection.py
```diff
@@ -2495,6 +2495,7 @@ def watch(
         start_after: Optional[Mapping[str, Any]] = None,
         comment: Optional[Any] = None,
         full_document_before_change: Optional[str] = None,
+        show_expanded_events: Optional[bool] = None,
```
blink1073, line 2498: Needs docs and versionadded.  Looks like we missed `full_document_before_change` in the docstring earlier as well.

pymongo/database.py
```diff
@@ -547,6 +547,7 @@ def watch(
         start_after: Optional[Mapping[str, Any]] = None,
         comment: Optional[Any] = None,
         full_document_before_change: Optional[str] = None,
+        show_expanded_events: Optional[bool] = None,
```
blink1073, line 550: Same comment as collection method

pymongo/mongo_client.py
```diff
@@ -895,6 +895,7 @@ def watch(
         start_after: Optional[Mapping[str, Any]] = None,
         comment: Optional[Any] = None,
         full_document_before_change: Optional[str] = None,
+        show_expanded_events: Optional[bool] = None,
```
blink1073, line 898: Here as well

blink1073 (changes requested):

pymongo/aggregation.py
```diff
@@ -40,6 +40,7 @@ def __init__(
         user_fields=None,
         result_processor=None,
         comment=None,
+        show_expanded_events=None,
```
ShaneHarvey, line 43: Since you're implementing PYTHON-3232 in this PR can you rename the title and close PYTHON-3373 as a duplicate of PYTHON-3232?

ShaneHarvey (changes requested):

blink1073:
Can you merge from master to pick up https://github.com/mongodb/mongo-python-driver/pull/1063?

ShaneHarvey (changes requested):
Let's add a changelog entry for show_expanded_events too.

pymongo/change_stream.py
```diff
@@ -175,6 +177,10 @@ def _change_stream_options(self):
 
         if self._start_at_operation_time is not None:
             options["startAtOperationTime"] = self._start_at_operation_time
+
+        if self._show_expanded_events is True:
```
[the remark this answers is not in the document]
ShaneHarvey: Yeah that's the point. `if x` works with anything truthy while `if x is True` doesn't. It's more verbose and less flexible.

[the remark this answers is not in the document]
ShaneHarvey: See
```
Don’t compare boolean values to True or False using ==:
# Correct:
if greeting:
# Wrong:
if greeting == True:
Worse:

# Wrong:
if greeting is True:
```
https://peps.python.org/pep-0008/

blink1073 (approved):
LGTM!

ShaneHarvey (approved):

PR #1072, opened 2022-10-10
PYTHON-3453 Test failure - Enterprise Auth Windows 64 Python 3.7

blink1073:


.evergreen/config.yml
```diff
@@ -490,7 +490,6 @@ functions:
     - command: shell.exec
       type: test
       params:
-        silent: true
         working_dir: "src"
         script: |
           # DO NOT ECHO WITH XTRACE (which PREPARE_SHELL does)
```
ShaneHarvey, line 495: Could you silently load the SASL etc. vars like we do for the "run atlas tests" below? This would avoid the concerns about xtrace and leaking creds.

ShaneHarvey (changes requested):

.evergreen/config.yml
blink1073: Done

PR #1076, opened 2022-10-13
PYTHON-3453 Fix handling of enterprise auth vars

blink1073:


ShaneHarvey (changes requested):
We have to use single quotes here to avoid shell expansion.

blink1073:
> We have to use single quotes here to avoid shell expansion.

Done

ShaneHarvey (approved):

PR #1081, opened 2022-10-18
PYTHON-3454 Specifying a generic type for a collection does not correctly enforce type safety when inserting data

juliusgeo:


test/test_mypy.py
```diff
@@ -18,13 +18,25 @@
 import os
 import tempfile
 import unittest
-from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List, Optional
 
 try:
-    from typing import TypedDict  # type: ignore[attr-defined]
+    from typing_extensions import NotRequired, TypedDict
```
ShaneHarvey, line 24: Doesn't this mean these tests will be skipped even when TypedDict exists because typing_extensions is not installed in CI?

test/test_mypy.py
```diff
@@ -304,6 +324,62 @@ def test_typeddict_document_type(self) -> None:
...
+    def test_typeddict_document_type_insertion(self) -> None:
+        client: MongoClient[Movie] = MongoClient()
+        coll: Collection[Movie] = client.test.test
+        mov = Movie(_id=ObjectId(), name="THX-1138", year=1971)
+        for meth, arg in [
+            (coll.insert_many, [mov]),
+            (coll.insert_one, mov),
+            (coll.bulk_write, [InsertOne(mov)]),
+        ]:
+            meth(arg)  # type:ignore[operator]
```
ShaneHarvey, line 337: How can we put a type ignore here? insert_many/insert_one/bulk_write working without raising a mypy error is the main purpose of this bug fix. I think we're loosing sight of what's important here because the PR is muddled by tests for find_one/_id/NotRequired. Can you open a new ticket for the NotRequired stuff? To fix the bug in PYTHON-3454 I believe we only need to mention schema validation should be configured on the server and add a test like this to confirm the bug is fixed:
```python
    def test_typeddict_document_type_insertion(self) -> None:
        coll: Collection[Movie] = self.db.test
        mov = Movie(name="THX-1138", year=1971)
        coll.insert_one(mov)
        coll.insert_many([mov])
        coll.bulk_write([InsertOne(mov)])
```

ShaneHarvey (changes requested):

test/test_mypy.py
ShaneHarvey: Let's remove everything about _id, NotRequired, typing_extensions, and all the other tests besides the one above. All that is extra stuff that's only tangentially related to the issue. We already have tests that find_one correctly yields the collection generic type: https://github.com/mongodb/mongo-python-driver/blob/4.3.2/test/test_mypy.py#L299-L305

ShaneHarvey: There are two complaints in [PYTHON-3454](https://jira.mongodb.org/browse/PYTHON-3454):
1) `collection.insert_one(Person({"name": "Harold", "age": 15}))` incorrectly raises a mypy error. This is fixed by using _DocumentType and tested by `test_typeddict_document_type_insertion` above.
2) `collection.insert_one({"name": 17, "age": "this isn't a number", "gender": 6})` does not raise a mypy error when it should. This is also fixed by using  _DocumentType and should also be tested (if we aren't doing it already).

This is why I see the _id, NotRequired, typing_extensions, and find test as all unrelated. They are independent from the 2 issues reported by the user.

ShaneHarvey: > Perhaps I am misunderstanding, but I don't think that the user wanted to be able to insert a dict and have that be type-checked.

They do want this. In fact that's all a TypeDict is, it's a regular dict that's been type checked. So both of these inserts are completely identical and should have the same behavior:
```python
collection: Collection[Person] = client.t.t
collection.insert_one(Person({"name": "Harold", "age": 15})) 
collection.insert_one({"name": "Harold", "age": 15}) 
```
And both of these line should also cause the same mypy error:
```python
collection: Collection[Person] = client.t.t
collection.insert_one(Person({"name": "Harold", "age": "WRONG TYPE"})) 
collection.insert_one({"name": "Harold", "age": "WRONG TYPE"}) 
```
Both of these issues will be fixed by this PR. 

I agree that the subtleties of _id + TypedDict are worth documenting but I would like you to open a new jira ticket for it that we can prioritize separately. It's a different issue from PYTHON-3454.

blink1073: In the original example the `TypeDict` did not have an explicit `_id` field.

blink1073: Not including the `_id` is still a valid choice, if it is not relevant to your application.

blink1073 (approved):
LGTM!

test/test_mypy.py
```diff
@@ -304,6 +327,28 @@ def test_typeddict_document_type(self) -> None:
...
+        coll.insert_many([mov])  # type:ignore[list-item]
+        coll.insert_many([movie])
+        bad_mov = {"name": "THX-1138", "year": "WRONG TYPE"}
+        bad_movie = Movie(name="THX-1138", year="WRONG TYPE")  # type:ignore[typeddict-item]
+        coll.insert_one(bad_mov)  # type:ignore[arg-type]
+        coll.insert_one({"name": "THX-1138", "year": "WRONG TYPE"})  # type:ignore[typeddict-item]
+        coll.insert_one(bad_movie)
+        coll.insert_many([bad_mov])  # type:ignore[list-item]
+        coll.insert_many(
+            [{"name": "THX-1138", "year": "WRONG TYPE"}]
```
blink1073, line 348: The auto-formatter got us here, you need to move the ignore comment

blink1073:
There's also some unused imports causing the lint to fail.

blink1073 (approved):

pymongo/operations.py
```diff
@@ -19,15 +19,15 @@
...
+from pymongo.typings import _CollationIn, _DocumentIn, _DocumentType, _Pipeline
 
 
 class InsertOne(object):
     """Represents an insert_one operation."""
 
     __slots__ = ("_doc",)
 
-    def __init__(self, document: _DocumentIn) -> None:
+    def __init__(self, document: Union[_DocumentIn, _DocumentType]) -> None:
```
ShaneHarvey, line 30: Can we defer this change to PYTHON-3493?

doc/examples/type_hints.rst
```diff
@@ -92,11 +92,13 @@ Note that when using :class:`~bson.son.SON`, the key and value types must be giv
...
 
-You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-defined schema for the data in a :class:`~pymongo.collection.Collection`:
+You can use :py:class:`~typing_extensions.TypedDict` (Python 3.8+) when using a well-defined schema for the data in a
+:class:`~pymongo.collection.Collection`. Note that all `schema validation`_ for inserts and updates is done on the server.
+These methods automatically add an "_id" field.
 
 .. doctest::
 
-  >>> from typing import TypedDict
+  >>> from typing_extensions import TypedDict
```
ShaneHarvey, line 101: Let's remove typing_extensions.

.github/workflows/test-python.yml
```diff
@@ -59,14 +59,15 @@ jobs:
           cache-dependency-path: 'setup.py'
       - name: Install dependencies
         run: |
-          python -m pip install -U pip mypy
+          python -m pip install -U pip mypy typing_extensions
```
ShaneHarvey, line 62: Let's remove typing_extensions.

test/test_mypy.py
```diff
@@ -14,22 +14,21 @@
...
 import tempfile
 import unittest
 from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List
 
 try:
-    from typing import TypedDict  # type: ignore[attr-defined]
-
     # Not available in Python 3.7
+    from typing_extensions import TypedDict
+
```
ShaneHarvey, line 25: Let's revert these changes and go back to using typing.

ShaneHarvey (changes requested):

test/test_mypy.py
blink1073: We will need to keep using `typing_extensions` in this file, or else split off a different file to run `mypy` on Python 3.7.  The reason it was working before is that we were inadvertently ignoring errors when running this file.

test/test_mypy.py
```diff
@@ -14,22 +14,17 @@
 
 """Test that each file in mypy_fails/ actually fails mypy, and test some
 sample client code that uses PyMongo typings."""
-
 import os
 import tempfile
 import unittest
 from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List
 
-try:
```
ShaneHarvey, line 23: We need to keep this in a try/except so that `python test/test_mypy.py` works even when typing_extensions is not installed.

ShaneHarvey (changes requested):

test/test_mypy.py
ShaneHarvey: > The reason it was working before is that we were inadvertently ignoring errors when running this file.

Can you explain? How was this fixed?

blink1073: We added a new invocation in GitHub Actions in this PR to run this file without disabling several error checks.

ShaneHarvey: I see thanks.

ShaneHarvey (approved):

PR #1095, opened 2022-10-25
PYTHON-2993 Deprecate MongoClient read only config option helpers and add client.options

blink1073:


pymongo/mongo_client.py
```diff
@@ -1142,22 +1146,33 @@ def max_pool_size(self):
         ``waitQueueTimeoutMS`` is set, a blocked operation will raise
         :exc:`~pymongo.errors.ConnectionFailure` after a timeout.
         By default ``waitQueueTimeoutMS`` is not set.
+
+        .. versionchanged:: 3.13
+           Deprecated.  Use ``client.options.pool_options.max_pool_size`` instead.
```
ShaneHarvey, line 1151: It would be nice to add the `Use ``client.options.pool_options.max_pool_size`` instead` to the warnings too.

doc/api/pymongo/mongo_client.rst
```diff
@@ -14,27 +14,22 @@
 
          Raises :class:`~pymongo.errors.InvalidName` if an invalid database name is used.
 
-      .. autoattribute:: event_listeners
```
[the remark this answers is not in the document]
blink1073: Was trying to reduce visibility, will move to the bottom

pymongo/mongo_client.py
blink1073: Updated

ShaneHarvey (approved):

PR #1096, opened 2022-10-25
PYTHON-2988 Deprecate MongoClient max_bson_size/max_message_size/max_write_batch_size

blink1073:


test/test_client.py
```diff
@@ -614,12 +614,10 @@ def test_init_disconnected(self):
 
         c = rs_or_single_client(connect=False)
         self.assertEqual(c.codec_options, CodecOptions())
-        self.assertIsInstance(c.max_bson_size, int)
```
ShaneHarvey, line 617: We should keep these tests. `c.max_bson_size` et al. should still work correctly even if it's deprecated.

ShaneHarvey (changes requested):

test/test_client.py
blink1073: Reverted

ShaneHarvey (approved):

PR #1097, opened 2022-10-26
PYTHON-2945 Raise Warning when username/password includes a plus sign (+)

blink1073:


pymongo/uri_parser.py
```diff
@@ -67,6 +67,8 @@ def parse_userinfo(userinfo):
     # No password is expected with GSSAPI authentication.
     if not user:
         raise InvalidURI("The empty string is not valid username.")
+    if '+' in user or '+' in passwd:
+        warnings.warn("Starting in 4.0 PyMongo will no longer decode plus signs in user names or passwords to spaces", stacklevel=2)
```
ShaneHarvey, line 71: "user names" -> "usernames"

Can we including something like this in the warning: "use quote() to escape the username/password instead of quote_plus()"

doc/examples/authentication.rst
```diff
@@ -11,17 +11,17 @@ Percent-Escaping Username and Password
...
 Python 2, to be used in a MongoDB URI. For example, in Python 3::
 
   >>> from pymongo import MongoClient
   >>> import urllib.parse
-  >>> username = urllib.parse.quote_plus('user')
+  >>> username = urllib.parse.quote('user')
   >>> username
   'user'
-  >>> password = urllib.parse.quote_plus('pass/word')
+  >>> password = urllib.parse.quote('pass*word')
```
ShaneHarvey, line 22: perhaps it would be worth throwing a plus and a space in the password too?

doc/changelog.rst
```diff
@@ -28,6 +28,9 @@ Deprecations
 - Deprecated :meth:`~pymongo.collection.Collection.map_reduce` and
   :meth:`~pymongo.collection.Collection.inline_map_reduce`.
   Use :meth:`~pymongo.collection.Collection.aggregate` instead.
+- Deprecated the use of a plus sign '+' in user name or password.
```
ShaneHarvey, line 31: "in the username and password when parsing MongoDB URIs."

doc/changelog.rst
```diff
@@ -28,6 +28,9 @@ Deprecations
...
+  Starting in 4.0 PyMongo will no longer decode plus signs in usernames or passwords to spaces.  Users are recommended to use :meth:`~urllib.parse.quote`
```
ShaneHarvey, line 32: Line length.

pymongo/uri_parser.py
ShaneHarvey: Also I'm curious about the stacklevel. What does the Warning look like for MongoClient(uri) and for parse_uri(uri)?

ShaneHarvey (changes requested):

PR #1104, opened 2022-11-01
PYTHON-3494 Improve Documentation Surrounding Type-Checking "_id"

juliusgeo:


doc/examples/type_hints.rst
```diff
@@ -94,24 +94,33 @@ Typed Collection
 
 You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-defined schema for the data in a
 :class:`~pymongo.collection.Collection`. Note that all `schema validation`_ for inserts and updates is done on the server.
-These methods automatically add an "_id" field.
+These methods automatically add an "_id" field. In the example below the "_id" field is
+marked by the :py:class:`~typing_extensions.NotRequired` notation to allow it to be accessed when reading from
+``result``. If it is simply not included in the definition, then it will be automatically added, but it will raise a
+type-checking error if you attempt to access it. Another option would be to generate the "_id" field yourself, and make
+it a required field. This would give the expected behavior, but would then also prevent you from relying on PyMongo to
+insert the "_id" field.
```
blink1073, line 102: We should document that `typing_extensions` has to be installed for `NotRequired` unless using Python 3.11+.

test/test_mypy.py
```diff
@@ -324,6 +336,33 @@ def test_typeddict_document_type_insertion(self) -> None:
         )
         coll.insert_many([bad_movie])
 
+    @only_type_check
+    def test_typeddict_explicit_document_type(self) -> None:
+        out = MovieWithId(_id=ObjectId(), name="THX-1138", year=1971)
```
blink1073, line 341: Shouldn't these be find operation results?

blink1073: Okay, I think instead we should have a single test that uses `NotRequired` and actually does execute a `find`

test/test_mypy.py
```diff
@@ -324,6 +336,41 @@ def test_typeddict_document_type_insertion(self) -> None:
...
+        assert out["foo"]  # type:ignore[typeddict-item]
+        # This should fail because _id is not included in our TypedDict definition.
+        assert type(out["_id"]) == ObjectId  # type:ignore[typeddict-item]
+
+    def test_typeddict_find_notrequired(self):
+        client: MongoClient[ImplicitMovie] = rs_or_single_client()
+        coll = client.test.test
+        coll.insert_one(ImplicitMovie(name="THX-1138", year=1971))
+        out = coll.find_one({})
+        if out is not None:
```
blink1073, line 371: This should assert out is not none instead to make sure the next line is run

test/test_mypy.py
blink1073: I don't think we need these tests since they aren't PyMongo-specific

blink1073 (changes requested):

test/test_mypy.py
blink1073: Fair enough

blink1073 (approved):
LGTM!

doc/examples/type_hints.rst
```diff
@@ -94,24 +94,34 @@ Typed Collection
 
 You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-defined schema for the data in a
 :class:`~pymongo.collection.Collection`. Note that all `schema validation`_ for inserts and updates is done on the server.
-These methods automatically add an "_id" field.
+These methods automatically add an "_id" field. In the example below the "_id" field is
+marked by the :py:class:`~typing_extensions.NotRequired` notation to allow it to be accessed when reading from
```
ShaneHarvey, line 98: I think the discussion of _id and NotRequired complicates what was a simple example and makes this section harder to read. What do you think about leaving this section as is and creating a new section to discuss modeling documents with TypedDict? There we could discuss these more advanced topics without complicating the simple intro example.

ShaneHarvey (changes requested):

.evergreen/run-tests.sh
```diff
@@ -195,6 +195,7 @@ $PYTHON -c 'import sys; print(sys.version)'
 
 # Run the tests with coverage if requested and coverage is installed.
 # Only cover CPython. PyPy reports suspiciously low coverage.
+$PYTHON -m pip install typing_extensions
```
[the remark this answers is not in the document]
blink1073: We could skip the test if `typing_extensions` is not installed with an `ImportWarning` guard.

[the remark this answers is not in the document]
blink1073: We could install it after the tests are run and then run `python setup.py test -s test.test_mypy`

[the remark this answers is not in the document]
ShaneHarvey: We shouldn't be testing typing_extensions. I would be fine with a test that uses NotRequired and only runs on 3.11+

doc/examples/type_hints.rst
```diff
@@ -92,28 +92,96 @@ Note that when using :class:`~bson.son.SON`, the key and value types must be giv
 Typed Collection
 ----------------
 
-You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-defined schema for the data in a
+You can use :py:class:`~typing_extensions.TypedDict` (Python 3.8+) when using a well-defined schema for the data in a
```
blink1073, line 95: This should be reverted to `typing.TypedDict`

doc/examples/type_hints.rst
blink1073: I think you added the other example but did not revert changes to this one.

doc/examples/type_hints.rst
```diff
@@ -92,28 +92,96 @@ Note that when using :class:`~bson.son.SON`, the key and value types must be giv
...
+  >>> result = collection.find_one({})
   >>> assert result is not None
   >>> assert result["year"] == 1993
   >>> # This will not be type checked, despite being present, because it is added by PyMongo.
   >>> assert type(result["_id"]) == ObjectId
 
+Modeling Document Types with TypedDict
+--------------------------------------
+
+You can use :py:class:`~typing_extensions.TypedDict` (Python 3.8+) to model structured data.
```
blink1073, line 128: Should also be `typing.TypedDict` and `typing.NotRequired` below.

doc/examples/type_hints.rst
```diff
@@ -92,28 +92,96 @@ Note that when using :class:`~bson.son.SON`, the key and value types must be giv
...
+  >>> assert type(result["_id"]) == ObjectId
+  >>> collection: Collection[NotRequiredMovie] = client.test.test
+  >>> # Note the lack of _id, similar to the first example
+  >>> inserted = collection.insert_one(NotRequiredMovie(name="Jurassic Park", year=1993))
+  >>> result = collection.find_one({})
+  >>> assert result is not None
+  >>> # This will be type checked, despite not being provided explicitly.
+  >>> assert type(result["_id"]) == ObjectId
+
+This same typing scheme works for all of the insert methods (`insert_one`, `insert_many`, and `bulk_write`). For `bulk_write`,
```
blink1073, line 181: I think this comment carried over from the other PR.

blink1073 (changes requested):

.github/workflows/test-python.yml
```diff
@@ -67,6 +67,7 @@ jobs:
           # Test overshadowed codec_options.py file
           mypy --install-types --non-interactive bson/codec_options.py
           mypy --install-types --non-interactive --disable-error-code var-annotated --disable-error-code attr-defined --disable-error-code union-attr --disable-error-code assignment --disable-error-code no-redef --disable-error-code index --allow-redefinition --allow-untyped-globals --exclude "test/mypy_fails/*.*" test
```
blink1073, line 69: We'll need to exclude test/test_mypy.py here as well.

test/test_mypy.py
```diff
@@ -33,19 +35,25 @@ class MovieWithId(TypedDict):
...
+            year: int
+
+    except ImportError:
+        NotRequired = None  # type: ignore[assignment]
+        ImplicitMovie = None  # type: ignore[assignment,misc]
 
 except ImportError:
-    NotRequired = None  # type: ignore[assignment]
+    TypedDict = None
+
```
blink1073, line 56: We need to set ImplicitMovie to None here too

doc/examples/type_hints.rst
```diff
@@ -98,27 +98,88 @@ These methods automatically add an "_id" field.
 
 .. doctest::
 
-  >>> from typing import TypedDict
+  >>> # The typing_extensions package must be installed unless using Python 3.11+.
+  >>> from typing_extensions import TypedDict
```
blink1073, line 102: We should use `:pyversion: >= 3.8` instead of `typing_extensions` in this example

.evergreen/run-tests.sh
```diff
@@ -221,6 +221,7 @@ if [ -z "$GREEN_FRAMEWORK" ]; then
     fi
 
     $PYTHON $COVERAGE_ARGS setup.py $C_EXTENSIONS test $TEST_ARGS $OUTPUT
+
```
blink1073, line 224: Unnecessary whitespace change

doc/examples/type_hints.rst
```diff
@@ -97,28 +97,89 @@ You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-define
...
+  >>> result = collection.find_one({})
   >>> assert result is not None
   >>> assert result["year"] == 1993
   >>> # This will not be type checked, despite being present, because it is added by PyMongo.
   >>> assert type(result["_id"]) == ObjectId
 
+Modeling Document Types with TypedDict
+--------------------------------------
+
+You can use :py:class:`~typing_extensions.TypedDict` (Python 3.8+) to model structured data.
```
blink1073, line 123: This paragraph should use `typing` for both, and mention that they can also be gotten on earlier Python versions

blink1073 (changes requested):

doc/examples/type_hints.rst
```diff
@@ -97,28 +97,89 @@ You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-define
...
   >>> client: MongoClient = MongoClient()
   >>> collection: Collection[Movie] = client.test.test
+  >>> collection.drop()
   >>> inserted = collection.insert_one(Movie(name="Jurassic Park", year=1993))
-  >>> result = collection.find_one({"name": "Jurassic Park"})
+  >>> result = collection.find_one({})
   >>> assert result is not None
   >>> assert result["year"] == 1993
   >>> # This will not be type checked, despite being present, because it is added by PyMongo.
   >>> assert type(result["_id"]) == ObjectId
```
ShaneHarvey, line 118: The phrases "will not be type checked" and "cannot be type checked" appear throughout this PR but it's not that these lines won't be typed checked, they will fail type checking. It might be clearer to express this by adding a type ignore comment:
```python
assert result["_id"]  # type:ignore[typeddict-item]
``` 

Separately, I think we can remove the ObjectId in this example, just accessing result["_id"] with the type:ignore seems like enough.

test/test_mypy.py
```diff
@@ -324,6 +356,45 @@ def test_typeddict_document_type_insertion(self) -> None:
...
+        # This should fail because the output is a Movie.
+        assert out["foo"]  # type:ignore[typeddict-item]
+        assert type(out["_id"]) == ObjectId
+
+    # This should work the same as the test above, but this time using NotRequired to allow
+    # automatic insertion of the _id field by insert_one.
+    @only_type_check
+    def test_typeddict_not_required_document_type(self) -> None:
+        if NotRequired is None or ImplicitMovie is None:
+            raise unittest.SkipTest("Python 3.11+ is required to use NotRequired.")
```
ShaneHarvey, line 372: This is not needed since the test is already skipped by only_type_check.

ShaneHarvey (changes requested):

doc/examples/type_hints.rst
```diff
@@ -97,28 +97,90 @@ You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-define
...
+Modeling Document Types with TypedDict
+--------------------------------------
+
+You can use :py:class:`~typing.TypedDict` (Python 3.8+) to model structured data.
+As noted above, PyMongo will automatically add an `_id` field if it is not present. This also applies to TypedDict.
+There are three approaches to this:
+
+  1. Do not specify `_id` at all. It will be inserted automatically, and can be retrieved at run-time, but will yield a type-checking error.
+
+  2. Specify `_id` explicitly. This will mean that every instance of your custom TypedDict class will have to passed a value for `_id`.
```
blink1073, line 129: "have to pass"

doc/examples/type_hints.rst
```diff
@@ -97,28 +97,90 @@ You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-define
...
   >>> assert type(result["_id"]) == ObjectId
 
+Modeling Document Types with TypedDict
+--------------------------------------
+
+You can use :py:class:`~typing.TypedDict` (Python 3.8+) to model structured data.
+As noted above, PyMongo will automatically add an `_id` field if it is not present. This also applies to TypedDict.
+There are three approaches to this:
+
+  1. Do not specify `_id` at all. It will be inserted automatically, and can be retrieved at run-time, but will yield a type-checking error.
```
blink1073, line 127: "type-checking error unless explicitly ignored"

doc/examples/type_hints.rst
```diff
@@ -97,28 +97,90 @@ You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-define
...
+
+  2. Specify `_id` explicitly. This will mean that every instance of your custom TypedDict class will have to passed a value for `_id`.
+
+  3. Make use of :py:class:`~typing.NotRequired`. This has the flexibility of 1, but with the ability to access the `_id` field without causing a type-checking error.
```
blink1073, line 131: "flexibility of option 1"

test/test_mypy.py
```diff
@@ -20,14 +20,40 @@
 from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List
 
 try:
-    from typing_extensions import TypedDict
+    if TYPE_CHECKING:
```
blink1073, line 23: Per our conversation yesterday, we should set all of these to `dict` when not type-checking to avoid the import error nesting.

blink1073 (changes requested):

blink1073 (approved):
LGTM!

test/test_mypy.py
```diff
@@ -20,14 +20,28 @@
...
+        _id: ObjectId
+        name: str
+        year: int
+
+    class ImplicitMovie(TypedDict):
+        _id: NotRequired[ObjectId]
+        name: str
+        year: int
+
+except ImportError as exc:
```
blink1073, line 41: Oh, you need to set TypedDict and NotRequired to `None` here

doc/examples/type_hints.rst
```diff
@@ -97,28 +97,89 @@ You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-define
 These methods automatically add an "_id" field.
 
 .. doctest::
+  :pyversion: >= 3.8
 
   >>> from typing import TypedDict
   >>> from pymongo import MongoClient
   >>> from pymongo.collection import Collection
+  >>> from bson import ObjectId
```
blink1073, line 105: Not needed anymore

doc/examples/type_hints.rst
```diff
@@ -97,28 +97,90 @@ You can use :py:class:`~typing.TypedDict` (Python 3.8+) when using a well-define
...
   >>> class Movie(TypedDict):
   ...       name: str
   ...       year: int
   ...
   >>> client: MongoClient = MongoClient()
   >>> collection: Collection[Movie] = client.test.test
+  >>> collection.drop()
   >>> inserted = collection.insert_one(Movie(name="Jurassic Park", year=1993))
-  >>> result = collection.find_one({"name": "Jurassic Park"})
+  >>> result = collection.find_one({})
```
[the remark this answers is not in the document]
blink1073: This one needs to be fixed as well

test/test_mypy.py
```diff
@@ -68,6 +84,8 @@ class TestMypyFails(unittest.TestCase):
     def ensure_mypy_fails(self, filename: str) -> None:
         if api is None:
             raise unittest.SkipTest("Mypy is not installed")
+        if TypedDict is None:
+            raise unittest.SkipTest("TypedDict is supported on Python 3.8+ only")
```
ShaneHarvey, line 88: This seems redundant given that typeddict_client is already skipped in test_mypy_failures. Also this skip is overeager and will skip all the files on 3.7 which we don't want.

test/test_mypy.py
```diff
@@ -277,6 +295,10 @@ def foo(self):
 
 
 class TestDocumentType(unittest.TestCase):
+    def setUp(self) -> None:
+        if TypedDict is None:
+            raise unittest.SkipTest("Python 3.8+ is required to use TypedDict.")
+
```
ShaneHarvey, line 301: This skip is overeager. It will skip the below tests that don't even use TypedDict. Can we remove this?

doc/examples/type_hints.rst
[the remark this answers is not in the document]
ShaneHarvey: ping.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #1105, opened 2022-11-02
PYTHON-3396 Support the Azure VM-assigned Managed Identity for Automatic KMS Credentials

blink1073:
Companion to https://github.com/mongodb/libmongocrypt/pull/484

Also fixes a bug in [PYTHON-3367](https://jira.mongodb.org/browse/PYTHON-3367) where we we were running the test file but not actually running the GCP tests.


.evergreen/config.yml
```diff
@@ -1072,6 +1072,49 @@ task_groups:
...
+          export AZUREKMS_CLIENTID="${testazurekms_clientid}"
+          export AZUREKMS_TENANTID="${testazurekms_tenantid}"
+          export AZUREKMS_SECRET="${testazurekms_secret}"
+          export AZUREKMS_DRIVERS_TOOLS="$DRIVERS_TOOLS"
+          export AZUREKMS_RESOURCEGROUP="${testazurekms_resourcegroup}"
+          export AZUREKMS_PUBLICKEYPATH="/tmp/testazurekms_publickey"
+          export AZUREKMS_PRIVATEKEYPATH="/tmp/testazurekms_privatekey"
+          export AZUREKMS_SCOPE="${testazurekms_scope}"
+          export AZUREKMS_VMNAME_PREFIX="PYTHON_DRIVER"
+          $DRIVERS_TOOLS/.evergreen/csfle/azurekms/create-and-setup-vm.sh
```
ShaneHarvey, line 1100: It looks like these exports will be echoed (by xtrace) and appear in the logs. Let's use the secure `silent: true` + source approach used by the other FLE tests: https://github.com/mongodb/mongo-python-driver/blob/228edd21f858fe20c3dd0bffdc6aff054a5726a0/.evergreen/config.yml#L408-L428

.evergreen/config.yml
```diff
@@ -1072,6 +1072,49 @@ task_groups:
...
+  - name: testazurekms_task_group
+    setup_group:
+    - func: fetch source
+    - func: prepare resources
+    - func: fix absolute paths
+    - func: make files executable
+    - command: shell.exec
+      params:
+        shell: bash
+        script: |-
```
ShaneHarvey, line 1084: `|-` -> `|`. Is the `-` important? We use `|` everywhere else.

.evergreen/run-mongodb-fle-azure-auto.sh
```diff
@@ -0,0 +1,34 @@
...
+
+authtest () {
+    echo "Running Azure Credential Acquisition Test with $PYTHON"
+    $PYTHON --version
+    $PYTHON -m pip install --upgrade wheel setuptools pip
+    $PYTHON -m pip install https://github.com/mongodb/libmongocrypt/archive/refs/heads/master.zip#subdirectory=bindings/python
+    curl -O https://s3.amazonaws.com/mciuploads/libmongocrypt/all/master/latest/libmongocrypt-all.tar.gz
+    mkdir libmongocrypt-all && tar xzf libmongocrypt-all.tar.gz -C libmongocrypt-all
+    $PYTHON -m pip install '.'
+    PYMONGOCRYPT_LIB=$(pwd)/libmongocrypt-all/debian10/nocrypto/lib/libmongocrypt.so TEST_FLE_AZURE_AUTO=1 $PYTHON test/test_on_demand_csfle.py
```
ShaneHarvey, line 31: This file looks almost identical to the gcp one.  Can we reuse that file rather than introducing another? In fact, I believe we could even reuse run-tests.sh with some minor changes to the `TEST_ARGS` logic.

ShaneHarvey (changes requested):

.evergreen/config.yml
blink1073: Done

.evergreen/run-mongodb-fle-azure-auto.sh
ShaneHarvey: Any thoughts about the second suggestion to use run-tests.sh? That file already knows how to download and setup libmongocrypt.

blink1073: Done

.evergreen/config.yml
```diff
@@ -2020,7 +2020,7 @@ tasks:
             export AZUREKMS_RESOURCEGROUP=${testazurekms_resourcegroup}
             export AZUREKMS_VMNAME=${AZUREKMS_VMNAME}
             export AZUREKMS_PRIVATEKEYPATH=/tmp/testazurekms_privatekey
-            AZUREKMS_CMD="MONGODB_URI='mongodb://localhost:27017' KEY_NAME='${testazurekms_keyname}' KEY_VAULT_ENDPOINT='${testazurekms_keyvaultendpoint}' SUCCESS=true TEST_FLE_AZURE_AUTO=1 ./.evergreen/run-mongodb-fle-auto.sh" \
+            AZUREKMS_CMD="MONGODB_URI='mongodb://localhost:27017' KEY_NAME='${testazurekms_keyname}' KEY_VAULT_ENDPOINT='${testazurekms_keyvaultendpoint}' SUCCESS=true TEST_ENCRYPTION=1 TEST_FLE_AZURE_AUTO=1 ./.evergreen/run-tests.sh" \
```
ShaneHarvey, line 2023: You might need to pass the right LIBMONGOCRYPT_URL for the remote azure host here.

.evergreen/config.yml
```diff
@@ -1969,7 +1969,7 @@ tasks:
               export GCPKMS_PROJECT=${GCPKMS_PROJECT}
               export GCPKMS_ZONE=${GCPKMS_ZONE}
               export GCPKMS_INSTANCENAME=${GCPKMS_INSTANCENAME}
-              GCPKMS_CMD="SUCCESS=true TEST_FLE_GCP_AUTO=1 ./.evergreen/run-mongodb-fle-auto.sh mongodb://localhost:27017" $DRIVERS_TOOLS/.evergreen/csfle/gcpkms/run-command.sh
+              GCPKMS_CMD="SUCCESS=true TEST_ENCRYPTION=1 TEST_FLE_GCP_AUTO=1 ./.evergreen/run-tests.sh" $DRIVERS_TOOLS/.evergreen/csfle/gcpkms/run-command.sh
```
ShaneHarvey, line 1972: You might need to pass the right LIBMONGOCRYPT_URL for the remote gcp host here.

.evergreen/config.yml
blink1073: Fixed

.evergreen/config.yml
blink1073: Fixed

ShaneHarvey (approved):
> In this case, the final line break and any trailing empty lines are excluded from the scalar’s content

I don't understand how the final newline could change the behavior here but LGTM.

PR #1107, opened 2022-11-03
PYTHON-3501 Ensure Auth Environment Variables are Always Dynamic

blink1073:


test/auth_aws/test_auth_aws.py
```diff
@@ -111,6 +111,67 @@ def test_poisoned_cache(self):
         client.get_database().test.find_one()
         self.assertNotEqual(auth.get_cached_credentials(), None)
 
+    def test_environment_variables_ignored(self):
```
ShaneHarvey, line 114: Is this behavior outlined in the spec? Shouldn't the env vars take precedence even if we already have a previous cached credential?

ShaneHarvey: Discussed in person, this behavior happens because we rely on aws for the credential lookup. The code is like:
```
if cached_creds:
    return cached_creds
creds = get_aws_creds()
if should_cache(creds):
    cached_creds = creds
return creds
```

So if the app is always using env vars then there will never be a cached credential and dynamic env var changes will propagate immediately. However, if the app is using temp creds then later tries to switch to dynamic env var creds, then changes won't propagate until the cached creds are cleared.

ShaneHarvey (approved):

blink1073:
We updated the spec to clarify that changes to the environment variables should be isolated in case the tests are run in parallel.

ShaneHarvey:
> We updated the spec to clarify that changes to the environment variables should be isolated in case the tests are run in parallel.

That's fine but I just want to confirm that mock.patch doesn't actually add thread safety to these tests.

blink1073:
> That's fine but I just want to confirm that mock.patch doesn't actually add thread safety to these tests.

Confirmed.  Based on the spec that means we should use unit tests instead, which means those tests will need to go in `pymongo_auth_aws`.

ShaneHarvey:
I'm confused, I don't think we need to do anything regarding the parallel test issue because we don't run the python tests in parallel.

ShaneHarvey (approved):

blink1073:
> I'm confused, I don't think we need to do anything regarding the parallel test issue because we don't run the python tests in parallel. 

Yeah that's fair, I'll update the spec a bit to clarify.

PR #1113, opened 2022-11-11
PYTHON-3517 Add documentation for on-demand KMS providers

blink1073:


doc/changelog.rst
```diff
@@ -1,6 +1,23 @@
 Changelog
 =========
 
+Changes in Version 4.4
+----------------------
+
+Version 4.4 brings a number of improvements including:
+
+- Added support for :ref:`CSFLE on-demand credentials` for cloud KMS providers.
```
ShaneHarvey, line 9: I think this can go into the 4.3.3 release if we market it as "documented support" instead of "added support". We can say that it requires pymongocrypt>=1.4. I think this makes more sense because it will be supported by older versions of pymongo too.

blink1073: Sounds good

doc/changelog.rst
```diff
@@ -4,7 +4,7 @@ Changelog
 Changes in Version 4.3.3
 ------------------------
 
-Version 4.3.3 fixes a number of bugs:
+Version 4.3.3 documents supportfor :ref:`CSFLE on-demand credentials` for cloud KMS providers, and fixes the following bugs:
```
ShaneHarvey, line 7: supportfor -> support for

doc/examples/encryption.rst
```diff
@@ -713,6 +713,81 @@ To configure automatic *decryption* without automatic *encryption* set
...
+AWS, GCP, and Azure cloud environments.
+
+To enable the driver's behavior to obtain credentials from the environment, add the appropriate key ("aws", "gcp", or "azure") with an empty map to
+"kms_providers" in either :class:`~pymongo.encryption_options.AutoEncryptionOpts` or :class:`~pymongo.encryption.ClientEncryption` options.
+
+An application using AWS credentials would look like::
+
+    from pymongo import MongoClient
+    from pymongo.encryption import ClientEncryption
+    from pymongo.encryption_options import AutoEncryptionOpts
```
ShaneHarvey, line 736: AutoEncryptionOpts is imported but never used in these examples. It would be good to add one AutoEncryptionOpts example.

doc/examples/encryption.rst
```diff
@@ -713,6 +713,81 @@ To configure automatic *decryption* without automatic *encryption* set
...
+    client = MongoClient()
+    client_encryption = ClientEncryption(
+          kms_providers={"aws": {}},
```
ShaneHarvey, line 739: A comment on these lines would be good to highlight that the empty dict means creds are looked up on demand.

ShaneHarvey (changes requested):

doc/changelog.rst
blink1073: fixed

doc/examples/encryption.rst
blink1073: Done

doc/examples/encryption.rst
blink1073: Done

ShaneHarvey (approved):

PR #1117, opened 2022-11-14
PYTHON-3388 Propagate Original Error for Write Errors Labeled NoWritesPerformed

juliusgeo:


test/test_retryable_writes.py
```diff
@@ -64,6 +67,24 @@
...
+class InsertEventListener(EventListener):
+    def __init__(self):
+        self._retried = False
+        super().__init__()
+
+    def reset(self) -> None:
+        self._retried = False
+        super().reset()
+
+    def succeeded(self, event: CommandSucceededEvent):
```
blink1073, line 79: Needs a return type per mypy error

test/test_retryable_writes.py
```diff
@@ -581,8 +602,56 @@ def test_pool_paused_error_is_retryable(self):
         failed = cmd_listener.failed_events
         self.assertEqual(1, len(failed), msg)
 
+    @client_context.require_failCommand_fail_point
+    @client_context.require_replica_set
+    @client_knobs(heartbeat_frequency=0.05, min_heartbeat_interval=0.05)
+    def test_returns_original_error_code(
+        self,
+    ):  # TODO: Make this a real integration test where we stepdown the primary.
```
blink1073, line 610: Was this TODO meant to be here?  It looks like it was referring to the TestRetryableWritesTxnNumber class.

test/test_retryable_writes.py
```diff
@@ -581,8 +602,56 @@ def test_pool_paused_error_is_retryable(self):
...
+                "data": {
+                    "errorCode": 10107,
+                    "errorLabels": ["RetryableWriteError", "NoWritesPerformed"],
+                    "failCommands": ["insert"],
+                },
+            }
+        )
+        with self.assertRaises(WriteConcernError) as exc:
+            client.test.test.insert_one({"_id": 1})
+        assert exc.exception.code == 91
```
blink1073, line 646: Should use `self.assertEquals`

blink1073 (dismissed):
LGTM!

pymongo/mongo_client.py
```diff
@@ -836,6 +837,8 @@ def __init__(
             # This will be used later if we fork.
             MongoClient._clients[self._topology._topology_id] = self
 
+        self._indefinite_error: Optional[WriteConcernError] = None
```
ShaneHarvey, line 840: We can't use an instance variable to track this because it makes MongoClient not thread safe.

pymongo/mongo_client.py
```diff
@@ -1405,6 +1408,10 @@ def is_retrying():
                 # Add the RetryableWriteError label, if applicable.
                 _add_retryable_write_error(exc, max_wire_version)
                 retryable_error = exc.has_error_label("RetryableWriteError")
+                if self._indefinite_error and exc.has_error_label("NoWritesPerformed"):
+                    raise self._indefinite_error from exc
+                if isinstance(exc, WriteConcernError):
+                    self._indefinite_error = exc
```
ShaneHarvey, line 1414: _indefinite_error  is not needed. `last_error` already tracks the error that should be propagated.

ShaneHarvey (changes requested):

blink1073:
Mypy failure:

```python
pymongo/mongo_client.py:1416: error: Incompatible types in assignment
(expression has type "PyMongoError", variable has type
"Optional[WriteConcernError]")  [assignment]
                            self._indefinite_error = exc
```

pymongo/mongo_client.py
ShaneHarvey: Why do you think we need to track the last WriteConcernError? This change should not track WriteConcernErrors any differently than other errors. This change is about errors labeled with NoWritesPerformed which may or may not be a WriteConcernError.

pymongo/mongo_client.py
```diff
@@ -1408,10 +1409,11 @@ def is_retrying():
                 # Add the RetryableWriteError label, if applicable.
                 _add_retryable_write_error(exc, max_wire_version)
                 retryable_error = exc.has_error_label("RetryableWriteError")
-                if self._indefinite_error and exc.has_error_label("NoWritesPerformed"):
-                    raise self._indefinite_error from exc
-                if isinstance(exc, WriteConcernError):
-                    self._indefinite_error = exc
+                with self._indefinite_error_lock:
```
ShaneHarvey, line 1412: Adding a lock here does not help at all. It's still not thread safe.

pymongo/mongo_client.py
```diff
@@ -1408,12 +1408,17 @@ def is_retrying():
                 if retryable_error:
                     session._unpin()
                 if not retryable_error or (is_retrying() and not multiple_retries):
-                    raise
+                    if exc.has_error_label("NoWritesPerformed"):
+                        assert last_error is not None
```
blink1073, line 1412: I assume this is here for the type checker.  I think we should be more conservative and check `if exc.has_error_label("NoWritesPerformed") and last_error is not None`

pymongo/mongo_client.py
```diff
@@ -1408,12 +1408,17 @@ def is_retrying():
...
+                        raise last_error from exc
+                    else:
+                        raise
```
ShaneHarvey, line 1415: There's a missing condition here. If the current error is NoWritesPerformed and the last error is None, then we must raise the current error even though it's labelled NoWritesPerformed. This is True for all the cases above with `assert last_error is not None` too.

ShaneHarvey (changes requested):

blink1073 (approved):

ShaneHarvey (changes requested):
Can you add the test from https://github.com/mongodb/specifications/pull/1349?

blink1073:
The pre-commit failure was addressed in #1122

ShaneHarvey (approved):

PR #1118, opened 2022-11-17
PYTHON-3527 + PYTHON-3528 Fix no-server tests

ShaneHarvey:
PYTHON-3528 Skip TestCreateEntities when no server is running.
PYTHON-3527 Skip test_typeddict_find_notrequired when no server is running.

Also fixes the 150s TestCreateEntities issue.


blink1073 (approved):
Thanks!

PR #1119, opened 2022-11-17
PYTHON-3526 Fix mockup tests

ShaneHarvey:


ShaneHarvey:
Originally we discussed adding `test/mockupdb/__init__.py` to fix this issue but we can't do that otherwise unittest will start running these tests by default. Instead I decided to just ignore the mypy error.

blink1073 (approved):
Thanks!

PR #1121, opened 2022-11-22
PYTHON-3524 Support passing list of strings to create_index

blink1073:


pymongo/helpers.py
```diff
@@ -80,6 +80,8 @@ def _index_list(key_or_list, direction=None):
     Takes such a list, or a single key, or a single key and direction.
     """
     if direction is not None:
+        if not isinstance(key_or_list, str):
+            raise TypeError("Expected a string and a a direction")
```
ShaneHarvey, line 84: "and a a direction" -> "and a direction"

pymongo/helpers.py
```diff
@@ -88,7 +90,12 @@ def _index_list(key_or_list, direction=None):
             return list(key_or_list)
         elif not isinstance(key_or_list, (list, tuple)):
             raise TypeError("if no direction is specified, key_or_list must be an instance of list")
-        return key_or_list
+        values = []
```
ShaneHarvey, line 93: Should this handle the ItemsView case too? EG an ItemsView with a mix of string and tuples should be normalized to all tuples.

ShaneHarvey (changes requested):
Should we update the docs for sort+hint as well?

pymongo/helpers.py
blink1073: Fixed

pymongo/operations.py
```diff
@@ -435,7 +435,9 @@ def __init__(self, keys: _IndexKeyHint, **kwargs: Any) -> None:
 
         For use with :meth:`~pymongo.collection.Collection.create_indexes`.
 
-        Takes either a single key or a list of (key, direction) pairs.
+        Takes either a single key or a list containing (key, direction) pairs
+        or keys.  If no direction is given, :data:`~pymongo.ASCENDING` will
+        be assumed.
```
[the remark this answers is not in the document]
blink1073: Done

pymongo/helpers.py
blink1073: I don't follow, you mean if the value were to be defined as a dictionary and the key was given as `.items()`?

ShaneHarvey: Yes, see above on line 89.

ShaneHarvey:
Did you see this question:

> Should we update the docs and test sort+hint as well?

blink1073:
> Did you see this question:

No, this UX leaves a lot to be desired...

pymongo/helpers.py
blink1073: Done

blink1073:
I updated the docs for `create_index` and `sort`.  The `hint` docs refer back to `create_index` for allowed values.

pymongo/helpers.py
ShaneHarvey: Your change made me realize my mistake. abc.ItemsView is not supported as an element inside key_or_list. We actually don't need to do anything special here since ItemsView always yields (key, value) tuples. Sorry for the confusion. You should remove the ItemsView line you added.

blink1073: I added a test and update the type to allow for that option

pymongo/operations.py
```diff
@@ -24,7 +35,7 @@
 
 # Hint supports index name, "myIndex", or list of either strings or index pairs: [('x', 1), ('y', -1), 'z'']
 _IndexList = Sequence[Union[str, Tuple[str, Union[int, str, Mapping[str, Any]]]]]
-_IndexKeyHint = Union[str, _IndexList]
+_IndexKeyHint = Union[str, _IndexList, ItemsView[str, Any]]
```
ShaneHarvey, line 38: Is this needed? Shouldn't ItemsView should already be supported by _IndexList because it's a sequence of tuples?

blink1073: No, mypy treated them differently.  The test was giving the error:

```python
test/test_collection.py: note: In member "test_create_index" of class "TestCollection":
test/test_collection.py:320: error: Argument 1 to "create_index" of "Collection" has incompatible type
"dict_items[str, int]"; expected "Union[str, Sequence[Union[str, Tuple[str, Union[int, str, Mapping[str, Any]]]]]]"
[arg-type]
            db.test.create_index({'hello': 1}.items())
```

ShaneHarvey: Let's not change our type signature here. "supporting" ItemsView was a mistake, let's not go more out of our way to support it. This looks like a glaring mypy bug to me.

Can we add a type ignore to the test instead of further complicating the type signature?

blink1073: Done

ShaneHarvey (approved):

pymongo/operations.py
ShaneHarvey: Thanks!

PR #1123, opened 2022-11-24
Please fix bug

RoTorEx:
My interpreter is marks my code after method *inser* as unreachable but it is not truth cause a problem with returing value


RoTorEx:
> 

Thanks for the quick response, I was indeed mistaken, the thing is that the project uses a version of pymongo 3.11.3. And in my Python environment was the latest version >4, which is why the compiler could not recognize the method and made the code below unreachable. Thanks

PR #1124, opened 2022-12-01
PYTHON-3492 Test mongocryptd is not spawned when shared library is loaded

juliusgeo:


test/test_encryption.py
```diff
@@ -2335,6 +2335,31 @@ def test_02_success(self):
         self.client_encryption.create_data_key("aws", self.master_key)
 
 
+# https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#8-bypass-spawning-mongocryptd
+class TestBypassMongocryptWithSharedLibrary(EncryptionIntegrationTest):
```
ShaneHarvey, line 2339: Let's add this to the existing TestBypassSpawningMongocryptdProse class instead.

test/test_encryption.py
```diff
@@ -2335,6 +2335,31 @@ def test_02_success(self):
...
+            "keyvault.datakeys",
+            schema_map=schemas,
+            mongocryptd_uri="mongodb://localhost:27021/db?serverSelectionTimeoutMS=1000",
+            mongocryptd_spawn_args=[
+                "--pidfilepath=bypass-spawning-mongocryptd.pid",
+                "--port=27021",
+            ],
+            crypt_shared_lib_required=True,
+        )
+        encrypted_client = rs_or_single_client(auto_encryption_opts=opts)
```
ShaneHarvey, line 2354: Please rename to `client_encrypted` to match the prose test description.

test/test_encryption.py
```diff
@@ -2335,6 +2335,31 @@ def test_02_success(self):
         self.client_encryption.create_data_key("aws", self.master_key)
 
 
+# https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#8-bypass-spawning-mongocryptd
+class TestBypassMongocryptWithSharedLibrary(EncryptionIntegrationTest):
+    @unittest.skipUnless(os.environ.get("TEST_CRYPT_SHARED"), "crypt_shared lib is not installed")
+    def test_via_loading_shared_library(self):
+        schemas = {"db.,coll": json_data("external", "external-schema.json")}
```
ShaneHarvey, line 2342: "db.,coll" -> "db.coll"

Asserting that the document was actually encrypted will catch bugs like this.

test/test_encryption.py
```diff
@@ -2335,6 +2335,31 @@ def test_02_success(self):
...
+        opts = AutoEncryptionOpts(
+            KMS_PROVIDERS,
+            "keyvault.datakeys",
+            schema_map=schemas,
+            mongocryptd_uri="mongodb://localhost:27021/db?serverSelectionTimeoutMS=1000",
+            mongocryptd_spawn_args=[
+                "--pidfilepath=bypass-spawning-mongocryptd.pid",
+                "--port=27021",
```
ShaneHarvey, line 2350: This port is inconvenient since it's used by launch when starting a sharded cluster. Let's use 47021 instead.

test/test_encryption.py
```diff
@@ -2335,6 +2335,31 @@ def test_02_success(self):
...
+            ],
+            crypt_shared_lib_required=True,
+        )
+        encrypted_client = rs_or_single_client(auto_encryption_opts=opts)
+        encrypted_client.db.coll.insert_one({"unencrypted": "test"})
```
ShaneHarvey, line 2355: Can we assert that this field was actually encrypted?
```
self.assertEncrypted(self.client.db.coll.find_one({})["unencrypted"])
```

ShaneHarvey (changes requested):

test/test_encryption.py
ShaneHarvey: Oh I see, my mistake was that I assumed "unencrypted" was an encrypted field. Can you change the test to use the "encrypted" field instead of "unencrypted"? Encryption will still work because this test uses the shared library.

blink1073 (approved):
LGTM!

ShaneHarvey (approved):
LGTM!

PR #1125, opened 2022-12-02
PYTHON-3351 Provide access to raw result document when the server returns an error for a command

juliusgeo:


blink1073:
I think we need to update `process_error` in `unified_format.py` as well to verify we are handling `errorResponse`.

test/unified_format.py
```diff
@@ -994,6 +994,10 @@ def process_error(self, exception, spec):
         error_labels_contain = spec.get("errorLabelsContain")
         error_labels_omit = spec.get("errorLabelsOmit")
         expect_result = spec.get("expectResult")
+        error_response = spec.pop("errorResponse")
```
blink1073, line 997: Should be `.get()`

ShaneHarvey (approved):

blink1073 (approved):

PR #1126, opened 2022-12-12
Optional filter for count_documents

DeveloperAnonymous:
Hello, I think it'd make more sense that by default, if no filter is passed, it simply ["Count the number of documents in [the] collection"](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/collection.py#L1764)

### DIFF
--- a/pymongo/collection.py
+++ b/pymongo/collection.py
@@ -1756,7 +1756,7 @@ class Collection(common.BaseObject, Generic[_DocumentType]):
 
     def count_documents(
         self,
-        filter: Mapping[str, Any],
+        filter: Optional[Mapping[str, Any]] = None,
         session: Optional["ClientSession"] = None,
         comment: Optional[Any] = None,
         **kwargs: Any,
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Hi @DeveloperAnonymous, thanks for opening this PR. After looking into this I found that we intentionally designed the filter to be required for count_documents so I'm going to close this as "works as designed". The spec says:

> Applications must be required to pass a value for filter, but an empty document is supported

https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#countdocuments

Your application will probably be better off using estimated_document_count instead because it has much better performance. It's only an "estimate" in some rare cases after an unexpected shutdown and only for a brief time which you can read more about here: https://www.mongodb.com/docs/manual/reference/command/count/#accuracy-after-unexpected-shutdown 

We may update the documentation to make this rationale more clear as part of https://jira.mongodb.org/browse/DRIVERS-2518.

DeveloperAnonymous:
Hey @ShaneHarvey, sounds good and thanks for the details about how estimated_document_count works!

Have a great day 🙂

PR #1128, opened 2022-12-20
PYTHON-3542 Test Failure - test_iteration on PyPy 3.8+

blink1073:


test/test_client.py
```diff
@@ -217,10 +217,7 @@ def test_getattr(self):
 
     def test_iteration(self):
         client = self.client
-        if "PyPy" in sys.version:
-            msg = "'NoneType' object is not callable"
-        else:
-            msg = "'MongoClient' object is not iterable"
+        msg = "'MongoClient' object is not iterable"
```
ShaneHarvey, line 220: Won't this now fail on the old pypy versions in EVG?

ShaneHarvey (changes requested):

test/test_client.py
blink1073: Oh ha, landmines everywhere.  😅

blink1073: Fixed, and scheduled some pypy tasks

ShaneHarvey (approved):

PR #1131, opened 2023-01-04
PYTHON-3546 bson.CodecOptions docs missing unicode_decode_error_handler=ignore option in newer documentation

blink1073:


bson/codec_options.py
```diff
@@ -245,87 +245,92 @@ class _BaseCodecOptions(NamedTuple):
...
+        .. versionadded:: 3.8
+           `type_registry` attribute.
+
+        .. warning:: Care must be taken when changing
+           `unicode_decode_error_handler` from its default value ('strict').
+           The 'replace' and 'ignore' modes should not be used when documents
+           retrieved from the server will be modified in the client application
+           and stored back to the server.
+        """
+        return super().__init__()
```
ShaneHarvey, line 333: Don't we have to pass `*args, **kwargs` here?

ShaneHarvey (commented):
> The diff weirdly shows the entire block being changed--not sure what's up with that

In diffs like this it's helpful to use the "ignore whitespace changes" setting (w=1): https://github.com/mongodb/mongo-python-driver/pull/1131/files?diff=split&w=1

bson/codec_options.py
blink1073: Ah, no, now I remember.  `namedtuple.__init__()` "takes exactly one argument (the instance to initialize))".  The args and kwargs are handled in `__new__()`.

ShaneHarvey: Ah okay. That's not intuitive at all :)

PR #1132, opened 2023-01-10
DRIVERS-2369 Disable causal consistency in implicit sessions

juliusgeo:
As can be seen [here](https://github.com/mongodb/mongo-python-driver/blob/a43f320753a0b8710e23fb36ae7bb488ef790b41/pymongo/mongo_client.py#L1744) we already make sure our implicit sessions are already not causally consistent.


ShaneHarvey (approved):
LGTM!

PR #1136, opened 2023-01-12
PYTHON-3446 Do not connect to mongocryptd if shared library is loaded

juliusgeo:


test/test_encryption.py
```diff
@@ -1942,6 +1942,32 @@ def test_via_loading_shared_library(self):
         with self.assertRaises(ServerSelectionTimeoutError):
             no_mongocryptd_client.db.command("ping")
 
+    # https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#20-bypass-creating-mongocryptd-client-when-shared-library-is-loaded
+    @unittest.skipUnless(os.environ.get("TEST_CRYPT_SHARED"), "crypt_shared lib is not installed")
+    def test_client_via_loading_shared_library(self):
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
```
blink1073, line 1948: This test is missing the `listenerThread` portion of the linked spec.

blink1073 (changes requested):

test/test_encryption.py
juliusgeo: Yup, working on doing that right now.

test/test_encryption.py
```diff
@@ -1945,28 +1946,38 @@ def test_via_loading_shared_library(self):
...
-            client_encrypted = rs_or_single_client(auto_encryption_opts=opts)
-            self.addCleanup(client_encrypted.close)
-            client_encrypted.db.coll.drop()
-            client_encrypted.db.coll.insert_one({"encrypted": "test"})
-            sock.bind(("localhost", 27021))
+        def listen():
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+                addr = ("localhost", 27021)
+                sock.bind(addr)
+                sock.settimeout(2)
```
blink1073, line 1953: This is technically a race condition.  If we instead add a non-blocking poll on accept, and raise an error if we do accept, but use a queue to pass a "stop" sentinel into the thread, we can remove the race condition.

blink1073 (changes requested):

test/test_encryption.py
ShaneHarvey: I would also prefer to refactor this so that the test doesn't take 2 seconds. Ideally it should run as fast as possible.

ShaneHarvey: We can probably use https://docs.python.org/3/library/socketserver.html for this to  avoid manually writing the setup and shutdown logic.

test/test_encryption.py
```diff
@@ -1942,6 +1944,44 @@ def test_via_loading_shared_library(self):
         with self.assertRaises(ServerSelectionTimeoutError):
             no_mongocryptd_client.db.command("ping")
 
+    # https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#20-bypass-creating-mongocryptd-client-when-shared-library-is-loaded
+    @unittest.skipUnless(os.environ.get("TEST_CRYPT_SHARED"), "crypt_shared lib is not installed")
+    def test_client_via_loading_shared_library(self):
+        class TCPHandler(socketserver.BaseRequestHandler):
+            def handle(self):
+                assert not self.request.recv(1024)
```
ShaneHarvey, line 1952: We don't even need the recv here. If this method ever gets called at all then it means a connection to the mongocryptd port was established and the test must fail. Also can you confirm the behavior of raising an exception here? According to the docs:
> If the user-provided [handle()](https://docs.python.org/3/library/socketserver.html#socketserver.BaseRequestHandler.handle) method of the handler class raises an exception, the server’s [handle_error()](https://docs.python.org/3/library/socketserver.html#socketserver.BaseServer.handle_error) method will be called.
> ... The default action is to print the traceback to standard error and continue handling further requests.

test/test_encryption.py
```diff
@@ -1942,6 +1944,44 @@ def test_via_loading_shared_library(self):
...
+
+        def listen():
+            addr = ("localhost", 27021)
+            with socketserver.TCPServer(addr, TCPHandler) as sock:
+                sock.server_bind()
+                sock.server_activate()
```
ShaneHarvey, line 1958: These two method are already called automatically. To actually start handling requests we need to: 
> call the [handle_request()](https://docs.python.org/3/library/socketserver.html#socketserver.BaseServer.handle_request) or [serve_forever()](https://docs.python.org/3/library/socketserver.html#socketserver.BaseServer.serve_forever) method of the server object to process one or many requests.

test/test_encryption.py
```diff
@@ -1942,6 +1944,44 @@ def test_via_loading_shared_library(self):
...
 
+    # https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#20-bypass-creating-mongocryptd-client-when-shared-library-is-loaded
+    @unittest.skipUnless(os.environ.get("TEST_CRYPT_SHARED"), "crypt_shared lib is not installed")
+    def test_client_via_loading_shared_library(self):
+        class TCPHandler(socketserver.BaseRequestHandler):
+            def handle(self):
+                assert not self.request.recv(1024)
+
+        def listen():
+            addr = ("localhost", 27021)
```
ShaneHarvey, line 1955: This port is used when testing sharded clusters. Let's change to 47021 like we did in PYTHON-3492.

ShaneHarvey (changes requested):

test/test_encryption.py
juliusgeo: This is the error when the address is being used:
```
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/Users/julius/Work/mongo-python-driver/test/utils.py", line 983, in run
    super(ExceptionCatchingThread, self).run()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/julius/Work/mongo-python-driver/test/test_encryption.py", line 1956, in listen
    with socketserver.TCPServer(addr, TCPHandler) as sock:
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/socketserver.py", line 452, in __init__
    self.server_bind()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/socketserver.py", line 466, in server_bind
    self.socket.bind(self.server_address)
OSError: [Errno 48] Address already in use
```
This is when the address times out:
```
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/Users/julius/Work/mongo-python-driver/test/utils.py", line 983, in run
    super(ExceptionCatchingThread, self).run()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/julius/Work/mongo-python-driver/test/test_encryption.py", line 1957, in listen
    sock.server_bind()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/socketserver.py", line 466, in server_bind
    self.socket.bind(self.server_address)
OSError: [Errno 22] Invalid argument
```
I added an assertion to make this distinction clear.

test/test_encryption.py

test/test_encryption.py
juliusgeo: Those two methods seem to not be called automatically--commenting out either one causes no error to be thrown in the thread. Also, I do not think we need to process any requests--we just need to confirm that the port is not in use.

juliusgeo: Ah, no, the issue is that those two methods are called automatically, but in a try block, so the error is suppressed. We need to call them ourselves to get the error.

juliusgeo: `handle_request` does not raise an error, and `serve_forever` ignores the timeout so it also does not raise an error and instead just blocks forever. The only functions that seem to work are the ones that I have.

ShaneHarvey: We can't call bind or server_activate here. Those methods are already called automatically. Calling sock.server_bind() here fails with errno 22 because you can't bind to the same address twice.

ShaneHarvey (changes requested):

test/test_encryption.py
ShaneHarvey: We have to call either handle_request() or serve_forever() here or else the TCPServer will just immediately close and exit the with block.

juliusgeo: As I explained, they are called automatically *in a try block*, so the error produced is suppressed, thus we must do it ourselves to capture the exception. 
This same test ran with a port that is in use gives a different error, so I know that this is not just because of rebinding. Additionally, those two functions simply do not display the right behavior, in addition to being unnecessary--the connection code already tells us if the port is in use. I have tried the configuration you are suggesting and it just doesn't work.

ShaneHarvey: I believe you are conflating the behavior of TCP server and the TCP connections that the server spawns (the socketserver docs call these "requests" and each one gets handled with a new BaseRequestHandler instance). As written all this test does currently is assert that calling TCPServer.server_bind() twice raises an error:

```Python
>>> import socketserver
>>> with socketserver.TCPServer(("localhost", 47021), None) as sock:pass
... 
>>> with socketserver.TCPServer(("localhost", 47021), None) as sock:
...     sock.server_bind()
... 
  File "<stdin>", line 2, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/socketserver.py", line 466, in server_bind
    self.socket.bind(self.server_address)
OSError: [Errno 22] Invalid argument
```

Instead we need to call serve_forever() and implement a handler which causes the test to fail if a connection is established to the TCPServer. Here's an example:
```python
import socketserver
import threading
import time

# Setup server.
connection_established = [False]
class Handler(socketserver.BaseRequestHandler):
    def handle(self):
        connection_established[0] = True

server = socketserver.TCPServer(("localhost", 47021), Handler)
def serve():
    with server:
        server.serve_forever(poll_interval=.05)  # Short poll timeout to speed up the test

server_thread = threading.Thread(target=serve)
server_thread.start()
# Run test: connect=True will cause the test to fail, connect=False should pass
client = MongoClient("localhost", 47021, connect=False)
time.sleep(1)
# Shutdown server.
server.shutdown()
server_thread.join()
assert not connection_established[0], "a connection was established"
```

ShaneHarvey (changes requested):

test/test_encryption.py
juliusgeo: Try your example with a port that is in use, though, and it gives a different error:
```
>>> with socketserver.TCPServer(("localhost", 27017), None) as sock:
...     sock.server_bind()
... 
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/socketserver.py", line 452, in __init__
    self.server_bind()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/socketserver.py", line 466, in server_bind
    self.socket.bind(self.server_address)
OSError: [Errno 48] Address already in use
```
Thus, as written this test does actually test that the port is not in use, that's why I added the assert at the end for the errno.

ShaneHarvey: We don't want to test a port that's already in use. The TCPServer has to bind to a free port because we're using it as a pretend mongocryptd server (which would also bind to a free port to listen for incoming connections). I suggest you do some reading on sockets, this page is a great guide: https://beej.us/guide/bgnet/html/split/system-calls-or-bust.html

In particular: https://beej.us/guide/bgnet/html/split/system-calls-or-bust.html#acceptthank-you-for-calling-port-3490

Note the page goes over C-level sockets but Python is very close in API.

test/test_encryption.py
```diff
@@ -1942,6 +1944,47 @@ def test_via_loading_shared_library(self):
         with self.assertRaises(ServerSelectionTimeoutError):
             no_mongocryptd_client.db.command("ping")
 
+    # https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#20-bypass-creating-mongocryptd-client-when-shared-library-is-loaded
+    @unittest.skipUnless(os.environ.get("TEST_CRYPT_SHARED"), "crypt_shared lib is not installed")
+    def test_client_via_loading_shared_library(self):
+        connection_established = [False]
```
blink1073, line 1950: I think it would be more idiomatic to use a boolean, a lock, and a nonlocal, even if you are ensuring the thread is closed before accessing the value.

blink1073 (approved):
LGTM!

test/test_encryption.py
```diff
@@ -1942,6 +1944,51 @@ def test_via_loading_shared_library(self):
         with self.assertRaises(ServerSelectionTimeoutError):
             no_mongocryptd_client.db.command("ping")
 
+    # https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#20-bypass-creating-mongocryptd-client-when-shared-library-is-loaded
+    @unittest.skipUnless(os.environ.get("TEST_CRYPT_SHARED"), "crypt_shared lib is not installed")
+    def test_client_via_loading_shared_library(self):
+        connection_established = False
+        connection_established_lock = Lock()
```
ShaneHarvey, line 1951: The lock is not needed here. Reassigning a variable is already atomic so the lock can be removed.

test/test_encryption.py
```diff
@@ -1942,6 +1944,51 @@ def test_via_loading_shared_library(self):
...
+        listener_t = Thread(target=listener)
+        listener_t.start()
+        key_vault = client_context.client.keyvault.datakeys
+        key_vault.drop()
+        key_vault.create_index(
+            "keyAltNames",
+            unique=True,
+            partialFilterExpression={"keyAltNames": {"$exists": True}},
+        )
+        key_vault.insert_one(json_data("external", "external-key.json"))
```
ShaneHarvey, line 1974: Use create_key_vault() to ensure this uses w="majority". It probably makes sense update create_key_vault to create the index too.

ShaneHarvey (changes requested):

test/test_encryption.py

ShaneHarvey (approved):

PR #1140, opened 2023-01-23
PYTHON-3280 Support for Range Indexes

juliusgeo:


pymongo/encryption.py
```diff
@@ -682,6 +704,74 @@ def encrypt(
...
+            :class:`~bson.binary.Binary` with subtype 4 (
+            :attr:`~bson.binary.UUID_SUBTYPE`).
+          - `key_alt_name`: Identifies a key vault document by 'keyAltName'.
+          - `query_type` (str): **(BETA)** The query type to execute. See
+            :class:`QueryType` for valid options.
+          - `contention_factor` (int): **(BETA)** The contention factor to use
+            when the algorithm is :attr:`Algorithm.INDEXED`.  An integer value
+            *must* be given when the :attr:`Algorithm.INDEXED` algorithm is
+            used.
+          - `range_options`: An instance of EncryptionRangeOptions.
```
blink1073, line 740: We should use the `**BETA**` consistently here and the other places we mention the new feature.

test/test_encryption.py
```diff
@@ -2459,5 +2463,100 @@ def MongoClient(**kwargs):
...
+            bypass_query_analysis=True,
+        )
+        self.encrypted_client = rs_or_single_client(auto_encryption_opts=opts)
+        self.addCleanup(self.encrypted_client.close)
+
+    def run_test(self, name, range_opts, cast_func):
+        encrypted_fields = json_data("etc", "data", f"range-encryptedFields-{name}.json")
+        self.db.drop_collection("explicit_encryption")
+        self.db.create_collection("explicit_encryption", encryptedFields=encrypted_fields)
+        self.encrypted_client.db.explicit_encryption.insert_many(
```
blink1073, line 2495: Can you please add a "case" comment for each section?

pymongo/encryption.py
```diff
@@ -430,6 +441,9 @@ class QueryType(str, enum.Enum):
     EQUALITY = "equality"
     """Used to encrypt a value for an equality query."""
 
+    RANGEPREVIEW = "rangePreview"
+    """Used to encrypt a value for a Range query."""
```
ShaneHarvey, line 445: Range -> range

pymongo/encryption.py
```diff
@@ -682,6 +704,74 @@ def encrypt(
...
+                value=doc,
+                algorithm=algorithm,
+                key_id=key_id,
+                key_alt_name=key_alt_name,
+                query_type=query_type,
+                contention_factor=contention_factor,
+                range_opts=range_options,
+                is_expression=True,
             )
             return decode(encrypted_doc)["v"]  # type: ignore[index]
```
ShaneHarvey, line 776: I think we need to use a raw bson codec_options in order to preserve the BSON here.

pymongo/encryption.py
```diff
@@ -682,6 +704,74 @@ def encrypt(
                 key_alt_name=key_alt_name,
                 query_type=query_type,
                 contention_factor=contention_factor,
+                range_opts=range_options,
+            )
+            return decode(encrypted_doc)["v"]  # type: ignore[index]
+
+    def encrypt_expression(
```
ShaneHarvey, line 711: Ideally encrypt() and encrypt_expression() can be refactored to share the same logic since the only difference is the type of the return value and passing is_expression=True/False to pymongocrypt.

pymongo/encryption_options.py
```diff
@@ -219,3 +219,44 @@ def __init__(
...
+        precision: Optional[int] = None,
+    ) -> None:
+        """Options to configure encrypted queries using the rangePreview algorithm.
+
+        .. note:: Support for Range queries is in beta.
+           Backwards-breaking changes may be made before the final release.
+
+        :Parameters:
+          - `min`: An Extended JSON Strict expression for a numeric type as a BSON Document like ``{ "$numberDouble": "0" }``.
+          - `max`: An Extended JSON Strict expression as a BSON Document like ``{ "$numberDouble": "0" }``.
```
ShaneHarvey, line 241: These should not be mappings or JSON. The type is any BSON scalar value:
```
   // min is required if precision is set.
   min: Optional<BSONValue>,
   // max is required if precision is set.
   max: Optional<BSONValue>,
```

pymongo/encryption_options.py
```diff
@@ -219,3 +219,44 @@ def __init__(
...
         self._kms_ssl_contexts = _parse_kms_tls_options(kms_tls_options)
         self._bypass_query_analysis = bypass_query_analysis
+
+
+class EncryptionRangeOpts:
+    """Options to configure encrypted queries using the rangePreview algorithm."""
+
+    def __init__(
+        self,
+        sparsity: int64,
```
ShaneHarvey, line 229: I think it's fine if we say int here and still do the int -> int64 conversion internally.

pymongo/encryption_options.py
```diff
@@ -219,3 +219,44 @@ def __init__(
...
+    ) -> None:
+        """Options to configure encrypted queries using the rangePreview algorithm.
+
+        .. note:: Support for Range queries is in beta.
+           Backwards-breaking changes may be made before the final release.
+
+        :Parameters:
+          - `min`: An Extended JSON Strict expression for a numeric type as a BSON Document like ``{ "$numberDouble": "0" }``.
+          - `max`: An Extended JSON Strict expression as a BSON Document like ``{ "$numberDouble": "0" }``.
+          - `sparsity`: An integer.
```
ShaneHarvey, line 242: Sparsity should be first since it's the one required argument.

pymongo/encryption_options.py
```diff
@@ -219,3 +219,44 @@ def __init__(
         # Maps KMS provider name to a SSLContext.
         self._kms_ssl_contexts = _parse_kms_tls_options(kms_tls_options)
         self._bypass_query_analysis = bypass_query_analysis
+
+
+class EncryptionRangeOpts:
```
ShaneHarvey, line 224: Since EncryptionRangeOpts is always used with encrypt methods (ClientEncryption.encrypt/ClientEncryption.encryptExpression), can we rename it to "RangeOpts"?

Could we also rename the parameters from "range_options" -> "range_opts" to match the "opts" using in the class name?

ShaneHarvey (changes requested):

test/test_encryption.py
```diff
@@ -2494,5 +2494,195 @@ def MongoClient(**kwargs):
         client_encryption.close()
 
 
+# https://github.com/mongodb/specifications/blob/d4c9432/source/client-side-encryption/tests/README.rst#explicit-encryption
```
blink1073, line 2497: I think this link needs a more recent sha, I don't see these tests in that commit.

test/test_encryption.py
```diff
@@ -2494,5 +2494,195 @@ def MongoClient(**kwargs):
...
+                "greater than or equal to the minimum value and less than or equal to the maximum value",
+            ):
+                self.client_encryption.encrypt(
+                    cast_func(201),
+                    key_id=self.key1_id,
+                    algorithm=Algorithm.RANGEPREVIEW,
+                    contention_factor=0,
+                    range_opts=range_opts,
+                )
+            # Case 7.
```
blink1073, line 2627: Add a new line between cases

blink1073 (changes requested):

test/test_encryption.py

test/test_encryption.py
```diff
@@ -2494,5 +2494,196 @@ def MongoClient(**kwargs):
...
+        )
+        if use_expr:
+            find_payload = {"$expr": find_payload}
+        sorted_find = sorted(
+            self.encrypted_client.db.explicit_encryption.find(find_payload), key=lambda x: x["_id"]
+        )
+        for elem, expected in zip(sorted_find, expected_elems):
+            self.assertEqual(elem[f"encrypted{name}"], expected)
+
+    def run_test(self, name, range_opts, cast_func):
```
blink1073, line 2539: I think "run_test_cases" would be a clearer name for this function.

blink1073 (approved):
LGTM!

pymongo/encryption.py
```diff
@@ -416,6 +416,14 @@ class Algorithm(str, enum.Enum):
 
     .. versionadded:: 4.2
     """
+    RANGEPREVIEW = "RangePreview"
+    """Unindexed.
```
ShaneHarvey, line 420: Unindexed->RangePreview

pymongo/encryption_options.py
```diff
@@ -219,3 +219,44 @@ def __init__(
...
+          - `precision`: An integer, may only be set for double or decimal128 types.
+
+        .. versionadded:: 4.4
+        """
+        self.min = min
+        self.max = max
+        self.sparsity = sparsity
+        self.precision = precision
+
+    def as_doc(self):
```
ShaneHarvey, line 252: ReadConcern, WriteConcern, and Collection all use the "document" property for this:
```
    @property
    def document(self) -> Dict[str, Any]:
```
Can we use that same pattern here?

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #1141, opened 2023-01-24
PYTHON-3523 Resync unified test format tests for getnonce

ShaneHarvey:


test/test_auth.py
```diff
@@ -581,7 +581,9 @@ def test_scram_threaded(self):
         coll.insert_one({"_id": 1})
 
         # The first thread to call find() will authenticate
-        coll = rs_or_single_client().db.test
+        client = rs_or_single_client()
+        self.addCleanup(client.close)
+        coll = client.db.test
```
ShaneHarvey, line 586: Note this is unrelated to the ticket. Just a cleaned to avoid having the client left running.

blink1073 (approved):
Thanks!

PR #1143, opened 2023-01-25
PYTHON-3558 Missing docs for JSONOptions

juliusgeo:
Weirdly enough sphinx does not recognize the docstring if you put it in `__new__`, so I had to define an `__init__`  function for the docstring, but it does not do anything because the class is immutable.

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -265,7 +265,7 @@ class JSONOptions(CodecOptions):
         .. versionchanged:: 4.0
            Changed default value of `tz_aware` to False.
         """
-        return super().__init__(*args, **kwargs)
+        pass
 
     def __new__(
         cls: Type["JSONOptions"],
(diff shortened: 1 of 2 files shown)


bson/json_util.py
```diff
@@ -205,66 +205,68 @@ class JSONMode:
...
+           :const:`~bson.binary.UuidRepresentation.PYTHON_LEGACY` to
+           :const:`~bson.binary.UuidRepresentation.UNSPECIFIED`.
+
+        .. versionchanged:: 3.5
+           Accepts the optional parameter `json_mode`.
+
+        .. versionchanged:: 4.0
+           Changed default value of `tz_aware` to False.
+        """
+        pass
```
ShaneHarvey, line 268: Let's add in the super call to be consistent with CodecOptions.

ShaneHarvey (changes requested):

blink1073 (approved):

PR #1145, opened 2023-01-27
PYTHON-3357 Automatically create Queryable Encryption keys

juliusgeo:


pymongo/database.py
```diff
@@ -292,6 +292,147 @@ def get_collection(
...
+                f"{self.name}.{coll_name}"
+            ]
+        if ask_db:
+            colls = list(self.list_collections(filter={"name": coll_name}, **kwargs))
+            if colls and colls[0]["options"].get("encryptedFields"):
+                return colls[0]["options"]["encryptedFields"]
+        return None
+
+    @_csot.apply
+    def create_encrypted_collection(
```
ShaneHarvey, line 316: This should be defined on ClientEncryption, not database: https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#id13

pymongo/database.py
```diff
@@ -292,6 +292,147 @@ def get_collection(
             read_concern,
         )
 
+    def _get_encrypted_fields(self, kwargs, coll_name, ask_db):
```
ShaneHarvey, line 295: Are there any existing places where we could use this method to avoid duplication?

test/test_encryption.py
```diff
@@ -2687,5 +2687,79 @@ def test_int(self):
...
+            {"local": {"key": LOCAL_MASTER_KEY}},
+            key_vault.full_name,
+            bypass_query_analysis=True,
+        )
+        self.encrypted_client = rs_or_single_client(auto_encryption_opts=opts)
+        self.encrypted_client.drop_database("db")
+        self.db = self.encrypted_client.db
+        self.addCleanup(self.encrypted_client.close)
+
+    def test_simple_create(self):
```
ShaneHarvey, line 2716: Let's put the prose test numbers in the test name like we do in other prose tests, eg `test_01_simple_create`.

test/test_encryption.py
```diff
@@ -2687,5 +2687,79 @@ def test_int(self):
...
+        coll, _ = self.db.create_encrypted_collection(
+            name="testing1",
+            encryptedFields={"fields": [{"path": "ssn", "bsonType": "string", "keyId": None}]},
+            kms_provider="local",
+            client_encryption=self.client_encryption,
+        )
+        with self.assertRaisesRegex(WriteError, "schemaRulesNotSatisfied"):
```
ShaneHarvey, line 2723: Can we use an error code instead of a regex string here? The server's error message may change but the code should be stable.

pymongo/database.py
```diff
@@ -292,6 +292,147 @@ def get_collection(
...
+            for field in encrypted_fields["fields"]:
+                if isinstance(field, dict) and field.get("keyId") is None:
+                    try:
+                        field["keyId"] = client_encryption.create_data_key(
+                            kms_provider=kms_provider, **data_key_opts
+                        )
+                    except WriteError as exc:
+                        raise EncryptionError(
+                            f"Error occured while creating data key with encryptedFields={str(encrypted_fields)}"
+                        ) from exc
```
ShaneHarvey, line 423: Is this try/catch needed? create_data_key already raises an EncryptionError.

pymongo/database.py
```diff
@@ -292,6 +292,147 @@ def get_collection(
...
+        else:
+            raise EncryptionError(
+                "'encryptedFields' must be provided as a keyword argument to create_encrypted_collection"
+                "or initialized with AutoEncryptionOpts."
+            )
+
+        return self.create_collection(
+            name=name,
+            **kwargs,
+        ), kwargs.get("encryptedFields")
```
ShaneHarvey, line 434: "kwargs.get("encryptedFields")" -> "encrypted_fields"?

ShaneHarvey (changes requested):

test/test_encryption.py
juliusgeo: Done.

test/test_encryption.py
juliusgeo: Done.

pymongo/encryption.py
```diff
@@ -552,6 +554,128 @@ def __init__(
...
+            enabling pre- and post-images.
+
+        .. versionadded:: 4.4
+           Added the codec_options, read_preference, and write_concern options.
+
+        .. _create collection command:
+            https://mongodb.com/docs/manual/reference/command/create
+        """
+        encrypted_fields = database._get_encrypted_fields(kwargs, name, False)
+        if encrypted_fields:
```
ShaneHarvey, line 657: Spec says to create a copy of encrypted_fields. Although I wonder about that. If we find the encrypted_fields from auto_encryption_options without a keyId and generate the new data key but don't update the auto_encryption_options, then will auto encrypted operations still work? Maybe it will work for a time because of libmongocrypt's caching logic but what happens when the cache expires?

pymongo/encryption.py
```diff
@@ -552,6 +554,128 @@ def __init__(
...
+            for field in encrypted_fields["fields"]:
+                if isinstance(field, dict) and field.get("keyId") is None:
+                    try:
+                        if kms_provider:
+                            field["keyId"] = self.create_data_key(
+                                kms_provider=kms_provider, **{data_key_opts or {}}
+                            )
+                    except WriteError as exc:
```
ShaneHarvey, line 665: IIUC create_data_key will never raise WriteError, it only ever raises EncryptionError.

pymongo/encryption.py
```diff
@@ -552,6 +554,128 @@ def __init__(
...
+                        raise EncryptionError(
+                            f"Error occured while creating data key with encryptedFields={str(encrypted_fields)}"
+                        ) from exc
```
ShaneHarvey, line 668: Let's include the field that failed in the error as well.

pymongo/encryption.py
```diff
@@ -552,6 +554,128 @@ def __init__(
...
+                "or initialized with AutoEncryptionOpts."
+            )
+
+        return (
+            database.create_collection(
+                name=name,
+                **kwargs,
+            ),
+            encrypted_fields,
+        )
```
ShaneHarvey, line 682: I think this entire return could fit on one line.

pymongo/encryption.py
```diff
@@ -552,6 +554,128 @@ def __init__(
...
+                          "keyId": Binary.from_uuid(UUID('04104104-1041-0410-4104-104104104104')),
+                          "bsonType": "string"
+                      }
+                    ]
+                }
+          - `**kwargs` (optional): additional keyword arguments are the same as "create_collection".
+
+        All optional `create collection command`_ parameters should be passed
+        as keyword arguments to this method. Valid options include, but are not
+        limited to:
```
ShaneHarvey, line 608: Can we instead just link to the create_collection docs and say that this method supports those arguments as well? Similar to find_raw_batches and find.

pymongo/encryption.py
```diff
@@ -552,6 +554,128 @@ def __init__(
...
+                  "fields": [
+                      {
+                          "path": "firstName",
+                          "keyId": Binary.from_uuid(UUID('00000000-0000-0000-0000-000000000000')),
+                          "bsonType": "string",
+                          "queries": {"queryType": "equality"}
+                      },
+                      {
+                          "path": "ssn",
+                          "keyId": Binary.from_uuid(UUID('04104104-1041-0410-4104-104104104104')),
```
ShaneHarvey, line 599: We should note that keyId can be omitted to autogenerate the key.

test/test_encryption.py
```diff
@@ -2687,5 +2687,80 @@ def test_int(self):
         self.run_test_cases("Int", RangeOpts(min=0, max=200, sparsity=1), int)
 
 
+# https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#automatic-data-encryption-keys
+class TestAutomaticDecryptionKeys(EncryptionIntegrationTest):
+    @client_context.require_no_standalone
+    @client_context.require_version_min(6, 2, -1)
```
ShaneHarvey, line 2693: 6.2 -> 6.0

test/test_encryption.py
```diff
@@ -2687,5 +2687,80 @@ def test_int(self):
...
+        )
+        self.addCleanup(self.client_encryption.close)
+        opts = AutoEncryptionOpts(
+            {"local": {"key": LOCAL_MASTER_KEY}},
+            key_vault.full_name,
+            bypass_query_analysis=True,
+        )
+        self.encrypted_client = rs_or_single_client(auto_encryption_opts=opts)
+        self.encrypted_client.drop_database("db")
+        self.db = self.encrypted_client.db
```
ShaneHarvey, line 2713: The prose test does not say that DB should be an encrypted client. I think it should be a regular client.

pymongo/database.py
```diff
@@ -292,6 +292,28 @@ def get_collection(
...
+                f"{self.name}.{coll_name}"
+            )
+        ):
+            return self.client.options.auto_encryption_opts._encrypted_fields_map[
+                f"{self.name}.{coll_name}"
+            ]
+        if ask_db and self.client.options.auto_encryption_opts:
+            colls = list(
+                self.list_collections(filter={"name": coll_name}, session=session, comment=comment)
+            )
```
ShaneHarvey, line 312: Let's use collection.options() here.

pymongo/database.py
```diff
@@ -292,6 +292,28 @@ def get_collection(
...
+            if colls and colls[0]["options"].get("encryptedFields"):
+                return colls[0]["options"]["encryptedFields"]
```
ShaneHarvey, line 314: This can be:
```
            if colls:
                return colls[0]["options"].get("encryptedFields")

```

test/test_encryption.py
```diff
@@ -2687,5 +2687,80 @@ def test_int(self):
...
+            encryptedFields={"fields": [{"path": "ssn", "bsonType": "string", "keyId": None}]},
+            kms_provider="local",
+        )
+        key1_id = ef["fields"][0]["keyId"]
+        encrypted_value = self.client_encryption.encrypt(
+            "123-45-6789",
+            key_id=key1_id,
+            algorithm=Algorithm.UNINDEXED,
+        )
+        coll.insert_one({"ssn": encrypted_value})
```
ShaneHarvey, line 2762: The testing here is pretty light. Let's add tests for:
- that the collection has the correct db and coll name.
- that custom create_collection options are forwarded to the create_collection call.
- that encryptedFields is mutated or not (depending on what we decide is the right behavior).
- mixed missing/None/existing keyId fields in the same encryptedFields.
- the expected behavior when a create_data_key fails in the middle of a call.
- Any other cases, basically any case there's a loop or a branch in the code.

juliusgeo: Done!

pymongo/encryption.py
juliusgeo: Fixed.

pymongo/encryption.py
juliusgeo: It was clarified that we do in fact want to create a copy. This is now done.

pymongo/encryption.py
juliusgeo: Done.

pymongo/database.py
juliusgeo: Done.

pymongo/database.py
```diff
@@ -292,6 +292,147 @@ def get_collection(
...
+    def create_encrypted_collection(
+        self,
+        name: str,
+        client_encryption: Any,
+        kms_provider: Optional[str] = None,
+        data_key_opts: Any = {},
+        **kwargs: Any,
+    ) -> Collection[_DocumentType]:
+        """Create a new :class:`~pymongo.collection.Collection` in this
+        database with encryptedFields.
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/database.py
juliusgeo: Done.

pymongo/database.py
juliusgeo: Done.

pymongo/encryption.py
juliusgeo: Done.

pymongo/encryption.py
juliusgeo: Done.

pymongo/encryption.py
juliusgeo: Done.

pymongo/database.py
```diff
@@ -292,6 +292,147 @@ def get_collection(
...
+
+    @_csot.apply
+    def create_encrypted_collection(
+        self,
+        name: str,
+        client_encryption: Any,
+        kms_provider: Optional[str] = None,
+        data_key_opts: Any = {},
+        **kwargs: Any,
+    ) -> Collection[_DocumentType]:
```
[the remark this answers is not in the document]
juliusgeo: Done.

pymongo/database.py
```diff
@@ -292,6 +292,26 @@ def get_collection(
             read_concern,
         )
 
+    def _get_encrypted_fields(self, kwargs, coll_name, ask_db):
+        encrypted_fields = kwargs.get("encryptedFields")
+        if encrypted_fields:
+            return encrypted_fields.copy()
```
ShaneHarvey, line 298: These should be deep copies otherwise we'll still be mutating the underlying dictionaries.

pymongo/encryption.py
```diff
@@ -552,6 +554,95 @@ def __init__(
...
+                          "bsonType": "string"
+                      }
+                    ]
+                }
+
+            The "keyId" may be set to `None` to auto-generate the data keys.
```
ShaneHarvey, line 606: Here and above:
```
`None`->``None``
```

pymongo/encryption.py
```diff
@@ -552,6 +554,95 @@ def __init__(
...
+
+        Normally collection creation is automatic. This method should
+        only be used to specify options on
+        creation. :class:`~pymongo.errors.CollectionInvalid` will be
+        raised if the collection already exists.
+
+        :Parameters:
+          - `name`: the name of the collection to create
+          - `kms_provider`: the KMS provider to be used
+          - `data_key_opts` (dict): a dictionary containing additional arguments to the `create_data_key` method that will be
```
ShaneHarvey, line 575: Line length.

test/test_encryption.py
```diff
@@ -2687,5 +2687,162 @@ def test_int(self):
...
+
+    def test_copy_encrypted_fields(self):
```
ShaneHarvey, line 2762: This test doesn't catch the fact that we are still mistakenly mutating the encrypted_fields_map.

test/test_encryption.py
```diff
@@ -2687,5 +2687,162 @@ def test_int(self):
...
+                ]
+            }
+        }
+        opts = AutoEncryptionOpts(
+            {"local": {"key": LOCAL_MASTER_KEY}},
+            self.key_vault.full_name,
+            bypass_query_analysis=True,
+            encrypted_fields_map=encrypted_fields,
+        )
+        client = rs_or_single_client(auto_encryption_opts=opts)
```
ShaneHarvey, line 2780: use addCleanup to close the client.

test/test_encryption.py
```diff
@@ -2687,5 +2687,162 @@ def test_int(self):
...
+
+    def test_mixed_null_keyids(self):
+        key = self.client_encryption.create_data_key(kms_provider="local")
+        coll, ef = self.client_encryption.create_encrypted_collection(
+            database=self.db,
+            name="testing1",
+            encryptedFields={
+                "fields": [
+                    {"path": "ssn", "bsonType": "string", "keyId": None},
+                    {"path": "dob", "bsonType": "string", "keyId": key},
```
ShaneHarvey, line 2808: Can we add a field that omits keyId?

pymongo/encryption.py
```diff
@@ -552,6 +554,95 @@ def __init__(
...
+                if isinstance(field, dict) and field.get("keyId") is None:
+                    try:
+                        field["keyId"] = self.create_data_key(
+                            kms_provider=kms_provider,  # type:ignore[arg-type]
+                            **data_key_opts,
+                        )
+                    except EncryptionError as exc:
+                        raise EncryptionError(
+                            Exception(
+                                f"Error occurred while creating data key for field {field['path']} with encryptedFields={str(encrypted_fields)}"
```
ShaneHarvey, line 634: I'd recommend using repr-style formatting for path and encrypted_fields here.

pymongo/encryption.py
```diff
@@ -552,6 +554,95 @@ def __init__(
...
+          - `name`: the name of the collection to create
+          - `kms_provider`: the KMS provider to be used
+          - `data_key_opts` (dict): a dictionary containing additional arguments to the `create_data_key` method that will be
+            used to auto-generate the data keys if `keyId` is `None` such as::
+
+                {
+                   masterKey: Optional<Document>
+                   keyAltNames: Optional<Array[String]>
+                   keyMaterial: Optional<BinData>
+                }
```
ShaneHarvey, line 582: These params are actually snake_case not camelCase. What do you think about exposing these explicitly instead of via a dict? Like:
```
    def create_encrypted_collection(
        self,
        database: Database,
        name: str,
        kms_provider: Optional[str] = None,
        master_key: Optional[Mapping[str, Any]] = None,
        key_alt_names: Optional[Sequence[str]] = None,
        key_material: Optional[bytes] = None,
```

pymongo/encryption.py
```diff
@@ -552,6 +554,95 @@ def __init__(
...
+                            )
+                        ) from exc
+            kwargs["encryptedFields"] = encrypted_fields
+        else:
+            raise ConfigurationError(
+                "'encryptedFields' must be provided as a keyword argument to create_encrypted_collection"
+                "or initialized with AutoEncryptionOpts."
+            )
+
+        return database.create_collection(name=name, **kwargs), encrypted_fields
```
ShaneHarvey, line 644: New requirement just came in to return the encrypted_fields when create fails as well: https://github.com/mongodb/specifications/pull/1375

ShaneHarvey (changes requested):

pymongo/encryption.py
```diff
@@ -552,6 +554,95 @@ def __init__(
...
+        .. versionadded:: 4.4
+           Added the codec_options, read_preference, and write_concern options.
+
+        .. _create collection command:
+            https://mongodb.com/docs/manual/reference/command/create
+
+        """
+        if not data_key_opts:
+            data_key_opts = {}
+        encrypted_fields = database._get_encrypted_fields(kwargs, name, False)
```
kevinAlbs, line 622: https://github.com/mongodb/specifications/pull/1376 proposes not checking AutoEncryptionOptions in createEncryptedCollection. This may need to be updated when that is merged.

test/test_encryption.py
juliusgeo: Done.

test/test_encryption.py
juliusgeo: Done.

test/test_encryption.py
juliusgeo: Fixed.

pymongo/encryption.py
juliusgeo: Done.

pymongo/encryption.py
juliusgeo: Done.

pymongo/encryption.py
juliusgeo: Thanks for the heads-up!

pymongo/encryption.py
juliusgeo: Done.

pymongo/database.py
juliusgeo: Done and added fixed the test.

pymongo/encryption.py
juliusgeo: Done.

pymongo/encryption.py
juliusgeo: Done.

test/test_encryption.py
ShaneHarvey: Not done. Both these fields have `"keyId": None` but we should test when the "keyId" field is missing (not None).

test/test_encryption.py
```diff
@@ -2687,5 +2687,180 @@ def test_int(self):
...
+            {"local": {"key": LOCAL_MASTER_KEY}},
+            self.key_vault.full_name,
+            self.key_vault_client,
+            OPTS,
+        )
+        self.addCleanup(self.client_encryption.close)
+        self.client = rs_or_single_client()
+        self.client.drop_database("db")
+        self.db = self.client.db
+        self.addCleanup(self.client.close)
```
ShaneHarvey, line 2712: I think we can remove these last 4 lines and reuse the existing self.client (ClientContext.client).

test/test_encryption.py
```diff
@@ -2687,5 +2687,180 @@ def test_int(self):
...
+        )
+        coll.insert_one({"ssn": encrypted_value})
+
+    def test_copy_encrypted_fields(self):
+        encrypted_fields = {
+            "fields": [
+                {
+                    "path": "ssn",
+                    "bsonType": "string",
+                    "keyId": self.client_encryption.create_data_key(kms_provider="local"),
```
ShaneHarvey, line 2770: keyId here should be None so that create_encrypted_collection creates a key.

test/test_encryption.py
```diff
@@ -2687,5 +2687,180 @@ def test_int(self):
...
+                }
+            ]
+        }
+        _, ef = self.client_encryption.create_encrypted_collection(
+            database=self.db,
+            name="testing1",
+            kms_provider="local",
+            encrypted_fields=encrypted_fields,
+        )
+        self.assertIsNot(
```
ShaneHarvey, line 2780: Instead of assertIsNot, let's assert that keyId is still None in `encrypted_fields` and not None in `ef`.

ShaneHarvey (changes requested):

test/test_encryption.py
juliusgeo: Ah, that makes sense. Fixed.

test/test_encryption.py
juliusgeo: Done.

test/test_encryption.py
juliusgeo: Done.

test/test_encryption.py

pymongo/encryption.py
```diff
@@ -552,6 +555,98 @@ def __init__(
...
+        master_key: Optional[Mapping[str, Any]] = None,
+        key_alt_names: Optional[Sequence[str]] = None,
+        key_material: Optional[bytes] = None,
+        **kwargs: Any,
+    ) -> Tuple[Collection[_DocumentType], Mapping[str, Any]]:
+        """Create a collection with encryptedFields.
+
+        Normally collection creation is automatic. This method should
+        only be used to specify options on
+        creation. :class:`~pymongo.errors.CollectionInvalid` will be
```
ShaneHarvey, line 573: This is no longer true since we always raise EncryptionError now.

ShaneHarvey (changes requested):

pymongo/encryption.py
juliusgeo: Done.

pymongo/encryption.py
blink1073: The referenced PR has been merged

juliusgeo: I have already committed my fix for this.

blink1073: I think you also need to add the associated docstring clarification.

juliusgeo: The docstring now shows encryption_fields as a required argument and I removed the note in the error message relating to this.

pymongo/encryption.py
```diff
@@ -552,6 +555,101 @@ def __init__(
...
+                    raise EncryptionError(
+                        Exception(
+                            "Error occurred while creating data key for field %s with encryptedFields=%s"
+                            % (field["path"], encrypted_fields)
+                        )
+                    ) from exc
+        kwargs["encryptedFields"] = encrypted_fields
+        try:
+            return (
+                database.create_collection(check_exists=False, name=name, **kwargs),
```
ShaneHarvey, line 643: This should be passed via kwargs:
```
kwargs["encryptedFields"] = encrypted_fields
kwargs["check_exists"] = False
...
```

Otherwise this will fail:
```
db.create_encrypted_collection(..., check_exists=False)
```

test/test_encryption.py
ShaneHarvey: This is still not fixed.

test/test_encryption.py
```diff
@@ -2687,5 +2687,208 @@ def test_int(self):
...
+            ]
+        }
+        self.db.create_collection("testing1")
+        with self.assertRaisesRegex(
+            EncryptionError,
+            "while creating collection with encryptedFields=.*keyId.*Binary",
+        ):
+            self.client_encryption.create_encrypted_collection(
+                database=self.db,
+                name="testing1",  # type:ignore[arg-type]
```
ShaneHarvey, line 2870: Unused type:ignore

ShaneHarvey (changes requested):

pymongo/encryption.py
juliusgeo: Done.

test/test_encryption.py
juliusgeo: The next test contains the test case you mentioned--line 2828

test/test_encryption.py

test/test_encryption.py
juliusgeo: I will move it to this test because I think it makes more sense here. I also made the test use all of the fields that I added.

ShaneHarvey: Ack.

ShaneHarvey (approved):

pymongo/encryption.py
blink1073: I was referring to: "Drivers MUST document that `createEncryptedCollection` does not affect any
auto encryption settings on existing MongoClients that are already configured with 
auto encryption.  Users must configure auto encryption after creating the 
encrypted collection with the `createEncryptedCollection` helper."

blink1073 (approved):

PR #1146, opened 2023-01-31
PYTHON-3577 Fix test_aggregate_out on 4.0 replica set

ShaneHarvey:
By looking at the server logs I discovered that the 4.0 primary was blocked from running the agg $out operation because a previous test intentionally left a transaction open. I've updated the test to close the transaction on the server so the test runs in a few milliseconds instead of 1 minute now.

### DIFF
--- a/test/test_load_balancer.py
+++ b/test/test_load_balancer.py
@@ -123,12 +123,7 @@ class TestLB(IntegrationTest):
         session.start_transaction()
         client.test_session_gc.test.find_one({}, session=session)
         # Cleanup the transaction left open on the server.
-        session_id = session.session_id
-
-        def kill_session():
-            self.client.admin.command("killSessions", [session_id])
-
-        self.addCleanup(kill_session)
+        self.addCleanup(self.client.admin.command, "killSessions", [session.session_id])
         if client_context.load_balancer:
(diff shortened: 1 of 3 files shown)


test/test_load_balancer.py
```diff
@@ -122,6 +122,13 @@ def test_session_gc(self):
         session = client.start_session()
         session.start_transaction()
         client.test_session_gc.test.find_one({}, session=session)
+        # Cleanup the transaction left open on the server.
+        session_id = session.session_id
```
juliusgeo, line 126: This variable is only used once, and the name is pretty self-explanatory. Could it be combined with the next to do:
```
        def kill_session():
             self.client.admin.command("killSessions", [session.session_id])
```
Or, even better maybe would be:
```
self.addCleanup(lambda: self.client.admin.command("killSessions", [session.session_id]))
```

ShaneHarvey: No because session is intentionally garbage collected in this test.

blink1073 (approved):

test/test_load_balancer.py
ShaneHarvey: Oh yes it's possible on second thought. Done.

PR #1147, opened 2023-02-03
PYTHON-3589 createEncryptedCollection should not accept keyAltNames

juliusgeo:


test/test_encryption.py
```diff
@@ -2835,12 +2835,9 @@ def test_create_datakey_fails(self):
                     "fields": [
                         {"path": "address", "bsonType": "string", "keyId": key},
                         {"path": "dob", "bsonType": "string", "keyId": None},
-                        # Because this is the second one to use the altName "1", it will fail when creating the data_key.
```
juliusgeo, line 2838: Unfortunately, because of this removal of `key_alt_names` we cannot make it fail on the second data key creation anymore.

blink1073 (approved):

test/test_encryption.py
ShaneHarvey: Fair enough.

ShaneHarvey (approved):
LGTM!

ShaneHarvey (approved):

PR #1149, opened 2023-02-06
PYTHON-3596 Guarantee a document update in retryable writes tests

ShaneHarvey:
The server no longer performs a write when running `replace_one({}, {})` on an empty document so I've updated the test operations to ensure a new document needs to be written.


test/test_retryable_writes.py
```diff
@@ -490,6 +490,7 @@ def setUpClass(cls):
         }
 
     @client_context.require_version_min(4, 0)
+    @client_knobs(heartbeat_frequency=0.05, min_heartbeat_interval=0.05)
```
ShaneHarvey, line 493: This was just a slow test I noticed. Decreasing these knobs decreases the runtime from 10 seconds to a few milliseonds.

blink1073 (approved):

PR #1150, opened 2023-02-08
Feature/pipeline helper

caeser1996:
The PipelineHelper class provides a set of static methods that can be used to write and build queries in MongoDB. These methods correspond to the various stages in the MongoDB aggregation pipeline, such as $project, $match, $group, $sort, $skip, $limit, and $unwind.

By using these methods, you can simplify the process of writing and building aggregation queries in MongoDB. Instead of writing the pipeline stages manually, you can use the pre-defined methods to construct the stages and create the query.

For example, consider the following code that uses the PipelineHelper class to build a query that projects a subset of fields, matches documents based on certain criteria, groups the documents based on a specified field, and calculates the sum of another field.

```
from pymongo import PipelineHelper
query = [
    PipelineHelper.project({"field1": 1, "field2": 1}),
    PipelineHelper.match({"field1": {"$gt": 10}}),
    PipelineHelper.group({"_id": "$field2"}, {"sum_field1": {"$sum": "$field1"}}),
    PipelineHelper.sort({"sum_field1": -1})
]
client = MongoClient("mongodb://localhost:27017/")
db = client["test_db"]
collection = db["test_collection"]
pipeline = [
    PipelineHelper.match({"status": "A"}),
    PipelineHelper.group({"_id": "$status"}, {"count": {"$sum": 1}}),
    PipelineHelper.sort({"count": -1}),
    PipelineHelper.limit(1)
]
result = collection.aggregate(pipeline)
```

### DIFF
--- a/pymongo/__init__.py
+++ b/pymongo/__init__.py
@@ -14,6 +14,24 @@
 
 """Python driver for MongoDB."""
 
+from pymongo.write_concern import WriteConcern
+from pymongo.read_preferences import ReadPreference
+from pymongo.operations import (
+    DeleteMany,
+    DeleteOne,
+    IndexModel,
+    InsertOne,
+    ReplaceOne,
+    UpdateMany,
+    UpdateOne,
+)
+from pymongo.mongo_client import MongoClient
+from pymongo.pipeline_helper import PipelineHelper
+from pymongo.cursor import CursorType
+from pymongo.common import MAX_SUPPORTED_WIRE_VERSION, MIN_SUPPORTED_WIRE_VERSION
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
Thanks @caeser1996 but we're not going to merge this API. I'd recommend publishing this as it's own library if you (or others) want to reuse it in multiple projects. In the future we may add an api in a similar spirit to this but it would be a concerted effort across multiple driver languages and would likely look very different.

caeser1996:
I can write it for all other languages if required

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #1152, opened 2023-02-10
PYTHON-3562 Type annotation of `bson.json_utils.loads` is incorrect (only accepts `str`)

juliusgeo:


ShaneHarvey (changes requested):
Could we add a test for bytes if we don't have one already?

juliusgeo:
I added a super simple test that tests that we can load these types.

test/test_json_util.py
```diff
@@ -71,6 +71,11 @@ def round_trip(self, doc, **kwargs):
     def test_basic(self):
         self.round_trip({"hello": "world"})
 
+    def test_loads_bytes(self):
+        string = rb'{"hello": "world"}'
+        for type in [bytes, bytearray]:
+            json_util.loads(type(string))
```
ShaneHarvey, line 77: In this case using a list makes the code less understandable from a type perspective. I suggest:
```
            json_util.loads(b'{"hello": "world"}')
            json_util.loads(bytearrary(b'{"hello": "world"}'))
```

ShaneHarvey: (with assertions that the loaded value is as expected).

juliusgeo: Done.

ShaneHarvey (approved):

PR #1155, opened 2023-02-13
PYTHON-3533 Permit tlsDisableOCSPEndpointCheck in KMS TLS options

blink1073:


test/test_encryption.py
```diff
@@ -2146,6 +2145,14 @@ def test_04_kmip(self):
         ):
             self.client_encryption_invalid_hostname.create_data_key("kmip")
 
+    def test_05_tlsDisableOCSPEndpointCheck_is_permitted(self):
+        providers = {"aws": {"accessKeyId": "foo", "secretAccessKey": "bar"}}
+        options: dict = dict(aws=dict(tlsdisableocspendpointcheck=True))
```
ShaneHarvey, line 2150: Can this be:
```python
options = {"aws": "tlsDisableOCSPEndpointCheck": True}
```

so that it's consistent with "providers"?

test/test_encryption.py
```diff
@@ -2146,6 +2145,14 @@ def test_04_kmip(self):
...
+        encryption = ClientEncryption(
+            providers, "keyvault.datakeys", self.client, OPTS, kms_tls_options=options
+        )
+        encryption.close()
```
ShaneHarvey, line 2154: Can we test that the option actually gets parsed correctly? Perhaps by inspecting a private attribute?

ShaneHarvey (changes requested):

test/test_encryption.py
blink1073: Done

test/test_encryption.py
blink1073: No, it is not exposed, it gets consumed by an `SSLContext`.

ShaneHarvey: We can check the check_ocsp_endpoint property (assuming the OCSP reqs are installed): https://github.com/mongodb/mongo-python-driver/blob/864812d40093fd1502626b9cc45d62f97a29025a/pymongo/pyopenssl_context.py#L240

blink1073: Done

blink1073: Hmm, it seems as if adding the `ocsp` requirements caused several tests to fail, investigating...

blink1073: I updated a few regexes to account for `pyopenssl` messages.

ShaneHarvey (approved):

PR #1156, opened 2023-02-14
PYTHON-3381 Improve readability of sphinx docs

blink1073:
image

image

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -372,7 +372,6 @@ class JSONOptions(CodecOptions):
             >>> json_options.tz_aware
             False
 
-
         .. versionadded:: 3.12
         """
         opts = self._options_dict()
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -372,6 +372,7 @@ class JSONOptions(CodecOptions):
             >>> json_options.tz_aware
             False
 
+
         .. versionadded:: 3.12
         """
         opts = self._options_dict()
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
     args: [--line-length=100]
 
 - repo: https://github.com/PyCQA/isort
-  rev: 5.10.1
+  rev: 5.12.0
   hooks:
   - id: isort
     files: \.py$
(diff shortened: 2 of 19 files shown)


bson/json_util.py
```diff
@@ -372,6 +372,7 @@ def with_options(self, **kwargs: Any) -> "JSONOptions":
             >>> json_options.tz_aware
             False
 
+
```
juliusgeo, line 375: Here and above seem to be accidental changes.

blink1073: I needed them to avoid sphinx warnings about indentation.

ShaneHarvey:
One problem I noticed is that multiple classes show up twice in the docs now. For example:
- https://pymongo--1156.org.readthedocs.build/en/1156/api/bson/binary.html#bson.binary.Binary
- https://pymongo--1156.org.readthedocs.build/en/1156/api/bson/index.html#bson.Binary

ShaneHarvey:
It looks like the typehints do not show up in the method signature or in the Parameter list: https://pymongo--1156.org.readthedocs.build/en/1156/api/bson/index.html#bson.BSON.encode 

Why is that?

blink1073:
`Binary` was already duplicated in the docs, since it is re-exported.

blink1073:
So perhaps we should not use [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) unless we switch to a standard format.

ShaneHarvey:
> Binary was already duplicated in the docs, since it is re-exported.

Could you open a bug for this? Looks like it was caused by #994.

> So perhaps we should not use [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) unless we switch to a standard format.

SGTM, let's do that in another ticket?

bson/json_util.py
blink1073: Removed now that we're not using `sphinx-autodoc-typehints`.

ShaneHarvey (changes requested):
Love the new theme, thanks for finding it. Did you notice that you can toggle between light/dark themes in the top right?? I do still see a few problems though, like these two issues [on this page](https://pymongo--1156.org.readthedocs.build/en/1156/python3.html): 
Screen Shot 2023-02-15 at 12 00 07 PM

Could you take a look through the pages to see if you find anything else?

blink1073:
I fixed those and a few other places I saw.  I went through all of the docs pages except for the secondary API docs pages.

ShaneHarvey (approved):

PR #1158, opened 2023-02-16
PYTHON-3609 Stop using deprecated setDaemon api in test suite

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3609

setDaemon is the only deprecated api we're using.


blink1073 (approved):
LGTM!

PR #1161, opened 2023-02-18
PYTHON-3511 Remove Python 2 references from the docs

sushiboytr:
[Our Catalog for all of the issues we found and decided to solve!](https://docs.google.com/spreadsheets/d/1TMENw8XC2YLQ_FwavHHocwiJsNKWAb4rTMX8B5Sbni0/edit?usp=sharing)
https://jira.mongodb.org/browse/PYTHON-3511


juliusgeo:
Hi! Thank you so much for this PR. This is super helpful. One question I have is: have you done a search for all referents to `str`, because there may be documentation that does not explicitly mention Python 2 or Python 3 but may be out of date. The change from `str` to `bytes` might have been missed in some cases. I see from the list you used various `grep` strings, but there might be some false negatives as the search strings are quite specific, and I didn't see one for `bytes` or `str`.

juliusgeo:
One other small thing--in `bson/__init__.py:818`, technically all integers in Python are arbitrary precision, not longs. I think we want to reword that to something like "returns an int`.

sushiboytr:
> Hi! Thank you so much for this PR. This is super helpful. One question I have is: have you done a search for all referents to `str`, because there may be documentation that does not explicitly mention Python 2 or Python 3 but may be out of date. The change from `str` to `bytes` might have been missed in some cases. I see from the list you used various `grep` strings, but there might be some false negatives as the search strings are quite specific, and I didn't see one for `bytes` or `str`. 

I did not, I agree I should have thought about that and done a search for those as well. Give me a few days to take care of that and ill update the catalog and PR. I will also make that adjustment to the bson/__init__:818: comment as well. Thank you!

ShaneHarvey:
@sushiboytr are you still planning to work on this PR?

sushiboytr:
> @sushiboytr are you still planning to work on this PR?

@ShaneHarvey Yes I am! Sorry for the delay in updates, just been busy with school work. I am in the process of going thru and finding any references to basestring, str, and bytes and removing/updating any that correlate to a python 2 reference. You should receive a updated PR for review in 1-2 weeks.

ShaneHarvey (approved):

ShaneHarvey:
Thanks @sushiboytr!

PR #1165, opened 2023-03-01
PYTHON-3620 Ensure unittest-xml-reporting is Installed in Evergreen Tasks

blink1073:


ShaneHarvey (commented):
IIUC this will install unittest-xml-reporting even when testing with a system python. Are we okay with that? We usually try not to install packages onto the system as not to mess with other Evergreen tasks.

blink1073:
Should we always create a virtualenv then?

ShaneHarvey:
I think always using a venv would be a great idea but I suspect it will come with some challenges and end up being more work. How about for now we only install unittest-xml-reporting in places where we already create a virtualenv? Then we can open and backlog a ticket to always use venvs.

blink1073:
Done, and https://jira.mongodb.org/browse/PYTHON-3625

ShaneHarvey (approved):

PR #1166, opened 2023-03-01
PYTHON-3622 Improve Server Log Download in Evergreen

blink1073:


blink1073:
Example for a replica set test:

image

ShaneHarvey (approved):
nice

PR #1172, opened 2023-03-17
PYTHON-2468 Add pymongoexplain example to pymongo docs

juliusgeo:


pymongo/collection.py
```diff
@@ -2368,8 +2368,9 @@ def aggregate(
         :attr:`~pymongo.read_preferences.ReadPreference.PRIMARY` is used.
 
         .. note:: This method does not support the 'explain' option. Please
-           use :meth:`~pymongo.database.Database.command` instead. An
-           example is included in the :ref:`aggregate-examples` documentation.
+           use `PyMongoExplain <https://github.com/mongodb-labs/pymongoexplain>`_
```
ShaneHarvey, line 2371: use pypi here as well.

ShaneHarvey (changes requested):

pymongo/collection.py
juliusgeo: Done.

ShaneHarvey (approved):

PR #1173, opened 2023-03-23
PYTHON-3634 Windows crypt shared rewrap many data key timing out sometimes

juliusgeo:


pymongo/encryption.py
```diff
@@ -144,7 +145,10 @@ def kms_request(self, kms_context):
             while kms_context.bytes_needed > 0:
                 # CSOT: update timeout.
                 conn.settimeout(max(_csot.clamp_remaining(_KMS_CONNECT_TIMEOUT), 0))
-                data = conn.recv(kms_context.bytes_needed)
+                try:
+                    data = conn.recv(kms_context.bytes_needed)
+                except Exception:
+                    data = conn.recv(kms_context.bytes_needed)
```
ShaneHarvey, line 151: The connection needs to be closed and recreated after a timeout. Also we can't retry after any Exception, only after transient errors.

ShaneHarvey (changes requested):

pymongo/encryption.py
ShaneHarvey: Oh interesting. closing/recreating won't work since we might have already called kms_context.feed() with a partial result. I'm not sure we can even do this until DRIVERS-1541 is implemented in libmongocrypt.

blink1073 (approved):
LGTM!

pymongo/encryption.py
juliusgeo: I think that might be a wider scope than what is intended for this ticket. I think just retrying the WantRead error is good enough to reduce the incidence of test failures.

ShaneHarvey: Sure sounds fine to defer retry logic until [DRIVERS-1541](https://jira.mongodb.org/browse/DRIVERS-1541). However, this code is still flawed because it will exceed the CSOT timeout. It seems like you just want to double the default KMS socket_timeout from 10 seconds to 20 seconds. How about let's just do that explicitly rather adding any retry logic?

ShaneHarvey (changes requested):

pymongo/encryption.py
```diff
@@ -64,7 +64,7 @@
 from pymongo.write_concern import WriteConcern
 
 _HTTPS_PORT = 443
-_KMS_CONNECT_TIMEOUT = 10  # TODO: CDRIVER-3262 will define this value.
+_KMS_CONNECT_TIMEOUT = 20  # TODO: CDRIVER-3262 will define this value.
```
ShaneHarvey, line 67: [CDRIVER-3262](https://jira.mongodb.org/browse/SPEC-1394) settled on this being the same as the default mongoclient connect timeout. Could you update the comment and use `common.CONNECT_TIMEOUT` here?

ShaneHarvey (changes requested):

blink1073 (approved):

ShaneHarvey (approved):

PR #1175, opened 2023-03-27
PYTHON-3619 MacOS hosts are incredibly slow

blink1073:


blink1073:
All of the previously timing out builds finished in 41m or less.  Two of them failed with known CSOT errors.

PR #1178, opened 2023-04-03
PYTHON-3644 Test encryption KMS connections with stdlib ssl, not just pyopenssl

blink1073:


.evergreen/config.yml
```diff
@@ -2507,6 +2510,13 @@ axes:
         variables:
           test_encryption: true
         batchtime: 10080  # 7 days
+      - id: "encryption_pyopenssl"
+        display_name: "Encryption PyOpenSSL"
+        tags: ["encryption_tag"]
+        variables:
+          test_encryption: true
+          test_encryption_pyopenssl: true
+        batchtime: 10080  # 7 days
```
ShaneHarvey, line 2519: What if we only test 1 server version instead of the whole matrix here? We could use the `rules: if/then/add_tasks/remove_tasks` feature like we do here: https://github.com/mongodb/mongo-python-driver/blob/cbad35ec9d591b09c492f4c869c50fd203b55e44/.evergreen/config.yml#L2849

blink1073: Fixed, we now test `latest-replica-set` with this variant and all server types otherwise.

.evergreen/config.yml
```diff
@@ -2726,8 +2736,15 @@ buildvariants:
...
-
+  tasks: "test-latest-replica_set"
+  rules:
+    - if:
+        encryption: ["encryption", "encryption_crypt_shared"]
+        platform: ubuntu-18.04
+        auth-ssl: noauth-nossl
+        python-version: "*"
+      then:
+        add_tasks: *encryption-server-versions
```
ShaneHarvey, line 2747: Let's do the same for test-macos-encryption and tests-windows-encryption

ShaneHarvey (changes requested):

.evergreen/config.yml
blink1073: Done

ShaneHarvey (approved):

PR #1186, opened 2023-04-11
fix wrong urllib parse function being used to percent encode usernames and passwords, where before spaces would be incorrectly encoded to '+' instead of '%20' which is behaviour for a different usecase

M-Moon:
The example given for percent-encoding usernames and passwords for a MongoDB URI first mentions .quote() should be used, but proceeds to incorrectly use .quote_plus(), changing spaces to plus signs instead of their percent encoding equivalent.

The example should use .quote() to be clear and exemplify proper use.

### DIFF
--- a/doc/examples/authentication.rst
+++ b/doc/examples/authentication.rst
@@ -15,10 +15,10 @@ Username and password must be percent-escaped with
 
   >>> from pymongo import MongoClient
   >>> import urllib.parse
-  >>> username = urllib.parse.quote_plus('user')
+  >>> username = urllib.parse.quote('user')
   >>> username
   'user'
-  >>> password = urllib.parse.quote_plus('pass/word')
+  >>> password = urllib.parse.quote('pass/word')
   >>> password
   'pass%2Fword'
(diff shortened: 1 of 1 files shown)


blink1073:
Hi @M-Moon, these examples were changed to handle the case where the username or password contains an unescaped `/` character, which would get interpreted as the database name.  https://jira.mongodb.org/browse/PYTHON-2973.   I think the right thing to do is change the examples to use `quote(..., safe='')`

M-Moon:
Thanks for the response @blink1073, this makes sense. I was not aware of the issue so thanks for letting me know.

ShaneHarvey:
In this case the code example is correct but the paragraph is wrong. It should say:
```python
Username and password must be percent-escaped with
:py:func:`urllib.parse.quote_plus`, to be used in a MongoDB URI.
```

ShaneHarvey:
@M-Moon thank you for reporting this, I've fixed it in https://github.com/mongodb/mongo-python-driver/pull/1193!

PR #1188, opened 2023-04-11
PYTHON-3652 Bump maxWireVersion for MongoDB 7.0

blink1073:


test/test_topology.py
```diff
@@ -540,8 +541,8 @@ def test_wire_version(self):
                 HelloCompat.LEGACY_CMD: True,
                 "setName": "rs",
                 "hosts": ["a"],
-                "minWireVersion": 21,
-                "maxWireVersion": 22,
+                "minWireVersion": max_version + 1,
+                "maxWireVersion": max_version + 2,
```
ShaneHarvey, line 545: I understand your motivation here but I think it's good to keep this test separate from `common.MAX_SUPPORTED_WIRE_VERSION`, that way when we change MAX_SUPPORTED_WIRE_VERSION we know it has an effect.

ShaneHarvey (changes requested):

test/test_topology.py
blink1073: Done

ShaneHarvey (approved):

PR #1189, opened 2023-04-11
PYTHON-3162 Deprecate ServerDescription.election_tuple

blink1073:


ShaneHarvey (commented):
Could you add a changelog entry?

ShaneHarvey:
Perhaps we should document them then? I'm of the opinion that all public attributes are part of the api whether we document them or not. I'll open a new ticket for it though.

ShaneHarvey (approved):

PR #1198, opened 2023-04-28
PYTHON-3686 Pin mypy dependencies, don't upgrade pip

ShaneHarvey:


ShaneHarvey:
Wow, pinning to pip 23.0.1 fixes the errors on Python 3.7.

ShaneHarvey:
Oh I bet if this is related to pip's build via wheel by default change. My theory is that be mistakenly don't ship the codec_options.pyi file in the wheel.

blink1073:
Better to be explicit and add `recursive-include bson *.pyi` to `MANIFEST.in` either way.

ShaneHarvey:
I'll try, I do see that codec_options.pyi is already included in the wheel so I don't think it's that simple.

ShaneHarvey:
Fixed by https://github.com/mongodb/mongo-python-driver/pull/1199

PR #1196, opened 2023-04-26
PYTHON-2504 Apply one time pyupgrade --py37-plus v3.3.2 and ruff

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-2504

Successor to https://github.com/mongodb/mongo-python-driver/pull/882.

Ran this repeatedly until it stablized after 3 or 4 runs:
```
$ pyupgrade --py37-plus bson/*.py pymongo/*.py gridfs/*.py test/*.py tools/*.py test/*/*.py
```

I also ran ruff 0.0.265 like this (see https://beta.ruff.rs/docs/rules/):
```
$ ruff --fix-only --select ALL --fixable ALL --target-version py37 --line-length=100 --unfixable COM812,D400,D415,ERA001,RUF100,SIM108,D211,D212,SIM105,SIM,PT,ANN204,EM bson/*.py pymongo/*.py gridfs/*.py test/*.py test/*/*.py
```

The --unfixable codes suppress some unneeded or particularly noisy changes.

As noted in the previous PR:
> I tested out adding this to pre-commit but pyupgrade's formatting conflicts with black so it's more of a nuisance.

I'll add this to `.git-blame-ignore-revs` after the PR is merged.

PR #1194, opened 2023-04-22
PYTHON-3679 Support mypy 1.2

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3679

Adds workarounds/fixes for some small changes in:
- https://github.com/python/typeshed/pull/9694
- https://github.com/python/typeshed/pull/9194

PR #1199, opened 2023-04-28
PYTHON-3686 Consolidate CodecOptions Typings

blink1073:


ShaneHarvey (approved):
Nice! I know why the --strict tests are failing, I'll add that back after we merge https://github.com/mongodb/mongo-python-driver/pull/1196 and https://github.com/mongodb/mongo-python-driver/pull/1194.

PR #1200, opened 2023-04-28
PYTHON-3690 Do not install unittest-xml-reporting on MacOS EG Hosts

blink1073:


.evergreen/utils.sh
```diff
@@ -30,7 +30,12 @@ createvirtualenv () {
     fi
 
     python -m pip install --upgrade pip
-    python -m pip install --upgrade setuptools wheel unittest-xml-reporting
+    python -m pip install --upgrade setuptools wheel
+
+    # lxml only has wheels for macos 10.15+
+    if [ "$(uname -s)" != "Darwin" ]; then
+        python -m pip install unittest-xml-reporting
```
ShaneHarvey, line 37: Can we somehow still install the package on 11+? Maybe with:
```
python -m pip install unittest-xml-reporting || true
```

blink1073: Done

ShaneHarvey (approved):

PR #1202, opened 2023-05-03
PYTHON-3676 Migrate from python-snappy in favour of cramjam

blink1073:


doc/installation.rst
```diff
@@ -59,8 +59,8 @@ dependency can be installed automatically along with PyMongo::
 
   $ python3 -m pip install "pymongo[ocsp]"
 
-Wire protocol compression with snappy requires `python-snappy
-<https://pypi.org/project/python-snappy>`_::
+Wire protocol compression with snappy requires `cramjam
+<https://pypi.org/project/cramjam>`_::
```
ShaneHarvey, line 63: For backwards compat I think we should to still support python-snappy as a fallback (to avoid breaking snappy users who just upgrade pymongo). Let's say that either can be used but cramjam is preferred if it is installed.

pymongo/compression_support.py
```diff
@@ -136,6 +136,8 @@ def decompress(data, compressor_id):
     if compressor_id == SnappyContext.compressor_id:
         # python-snappy doesn't support the buffer interface.
         # https://github.com/andrix/python-snappy/issues/65
+        # cramjam also doesn't support the buffer interface.
+        # https://github.com/milesgranger/pyrus-cramjam/issues/102
         # This only matters when data is a memoryview since
         # id(bytes(data)) == id(data) when data is a bytes.
         return snappy.uncompress(bytes(data))
```
ShaneHarvey, line 143: Still need to use cramjam here and for compression.

ShaneHarvey (changes requested):

pymongo/compression_support.py
blink1073: Done, those changes got lost somewhere.

doc/installation.rst
blink1073: Done

PR #1216, opened 2023-05-16
PYTHON-3691 [Build Failure] test_client.TestClient.test_exhaust_network_error

blink1073:


pymongo/pyopenssl_context.py
```diff
@@ -107,6 +107,9 @@ def _call(self, call, *args, **kwargs):
             try:
                 return call(*args, **kwargs)
             except BLOCKING_IO_ERRORS as exc:
+                # Check for closed socket.
+                if self.fileno() == -1:
+                    raise
```
ShaneHarvey, line 112: When this case happens what traceback does the user see? I ask because we normally hide BLOCKING_IO_ERRORS type errors from the user and instead raise _socket.timeout like we do below.

ShaneHarvey (changes requested):

pymongo/pyopenssl_context.py
blink1073: I agree we should override the error, updated.

pymongo/pyopenssl_context.py
```diff
@@ -107,6 +107,9 @@ def _call(self, call, *args, **kwargs):
             try:
                 return call(*args, **kwargs)
             except BLOCKING_IO_ERRORS as exc:
+                # Check for closed socket.
+                if self.fileno() == -1:
+                    raise _socket.timeout("timed out")
```
ShaneHarvey, line 112: Discussed offline but I think we settled on this being:
```python
                if self.fileno() == -1:
                    if timeout and _time.monotonic() - start > timeout:
                        raise _socket.timeout("timed out")
                    raise SSLError("Underlying socket has been closed")
```

To match the stdlib ssl:
```c
_ssl__SSLSocket_write_impl(PySSLSocket *self, Py_buffer *b)
...

    sockstate = PySSL_select(sock, 1, timeout);
    if (sockstate == SOCKET_HAS_TIMED_OUT) {
        PyErr_SetString(PyExc_TimeoutError,
                        "The write operation timed out");
        goto error;
    } else if (sockstate == SOCKET_HAS_BEEN_CLOSED) {
        PyErr_SetString(get_state_sock(self)->PySSLErrorObject,
                        "Underlying socket has been closed.");
        goto error;
```
https://github.com/python/cpython/blob/ede89af/Modules/_ssl.c#L2349-L2357

ShaneHarvey (changes requested):

pymongo/pyopenssl_context.py
blink1073: Done

ShaneHarvey (approved):

PR #1219, opened 2023-05-24
PYTHON-3717 Speed up _type_marker check in BSON

thalassemia:
In _cbsonmodule.c, [the `_type_marker` function](https://github.com/mongodb/mongo-python-driver/blob/bda9e3a0bb533b6d0c6c5141feeecb77c4d31709/bson/_cbsonmodule.c#L431-L440) uses `PyObject_HasAttrString(object, "_type_marker")` and `PyObject_GetAttrString(object, "_type_marker")`. In my workloads (highly nested documents with many large array fields), these functions become severe bottlenecks to performance, because they each create new Python string objects by calling `PyUnicode_FromString("_type_marker")` internally every time they run.

 

This simple change improved my performance by more than double. One caveat is that this leaks the `TYPEMARKERSTR` object in the case that the `cbson` module is unloaded.

 

Also, correct me if I'm wrong, but I believe [these lines are redundant](https://github.com/mongodb/mongo-python-driver/blob/bda9e3a0bb533b6d0c6c5141feeecb77c4d31709/bson/_cbsonmodule.c#L453-L459) because the function returns type at the end regardless.

### DIFF
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -55,6 +55,7 @@ struct module_state {
     PyObject* DatetimeMS;
     PyObject* _min_datetime_ms;
     PyObject* _max_datetime_ms;
+    PyObject* _type_marker_str;
 };
 
 #define GETSTATE(m) ((struct module_state*)PyModule_GetState(m))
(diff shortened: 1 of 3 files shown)


blink1073:
Hi @thalassemia, thanks for working into this!  We should be able to add this value to the [module_state](https://github.com/mongodb/mongo-python-driver/blob/bda9e3a0bb533b6d0c6c5141feeecb77c4d31709/bson/_cbsonmodule.c#L40), and then clear it on [teardown](https://github.com/mongodb/mongo-python-driver/blob/bda9e3a0bb533b6d0c6c5141feeecb77c4d31709/bson/_cbsonmodule.c#L2957).  See [here](https://github.com/mongodb/mongo-python-driver/blob/bda9e3a0bb533b6d0c6c5141feeecb77c4d31709/bson/_cbsonmodule.c#L194) for an example of interacting with the module state.

blink1073 (approved):
Nice work, thank you!

PR #1225, opened 2023-06-08
PYTHON-3728 Simplify convert_codec_options signature

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3728

I'm convinced the Coverity issue is a false positive and that simplifying the convert_codec_options signature should help.

### DIFF
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -83,7 +83,7 @@ struct module_state {
 #define DATETIME_AUTO 4
 
 /* Converts integer to its string representation in decimal notation. */
-int cbson_long_long_to_str(long long num, char* str, size_t size) {
+extern int cbson_long_long_to_str(long long num, char* str, size_t size) {
     // Buffer should fit 64-bit signed integer
     if (size < 21) {
         PyErr_Format(
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -599,8 +599,7 @@ fail:
  * Return 1 on success. options->document_class is a new reference.
  * Return 0 on failure.
  */
-int convert_codec_options(PyObject* self, PyObject* options_obj, void* p) {
-    codec_options_t* options = (codec_options_t*)p;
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -263,12 +263,7 @@ def build_extension(self, ext):
     Extension(
         "pymongo._cmessage",
         include_dirs=["bson"],
-        sources=[
-            "pymongo/_cmessagemodule.c",
-            "bson/_cbsonmodule.c",
-            "bson/time64.c",
-            "bson/buffer.c",
-        ],
```
ShaneHarvey, line 271: I reverted this change made in PYTHON-3717 because I suspect it was unneeded.

bson/_cbsonmodule.c
```diff
@@ -599,8 +599,7 @@ int cbson_convert_type_registry(PyObject* registry_obj, type_registry_t* registr
  * Return 1 on success. options->document_class is a new reference.
  * Return 0 on failure.
  */
-int convert_codec_options(PyObject* self, PyObject* options_obj, void* p) {
-    codec_options_t* options = (codec_options_t*)p;
+int convert_codec_options(PyObject* self, PyObject* options_obj, codec_options_t* options) {
```
NoahStapp, line 602: To clarify--what was the reason for the `options` parameter being a generic void pointer before this change?

setup.py
blink1073: Looks like it is: `ImportError: /data/mci/881d33b0b5dec4eb9cb9c1f9be10aa61/src/pymongo/_cmessage.cpython-310-x86_64-linux-gnu.so: undefined symbol: cbson_long_long_to_str`

blink1073: Unless removing `extern` is what caused the error.

ShaneHarvey: I'm going to merge the codec options change and open a new ticket to investigate this issue.

ShaneHarvey:
Ready for another look.

blink1073 (approved):
LGTM!

PR #1229, opened 2023-06-12
PYTHON-3702 Stop using utcnow and utcfromtimestamp

NoahStapp:
Several places in the code currently depend on the UTC timestamp being naive, meaning there is no explicit attached timezone. As a result, I've used `.replace(tzinfo=None)` where necessary to strip the associated timezone information after generating a UTC timestamp. It's likely preferable to standardize our timestamps to always be aware since we're migrating away from implicit UTC timestamps.

### DIFF
--- a/bson/datetime_ms.py
+++ b/bson/datetime_ms.py
@@ -26,7 +26,7 @@ from bson.codec_options import DEFAULT_CODEC_OPTIONS, CodecOptions, DatetimeConv
 from bson.tz_util import utc
 
 EPOCH_AWARE = datetime.datetime.fromtimestamp(0, utc)
-EPOCH_NAIVE = datetime.datetime.utcfromtimestamp(0)
+EPOCH_NAIVE = datetime.datetime.fromtimestamp(0, tz=datetime.timezone.utc).replace(tzinfo=None)
 
 
 class DatetimeMS:
--- a/doc/examples/datetimes.rst
+++ b/doc/examples/datetimes.rst
@@ -25,10 +25,12 @@ time into MongoDB:
 
 .. doctest::
 
-   >>> result = db.objects.insert_one({"last_modified": datetime.datetime.utcnow()})
+   >>> result = db.objects.insert_one(
+   ...     {"last_modified": datetime.datetime.now(tz=timezone.utc)}
+   ... )
 
-Always use :meth:`datetime.datetime.utcnow`, which returns the current time in
-UTC, instead of :meth:`datetime.datetime.now`, which returns the current local
+Always use :meth:`datetime.datetime.now(tz=timezone.utc)`, which explicitly returns the current time in
+UTC, instead of :meth:`datetime.datetime.now`, with no arguments, which returns the current local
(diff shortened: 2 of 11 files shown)


blink1073 (approved):
LGTM, thanks!

bson/datetime_ms.py
```diff
@@ -26,7 +26,7 @@
 from bson.tz_util import utc
 
 EPOCH_AWARE = datetime.datetime.fromtimestamp(0, utc)
-EPOCH_NAIVE = datetime.datetime.utcfromtimestamp(0)
+EPOCH_NAIVE = datetime.datetime.fromtimestamp(0, tz=datetime.timezone.utc).replace(tzinfo=None)
```
ShaneHarvey, line 29: Above we use `utc` from:
```python
from bson.tz_util import utc

```

I'd suggest we use that instead of `datetime.timezone.utc` for consistency. We could open a new ticket to deprecate/remove tz_util since datetime.timezone has existed since Python 3.2: https://docs.python.org/3.6/library/datetime.html#datetime.timezone

doc/examples/datetimes.rst
```diff
@@ -25,10 +25,12 @@ time into MongoDB:
 
 .. doctest::
 
-   >>> result = db.objects.insert_one({"last_modified": datetime.datetime.utcnow()})
+   >>> result = db.objects.insert_one(
+   ...     {"last_modified": datetime.datetime.now(tz=timezone.utc)}
+   ... )
 
-Always use :meth:`datetime.datetime.utcnow`, which returns the current time in
-UTC, instead of :meth:`datetime.datetime.now`, which returns the current local
```
ShaneHarvey, line 31: Hmm would it be worth adding a note in the changelog about this? Like "Previously PyMongo's docs recommended using utcnow. utcnow is deprecated in Python 3.12. Instead , we now recommend using X."

I'm thinking we can help point users to the right workaround since pymongo apps will likely have copied this pattern of using utcnow and will start seeing deprecation warnings in 3.12.

ShaneHarvey (commented):
Nice work! Adding two inline comments.

bson/datetime_ms.py
NoahStapp: To clarify, you're suggesting we use the `utc` `bson.tz_util` specifically in the code above? Makes sense to me, although I agree we should deprecate tz_util to avoid duplicating the datetime.timezone functionality that has been present for years now.

doc/examples/datetimes.rst
NoahStapp: Agreed, I imagine a lot of users will start seeing deprecation warnings in 3.12. Adding a note in the changelog with a link to the Python PR change explanation is a good idea.

bson/datetime_ms.py
ShaneHarvey: Yes I'm suggesting we use `bson.tz_util.utc` only on this line (and not elsewhere in the docs/tests).

PR #1231, opened 2023-06-13
PYTHON-3735 Add types to PyMongo auth module

NoahStapp:


pymongo/auth.py
```diff
@@ -557,15 +573,15 @@ def speculate_command(self):
 
 
 class _X509Context(_AuthContext):
-    def speculate_command(self):
+    def speculate_command(self) -> SON[str, int | str]:
```
ShaneHarvey: `|` was introduced in Python 3.10 so we'll need to use `Union` instead. Actually it might be better to use `Dict[str, Any]` or `MutableMapping[str, Any]` here (unless it's important that this is actually SON).

ShaneHarvey (changes requested):

pymongo/auth.py
```diff
@@ -557,15 +575,15 @@ def speculate_command(self):
...
+    def speculate_command(self) -> Optional[SON[str, Any]]:
         cmd = SON([("authenticate", 1), ("mechanism", "MONGODB-X509")])
         if self.credentials.username is not None:
             cmd["user"] = self.credentials.username
         return cmd
 
 
 class _OIDCContext(_AuthContext):
-    def speculate_command(self):
+    def speculate_command(self) -> Optional[SON[str, Any]]:
```
ShaneHarvey, line 586: When would speculate_command here return None? @blink1073 is there a way to avoid that? It's unexpected because pool.py doesn't handle this case:
```
            if auth_ctx:
                cmd["speculativeAuthenticate"] = auth_ctx.speculate_command()
```

pymongo/auth.py
NoahStapp: Got it, any objections to using MutableMapping and Mapping in general to standardize collection type annotations?

pymongo/auth.py
NoahStapp: The parent class `_AuthContext`'s `speculate_command` simply raises a `NotImplementedError`, which causes mypy to complain if the subclass `speculate_command` does not have an `Optional` return type.

ShaneHarvey: The problem is not quite the NotImplementedError in the parent class, that part is fine because exceptions don't affect the return type. The problem is that _OIDCContext's speculate_command implementation can return None which is unexpected:
```python
        cmd = authenticator.auth_start_cmd(False)
        if cmd is None:
            return None # <- Uh?
        ...
```

NoahStapp: Ahh, got it. 

Looks like the actual code that can return None is here, in `auth_odic.py`'s `get_current_token`:

```        
if not use_callbacks and not current_valid_token:
	return None
```

ShaneHarvey: Hmm I see. I suspect we do actually need these changes to make the return Optional, but we need Steve's to confirm since he owns the OIDC feature. If so we also need to update pool.py to handle a None return.

blink1073: Yes, we do not want to send a speculate command if there is no current token, and we don't want to call a callback either, so Optional is appropriate.

pymongo/auth.py
blink1073: +1 to using more general types where appropriate, including `Sequence`.

pymongo/auth.py
NoahStapp: Do we want a separate ticket to handle the pool.py updates?

ShaneHarvey: I think we can do it in this PR. If speculate_command() returns None then we should skip adding the "speculativeAuthenticate" field.

pymongo/auth.py
```diff
@@ -13,15 +13,17 @@
 # limitations under the License.
 
 """Authentication helpers."""
+from __future__ import annotations
```
ShaneHarvey, line 16: I like this. What do you think about opening a new ticket to add this import to all files so we don't have to deal with forward references as strings (like we do in SON)?

NoahStapp: I used autotyping to start, but even with the --aggressive flag it didn't do a ton beyond the very simple functions, so most of it was by hand. 

Adding that import everywhere is a good idea, I'll file a ticket.

PR #1233, opened 2023-06-14
PYTHON-3702 bson datetime utc import cleanup

NoahStapp:


bson/datetime_ms.py
```diff
@@ -26,7 +26,7 @@
 from bson.tz_util import utc
 
 EPOCH_AWARE = datetime.datetime.fromtimestamp(0, utc)
-EPOCH_NAIVE = datetime.datetime.fromtimestamp(0, tz=datetime.timezone.utc).replace(tzinfo=None)
+EPOCH_NAIVE = datetime.datetime.fromtimestamp(0, tz=utc).replace(tzinfo=None)
```
ShaneHarvey, line 29: How about EPOCH_AWARE.replace(tzinfo=None)?

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM! I believe the test failure is unrelated.

PR #1240, opened 2023-06-15
PYTHON-2963 Add tox config in preparation for migration from setup.py

NoahStapp:
This change is still using `python setup.py test` inside tox to run the unit tests: migration to pytest will be done in https://jira.mongodb.org/browse/PYTHON-3727, and migration to pyproject.toml will be done in https://jira.mongodb.org/browse/PYTHON-2965.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,45 @@
+[tox]
+requires =
+    tox>=4
+envlist =
+    # Test using the system Python.
+    test,
+    # Run pre-commit on all files.
+    lint,
+    # Run pre-commit on all files, including stages that require manual fixes.
+    lint-manual,
+    # Typecheck all files.
+    typecheck
+
+[testenv:test]
+description = run unit tests
+commands =
+    python --version
+    python setup.py test {posargs}
+
+[testenv:lint]
+description = run pre-commit
+deps =
+    pre-commit
+commands =
+    pre-commit run --all-files
+
+[testenv:lint-manual]
+description = run all pre-commit stages, including those that require manual fixes
+deps =
+    pre-commit
+commands =
+    pre-commit run --all-files
+    pre-commit run --all-files --hook-stage manual flake8
+    pre-commit run --all-files --hook-stage manual doc8
+
+[testenv:typecheck]
+description = run mypy to typecheck
+deps =
+    mypy
+    zstandard
+    certifi; platform_system == "win32" or platform_system == "Darwin"
+commands =
+    mypy --install-types --non-interactive bson gridfs tools pymongo
+    mypy --install-types --non-interactive --disable-error-code var-annotated --disable-error-code attr-defined --disable-error-code union-attr --disable-error-code assignment --disable-error-code no-redef --disable-error-code index --allow-redefinition --allow-untyped-globals --exclude "test/mypy_fails/*.*" test
+    mypy --install-types --non-interactive test/test_typing.py test/test_typing_strict.py
--- a/tox.ini
+++ b/tox.ini
@@ -57,10 +57,9 @@ commands =
     pyright test/test_typing.py test/test_typing_strict.py
 
 [testenv:typecheck-pyright-strict]
-description = run pyright to typecheck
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -0,0 +1,45 @@
+[tox]
+requires =
+    tox>=4
+envlist =
```
NoahStapp, line 4: The lack of specified Python versions here will cause tox to run using the default system interpreter.

tox.ini
```diff
@@ -0,0 +1,45 @@
...
+    pre-commit run --all-files
+
+[testenv:lint-manual]
+description = run all pre-commit stages, including those that require manual fixes
+deps =
+    pre-commit
+commands =
+    pre-commit run --all-files
+    pre-commit run --all-files --hook-stage manual flake8
+    pre-commit run --all-files --hook-stage manual doc8
```
ShaneHarvey, line 34: `pre-commit run --all-files --hook-stage manual` will run all the pre-commit checks in one command.

ShaneHarvey (changes requested):
If the plan is to migrate to tox in future PRs then could you update the title to something like "Add tox config"?

tox.ini
NoahStapp: Good catch, thanks!

ShaneHarvey (approved):
LGTM!

blink1073 (changes requested):
We should update the `github` workflow as part of this PR to use `tox`.

blink1073:
Note that moving our evergreen tests to `tox` will involve adding new environments for things like `encryption`, etc, and move things out of the EG files like https://github.com/mongodb/mongo-python-driver/blob/master/.evergreen/run-tests.sh.   @NoahStapp, can you please add a note to the Epic that we'll need tickets for each of those files to migrate to tox?

NoahStapp:
> Note that moving our evergreen tests to `tox` will involve adding new environments for things like `encryption`, etc, and move things out of the EG files like [`master`/.evergreen/run-tests.sh](https://github.com/mongodb/mongo-python-driver/blob/master/.evergreen/run-tests.sh?rgh-link-date=2023-06-15T18%3A54%3A18Z). @NoahStapp, can you please add a note to the Epic that we'll need tickets for each of those files to migrate to tox?

Sure, will do.

NoahStapp:
> We should update the `github` workflow as part of this PR to use `tox`.

For all of the workflow jobs?

tox.ini
```diff
@@ -0,0 +1,89 @@
...
+description = run pyright to typecheck
+deps =
+    mypy
+    pyright==1.1.290
+allowlist_externals=echo
+commands =
+    echo '{"strict": ["tests/test_typing_strict.py"]}' >> pyrightconfig.json
+    pyright test/test_typing_strict.py
+
+[testenv:typecheck]
```
blink1073, line 69: I don't think we need this environment, we can call the individual ones.

NoahStapp: This is intended to replace needing an alias to run all of the typechecks in a single command.

blink1073: Hmm, it is possible then to compose this to use the deps and commands of the other envs?

NoahStapp: Figured it out, fixed!

tox.ini
```diff
@@ -0,0 +1,77 @@
...
+commands =
+    pyright test/test_typing.py test/test_typing_strict.py
+
+[testenv:typecheck-pyright-strict]
+description = run pyright with strict mode to typecheck
+deps =
+    {[testenv:typecheck-pyright]deps}
+allowlist_externals=echo
+commands =
+    echo '{"strict": ["tests/test_typing_strict.py"]}' >> pyrightconfig.json
```
ShaneHarvey, line 65: `>>` should be `>` IIUC. Otherwise we'd repeatedly append to the same file.

ShaneHarvey (changes requested):

blink1073 (approved):
Thanks!

ShaneHarvey (approved):

PR #1246, opened 2023-06-16
PYTHON-3740 add types to helpers.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3740

### DIFF
--- a/pymongo/cursor.py
+++ b/pymongo/cursor.py
@@ -274,6 +274,7 @@ class Cursor(Generic[_DocumentType]):
         self.__show_record_id = show_record_id
         self.__allow_disk_use = allow_disk_use
         self.__snapshot = snapshot
+        self.__hint: Union[str, SON[str, Any], None]
         self.__set_hint(hint)
 
         # Exhaust cursor support
@@ -437,7 +438,7 @@ class Cursor(Generic[_DocumentType]):
 
     def __query_spec(self):
         """Get the spec to use for a query."""
-        operators = {}
+        operators: Dict[str, Any] = {}
         if self.__ordering:
             operators["$orderby"] = self.__ordering
         if self.__explain:
--- a/pymongo/helpers.py
+++ b/pymongo/helpers.py
@@ -17,10 +17,25 @@
 import sys
 import traceback
 from collections import abc
-from typing import Any, List, NoReturn
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Mapping,
+    NoReturn,
+    Optional,
+    Sequence,
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+)
 
 from bson.son import SON
 from pymongo import ASCENDING
+from pymongo.cursor import _Hint
 from pymongo.errors import (
     CursorNotFound,
     DuplicateKeyError,
@@ -33,9 +48,10 @@ from pymongo.errors import (
     _wtimeout_error,
 )
 from pymongo.hello import HelloCompat
+from pymongo.operations import _IndexList
 
 # From the SDAM spec, the "node is shutting down" codes.
-_SHUTDOWN_CODES = frozenset(
+_SHUTDOWN_CODES: frozenset = frozenset(
     [
         11600,  # InterruptedAtShutdown
         91,  # ShutdownInProgress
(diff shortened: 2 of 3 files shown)


blink1073 (approved):
Nice work, thanks!

pymongo/helpers.py
```diff
@@ -93,15 +115,15 @@ def _index_list(key_or_list, direction=None):
...
+        values: List[Tuple[str, int]] = []
         for item in key_or_list:
             if isinstance(item, str):
                 item = (item, ASCENDING)
             values.append(item)
         return values
 
 
-def _index_document(index_list):
+def _index_document(index_list: _IndexList) -> SON[str, Any]:
```
NoahStapp, line 126: Is it important that this return type be the specific `SON` type? If not, using a more general type like `MutableMapping` is a good idea for consistency.

pymongo/helpers.py
```diff
@@ -134,13 +156,19 @@ def _index_document(index_list):
 
 
 def _check_command_response(
-    response, max_wire_version, allowable_errors=None, parse_write_concern_error=False
-):
+    response: Dict[str, Any],
```
NoahStapp, line 159: We should try to use `Mapping` instead of `Dict` where applicable, and use `dict` where not applicable as `Dict` is deprecated as of 3.9: https://docs.python.org/3/library/typing.html#typing.Dict.

NoahStapp (changes requested):

pymongo/helpers.py
sleepyStick: when this function is called on line 938 in `cursor.py`, `self.__hint` is expecting `Union[str, SON[str, Any], None]` so I think it actually does need to be type `SON`

pymongo/cursor.py
```diff
@@ -437,7 +438,7 @@ def close(self) -> None:
 
     def __query_spec(self):
         """Get the spec to use for a query."""
-        operators = {}
+        operators: Dict[str, Any] = {}
```
NoahStapp, line 441: Use `dict` instead of `Dict` due to `Dict` being deprecated in 3.9.

NoahStapp (changes requested):

pymongo/helpers.py
```diff
@@ -222,7 +250,7 @@ def _get_wce_doc(result):
     return wce
 
 
-def _check_write_command_response(result):
+def _check_write_command_response(result: Dict[str, Any]) -> None:
```
NoahStapp, line 253: Use `dict` instead of `Dict`.

pymongo/helpers.py
NoahStapp: Does `self.__hint` actually need the type to be `SON` or is it just expecting that but could use any `MutableMapping`?

pymongo/cursor.py
ShaneHarvey: `dict` doesn't support generics until 3.9 so we can only use Dict or MutableMapping here.

pymongo/helpers.py
sleepyStick: Ah I see, I think you're right. I forgot that I was the one that added the type for `self.__hint`

pymongo/cursor.py
ShaneHarvey: That brings up a good question. Should we support type checking in all versions of python we support. My intuition says yes because a user might want to run mypy with 3.7. I think it would be odd to say you have to run mypy with a different version of python than your app actually uses.

blink1073: Yes, we test with Python 3.7 and 3.10 specifically in GitHub Actions.

pymongo/helpers.py
sleepyStick: talked in person; decided to stick with more specific return types

pymongo/cursor.py
sleepyStick: talked in person, decided we should stick with `dict` for now

pymongo/helpers.py
```diff
@@ -210,7 +238,7 @@ def _raise_write_concern_error(error: Any) -> NoReturn:
     raise WriteConcernError(error.get("errmsg"), error.get("code"), error)
 
 
-def _get_wce_doc(result):
+def _get_wce_doc(result: Dict[str, Any]) -> Optional[Dict[str, Any]]:
```
NoahStapp, line 241: The argument type should be `Mapping[str, Any]` to be consistent with https://docs.python.org/3/library/typing.html#typing.Dict.

pymongo/helpers.py
NoahStapp, line 253: The argument type should be `Mapping[str, Any]`.

NoahStapp (changes requested):

ShaneHarvey (approved):
LGTM after Noah's last two comments are addressed.

ShaneHarvey (approved):
LGTM!

PR #1248, opened 2023-06-20
PYTHON-3750 add types to server.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3750

### DIFF
--- a/pymongo/server.py
+++ b/pymongo/server.py
@@ -13,22 +13,43 @@
 # permissions and limitations under the License.
 
 """Communicate with one MongoDB server in a topology."""
+from __future__ import annotations
 
 from datetime import datetime
+from typing import TYPE_CHECKING, Any, Callable, List, Mapping, Optional, Tuple, Union
 
 from bson import _decode_all_selective
 from pymongo.errors import NotPrimaryError, OperationFailure
 from pymongo.helpers import _check_command_response, _handle_reauth
-from pymongo.message import _convert_exception, _OpMsg
+from pymongo.message import _convert_exception, _GetMore, _OpMsg, _OpReply, _Query
 from pymongo.response import PinnedResponse, Response
+
+if TYPE_CHECKING:
+    from contextlib import _GeneratorContextManager
+    from queue import Queue
+    from weakref import ReferenceType
+
+    from bson.codec_options import CodecOptions
+    from bson.objectid import ObjectId
+    from pymongo.mongo_client import _MongoClientErrorHandler
+    from pymongo.monitor import Monitor
+    from pymongo.monitoring import _EventListeners
+    from pymongo.pool import Pool, SocketInfo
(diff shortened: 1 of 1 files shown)


NoahStapp (approved):
LGTM!

blink1073 (approved):

pymongo/server.py
```diff
@@ -159,7 +186,10 @@ def run_operation(self, sock_info, operation, read_preference, listeners, unpack
             elif operation.name == "explain":
                 res = docs[0] if docs else {}
             else:
-                res = {"cursor": {"id": reply.cursor_id, "ns": operation.namespace()}, "ok": 1}
+                res = {
+                    "cursor": {"id": reply.cursor_id, "ns": operation.namespace()},
+                    "ok": 1,
+                }
```
ShaneHarvey, line 192: Was this change needed or can it be undone? (we prefer not to reformat unnecessarily to avoid code churn).

test/test_server.py
```diff
@@ -15,21 +15,20 @@
...
-sys.path[0:0] = [""]
-
 from test import unittest
 
 from pymongo.hello import Hello
 from pymongo.server import Server
 from pymongo.server_description import ServerDescription
 
+sys.path[0:0] = [""]
+
```
ShaneHarvey, line 25: Can you undo this change? We do this before importing test/pymongo so that we can run the test file directly like `python test/test_server.py`.

ShaneHarvey (changes requested):

pymongo/server.py
sleepyStick: I can undo it :)

test/test_server.py
```diff
@@ -14,10 +14,6 @@
 
 """Test the server module."""
 
-import sys
-
-sys.path[0:0] = [""]
-
```
ShaneHarvey, line 20: We want to keep these lines.

pymongo/server.py
```diff
@@ -13,22 +13,42 @@
...
     def __init__(
-        self, server_description, pool, monitor, topology_id=None, listeners=None, events=None
-    ):
+        self,
+        server_description: ServerDescription,
+        pool: Pool,
+        monitor: Monitor,
+        topology_id: Optional[ObjectId] = None,
+        listeners: Optional[_EventListeners] = None,
+        events: Optional[ReferenceType[Optional[Queue]]] = None,
```
ShaneHarvey, line 50: Should this be: `Optional[ReferenceType[Queue]]`?

sleepyStick: When its `Optional[ReferenceType[Queue]]`, I get an incompatible type elsewhere...so I think it does needs to be `Optional[ReferenceType[Optional[Queue]]]`

ShaneHarvey (approved):

PR #1249, opened 2023-06-20
PYTHON-3727 Use tox for unit tests and switch to pytest

NoahStapp:


blink1073:
I think more has to be done to make us pytest-compatible, lots of the tests are getting skipped that shouldn't be.

Note that the `setup.py` entrypoint does some extra work [here](https://github.com/mongodb/mongo-python-driver/blob/bc66d83efc5e57c5685cdd621aa3bf74ba44f2a7/setup.py#LL109C34-L109C34).

I believe the equivalent of the [setup](https://github.com/mongodb/mongo-python-driver/blob/bc66d83efc5e57c5685cdd621aa3bf74ba44f2a7/test/__init__.py#L1159) and [teardown](https://github.com/mongodb/mongo-python-driver/blob/bc66d83efc5e57c5685cdd621aa3bf74ba44f2a7/test/__init__.py#L1216) functions will need to be done in a `conftest.py` [fixture](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html?highlight=conftest#scope-sharing-fixtures-across-classes-modules-packages-or-session):

```python
def pymongo_test_fixture(scope='session', autouse=True):
      # do setup stuff
      yield
      # do teardown stuff
```

NoahStapp:
> 

Latest commit should address this, I've added a session-level autouse fixture to perform the `setup` and `teardown` found in the test module `__init__`.

> I think more has to be done to make us pytest-compatible, lots of the tests are getting skipped that shouldn't be.
> 
> Note that the `setup.py` entrypoint does some extra work [here](https://github.com/mongodb/mongo-python-driver/blob/bc66d83efc5e57c5685cdd621aa3bf74ba44f2a7/setup.py#LL109C34-L109C34).
> 
> I believe the equivalent of the [setup](https://github.com/mongodb/mongo-python-driver/blob/bc66d83efc5e57c5685cdd621aa3bf74ba44f2a7/test/__init__.py#L1159) and [teardown](https://github.com/mongodb/mongo-python-driver/blob/bc66d83efc5e57c5685cdd621aa3bf74ba44f2a7/test/__init__.py#L1216) functions will need to be done in a `conftest.py` [fixture](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html?highlight=conftest#scope-sharing-fixtures-across-classes-modules-packages-or-session):
> 
> ```python
> def pymongo_test_fixture(scope='session', autouse=True):
>       # do setup stuff
>       yield
>       # do teardown stuff
> ```

Latest commit should address this, I've added a session-level autouse fixture to perform the `setup` and `teardown` found in the test module `__init__`.

blink1073 (approved):
Very nice!

tox.ini
```diff
@@ -24,10 +26,23 @@ envlist =
...
+commands =
+    python --version
+    pytest -v -rs {posargs}
+
+[testenv:test-encryption]
+description = run base unit tests with encryption enabled
+deps =
+    {[testenv:test]deps}
+    pymongocrypt>=1.6.0,<2.0.0
+    pymongo-auth-aws<2.0.0
```
ShaneHarvey, line 41: is it possible to refer to extras here so we don't need to duplicate the dependencies? like `.[encryption]`?

blink1073: Yeah that should work

NoahStapp: Fixed, thanks!

ShaneHarvey (approved):
Nice LGTM!

PR #1250, opened 2023-06-20
PYTHON-3752 add types to bulk.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3752

### DIFF
--- a/pymongo/bulk.py
+++ b/pymongo/bulk.py
@@ -232,9 +232,7 @@ class _Bulk:
         cmd: Dict[str, Any] = dict(
             [("q", selector), ("u", update), ("multi", multi), ("upsert", upsert)]
         )
-        collation: Optional[Dict[str, Any]] = validate_collation_or_none(  # type: ignore[no-redef]
-            collation
-        )
+        collation: Optional[Dict[str, Any]] = validate_collation_or_none(collation)
         if collation is not None:
             self.uses_collation = True
             cmd["collation"] = collation
@@ -260,9 +258,7 @@ class _Bulk:
         """Create a replace document and add it to the list of ops."""
         validate_ok_for_replace(replacement)
         cmd = SON([("q", selector), ("u", replacement), ("multi", False), ("upsert", upsert)])
-        collation: Optional[Dict[str, Any]] = validate_collation_or_none(  # type: ignore[no-redef]
-            collation
-        )
+        collation: Optional[Dict[str, Any]] = validate_collation_or_none(collation)
         if collation is not None:
             self.uses_collation = True
             cmd["collation"] = collation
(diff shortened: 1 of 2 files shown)


pymongo/bulk.py
```diff
@@ -183,18 +215,22 @@ def add_insert(self, document):
...
-        hint=None,
-    ):
+        selector: MutableMapping[str, Any],
+        update: Union[
+            MutableMapping[str, Any],
+            List[MutableMapping[str, Any]],
+        ],
+        multi: bool = False,
+        upsert: bool = False,
+        collation: Optional[Dict[str, Any]] = None,
```
NoahStapp, line 225: `collation` should have type `Mapping` since it's an argument.

pymongo/bulk.py
```diff
@@ -209,11 +245,17 @@ def add_update(
             self.is_retryable = False
         self.ops.append((_UPDATE, cmd))
 
-    def add_replace(self, selector, replacement, upsert=False, collation=None, hint=None):
+    def add_replace(
+        self,
+        selector: MutableMapping[str, Any],
+        replacement: MutableMapping[str, Any],
+        upsert: bool = False,
+        collation: Optional[Dict[str, Any]] = None,
```
NoahStapp, line 253: collation should have type `Mapping` since it's an argument.

pymongo/bulk.py
```diff
@@ -222,10 +264,15 @@ def add_replace(self, selector, replacement, upsert=False, collation=None, hint=
             cmd["hint"] = hint
         self.ops.append((_UPDATE, cmd))
 
-    def add_delete(self, selector, limit, collation=None, hint=None):
+    def add_delete(
+        self,
+        selector: MutableMapping[str, Any],
+        limit: int,
+        collation: Optional[Dict[str, Any]] = None,
```
NoahStapp, line 271: collation should have type `Mapping` since it's an argument.

pymongo/bulk.py
```diff
@@ -149,31 +181,31 @@ def __init__(self, collection, ordered, bypass_document_validation, comment=None
...
+        self.is_retryable: bool = True
+        self.retrying: bool = False
+        self.started_retryable_write: bool = False
         # Extra state so that we know where to pick up on a retry attempt.
         self.current_run = None
         self.next_run = None
 
     @property
-    def bulk_ctx_class(self):
+    def bulk_ctx_class(self) -> Type[_BulkWriteContext]:
```
NoahStapp, line 201: Should this be just `_BulkWriteContext` instead since we're using future annotations?

NoahStapp (changes requested):

pymongo/bulk.py
blink1073: No, it needs to be the class, not an instance

sleepyStick: I don't think so. The function is returning the class itself..not an instance of the class. If it were an instance of the class then I think it would be `_BulkWriteContext` but since the return statement is like `return _BulkWriteContext`, I think the return type has to be `Type[_BulkWriteContext]`

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM!

PR #1252, opened 2023-06-21
PYTHON-2965 Migrate to a PEP517 compliant build system

blink1073:


setup.py
```diff
@@ -4,10 +4,6 @@
 import sys
 import warnings
 
-if sys.version_info[:3] < (3, 7):
-    raise RuntimeError("Python version >= 3.7 required.")
```
ShaneHarvey, line 8: Do we still need this as long as setup.py exists?

setup.py
```diff
@@ -25,29 +21,6 @@
...
-f = open("README.rst")
-try:
-    try:
-        readme_content = f.read()
-    except BaseException:
-        readme_content = ""
-finally:
-    f.close()
-
 # PYTHON-654 - Clang doesn't support -mno-fused-madd but the pythons Apple
```
ShaneHarvey, line 24: How does the new build process handle this "-mno-fused-madd" issue? Or is it no longer a concern.

setup.py
blink1073: No, setuptools will use the `python_requires` config from `pyproject.toml`.

setup.py
blink1073: Actually, we don't need to use the stdlib `distutils` at all, and the `distutils` that ships with `setuptools` does not add this flag.  I cleaned up this block.

ShaneHarvey (approved):

PR #1263, opened 2023-06-26
PYTHON-3588 Expose an API to create a cursor from a command response

NoahStapp:
Any failing tests should be due to unsupported use of cursorType as per the current API spec.

### DIFF
--- a/pymongo/command_cursor.py
+++ b/pymongo/command_cursor.py
@@ -19,7 +19,7 @@ from collections import deque
 from typing import TYPE_CHECKING, Any, Generic, Iterator, Mapping, NoReturn, Optional
 
 from bson import _convert_raw_document_lists_to_streams
-from pymongo.cursor import _CURSOR_CLOSED_ERRORS, _SocketManager
+from pymongo.cursor import _CURSOR_CLOSED_ERRORS, CursorType, _SocketManager
 from pymongo.errors import ConnectionFailure, InvalidOperation, OperationFailure
 from pymongo.message import _CursorAddress, _GetMore, _RawBatchGetMore
 from pymongo.response import PinnedResponse
@@ -41,6 +41,7 @@ class CommandCursor(Generic[_DocumentType]):
         cursor_info: Mapping[str, Any],
         address: Optional[_Address],
         batch_size: int = 0,
+        cursor_type: int = CursorType.NON_TAILABLE,
         max_await_time_ms: Optional[int] = None,
         session: Optional[ClientSession] = None,
         explicit_session: bool = False,
@@ -71,6 +72,15 @@ class CommandCursor(Generic[_DocumentType]):
 
         if not isinstance(max_await_time_ms, int) and max_await_time_ms is not None:
             raise TypeError("max_await_time_ms must be an integer or None")
+
+        if cursor_type not in (
+            CursorType.NON_TAILABLE,
+            CursorType.TAILABLE,
+            CursorType.TAILABLE_AWAIT,
+        ):
+            raise ValueError("not a valid value for cursor_type")
+        else:
+            self.__cursor_type = cursor_type
 
     def __del__(self) -> None:
         self.__die()
--- a/pymongo/command_cursor.py
+++ b/pymongo/command_cursor.py
@@ -292,7 +292,7 @@ class CommandCursor(Generic[_DocumentType]):
 
     __next__ = next
 
-    def _try_next(self, get_more_allowed) -> Optional[_DocumentType]:
+    def _try_next(self, get_more_allowed: bool) -> Optional[_DocumentType]:
         """Advance the cursor blocking for at most one getMore command."""
         if not len(self.__data) and not self.__killed and get_more_allowed:
             self._refresh()
--- a/pymongo/command_cursor.py
+++ b/pymongo/command_cursor.py
@@ -301,6 +301,9 @@ class CommandCursor(Generic[_DocumentType]):
         else:
             return None
 
+    def try_next(self):
+        return self._try_next(get_more_allowed=True)
+
     def __enter__(self) -> "CommandCursor[_DocumentType]":
         return self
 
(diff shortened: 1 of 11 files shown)


ShaneHarvey (changes requested):

pymongo/database.py
```diff
@@ -833,6 +833,114 @@ def command(
                 **kwargs,
             )
 
+    @_csot.apply
```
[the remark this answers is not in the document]
NoahStapp: The existing _try_next method simply waits for data to become available or the cursor to be killed. I added a try_next() wrapper around it for testing and it caused this test to hang. What behavior do we expect from the try_next() method?

test/unified_format.py
```diff
@@ -1087,6 +1087,25 @@ def _databaseOperation_runCommand(self, target, **kwargs):
...
+    def _databaseOperation_runCursorCommand(self, target, **kwargs):
+        self.__raise_if_unsupported("runCursorCommand", target, Database)
+        # Ensure the first key is the command name.
+        ordered_command = SON([(kwargs.pop("command_name"), 1)])
+        ordered_command.update(kwargs["command"])
+        kwargs["command"] = ordered_command
+        if "maxTimeMS" in kwargs:
+            kwargs["max_time_ms"] = kwargs["maxTimeMS"]
+            del kwargs["maxTimeMS"]
+        return list(target.cursor_command(**kwargs))
```
ShaneHarvey, line 1099: What to think about refactoring this to share the main logic with  _databaseOperation_createCommandCursor?

test/unified_format.py
```diff
@@ -1087,6 +1087,25 @@ def _databaseOperation_runCommand(self, target, **kwargs):
...
+
+    def _databaseOperation_createCommandCursor(self, target, **kwargs):
+        self.__raise_if_unsupported("createCommandCursor", target, Database)
+        # Ensure the first key is the command name.
+        ordered_command = SON([(kwargs.pop("command_name"), 1)])
+        ordered_command.update(kwargs["command"])
+        kwargs["command"] = ordered_command
+        return target.cursor_command(**kwargs)
```
ShaneHarvey, line 1107: Let's add support the cursorType field for now (while DRIVERS-2664 is discussed). Like:
```
ctype = kwargs.pop("cursor_type", "nonTailable")
if ctype == "tailable":
    ordered_command["tailable"] = True
elif ctype == "tailableAwait":
    ordered_command["tailable"] = True
    ordered_command["awaitData"] = True
elif ctype != "nonTailable"
    self.fail(f"unknown cursorType: {ctype}")
```

pymongo/database.py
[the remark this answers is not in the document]
ShaneHarvey: I beleive that's because of the missing cursorType handling. For now let's add it in the unified test runner like I describe here: https://github.com/mongodb/mongo-python-driver/pull/1263#discussion_r1244405528

test/unified_format.py
NoahStapp: Done, it now just calls `_databaseOperation_createCommandCursor` and wraps the result in a `list()`

test/unified_format.py
```diff
@@ -1252,6 +1287,8 @@ def run_entity_operation(self, spec):
             method_name = f"_changeStreamOperation_{opname}"
         elif isinstance(target, NonLazyCursor):
             method_name = f"_cursor_{opname}"
+        elif isinstance(target, CommandCursor):
```
ShaneHarvey, line 1290: What do you think about instead adding CommandCursor to the `_cursor_` elif above?:
```python
        elif isinstance(target, (NonLazyCursor, CommandCursor)):
            method_name = f"_cursor_{opname}"
```

NoahStapp: Pushed too early sorry, I added CommandCursor as a valid type for all of the `_cursor` test functions.

test/test_database.py
```diff
@@ -407,6 +408,30 @@ def test_command_with_regex(self):
...
+        cursor = db.cursor_command("find", "test")
+        for x in cursor:
+            for (k, _v) in x.items():
+                self.assertEqual(k, "_id")
+                break
+
+    def test_cursor_command_invalid(self):
+        # All commands that do not return a cursor should cause a TypeError to be raised.
+        db = self.client.pymongo_test
+        db.test.drop()
```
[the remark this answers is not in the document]
NoahStapp: As in completely remove this test, or change it to not drop the database?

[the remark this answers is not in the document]
ShaneHarvey: Yes, change it to not drop the db.

test/test_database.py
```diff
@@ -407,6 +410,26 @@ def test_command_with_regex(self):
         for doc in result["cursor"]["firstBatch"]:
             self.assertTrue(isinstance(doc["r"], Regex))
 
+    def test_cursor_command(self):
+        db = self.client.pymongo_test
+        db.test.drop()
+        db.test.insert_one(SON([("hello", "world"), ("_id", 5)]))
+        db.test.insert_one(SON([("hello2", "world2"), ("_id", 6)]))
+        db.test.insert_one(SON([("hello3", "world3"), ("_id", 7)]))
```
ShaneHarvey, line 418: This is great but we can do it more efficiently (make tests faster) by using insert_many() to bulk insert multiple documents in one command:
```python
docs = [{"_id": i, "doc": i} for i in range(3)]
db.test.insert_many(docs)
...
actual_docs = list(cursor)
self.assertEqual(docs, actual_docs)
```

pymongo/command_cursor.py
```diff
@@ -301,6 +301,9 @@ def _try_next(self, get_more_allowed):
         else:
             return None
 
+    def try_next(self):
```
ShaneHarvey, line 304: Let's add a docstring here similar to ChangeStream.try_next. We should also mention the new cursor_command() and try_next() methods in the changelog.

ShaneHarvey (changes requested):

pymongo/command_cursor.py
NoahStapp: Got it, added both.

ShaneHarvey (approved):
LGTM!

PR #1269, opened 2023-06-27
PYTHON-3871 fix type for server_selector

sleepyStick:


pymongo/settings.py
```diff
@@ -111,7 +125,7 @@ def server_selection_timeout(self) -> int:
         return self._server_selection_timeout
 
     @property
-    def server_selector(self) -> Optional[_ServerSelector]:
+    def server_selector(self) -> Union[_ServerSelector, Callable[[Selection], Selection], None]:
```
ShaneHarvey, line 128: These types are correct, it's actually TopologySettings.server_selector which is incorrect.

ShaneHarvey (changes requested):

pymongo/settings.py
ShaneHarvey: To elaborate, `def server_selector(self) -> Optional[_ServerSelector]:` is correct here. server_selector in this context refers to the server_elector argument to MongoClient:
```
          - `server_selector`: (callable or None) Optional, user-provided
            function that augments server selection rules. The function should
            accept as an argument a list of
            :class:`~pymongo.server_description.ServerDescription` objects and
            return a list of server descriptions that should be considered
            suitable for the desired operation.
```

TopologySettings.server_selector actually has an incorrect type right now which is what is causing the current mypy errors.

sleepyStick: I thought this method was acting as `TopologySettings.server_selector`, given that it returns `self._server_selector` where `self` is `TopologySettings`. If this isn't `TopologySettings.server_selector`, can you point me to where `TopologySettings.server_selector` is?

ShaneHarvey: Oh sorry, I meant to say **ClientOptions.server_selector** actually has an incorrect type right now.

sleepyStick: Makes sense. Thank you! And changed!

ShaneHarvey (approved):
LGTM!

PR #1270, opened 2023-06-27
PYTHON-3774 Update Evergreen run-atlas-tests to use tox + pytest

NoahStapp:
The Atlas test on PyPy 3.7 fails due to an ASCII encoding error that I haven't been able to solve with Python interpreter environment variables, see [here for the error](https://spruce.mongodb.com/task/mongo_python_driver_atlas_connect__platform~rhel8_python_version~pypy3.7_atlas_connect_patch_8b2320440863ccd87b50713a2dafdf0033016e4f_649b5003850e61a3dcfd1ea7_23_06_27_21_09_42/logs?execution=0):


ShaneHarvey (commented):
For the Unicode error could you try this?:

> If anyone is still hitting this issue, it's probably caused by an error messages emitted by pip during preparation of the testing environment (in my case, it was a failing python-mysql build).
> 
> In my particular case I was able to get to the actual error message by exporting LANG=en_US.UTF-8
> 
> ref: https://stackoverflow.com/a/43385201/574084

https://github.com/tox-dev/tox/issues/1369#issuecomment-514623340

.evergreen/config.yml
```diff
@@ -553,12 +553,16 @@ functions:
...
           rm -f ./prepare_atlas_connectivity.sh
 
-          PYTHON_BINARY=${PYTHON_BINARY} bash ${PROJECT_DIRECTORY}/.evergreen/run-atlas-tests.sh
+          set -o errexit
+          set -o xtrace
+          cd ${PROJECT_DIRECTORY}
+          alias python=${PYTHON_BINARY}
+          python -m pip install --upgrade pip
+          python -m pip install tox
+          python -m tox -e test-atlas
```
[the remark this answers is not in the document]
NoahStapp: If we only test on a toolchain host that sounds good to me.

NoahStapp:
> For the Unicode error could you try this?:
> 
> > If anyone is still hitting this issue, it's probably caused by an error messages emitted by pip during preparation of the testing environment (in my case, it was a failing python-mysql build).
> > In my particular case I was able to get to the actual error message by exporting LANG=en_US.UTF-8
> > ref: [stackoverflow.com/a/43385201/574084](https://stackoverflow.com/a/43385201/574084)
> 
> [tox-dev/tox#1369 (comment)](https://github.com/tox-dev/tox/issues/1369#issuecomment-514623340)

Still failed sadly even with that setting in the tox env: https://evergreen.mongodb.com/task_log_raw/mongo_python_driver_atlas_connect__platform~rhel8_python_version~pypy3.7_atlas_connect_patch_8b2320440863ccd87b50713a2dafdf0033016e4f_649c59cc2a60ed590b378851_23_06_28_16_03_41/0?type=ALL

NoahStapp:
Had to use LC_ALL=en_US.UTF-8 for it to work.

.evergreen/config.yml
```diff
@@ -555,12 +555,14 @@ functions:
       params:
         working_dir: "src"
         script: |
-          # Disable xtrace (just in case it was accidentally set).
-          set +x
```
ShaneHarvey, line 559: Can this be added back now?

.evergreen/config.yml
```diff
@@ -555,12 +555,14 @@ functions:
...
           . ./prepare_atlas_connectivity.sh
           rm -f ./prepare_atlas_connectivity.sh
 
-          PYTHON_BINARY=${PYTHON_BINARY} bash ${PROJECT_DIRECTORY}/.evergreen/run-atlas-tests.sh
+          set -o errexit
+          set -o xtrace
+          cd ${PROJECT_DIRECTORY}
```
ShaneHarvey, line 563: This cd is not needed IIUC.

ShaneHarvey (commented):
Now that you've figured out how to workaround the unicode error, would it be possible to identify the problematic file itself? Maybe via a spawn host? It would be interesting to know what filename has the biohazard symbol in it.

.evergreen/config.yml
NoahStapp: I believe we want xtrace since we are running tox from here directly instead of from a separate bash file like before.

ShaneHarvey: This is here to avoid leaking info on the next line: `. ./prepare_atlas_connectivity.sh`

NoahStapp: Ah I see, makes sense!

.evergreen/config.yml
```diff
@@ -555,12 +555,16 @@ functions:
...
+          # Disable xtrace for security reasons (just in case it was accidentally set).
           set +x
+
           . ./prepare_atlas_connectivity.sh
           rm -f ./prepare_atlas_connectivity.sh
 
-          PYTHON_BINARY=${PYTHON_BINARY} bash ${PROJECT_DIRECTORY}/.evergreen/run-atlas-tests.sh
+          set -o errexit
+          set -o xtrace
+          export LC_ALL=en_US.UTF-8
```
ShaneHarvey, line 566: Can you add a comment for why this line is needed so that our future selves remember?

.evergreen/config.yml
```diff
@@ -555,12 +555,16 @@ functions:
...
+          ${PYTHON_BINARY} -m tox -e test-atlas
```
ShaneHarvey, line 567: Should this use the new run-tox script now?

NoahStapp: I was planning to do that in a tiny future PR to separate out the two separate tickets involved since there's a lot of changes between these changes and the tox.sh script.

ShaneHarvey (approved):

PR #1277, opened 2023-06-27
PYTHON-3789 Use tox for Evergreen mockupdb tests

NoahStapp:
We should decide if we want to continue having Evergreen use a bash script to setup + run each test suite using tox, or if we want to put the contents of that script into the config.yml file directly.


ShaneHarvey (changes requested):

ShaneHarvey:
> We should decide if we want to continue having Evergreen use a bash script to setup + run each test suite using tox, or if we want to put the contents of that script into the config.yml file directly.

I don't think we need a bash script for things as simple as "python -m tox -e "test-*" but if some of the tests get more complicated than that we might still need a script.

.evergreen/config.yml
```diff
@@ -357,7 +357,10 @@ functions:
         script: |
           set -o xtrace
           ${PREPARE_SHELL}
-          PYTHON_BINARY=${PYTHON_BINARY} bash ${PROJECT_DIRECTORY}/.evergreen/run-mockupdb-tests.sh
+
+          cd ${PROJECT_DIRECTORY}
```
ShaneHarvey, line 361: I think this `cd` can be removed because we're already in the `working_dir: "src"`.

.evergreen/config.yml
```diff
@@ -357,7 +357,10 @@ functions:
...
+          alias python=${PYTHON_BINARY}
```
ShaneHarvey, line 362: Is this `alias` needed or can we just do `$PYTHON_BINARY -m tox ...`?

ShaneHarvey: Oh so the alias is used within the tox config? Interesting.

NoahStapp: Yes, tox is configured to use whatever Python `python` points to so as long as we correctly set the path of `python` a host it should behave as expected.

ShaneHarvey (approved):

PR #1278, opened 2023-06-27
PYTHON-3788 add types to server_selectors.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3788

### DIFF
--- a/pymongo/server_selectors.py
+++ b/pymongo/server_selectors.py
@@ -13,15 +13,22 @@
 # permissions and limitations under the License.
 
 """Criteria to select some ServerDescriptions from a TopologyDescription."""
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, List, Optional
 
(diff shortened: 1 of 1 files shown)


pymongo/server_selectors.py
```diff
@@ -36,54 +43,60 @@ def from_topology_description(cls, topology_description):
...
+    def __bool__(self) -> bool:
         return bool(self.server_descriptions)
 
-    def __getitem__(self, item):
+    def __getitem__(self, item: int) -> ServerDescription:
         return self.server_descriptions[item]
 
 
-def any_server_selector(selection):
+def any_server_selector(selection: Selection) -> Selection:
```
ShaneHarvey, line 89: What do you think about making this generic since it just returns whatever is provided?

ShaneHarvey (changes requested):

PR #1280, opened 2023-06-28
PYTHON-3790 Standardize Evergreen tox setup

NoahStapp:


.evergreen/config.yml
```diff
@@ -358,8 +358,7 @@ functions:
           set -o xtrace
           ${PREPARE_SHELL}
 
-          alias python=${PYTHON_BINARY}
-          python -m tox -e test-mockupdb
+          bash -i ${PROJECT_DIRECTORY}/.evergreen/tox.sh ${PYTHON_BINARY} -e test-mockupdb
```
ShaneHarvey, line 361: What does `-i` do?

.evergreen/tox.sh
```diff
@@ -0,0 +1,31 @@
...
+
+PYTHON_BINARY="$1"
+
+if $PYTHON_BINARY -m tox --version; then
+    alias run_tox='$PYTHON_BINARY -m tox'
+elif python3 -m tox --version; then
+    alias run_tox='python3 -m tox'
+elif tox --version; then
+    alias run_tox=tox
+else # No toolchain present, set up venv before installing tox
```
ShaneHarvey, line 12: Can we use createvirtualenv from utils.sh?

.evergreen/config.yml
NoahStapp: Makes the script execution interactive, so aliases defined inside of the script will correctly propagate to subprocesses created inside of it.

.evergreen/tox.sh
NoahStapp: Currently createvirtualenv uses this flow to determine its venv executable:

```
if $PYTHON -m virtualenv --version; then
        VIRTUALENV="$PYTHON -m virtualenv"
elif $PYTHON -m venv -h > /dev/null; then
        # System virtualenv might not be compatible with the python3 on our path
        VIRTUALENV="$PYTHON -m venv"
```

can we remove the first check to migrate off of virtualenv, which is deprecated since Python 3.6?

.evergreen/config.yml
ShaneHarvey: Could you add a comment explaining that subtlety?

ShaneHarvey: Are you sure that the alias is needed at all? In my testing, tox sets "python" to the interpreter that tox was installed in, so there should be no need for an alias IIUC.

.evergreen/tox.sh
```diff
@@ -0,0 +1,31 @@
...
+elif python3 -m tox --version; then
+    run_tox() {
+      python3 -m tox "$@"
+    }
+elif tox --version; then
+    run_tox() {
+      tox "$@"
+    }
+else # No toolchain present, set up venv before installing tox
+    createvirtualenv PYTHON_BINARY env
```
ShaneHarvey, line 21: PYTHON_BINARY -> "$PYTHON_BINARY"

Also need to run `python -m pip install tox`

.evergreen/tox.sh
```diff
@@ -0,0 +1,31 @@
...
+    run_tox() {
+      $PYTHON_BINARY -m tox "$@"
```
ShaneHarvey, line 23: $PYTHON_BINARY -> python

At this point we'll be running in a venv.

.evergreen/utils.sh
```diff
@@ -31,8 +31,7 @@ createvirtualenv () {
 
     python -m pip install --upgrade pip
     python -m pip install --upgrade setuptools wheel
-    # lxml only has wheels for macos 10.15+
-    python -m pip install unittest-xml-reporting || true
```
ShaneHarvey, line 35: Let's avoid changing this for now since we don't always need tox where createvirtualenv is used. Instead let's install tox in tox.sh.

ShaneHarvey (changes requested):

.evergreen/tox.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if $PYTHON_BINARY -m tox --version; then
+    run_tox() {
+      $PYTHON_BINARY -m tox "$@"
+    }
+elif python3 -m tox --version; then
+    run_tox() {
+      python3 -m tox "$@"
+    }
+elif tox --version; then
```
ShaneHarvey, line 16: I just realized that we can't use `python3 -m tox` or `tox --version` because those would end up using an incorrect python version (it wouldn't use $PYTHON_BINARY). Let's remove the two elifs.

ShaneHarvey (changes requested):

.evergreen/tox.sh
```diff
@@ -0,0 +1,24 @@
...
+    python -m pip install tox
+    run_tox() {
+      python -m tox "$@"
+    }
+fi
+
+run_tox "${@:2}"
+
+if command -v deactivate &> /dev/null; then
+    trap "deactivate; rm -rf env" EXIT HUP
```
ShaneHarvey, line 23: let's move this to under the createvirtualenv line and update the `rm` to  "toxenv", like this:
```
createvirtualenv "$PYTHON_BINARY" toxenv
trap "deactivate; rm -rf env" EXIT HUP
```

ShaneHarvey: (This is the pattern we use in run-tests.sh and elsewhere.)

ShaneHarvey (approved):

PR #1285, opened 2023-06-30
PYTHON-3777 add types to command_cursor.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3777

### DIFF
--- a/pymongo/command_cursor.py
+++ b/pymongo/command_cursor.py
@@ -26,20 +26,12 @@ from typing import (
     NoReturn,
     Optional,
     Union,
-    cast,
 )
 
 from bson import _convert_raw_document_lists_to_streams
 from pymongo.cursor import _CURSOR_CLOSED_ERRORS, _SocketManager
 from pymongo.errors import ConnectionFailure, InvalidOperation, OperationFailure
-from pymongo.message import (
-    _CursorAddress,
-    _GetMore,
-    _OpMsg,
-    _OpReply,
-    _Query,
-    _RawBatchGetMore,
-)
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
LGTM!

pymongo/command_cursor.py
```diff
@@ -51,7 +71,7 @@ def __init__(
         self.__collection: Collection[_DocumentType] = collection
         self.__id = cursor_info["id"]
         self.__data = deque(cursor_info["firstBatch"])
-        self.__postbatchresumetoken = cursor_info.get("postBatchResumeToken")
+        self.__postbatchresumetoken: int = cast(int, cursor_info.get("postBatchResumeToken"))
```
ShaneHarvey, line 74: postBatchResumeToken should be an optional mapping, not an int.

pymongo/command_cursor.py
```diff
@@ -158,7 +178,7 @@ def _maybe_pin_connection(self, sock_info):
             else:
                 self.__sock_mgr = sock_mgr
 
-    def __send_message(self, operation):
+    def __send_message(self, operation: Union[_Query, _GetMore]) -> None:
```
ShaneHarvey, line 181: operation can only be _GetMore here, not _Query.

ShaneHarvey (changes requested):

pymongo/command_cursor.py
sleepyStick: Changed!

pymongo/command_cursor.py
sleepyStick: thanks for catching that!

ShaneHarvey (approved):

PR #1290, opened 2023-07-05
PYTHON-3780 add types to cursor.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3780

### DIFF
--- a/pymongo/cursor.py
+++ b/pymongo/cursor.py
@@ -142,10 +142,9 @@ class CursorType:
 class _SocketManager:
     """Used with exhaust cursors to ensure the socket is returned."""
 
-    def __init__(self, sock: SocketInfo, more_to_come: bool):
+    def __init__(self, sock: Optional[SocketInfo], more_to_come: bool):
         self.sock = sock
         self.more_to_come = more_to_come
-        self.closed = False
         self.lock = _create_lock()
 
     def update_exhaust(self, more_to_come: bool) -> None:
@@ -153,10 +152,9 @@ class _SocketManager:
 
     def close(self) -> None:
         """Return this instance's socket to the connection pool."""
-        if not self.closed:
-            self.closed = True
+        if self.sock:
             self.sock.unpin()
-            self.sock = None  # type: ignore[assignment]
+            self.sock = None
 
 
 _Sort = Sequence[Union[str, Tuple[str, Union[int, str, Mapping[str, Any]]]]]
--- a/pymongo/cursor.py
+++ b/pymongo/cursor.py
@@ -26,10 +26,12 @@ from typing import (
     Iterable,
     List,
     Mapping,
+    MutableMapping,
     NoReturn,
     Optional,
     Sequence,
     Tuple,
+    TypeVar,
     Union,
     cast,
     overload,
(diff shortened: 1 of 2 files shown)


pymongo/cursor.py
```diff
@@ -131,33 +142,27 @@ class CursorType:
 class _SocketManager:
     """Used with exhaust cursors to ensure the socket is returned."""
 
-    def __init__(self, sock, more_to_come):
+    def __init__(self, sock: SocketInfo, more_to_come: bool):
         self.sock = sock
```
blink1073, line 146: We should instead make this `Optional[None]` instead of adding the ignore assignment so the other methods can handle when it is `None`

pymongo/cursor.py
```diff
@@ -1277,23 +1287,35 @@ def __deepcopy__(self, memo: Any) -> Any:
         """
         return self._clone(deepcopy=True)
 
-    def _deepcopy(self, x, memo=None):
+    @overload
+    def _deepcopy(self, x: Iterable, memo: Optional[Dict[int, Union[List, Dict]]] = None) -> List:
```
blink1073, line 1291: Can these be `Mapping` instead of `Dict`?

blink1073 (changes requested):

pymongo/cursor.py
sleepyStick: I don't think so because `copy.deepcopy` takes in a `Dict` and we call that on line's 1325 and 1331

pymongo/cursor.py
sleepyStick: If I make it `Optional[SockInfo]`, then we'd have to add an ignore on line 158 because we'd try to call `unpin` on something that can be `None`. Is that preferred?

ShaneHarvey: Perhaps "if not self.closed" can be removed and replaced with `if self.sock`?

pymongo/cursor.py
blink1073: 👍🏼

pymongo/cursor.py
blink1073: I still think `Optional[SockInfo]` is the right type to use (haha on my typo earlier).

sleepyStick: yeah, so now that the if statement doesn't rely on `close`, sock is `Optional[SockInfo]`

blink1073: To be clear, the parameter should still be `SockInfo`, but the property should be `Optional[SockInfo]`

sleepyStick: ahh I see I see, sorry about that. Fixed it now!

pymongo/cursor.py
```diff
@@ -684,7 +687,7 @@ def __getitem__(self, index: int) -> _DocumentType:
     def __getitem__(self, index: slice) -> "Cursor[_DocumentType]":
         ...
 
-    def __getitem__(self, index):
+    def __getitem__(self, index: Union[int, slice]) -> Union[_DocumentType, Cursor[_DocumentType]]:
```
ShaneHarvey, line 690: Are you sure we need to add type hints here? The examples in the docs only add hints to the `@overload` lines, for example: https://docs.python.org/3/library/typing.html#typing.overload

sleepyStick: Uh good question, I added it because I've been making sure that everything is typed by running mypy with the `--disallow-untyped-defs` flag and mypy will complain that its an untyped def. I can remove it if that is preferred.

PR #1292, opened 2023-07-05
PYTHON-3797 cache commonly used strings

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3797

### DIFF
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -71,9 +71,9 @@ struct module_state {
     PyObject* _replace_str;
     PyObject* _astimezone_str;
     PyObject* _id_str;
-    PyObject* _ref_str;
-    PyObject* __id_str;
-    PyObject* _db_str;
+    PyObject* _dollar_ref_str;
+    PyObject* _dollar_id_str;
+    PyObject* _dollar_db_str;
     PyObject* _tzinfo_str;
 };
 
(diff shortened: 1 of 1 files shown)


bson/_cbsonmodule.c
```diff
@@ -70,6 +70,11 @@ struct module_state {
     PyObject* _bid_str;
     PyObject* _replace_str;
     PyObject* _astimezone_str;
+    PyObject* _id_str;
+    PyObject* _$ref_str;
```
blink1073, line 74: While these are valid identifiers, they are a bit confusing (and also tripped up the syntax highlighting), can we remove the dollar signs?

blink1073 (changes requested):

bson/_cbsonmodule.c
sleepyStick: yes, fixed!

bson/_cbsonmodule.c
```diff
@@ -70,6 +70,11 @@ struct module_state {
     PyObject* _bid_str;
     PyObject* _replace_str;
     PyObject* _astimezone_str;
+    PyObject* _id_str;
+    PyObject* _ref_str;
```
blink1073, line 74: Hmm, now the `__id_str` stands out as odd.  Perhaps use `_dollar_<foo>_str` for each of them?

sleepyStick: good idea! just changed it

ShaneHarvey (approved):
Nice work, LGTM!

PR #1293, opened 2023-07-06
PYTHON-3800 Add test for SRV URI with uppercase hostname

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3800

This doesn't fully test to issue but it's a start. What we eventually want is to test the case where the hosts returned in the SRV record have uppercase letters but that will need to wait for https://jira.mongodb.org/browse/DRIVERS-2668.

### DIFF
--- a/test/test_dns.py
+++ b/test/test_dns.py
@@ -21,7 +21,7 @@ import sys
 
 sys.path[0:0] = [""]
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thanks!

PR #1295, opened 2023-07-07
PYTHON-3802 add types to database.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3802

### DIFF
--- a/bson/codec_options.py
+++ b/bson/codec_options.py
@@ -485,7 +485,7 @@ else:
             return CodecOptions(**opts)
 
 
-DEFAULT_CODEC_OPTIONS: "CodecOptions[Dict[str, Any]]" = CodecOptions()
+DEFAULT_CODEC_OPTIONS: "CodecOptions[Mapping[str, Any]]" = CodecOptions()
 
 
 def _parse_codec_options(options: Any) -> CodecOptions:
--- a/bson/codec_options.py
+++ b/bson/codec_options.py
@@ -485,7 +485,7 @@ else:
             return CodecOptions(**opts)
 
 
-DEFAULT_CODEC_OPTIONS: "CodecOptions[Mapping[str, Any]]" = CodecOptions()
+DEFAULT_CODEC_OPTIONS: "CodecOptions[Dict[str, Any]]" = CodecOptions()
 
 
 def _parse_codec_options(options: Any) -> CodecOptions:
--- a/pymongo/database.py
+++ b/pymongo/database.py
@@ -696,12 +696,12 @@ class Database(common.BaseObject, Generic[_DocumentType]):
         check: bool = True,
         allowable_errors: Optional[Sequence[Union[str, int]]] = None,
         read_preference: Optional[_ServerMode] = ReadPreference.PRIMARY,
-        codec_options: CodecOptions[Dict[str, Any]] = DEFAULT_CODEC_OPTIONS,
+        codec_options: CodecOptions[Mapping[str, Any]] = DEFAULT_CODEC_OPTIONS,
         write_concern: Optional[WriteConcern] = None,
         parse_write_concern_error: bool = False,
         session: Optional[ClientSession] = None,
         **kwargs: Any,
-    ) -> Dict[str, Any]:
+    ) -> Mapping[str, Any]:
         ...
 
     @overload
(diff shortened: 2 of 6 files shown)


bson/codec_options.py
```diff
@@ -485,7 +485,7 @@ def with_options(self, **kwargs: Any) -> "CodecOptions":
             return CodecOptions(**opts)
 
 
-DEFAULT_CODEC_OPTIONS: "CodecOptions[Mapping[str, Any]]" = CodecOptions()
```
blink1073, line 488: I think all of the places you changed to "Dict" can be replaced with "Mapping", if you use a cast in `_drop_helper` to convert the result from `_command` to `Dict`.

sleepyStick: I agree that would work. Why is `Mapping` preferred over `Dict`? When looking into the `CodecOptions` class, it looks like the default is a dict.

blink1073: In general we want to use read-only interfaces where possible.  The underlying implementation is a `dict`, but it should be considered read-only for most use cases.

blink1073 (approved):
LGTM!

pymongo/database.py
```diff
@@ -942,20 +1015,25 @@ def cursor_command(
...
-        **kwargs,
-    ):
+        command: Union[str, MutableMapping[str, Any]],
+        value: int = 1,
+        check: bool = True,
+        allowable_errors: Optional[Sequence[Union[str, int]]] = None,
+        read_preference: Optional[_ServerMode] = None,
+        codec_options: CodecOptions = DEFAULT_CODEC_OPTIONS,
+        session: Optional[ClientSession] = None,
+        **kwargs: Any,
```
ShaneHarvey, line 1025: I think all of these arguments can be removed except `command` and `session`. _retryable_read_command is only used once: `admin._retryable_read_command(cmd, session=session)`

pymongo/database.py
```diff
@@ -680,20 +687,56 @@ def watch(
...
+    @overload
+    def _command(
+        self,
+        sock_info: SocketInfo,
+        command: Union[str, MutableMapping[str, Any]],
+        value: int = 1,
+        check: bool = True,
+        allowable_errors: Optional[Sequence[Union[str, int]]] = None,
+        read_preference: Optional[_ServerMode] = ReadPreference.PRIMARY,
+        codec_options: CodecOptions[Mapping[str, Any]] = DEFAULT_CODEC_OPTIONS,
```
ShaneHarvey, line 699: How is this different from `codec_options: CodecOptions[_CodecDocumentType] = ...,` below? Should this be `codec_options: None`?

pymongo/database.py
```diff
@@ -713,6 +756,36 @@ def _command(
...
+        command: Union[str, MutableMapping[str, Any]],
+        value: Any = 1,
+        check: bool = True,
+        allowable_errors: Optional[Sequence[Union[str, int]]] = None,
+        read_preference: Optional[_ServerMode] = None,
+        codec_options: None = None,
+        session: Optional[ClientSession] = None,
+        comment: Optional[Any] = None,
+        **kwargs: Any,
+    ) -> Mapping[str, Any]:
```
ShaneHarvey, line 771: Should this be `Dict[str, Any]`? I thought our guiding principle was "accept abstract types as input but return concrete types".

pymongo/database.py
```diff
@@ -1079,19 +1168,24 @@ def list_collection_names(
...
         with self.__client._socket_for_writes(session) as sock_info:
-            return self._command(
-                sock_info,
-                command,
-                allowable_errors=["ns not found", 26],
-                write_concern=self._write_concern_for(session),
-                parse_write_concern_error=True,
-                session=session,
+            return cast(
+                Dict[str, Any],
```
ShaneHarvey, line 1180: Cast no longer needed if we default _command to return Dict (the concrete type).

test/test_typing.py
```diff
@@ -479,7 +490,7 @@ class TestCommandDocumentType(unittest.TestCase):
     @only_type_check
     def test_default(self) -> None:
         client: MongoClient = MongoClient()
-        result: Dict = client.admin.command("ping")
+        result: MutableMapping[str, Any] = cast(MutableMapping, client.admin.command("ping"))
```
ShaneHarvey, line 493: This highlights the problem with returning Mapping, it will be annoying for users to have to call cast().

ShaneHarvey (changes requested):

bson/codec_options.py
ShaneHarvey: I think we should prefer to return Dict when we actually do return Dicts. Like db.command("ping") ideally returns a Dict, since that's what it actually returns.

blink1073: I guess it comes down to whether we want to expose the fact that the return value is mutable.  In the case of `command`, I agree, we could expose this fact.

sleepyStick: Originally, I changed them to `Dict` because I thought it would be better to use the more specific type, especially when we know for sure that it is a `Dict`

pymongo/database.py
sleepyStick: makes sense. removed

pymongo/database.py
ShaneHarvey: I'm confused because later in database.command we use `codec_options: None = None,` in the first overload. Shouldn't we do the same here?

sleepyStick: In database.command, codec_options can be `None` and if it is none, it'll pass in `DEFAULT_CODEC_OPTIONS` into `_command`. The default value for `codec_options` in `_command` is also not `None` so I don't think it would make sense to have an overloaded version take in `None`

ShaneHarvey (approved):
LGTM!

pymongo/database.py
ShaneHarvey: I see, in that case this LGTM.

PR #1296, opened 2023-07-07
PYTHON-3803 add types to encryption.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3803

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -53,6 +53,7 @@ bytes [#bytes]_                          binary         both
 .. [#bytes] The bytes type is encoded as BSON binary with
    subtype 0. It will be decoded back to bytes.
 """
+from __future__ import annotations
 
 import datetime
 import itertools
@@ -84,6 +85,7 @@ from typing import (
     TypeVar,
     Union,
     cast,
+    overload,
 )
 
 from bson.binary import (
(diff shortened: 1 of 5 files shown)


bson/__init__.py
```diff
@@ -1025,9 +1025,21 @@ def encode(
     return _dict_to_bson(document, check_keys, codec_options)
 
 
+@overload
+def decode(data: "_ReadableBuffer", codec_options: None = None) -> Dict[str, Any]:
```
ShaneHarvey, line 1029: Nice! This is great! I think we want to apply this same overload pattern throughout the codebase but I think that's too much for this PR. Could you open a new ticket for it? Eg decode_all, decode_iter, and anywhere else we use _DocumentType.

bson/raw_bson.py
```diff
@@ -62,7 +62,7 @@
 
 def _inflate_bson(
     bson_bytes: bytes, codec_options: CodecOptions, raw_array: bool = False
-) -> Mapping[Any, Any]:
+) -> MutableMapping[Any, Any]:
```
ShaneHarvey, line 65: Can we use dict here?

ShaneHarvey (changes requested):

bson/__init__.py
sleepyStick: sounds good, made the ticket

bson/raw_bson.py
sleepyStick: yes!

ShaneHarvey (approved):

PR #1300, opened 2023-07-10
PYTHON-3818 Create unique key vault index in auto QE example

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3818


blink1073 (approved):
👍🏼

PR #1301, opened 2023-07-11
PYTHON-3817 Optimize BSON encoding of standard Python dict

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3817


ShaneHarvey (approved):
LGTM! Any rough timeit numbers on the perf improvement here? Besides deeply nested documents this should also yield benefits to flat documents with many fields.

PR #1306, opened 2023-07-13
PYTHON-3829 Certificate typecast in encryption breaking build

NoahStapp:
Successful patch: https://spruce.mongodb.com/version/64b04375306615a3d71adc10/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/pymongo/ocsp_support.py
+++ b/pymongo/ocsp_support.py
@@ -148,8 +148,8 @@ def _verify_signature(
             key.verify(signature, data, algorithm)  # type: ignore[arg-type]
         elif isinstance(key, _EllipticCurvePublicKey):
             key.verify(signature, data, _ECDSA(algorithm))  # type: ignore[arg-type]
-        elif isinstance(key, _X25519PublicKey) or isinstance(
-            key, _X448PublicKey
+        elif isinstance(
+            key, (_X25519PublicKey, _X448PublicKey)
         ):  # Curve25519 and Curve448 keys do not require verification
             return 1
         else:
--- a/pymongo/ocsp_support.py
+++ b/pymongo/ocsp_support.py
@@ -19,7 +19,7 @@ import logging as _logging
 import re as _re
 from datetime import datetime as _datetime
 from datetime import timezone
-from typing import TYPE_CHECKING, Iterable, List, Optional, Type, Union, cast
+from typing import TYPE_CHECKING, Iterable, List, Optional, Type, Union
 
 from cryptography.exceptions import InvalidSignature as _InvalidSignature
 from cryptography.hazmat.backends import default_backend as _default_backend
@@ -30,6 +30,12 @@ from cryptography.hazmat.primitives.asymmetric.ec import (
 )
 from cryptography.hazmat.primitives.asymmetric.padding import PKCS1v15 as _PKCS1v15
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey as _RSAPublicKey
+from cryptography.hazmat.primitives.asymmetric.x448 import (
+    X448PublicKey as _X448PublicKey,
+)
+from cryptography.hazmat.primitives.asymmetric.x25519 import (
+    X25519PublicKey as _X25519PublicKey,
+)
 from cryptography.hazmat.primitives.hashes import SHA1 as _SHA1
 from cryptography.hazmat.primitives.hashes import Hash as _Hash
 from cryptography.hazmat.primitives.serialization import Encoding as _Encoding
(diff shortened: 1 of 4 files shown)


pymongo/ocsp_support.py
```diff
@@ -30,6 +30,12 @@
 )
 from cryptography.hazmat.primitives.asymmetric.padding import PKCS1v15 as _PKCS1v15
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey as _RSAPublicKey
+from cryptography.hazmat.primitives.asymmetric.x448 import (
+    X448PublicKey as _X448PublicKey,
```
ShaneHarvey, line 34: This is new in cryptography 2.5 so we would need to bump our min version: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/x448/#cryptography.hazmat.primitives.asymmetric.x448.X448PublicKey

We have:
```
.evergreen/test-encryption-requirements.txt:2:cryptography>=2
```

And also in pymongocrypt:
```
bindings/python/setup.py:58:        "cryptography>=2.0",
```

ShaneHarvey: We might need to add cryptography>2.5 as an explicit dependency for OCSP.

pymongo/ocsp_support.py
```diff
@@ -132,6 +148,10 @@ def _verify_signature(
             key.verify(signature, data, algorithm)  # type: ignore[arg-type]
         elif isinstance(key, _EllipticCurvePublicKey):
             key.verify(signature, data, _ECDSA(algorithm))  # type: ignore[arg-type]
+        elif isinstance(key, _X25519PublicKey) or isinstance(
+            key, _X448PublicKey
+        ):  # Curve25519 and Curve448 keys do not require verification
```
ShaneHarvey, line 153: Neat! Can you give a reference for this?

ShaneHarvey (changes requested):

pymongo/ocsp_support.py
NoahStapp: For Curve448: https://crypto.stackexchange.com/questions/91710/how-do-i-validate-curve448-public-keys

For Curve25519: https://crypto.stackexchange.com/questions/87709/curve25519-key-validation and https://cr.yp.to/ecdh.html

pymongo/ocsp_support.py
NoahStapp: Is bumping cryptography to >2.5 acceptable for OCSP?

ShaneHarvey: It should be fine given the version reqs here: https://github.com/pyca/pyopenssl/blob/17.2.0/setup.py#L99

ShaneHarvey: Either this or you could try/except importerror and add a fallback.

NoahStapp: So we don't need any changes for anything but the test-encryption-requirements + pymongocrypt?

ShaneHarvey: I see two options:
1) add cryptography>=2.5 as an explicit dependency for the OCSP extra. Also update test-encryption-requirements and pymongocrypt.
2) add a try/except importerror and add a fallback for this import (and maybe the other one)

ShaneHarvey (changes requested):

pymongo/ocsp_support.py
NoahStapp: Option 1 seems cleaner to me, especially since 2.0 is over 5 years old.

ShaneHarvey: SGTM! could you also mention this dependency change in the changelog?

NoahStapp: Done, and PR opened for pymongocrypt change: https://github.com/mongodb/libmongocrypt/pull/668

pymongo/ocsp_support.py
```diff
@@ -132,6 +148,10 @@ def _verify_signature(
             key.verify(signature, data, algorithm)  # type: ignore[arg-type]
         elif isinstance(key, _EllipticCurvePublicKey):
             key.verify(signature, data, _ECDSA(algorithm))  # type: ignore[arg-type]
+        elif isinstance(key, _X25519PublicKey) or isinstance(
```
ShaneHarvey, line 151: You can pass a tuple with both classes to combine these two isinstance calls into one.

ShaneHarvey (commented):
Nice work! One minor nit

ShaneHarvey (approved):

PR #1308, opened 2023-07-13
PYTHON-3831 remove _DocumentIn

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3831

### DIFF
--- a/pymongo/monitoring.py
+++ b/pymongo/monitoring.py
@@ -187,12 +187,12 @@ from __future__ import annotations
 
 import datetime
 from collections import abc, namedtuple
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, Mapping, Optional
 
 from bson.objectid import ObjectId
 from pymongo.hello import Hello, HelloCompat
 from pymongo.helpers import _handle_exception
-from pymongo.typings import _Address, _DocumentOut
+from pymongo.typings import _Address
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


pymongo/monitoring.py
```diff
@@ -187,12 +187,12 @@ def connection_checked_in(self, event):
 
 import datetime
 from collections import abc, namedtuple
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, Mapping, Optional
 
 from bson.objectid import ObjectId
 from pymongo.hello import Hello, HelloCompat
 from pymongo.helpers import _handle_exception
-from pymongo.typings import _Address, _DocumentOut
```
ShaneHarvey, line 195: I believe we need to keep using _DocumentOut in this file since `MutableMapping|RawBSONDocument` is the more specific type and we want to return the most specific types we can.

ShaneHarvey (changes requested):

pymongo/monitoring.py
sleepyStick: Makes sense, I added that back in. 

I had initially removed it because I thought the `_DocumentOut` types were somehow associated with the `_DocumentIn` type. like if a user passed us a Mapping, we'd then give back a Mapping. But if we create a new object and give that back (which i'm guessing is what we do), then it would make sense to keep `_DocumentOut`

ShaneHarvey (approved):
LGTM!

PR #1312, opened 2023-07-17
PYTHON-3806 add types to message.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3806


bson/__init__.py
```diff
@@ -1138,7 +1138,7 @@ def decode_all(
     return _decode_all(data, opts)  # type:ignore[arg-type]
 
 
-def _decode_selective(rawdoc: Any, fields: Any, codec_options: Any) -> Mapping[Any, Any]:
+def _decode_selective(rawdoc: Any, fields: Any, codec_options: Any) -> "_DocumentOut":
```
ShaneHarvey, line 1141: Is this change needed? IIUC this method always returns MutableMapping.

gridfs/grid_file.py
```diff
@@ -854,7 +854,7 @@ class GridOutCursor(Cursor):
     def __init__(
         self,
         collection: Collection,
-        filter: Optional[Mapping[str, Any]] = None,
+        filter: Optional[MutableMapping[str, Any]] = None,
```
ShaneHarvey, line 857: Filter should be Mapping, not MutableMapping.

pymongo/cursor.py
```diff
@@ -170,7 +172,7 @@ class Cursor(Generic[_DocumentType]):
     def __init__(
         self,
         collection: Collection[_DocumentType],
-        filter: Optional[Mapping[str, Any]] = None,
+        filter: Optional[MutableMapping[str, Any]] = None,
```
ShaneHarvey, line 175: Filter should be Mapping, not MutableMapping.

pymongo/command_cursor.py
```diff
@@ -386,13 +388,13 @@ def _unpack_response(
...
         legacy_response: bool = False,
-    ) -> List[Mapping[str, Any]]:
+    ) -> List[_DocumentOut]:
         raw_response = response.raw_response(cursor_id, user_fields=user_fields)
         if not legacy_response:
             # OP_MSG returns firstBatch/nextBatch documents as a BSON array
             # Re-assemble the array of documents into a document stream
             _convert_raw_document_lists_to_streams(raw_response[0])
-        return raw_response
+        return cast(List[_DocumentOut], raw_response)
```
ShaneHarvey, line 397: We don't need this cast. Instead let's say this returns Mapping and add a # type ignore. The RawBatch classes will never be compatible with the super class because they yield raw bytes, not decoded documents.

pymongo/errors.py
```diff
@@ -145,9 +160,9 @@ class NotPrimaryError(AutoReconnect):
     .. versionadded:: 3.12
     """
 
-    def __init__(
-        self, message: str = "", errors: Optional[Union[Mapping[str, Any], List]] = None
-    ) -> None:
+    details: _DocumentOut
+
+    def __init__(self, message: str = "", errors: Optional[_DocumentOut] = None) -> None:
```
ShaneHarvey, line 165: Why did we need to change this from Mapping to _DocumentOut?

pymongo/errors.py
```diff
@@ -183,15 +198,15 @@ def __init__(
...
+        details: Optional[_DocumentOut] = None,
         max_wire_version: Optional[int] = None,
     ) -> None:
         error_labels = None
         if details is not None:
             error_labels = details.get("errorLabels")
         super().__init__(_format_detailed_error(error, details), error_labels=error_labels)
         self.__code = code
-        self.__details = details
+        self.__details = details or {}
```
ShaneHarvey, line 209: I'm concerned about this change. Technically it's breaking since details used to default to `None` but now will be `{}`.

pymongo/network.py
```diff
@@ -84,7 +85,7 @@ def command(
     user_fields: Optional[Mapping[str, Any]] = None,
     exhaust_allowed: bool = False,
     write_concern: Optional[WriteConcern] = None,
-) -> Dict[str, Any]:
+) -> _DocumentOut:
```
ShaneHarvey, line 88: Can you make this generic via `codec_options`?

pymongo/network.py
```diff
@@ -267,7 +269,9 @@ def receive_message(
         data = _receive_data_on_socket(sock_info, length - 16, deadline)
 
     try:
-        unpack_reply = _UNPACK_REPLY[op_code]
+        unpack_reply: Callable[[bytes], _OpReply] = cast(
```
ShaneHarvey, line 272: How about let's add these type hints to _UNPACK_REPLY? Then we can remove the cast here.

pymongo/pool.py
```diff
@@ -662,7 +662,7 @@ def __init__(self, sock, pool, address, id):
             # This is a Monitor connection.
             self.cancel_context = _CancellationContext()
         self.opts = pool.opts
-        self.more_to_come = False
+        self.more_to_come: Union[int, bool] = False
```
ShaneHarvey, line 665: why int? IIUC this should be bool

pymongo/response.py
```diff
@@ -35,7 +35,7 @@ def __init__(
         request_id: int,
         duration: Optional[timedelta],
         from_command: bool,
-        docs: List[Mapping[str, Any]],
+        docs: List[_DocumentOut],
```
ShaneHarvey, line 38: Why did we need this change?

test/test_grid_file.py
```diff
@@ -40,9 +35,12 @@
     GridOutCursor,
 )
 from pymongo import MongoClient
+from pymongo.database import Database
 from pymongo.errors import ConfigurationError, ServerSelectionTimeoutError
 from pymongo.message import _CursorAddress
 
+sys.path[0:0] = [""]
```
ShaneHarvey, line 42: this sys line needs to be above the `test` imports. Otherwise you can't run the test like this:
```
python test/test_grid_file.py
```

ShaneHarvey (changes requested):

pymongo/response.py
sleepyStick: At some point, docs was being a `List[_DocumentOut]` and because lists are invariant, I changed it. Would it be better if I change the `List`, to be a `Sequence`?

pymongo/errors.py
sleepyStick: Hm, that's a fair concern. What if I undo the changes in this file and then add type ignores for now?

pymongo/pool.py
ShaneHarvey: Let's make _OpMsg.more_to_come a bool as well.

pymongo/errors.py
ShaneHarvey: Yes that sounds safer.

ShaneHarvey (changes requested):
There's a few conflicts here so I'm going to wait until those are resolved.

pymongo/pool.py
sleepyStick: Sounds good

pymongo/errors.py
sleepyStick: Sounds good, just changed it back

pymongo/errors.py
sleepyStick: If i recall correctly, this had to do with the change below. since i'm undoing that change, this is also Mapping again

ShaneHarvey (approved):

ShaneHarvey:
Looks like small conflict with #1332.

blink1073 (approved):
LGTM!

PR #1315, opened 2023-07-18
PYTHON-3807 add types to mongo_client.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3807


pymongo/collection.py
```diff
@@ -2250,7 +2248,7 @@ def list_indexes(
         self,
         session: Optional[ClientSession] = None,
         comment: Optional[Any] = None,
-    ) -> CommandCursor[MutableMapping[str, Any]]:
+    ) -> CommandCursor[_DocumentType]:
```
ShaneHarvey, line 2251: I believe list_indexes always returns MutableMapping. Could you confirm?

_DocumentType is intended to represent documents in the collection, not index documents, so let's avoid using it here.

pymongo/collection.py
```diff
@@ -886,8 +881,11 @@ def _update(
                 comment=comment,
             )
 
-        return self.__database.client._retryable_write(
-            (write_concern or self.write_concern).acknowledged and not multi, _update, session
+        return cast(
+            Mapping[str, Any],
```
ShaneHarvey, line 885: This cast shouldn't be needed. Is the return type of `_update` inaccurate?

pymongo/collection.py
```diff
@@ -2351,7 +2349,10 @@ def index_information(
         .. versionchanged:: 3.6
            Added ``session`` parameter.
         """
-        cursor = self.list_indexes(session=session, comment=comment)
+        cursor = cast(
+            CommandCursor[MutableMapping[str, Any]],
```
ShaneHarvey, line 2353: We don't need this cast after we fix list_indexes.

ShaneHarvey (changes requested):

bson/codec_options.py
```diff
@@ -245,6 +245,31 @@ class CodecOptions(Tuple, Generic[_DocumentType]):
...
+        #     document_class: Type[_DocumentType],
+        #     tz_aware: bool = ...,
+        #     uuid_representation: Optional[int] = ...,
+        #     unicode_decode_error_handler: Optional[str] = ...,
+        #     tzinfo: Optional[datetime.tzinfo] = ...,
+        #     type_registry: Optional[TypeRegistry] = ...,
+        #     datetime_conversion: Optional[int] = ...,
+        # ) -> "CodecOptions[_DocumentType]":
+        #     ...
+
```
ShaneHarvey, line 272: Should this be removed?

ShaneHarvey (approved):

bson/codec_options.py
sleepyStick: yes! thanks for catching that!

ShaneHarvey (approved):

PR #1317, opened 2023-07-24
PYTHON-3801 gRPC POC phase 1

NoahStapp:
gRPC POC phase 1 work. Partially rip-and-replace, replacing TCP-specific interfaces with gRPC-specific interfaces, but mostly interchangeable between the two protocols in preparation for POC phase 2 work.


ShaneHarvey (changes requested):
Could you flesh this out a bit more by adding a `MongoClient(grpc=True)` flag which then configures the Pool setting? That way we can use this branch to compare TCP vs gRPC performance?

NoahStapp:
> Could you flesh this out a bit more by adding a `MongoClient(grpc=True)` flag which then configures the Pool setting? That way we can use this branch to compare TCP vs gRPC performance?

Done! I added a minimal working example in tests/grpc_example.py that depends on the grpc-enabled serverless proxy running locally.

ShaneHarvey:
Looks like grpcio still does not support pypy: https://github.com/grpc/grpc/issues/4221

NoahStapp:
> Looks like grpcio still does not support pypy: [grpc/grpc#4221](https://github.com/grpc/grpc/issues/4221) That's a bit disappointing. Perhaps we could work on that in the future?
> 
> Edit: looks like the pypy error might be fixed by installing cython 0.29.36: [grpc/grpc#29517 (comment)](https://github.com/grpc/grpc/issues/29517#issuecomment-1631282367)

This fix didn't work for me, even putting cython 0.29.36 in our pyproject.toml `[build-system]` `requires` section. We'll have to investigate further in a future POC phase.

pymongo/aggregation.py
```diff
@@ -29,7 +29,7 @@
     from pymongo.collection import Collection
     from pymongo.command_cursor import CommandCursor
     from pymongo.database import Database
-    from pymongo.pool import SocketInfo
```
ShaneHarvey, line 32: Could you open a new ticket to do this SocketInfo->Connection renaming as it's own PR? It makes this POC difficult to review.

ShaneHarvey:
LGTM!

PR #1327, opened 2023-07-25
PYTHON-694 Test mod_wsgi sub interpreters 

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-694 

Tasks:
- [x] Tests mod_wsgi sub interpreters.
- [x] Tests mod_wsgi embedded mode.
- [x] Test decoding all BSON types.
- [x] Test encoding all BSON types.
- [x] Test that decoded types created from C are valid in Python, ie `assert isinstance(doc["regex"], Regex)`.
- [x] Update Readme to document why multiple URLs are used.
- [x] Fix mypy error:
```
      test/mod_wsgi_test/mod_wsgi_test.py:71: error: Cannot find implementation or
library stub for module named "mod_wsgi"  [import]
        from mod_wsgi import version as mod_wsgi_version
```

References:
- https://emptysqua.re/blog/python-c-extensions-and-mod-wsgi/
- https://modwsgi.readthedocs.io/en/develop/user-guides/checking-your-installation.html?#sub-interpreter-being-used

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -347,7 +347,9 @@ functions:
         script: |
           set -o xtrace
           ${PREPARE_SHELL}
-          PYTHON_BINARY=${PYTHON_BINARY} MOD_WSGI_VERSION=${MOD_WSGI_VERSION} PROJECT_DIRECTORY=${PROJECT_DIRECTORY} bash ${PROJECT_DIRECTORY}/.evergreen/run-mod-wsgi-tests.sh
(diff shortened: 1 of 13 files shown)


ShaneHarvey:
Somewhat confusingly, these tests pass even with DatetimeMS which should be expected to fail due to the bug described in https://jira.mongodb.org/browse/PYTHON-3847. I'm a bit perplexed here but we can merge this work regardless. Perhaps Python or mod_wsgi was changed to make Python classes work between sub interpreters?

blink1073 (approved):
Nice work!

PR #1328, opened 2023-07-26
PYTHON-3808 add types to monitor.py

sleepyStick:


pymongo/srv_resolver.py
```diff
@@ -56,7 +56,7 @@ def _resolve(*args: Any, **kwargs: Any) -> resolver.Answer:
 class _SrvResolver:
     def __init__(
         self,
-        fqdn: str,
+        fqdn: Optional[str],
```
ShaneHarvey, line 59: Is fqdn really optional?

sleepyStick: I thought it was based on line 335 of monitor.py which passes `SrvMonitor._fqdn` into _SrvResolver as `fqdn` and `SrvMonitor._fqdn` is defined by taking `TopologySettings.fqdn` which is an `Optional[str]`

ShaneHarvey: It's optional on TopologySettings but it's required for SrvMonitor and _SrvResolver.

sleepyStick: I see, just fixed it!

pymongo/monitor.py
```diff
@@ -29,24 +31,31 @@
 from pymongo.server_description import ServerDescription
 from pymongo.srv_resolver import _SrvResolver
 
+if TYPE_CHECKING:
+    from OpenSSL.SSL import Connection
```
ShaneHarvey, line 35: This should be a pymongo.pool.Connection.

ShaneHarvey (approved):
One last comment on Connection but otherwise LGTM!

pymongo/monitor.py
sleepyStick: makes sense, fixed!

ShaneHarvey (approved):

PR #1332, opened 2023-07-28
PYTHON-3809 add types to monitoring.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3809

### DIFF
--- a/pymongo/client_options.py
+++ b/pymongo/client_options.py
@@ -152,7 +152,7 @@ def _parse_pool_options(
     connect_timeout = options.get("connecttimeoutms", common.CONNECT_TIMEOUT)
     socket_timeout = options.get("sockettimeoutms")
     wait_queue_timeout = options.get("waitqueuetimeoutms", common.WAIT_QUEUE_TIMEOUT)
-    event_listeners = cast(Sequence[_EventListener], options.get("event_listeners"))
+    event_listeners = cast(Optional[Sequence[_EventListener]], options.get("event_listeners"))
     appname = options.get("appname")
     driver = options.get("driver")
     server_api = options.get("server_api")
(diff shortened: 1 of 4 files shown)


pymongo/server.py
```diff
@@ -191,9 +192,9 @@ def run_operation(
             # Must publish in find / getMore / explain command response
             # format.
             if use_cmd:
-                res = docs[0]
+                res: _DocumentOut = docs[0]  # type: ignore[assignment]
             elif operation.name == "explain":
-                res = docs[0] if docs else {}
+                res = docs[0] if docs else {}  # type: ignore[assignment]
```
NoahStapp, line 197: Should this also be `res: _DocumentOut`?

NoahStapp (changes requested):

pymongo/server.py
sleepyStick: I believe that's implied here because of line 195

NoahStapp: Got it, thanks!

NoahStapp (approved):

pymongo/client_options.py
```diff
@@ -152,7 +152,7 @@ def _parse_pool_options(
     connect_timeout = options.get("connecttimeoutms", common.CONNECT_TIMEOUT)
     socket_timeout = options.get("sockettimeoutms")
     wait_queue_timeout = options.get("waitqueuetimeoutms", common.WAIT_QUEUE_TIMEOUT)
-    event_listeners = options.get("event_listeners")
+    event_listeners = cast(Sequence[_EventListener], options.get("event_listeners"))
```
ShaneHarvey, line 155: I believe this is `Optional[Sequence[_EventListener]]`.

ShaneHarvey (changes requested):

pymongo/client_options.py
sleepyStick: fixed, thanks for catching that!

ShaneHarvey (approved):
LGTM

PR #1334, opened 2023-08-01
PYTHON-3878 Prototype using Greenlet to support Async PyMongo without…

NoahStapp:


test/test_auth.py
```diff
@@ -16,9 +16,10 @@
 
 import os
 import sys
-import threading
 from urllib.parse import quote_plus
 
+from pymongo.io import threading
```
ShaneHarvey, line 21: We need to change all the test imports back to use stdlib threading. We can't ask all users to change what threading module they use since that would be a huge breaking change.

ShaneHarvey (changes requested):

test/test_auth.py
blink1073: We talked in the catch up meeting today about making our asyncio an explicit opt-in, similar to `gevent.monkey_patch()`.

ShaneHarvey: Even if we make it opt in, apps still need to be able to use the sync client from standard threads.

blink1073: Right, with the patch in place, `import threading` will be the patched version.

ShaneHarvey: I think that approach is a non-starter. We can't expect users to patch the stndlib just to use async pymongo.

blink1073: There isn't another way to ensure that the versions of threading, time, etc. that they import will work short of us telling them to import from us, similar to using [gevent](http://www.gevent.org/intro.html#monkey-patching).

If that is a non-starter then we're back to the drawing board.

ShaneHarvey: I don't understand. What happens when a stdlib thread runs a pymongo operation that internally uses greenletio? Are you saying that doesn't work?

blink1073: Correct, I verified that it blocks all other execution.

blink1073: Actually wait, what I actually did was use `time.time`.  Let me dig in again.

ShaneHarvey: Can you expand on "blocks all other execution"?

ShaneHarvey: I changed test_threading.py like this:
```
-from pymongo.io import threading
+import threading
```

And the tests pass and run in the same amount of time.

blink1073: If I change to stdlib import:

```python
  File "/Library/Frameworks/Python.Framework/Versions/3.10/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
starting sync!
waiting: 1691100697
    self.run()
  File "/Library/Frameworks/Python.Framework/Versions/3.10/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/steve.silvester/workspace/mongo-python-driver/green_test.py", line 38, in async_main_entry
    await_(async_main())
  File "/Users/steve.silvester/workspace/.venvs/mongo-python-driver/lib/python3.10/site-packages/greenletio/core.py", line 148, in await_
    parent = greenlet.getcurrent().parent or bridge.start()
  File "/Users/steve.silvester/workspace/.venvs/mongo-python-driver/lib/python3.10/site-packages/greenletio/core.py", line 39, in start
    if asyncio.get_event_loop().is_running():
  File "/Library/Frameworks/Python.Framework/Versions/3.10/lib/python3.10/asyncio/events.py", line 671, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'Thread-2 (async_main_entry)'.
/Library/Frameworks/Python.Framework/Versions/3.10/lib/python3.10/threading.py:1018: RuntimeWarning: coroutine 'test_database.<locals>.async_main' was never awaited
  self._invoke_excepthook(self)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
find:    1691100697
Exception in thread Thread-1 (main):
  File "/Library/Frameworks/Python.Framework/Versions/3.10/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/Library/Frameworks/Python.Framework/Versions/3.10/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/steve.silvester/workspace/mongo-python-driver/green_test.py", line 66, in main
    time.sleep(1)
  File "/Users/steve.silvester/workspace/.venvs/mongo-python-driver/lib/python3.10/site-packages/greenletio/green/time.py", line 10, in sleep
    await_(asyncio.sleep(seconds))
  File "/Users/steve.silvester/workspace/.venvs/mongo-python-driver/lib/python3.10/site-packages/greenletio/core.py", line 152, in await_
    return parent.switch(coro_or_fn)
greenlet.error: cannot switch to a different thread
/Library/Frameworks/Python.Framework/Versions/3.10/lib/python3.10/threading.py:1018: RuntimeWarning: coroutine 'sleep' was never awaited
  self._invoke_excepthook(self)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

We *might* be able to adapt `greenletio` to handle this case.

NoahStapp: I ran into the above error when I was using only `greenletio.green.socket` and stdlib for the rest, are you using all `greenletio` besides `thread`?

blink1073: Yeah, in your case you were using sync things in a thread, I verified that works.  It's when you want to do async things in a thread that things break down.  Which, you wouldn't really do anyway, so maybe we're okay?

blink1073: > Can you expand on "blocks all other execution"?

I was trying to mix `time.sleep(1)` with greenletio and it would not run the tasks concurrently.

NoahStapp: If you use sync pymongo in a thread while `greenletio` is present, things break?

blink1073: You can use sync pymongo in a thread with `greentletio` and stdlib threading, or async pymongo in a thread with `greenletio` threading.

ShaneHarvey: I believe those errors are expected. You can't call `await_` from a thread without setting up a loop just like you can't await an `async def` method outside of a running eventloop.

ShaneHarvey: The code appears to work fine with standard time+threading with this addition:
```python
    def async_main_entry():
        asyncio.set_event_loop(asyncio.new_event_loop())
        await_(async_main())
```

blink1073: Oh cool, yeah, that makes sense.

pymongo/asyncio/database.py
```diff
@@ -0,0 +1,53 @@
...
+        value: Any = 1,
+        check: bool = True,
+        allowable_errors: Optional[Sequence[Union[str, int]]] = None,
+        read_preference: Optional[_ServerMode] = None,
+        codec_options: Optional[bson.codec_options.CodecOptions[_CodecDocumentType]] = None,
+        session: Optional[ClientSession] = None,
+        comment: Optional[Any] = None,
+        **kwargs: Any,
+    ) -> Union[None, None, Union[Dict[str, Any], _CodecDocumentType]]:
+        return await async_(self.database.command)(
```
ShaneHarvey, line 43: I believe this can actually be simplified to “return self.database.command(“.

ShaneHarvey (commented):
I have a few more concerns with this approach. In particular how will we deal with cases where we need to use a 3rd party library to perform I/O? dnspython and requests are two examples right now and we could add more of these in the future. I understand that we can wrap the imports of these libraries in greentletio's patch_blocking:
```python
with patch_blocking():
    import requests
    import dns
```

But does that approach still work if the user has already imported requests/dnspython *before* importing pymongo? ie:
```python
# App imports pymongo's deps before pymongo itself so module is already loaded:
import requests
import dns
# Does patch_blocking still work in this case?
import pymongo
```

NoahStapp:
> I have a few more concerns with this approach. In particular how will we deal with cases where we need to use a 3rd party library to perform I/O? dnspython and requests are two examples right now and we could add more of these in the future. I understand that we can wrap the imports of these libraries in greentletio's patch_blocking:
> 
> ```python
> with patch_blocking():
>     import requests
>     import dns
> ```
> 
> But does that approach still work if the user has already imported requests/dnspython _before_ importing pymongo? ie:
> 
> ```python
> # App imports pymongo's deps before pymongo itself so module is already loaded:
> import requests
> import dns
> # Does patch_blocking still work in this case?
> import pymongo
> ```

Greenletio's `patch_blocking` directly accesses `sys.modules`: https://github.com/miguelgrinberg/greenletio/blob/main/src/greenletio/patcher.py. According to https://stackoverflow.com/questions/73060129/how-are-changes-to-sys-modules-propagated, I would expect this patching to still work, but only in the scope of `pymongo`.

ShaneHarvey:
Great, could we add a test for it to confirm the behavior?

NoahStapp:
> Great, could we add a test for it to confirm the behavior?

Done, see the file `green_import_test.py`. The thread does end with the error

```
RuntimeError: There is no current event loop in thread 'Thread-4 (test_database)'.
/Users/nstapp/.pyenv/versions/3.11.3/lib/python3.11/threading.py:1040: RuntimeWarning: coroutine 'Thread.start.<locals>.bootstrap' was never awaited
  self._invoke_excepthook(self)
```

which suggests there is more investigation to be done.

blink1073:
I think what we'll need is a decorator added to our async methods that does something like:

```python
try:
   asyncio.get_event_loop()
except RuntimeError:
   asyncio.set_event_loop(asyncio.new_event_loop())
```

ShaneHarvey:
I don't think we want to be in the business of setting up the event loop. The user should need to set up the loop themselves if they want to call the async apis, just like they would need if they used any other asyncio apis.

NoahStapp:
The latest commit's `green_import_test.py` tests the how import ordering interacts with `patch_blocking`. The result is the following:

```
import time
import green_test
...

time.sleep is builtin
Source for green_test.time.sleep: /Users/nstapp/.pyenv/versions/3.11.3/lib/python3.11/site-packages/greenletio/green/time.py
```
If `time.sleep` is inspected inside `green_test` itself, the expected result of `time.sleep` being patched occurs. 

This suggests that using `patch_blocking` in pymongo will not affect any duplicated imports in client code.

blink1073:
What about a delayed import?

```python
def inner():
    import time
	...
```

NoahStapp:
> What about a delayed import?
> 
> ```python
> def inner():
>     import time
> 	...
> ```

Same result, `time.sleep` is builtin, `green_test.time.sleep` is patched.

ShaneHarvey:
@NoahStapp could you create a patch build on [mongo-python-driver-perf](https://spruce.mongodb.com/commits/mongo-python-driver-perf) that runs the benchmarks with greenletio installed to see if there's any perf difference for sync code?

ShaneHarvey:
I believe the import issue is more subtle when we consider a third party library like dns. If pymongo imports dns in a patch_blocking call then we'll have this behavior:
```python 
import dns as dns1  # dns imported for the first time.

# Simulate import pymongo
import greenletio
with greenletio.patch_blocking():
    import dns as dns2  

assert dns1 is dns2  # dns2 will be the same as dns1
```

The only way to deal with this problem is avoid patch_blocking and explicitly use the asyncio dns apis. Alternatively we _could_ reimport dns ourselves (although I'm not a fan since mucking around python internals is frowned upon because it can have unintended side effects):
```python
import dns

# Always reimport dns
import sys
old_dns = sys.modules.pop('dns', None)
try:
    with greenletio.patch_blocking():
        import dns as dns3
finally:
    if old_dns is not None:
        sys.modules['dns'] = old_dns

assert dns3 is not dns1  # dns3 will be newly imported
```

NoahStapp:
> I believe the import issue is more subtle when we consider a third party library like dns. If pymongo imports dns in a patch_blocking call then we'll have this behavior:
> 
> ```python
> import dns as dns1  # dns imported for the first time.
> 
> # Simulate import pymongo
> import greenletio
> with greenletio.patch_blocking():
>     import dns as dns2  
> 
> assert dns1 is dns2  # Oops! dns2 will be the same as dns1
> ```
> 
> The only way to deal with this problem is avoid patch_blocking and explicitly use the asyncio dns apis. Alternatively we _could_ reimport dns ourselves (although I'm not a fan since mucking around python internals is frowned upon because it can have unintended side effects):
> 
> ```python
> import dns
> 
> # Always reimport dns
> import sys
> old_dns = sys.modules.pop('dns', None)
> try:
>     with greenletio.patch_blocking():
>         import dns as dns3
> finally:
>     if old_dns is not None:
>         sys.modules['dns'] = old_dns
> 
> assert dns3 is not dns1  # dns3 will be newly imported
> ```

Isn't this expected behavior for `patch_blocking` with third-party libraries? It's still the normal `dns` library, just with its blocking dependencies replaced with `greenletio` versions. Can we verify if those dependencies are patched correctly?

ShaneHarvey:
The problem is that patch_blocking does not work at all if the app has already imported dns (the first example above).

NoahStapp:
> The problem is that patch_blocking does not work at all if the app has already imported dns (the first example above).

My question is does a successful use of `patch_blocking` make dns2 != dns1? I'm not familiar with how equality of module objects works

PR #1335, opened 2023-08-01
PYTHON-3893 Add BSON Binary Data subtype Sensitive

NoahStapp:


bson/binary.py
```diff
@@ -183,6 +183,13 @@ class UuidRepresentation:
 .. versionadded:: 4.0
 """
 
+SENSITIVE_SUBTYPE = 8
+"""BSON binary subtype for sensitive data.
+
+.. versionadded:: 7.0
```
blink1073, line 189: This should be 4.5 (the pymongo version).

blink1073 (changes requested):

bson/binary.py
```diff
@@ -183,6 +183,13 @@ class UuidRepresentation:
 .. versionadded:: 4.0
 """
 
+SENSITIVE_SUBTYPE = 8
+"""BSON binary subtype for sensitive data.
```
ShaneHarvey, line 187: Are users ever expected to use this?

blink1073: The spec doesn't mention it, but other drivers expose these constants, e.g. https://github.com/mongodb/mongo-go-driver/blob/4acd55139b846e6bfdfe316770a5b1b646d5ff96/bson/types.go#L47.

ShaneHarvey (approved):

blink1073 (approved):

PR #1339, opened 2023-08-03
PYTHON-3866 add types to topology_description.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3866

### DIFF
--- a/pymongo/topology_description.py
+++ b/pymongo/topology_description.py
@@ -25,6 +25,7 @@ from typing import (
     NamedTuple,
     Optional,
     Tuple,
+    Union,
     cast,
 )
 
@@ -555,8 +556,12 @@ def _update_rs_from_primary(
         return _check_has_primary(sds), replica_set_name, max_set_version, max_election_id
 
     if server_description.max_wire_version is None or server_description.max_wire_version < 17:
-        new_election_tuple = server_description.set_version, server_description.election_id
-        max_election_tuple = max_set_version, max_election_id
+        new_election_tuple: Union[
+            Tuple[Optional[int], Optional[ObjectId]], Tuple[Optional[ObjectId], Optional[int]]
+        ] = (server_description.set_version, server_description.election_id)
+        max_election_tuple: Union[
+            Tuple[Optional[int], Optional[ObjectId]], Tuple[Optional[ObjectId], Optional[int]]
+        ] = (max_set_version, max_election_id)
         if None not in new_election_tuple:
             if None not in max_election_tuple and new_election_tuple < max_election_tuple:
                 # Stale primary, set to type Unknown.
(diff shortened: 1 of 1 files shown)


NoahStapp (approved):
LGTM!

pymongo/topology_description.py
```diff
@@ -536,8 +556,12 @@ def _update_rs_from_primary(
...
 
     if server_description.max_wire_version is None or server_description.max_wire_version < 17:
-        new_election_tuple = server_description.set_version, server_description.election_id
-        max_election_tuple = max_set_version, max_election_id
+        new_election_tuple: Union[
+            Tuple[Optional[int], Optional[ObjectId]], Tuple[Optional[ObjectId], Optional[int]]
+        ] = (server_description.set_version, server_description.election_id)
+        max_election_tuple: Union[
+            Tuple[Optional[int], Optional[ObjectId]], Tuple[Optional[ObjectId], Optional[int]]
+        ] = (max_set_version, max_election_id)
```
ShaneHarvey, line 564: Why do these need explicit hints? It makes this code significantly less readable.

pymongo/topology_description.py
```diff
@@ -253,14 +265,16 @@ def heartbeat_frequency(self) -> int:
...
             return []
         # Round trip time in seconds.
-        fastest = min(s.round_trip_time for s in selection.server_descriptions)
+        fastest = min(cast(float, s.round_trip_time) for s in selection.server_descriptions)
         threshold = self._topology_settings.local_threshold_ms / 1000.0
         return [
-            s for s in selection.server_descriptions if (s.round_trip_time - fastest) <= threshold
+            s
+            for s in selection.server_descriptions
+            if (cast(float, s.round_trip_time) - fastest) <= threshold
```
ShaneHarvey, line 277: Hmm I wonder if we could use `float("inf")` as the round trip time for unknown servers rather than None. Could you open a ticket for that idea?

pymongo/topology_description.py
sleepyStick: Um the other alternative that I see is to just type it as `Tuple` and not specify the contents. By default, mypy will assume `new_election_tuple` to be a tuple of `Optional[int]`, `Optional[OblectId]`. however, it gets reassigned to be a tuple of `Optional[ObjectId]`, `Optional[int]` which mypy then complains about. 

I've changed it to now just be `Tuple`

pymongo/topology_description.py
sleepyStick: yeah!

pymongo/topology_description.py
ShaneHarvey: Oh I see the problem is that mypy considers the two vars to be overlapping even though they are never used in the same scope. Yeah I think bare Tuple is fine.

ShaneHarvey (approved):

PR #1341, opened 2023-08-03
PYTHON-3814 add types to pyopenssl_context.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3814

### DIFF
--- a/pymongo/pyopenssl_context.py
+++ b/pymongo/pyopenssl_context.py
@@ -236,7 +236,9 @@ class SSLContext:
             # and OpenSSL's SSL_CTX_set_verify let's you pass NULL
             # for the callback option. It's weird that PyOpenSSL requires
             # this.
-            # This is optional in pyopenssl >= 20 and can be removed once minimum supported version is bumped
+            # This is optional in pyopenssl >= 20 and can be removed once minimum
+            # supported version is bumped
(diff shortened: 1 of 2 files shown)


pymongo/pyopenssl_context.py
```diff
@@ -198,66 +209,77 @@ def __init__(self, protocol):
...
+            errnum: int,
+            errdepth: int,
+            retcode: int,
+        ) -> bool:
             # It seems we don't need to do anything here. Twisted doesn't,
             # and OpenSSL's SSL_CTX_set_verify let's you pass NULL
             # for the callback option. It's weird that PyOpenSSL requires
             # this.
-            return retcode
+            return bool(retcode)
```
ShaneHarvey, line 239: Can you add a comment that this is now optional in pyopenssl >=20 and can be removed when we bump our min supported version? 

> Make verification callback optional in Context.set_verify. If omitted, OpenSSL’s default verification is used. [#933](https://github.com/pyca/pyopenssl/pull/933)

See: https://www.pyopenssl.org/en/latest/changelog.html#id47

sleepyStick: yes, done

ShaneHarvey (approved):

PR #1344, opened 2023-08-04
PYTHON-3739 Refactor retryable reads and writes logic to avoid duplication

Jibola:
first draft commit; consolidated _retryable_(read|write) to call _retry_internal

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -1410,6 +1410,67 @@ class MongoClient(common.BaseObject, Generic[_DocumentType]):
         )
         return self._retry_internal(retryable, func, session, bulk)
 
+    def _generate_retriable(
+        self,
+        retryable: bool,
+        func: Callable[[Optional[ClientSession], Connection, bool], T],
+        session: Optional[ClientSession],
+        writeable: bool,
+        address: Optional[_Address],
+        read_pref: Optional[_ServerMode],
+        last_error: Optional[Exception],
+        is_retrying: Callable[[], bool]
+    ) -> T:
+        """Generates a callable with func()'s return value on success
+
+        Identifying func() as not "writeable" will require a read_pref
+        """
+        local_read_pref = read_pref
+
+        def _write_wrap():
+            local_retryable = retryable
+            try:
+                max_wire_version = 0
+                server = self._select_server(writable_server_selector, session)
+                supports_session = (
+                    session is not None and server.description.retryable_writes_supported
+                )
+                with self._checkout(server, session) as conn:
+                    max_wire_version = conn.max_wire_version
+                    if local_retryable and not supports_session:
+                        if is_retrying():
+                            # A retry is not possible because this server does
+                            # not support sessions raise the last error.
+                            assert last_error is not None
+                            raise last_error
+                        local_retryable = False
+                    return func(session, conn, local_retryable)
+            except PyMongoError as exc:
+                if not local_retryable:
+                    raise
+                assert session
+                # Add the RetryableWriteError label, if applicable.
+                _add_retryable_write_error(exc, max_wire_version)
+                raise
+
+        def _read_wrap():
+            if local_read_pref is None:
+                raise PyMongoError("read_pref is required on retriable read connection attempt")
+
+            server = self._select_server(local_read_pref, session, address=address)
+            with self._conn_from_server(local_read_pref, server, session) as (
+                conn,
+                read_pref,
+            ):
+                if is_retrying() and not retryable:
+                    # A retry is not possible because this server does
+                    # not support retryable reads, raise the last error.
+                    assert last_error is not None
+                    raise last_error
+                return func(session, server, conn, read_pref)
+
+        return _write_wrap if writeable else _read_wrap
+
     @_csot.apply
     def _retry_internal(
         self,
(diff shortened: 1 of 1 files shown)


pymongo/mongo_client.py
```diff
@@ -1410,30 +1410,104 @@ def _retry_with_session(
...
+
+        def _write_wrap():
+            local_retryable = retryable
+            try:
+                max_wire_version = 0
+                server = self._select_server(writable_server_selector, session)
+                supports_session = (
+                    session is not None and server.description.retryable_writes_supported
+                )
+                with self._checkout(server, session) as conn:
```
ShaneHarvey, line 1438: Can we use _conn_from_server here?

pymongo/mongo_client.py
```diff
@@ -1506,54 +1580,21 @@ def _retryable_read(
...
-                last_error = exc
-            except OperationFailure as exc:
-                if not retryable or (retrying and not multiple_retries):
-                    raise
-                if exc.code not in helpers._RETRYABLE_ERROR_CODES:
-                    raise
-                retrying = True
-                last_error = exc
+        return self._retry_internal(
+            self,
```
ShaneHarvey, line 1589: Extra self here.

pymongo/mongo_client.py
```diff
@@ -1467,12 +1528,25 @@ def is_retrying() -> bool:
...
+                retrying = True
+                last_error = exc
+            except OperationFailure as exc:
+                # This exception is endemic in read-sponsored issues``
+                if not retryable or (retrying and not multiple_retries):
+                    raise
+                if exc.code not in helpers._RETRYABLE_ERROR_CODES:
+                    raise
+                retrying = True
+                last_error = exc
```
ShaneHarvey, line 1544: This case is missing the "RetryableWriteError" logic from the PyMongoError case.

ShaneHarvey (changes requested):

Jibola (commented):
Thanks for the first review @ShaneHarvey ! 

Also changed the abstraction to be more object-oriented rather than functional and I think that helped quite a bit with readability.

pymongo/mongo_client.py
```diff
@@ -2304,6 +2205,159 @@ def __exit__(
...
+                assert self._session
+                if exc.has_error_label("RetryableWriteError"):
+                    self._session._unpin()
+                elif self._is_retrying() and not self._multiple_retries:
+                    if exc.has_error_label("NoWritesPerformed") and self._last_error:
+                        raise self._last_error from exc
+                    else:
+                        raise
+                if self._bulk:
+                    self._bulk.retrying = True
```
ShaneHarvey, line 2291: self._bulk.retrying needs to be set when retrying any exception not just PyMongoError.

ShaneHarvey (changes requested):

ShaneHarvey (changes requested):
Since we're changing _retry_internal we need to update ClientSession because it also calls this method:
```
pymongo/client_session.py:847: error: Argument 1 to "_retry_internal" of
"MongoClient" has incompatible type "bool"; expected
"Callable[[Optional[ClientSession], Connection, bool], Dict[str, Any]]" 
[arg-type]
            return self._client._retry_internal(True, func, self, None)
                                                ^~~~
```

pymongo/mongo_client.py
```diff
@@ -2304,6 +2205,159 @@ def __exit__(
...
+        self._server: Server = None
+
+    def run(self) -> T:
+        """Runs the supplied func() and attempts a retry
+
+        :raises self._last_error: Exception raised
+        :return: Result of the func() call
+        :rtype: T
+        """
+        if self._in_transaction() and self._is_write:
```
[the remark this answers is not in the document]
Jibola: Nice catch there!  I'll also go ahead and reattach that comment block

pymongo/mongo_client.py
```diff
@@ -2304,6 +2204,166 @@ def __exit__(
...
+                        if self._is_not_retry_eligible() or (
+                            exc_code and exc_code not in helpers._RETRYABLE_ERROR_CODES
+                        ):
+                            raise
+                        self._retrying = True
+                        self._last_error = exc
+                    else:
+                        raise
+
+                # Assumed write operation henceforth
```
blink1073, line 2284: It looks like we can get to this point if we get to the `self._retrying = True` case for a read above.

blink1073 (changes requested):

pymongo/mongo_client.py
Jibola: Ah, you're right. Looks like I'll just add an explicit if conditional `if not self._is_write` above line 2285 to avoid this.

pymongo/mongo_client.py
```diff
@@ -2307,6 +2259,189 @@ def __exit__(
...
+                    if self._bulk:
+                        self._bulk.retrying = True
+                    else:
+                        self._retrying = True
+                    if not exc.has_error_label("NoWritesPerformed"):
+                        self._last_error = exc
+                    if self._last_error is None:
+                        self._last_error = exc
+
+    def _is_not_retry_eligible(self) -> bool:
```
blink1073, line 2360: This name confused me, perhaps `_is_not_eligible_for_retry`?

pymongo/mongo_client.py
```diff
@@ -1507,65 +1472,52 @@ def _retryable_read(
...
     def _retryable_write(
         self,
         retryable: bool,
-        func: Callable[[Optional[ClientSession], Connection, bool], T],
+        func: _WriteCall[T],
         session: Optional[ClientSession],
+        bulk: Optional[_Bulk] = None,
     ) -> T:
-        """Internal retryable write helper."""
+        """Execute an operation with at most one consecutive retries on write
```
blink1073, line 1506: Won't there be more than one attempt if `csot` is enabled?

blink1073 (changes requested):

pymongo/mongo_client.py
Jibola: Admittedly, I did not assume any retries occurred more than once. I grifted the docstring from [`_retry_with_session`](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/mongo_client.py#L1403-L1408) which led me to assume retry was only a "one and done" scenario. 

Thanks for raising the question, though. I ended up reviewing the `_retry_internal` call--you're definitely right. If `csot` is enabled, we check for the `_csot.get_timeout()` and that becomes the determinant for retries.

I'll go ahead and change _all_ the docstrings to reflect that.

PR #1346, opened 2023-08-08
PYTHON-3867 add types to topology.py

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3867

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -1907,7 +1907,7 @@ class MongoClient(common.BaseObject, Generic[_DocumentType]):
         session_time = session.cluster_time if session else None
         if topology_time and session_time:
             if topology_time["clusterTime"] > session_time["clusterTime"]:
-                cluster_time: ClusterTime = topology_time
+                cluster_time: Optional[ClusterTime] = topology_time
             else:
                 cluster_time = session_time
         else:
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -1096,7 +1096,7 @@ class MongoClient(common.BaseObject, Generic[_DocumentType]):
         return self._server_property("address")
 
     @property
-    def primary(self) -> Optional[_Address]:
+    def primary(self) -> Optional[Tuple[str, int]]:
         """The (host, port) of the current primary of the replica set.
 
         Returns ``None`` if this client is not connected to a replica set,
@@ -1106,7 +1106,7 @@ class MongoClient(common.BaseObject, Generic[_DocumentType]):
         .. versionadded:: 3.0
            MongoClient gained this property in version 3.0.
         """
-        return self._topology.get_primary()
+        return self._topology.get_primary()  # type: ignore[return-value]
 
     @property
     def secondaries(self) -> Set[_Address]:
@@ -2271,9 +2271,8 @@ class _MongoClientErrorHandler:
     def handle(
         self, exc_type: Optional[Type[BaseException]], exc_val: Optional[BaseException]
     ) -> None:
-        if self.handled or exc_type is None:
+        if self.handled or exc_val is None:
             return
-        assert exc_val is not None
         self.handled = True
         if self.session:
             if isinstance(exc_val, ConnectionFailure):
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -98,6 +98,7 @@ from pymongo.settings import TopologySettings
 from pymongo.topology import Topology, _ErrorContext
 from pymongo.topology_description import TOPOLOGY_TYPE, TopologyDescription
 from pymongo.typings import (
+    ClusterTime,
     _Address,
     _CollationIn,
     _DocumentType,
@@ -1906,7 +1907,7 @@ class MongoClient(common.BaseObject, Generic[_DocumentType]):
         session_time = session.cluster_time if session else None
         if topology_time and session_time:
             if topology_time["clusterTime"] > session_time["clusterTime"]:
-                cluster_time: Optional[Mapping[str, Any]] = topology_time
+                cluster_time: ClusterTime = topology_time
             else:
                 cluster_time = session_time
         else:
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -2268,9 +2268,12 @@ class _MongoClientErrorHandler:
         self.service_id = conn.service_id
         self.completed_handshake = completed_handshake
 
-    def handle(self, exc_type: Optional[Type[Exception]], exc_val: Optional[Exception]) -> None:
+    def handle(
+        self, exc_type: Optional[Type[BaseException]], exc_val: Optional[BaseException]
+    ) -> None:
         if self.handled or exc_type is None:
             return
+        assert exc_val is not None
         self.handled = True
         if self.session:
             if isinstance(exc_val, ConnectionFailure):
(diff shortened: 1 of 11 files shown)


pymongo/mongo_client.py
```diff
@@ -2273,6 +2273,7 @@ def handle(
     ) -> None:
         if self.handled or exc_type is None:
             return
+        assert exc_val is not None
```
ShaneHarvey, line 2276: I suggest switching the order of the conditions in the if statement instead of adding an assert.

test/test_read_preferences.py
```diff
@@ -265,7 +265,7 @@ def test_nearest(self):
 
         not_used = data_members.difference(used)
         latencies = ", ".join(
-            "%s: %dms" % (server.description.address, server.description.round_trip_time)
+            "%s: %dms" % (server.description.address, server.description.round_trip_time)  # type: ignore[str-format]
```
ShaneHarvey, line 268: Is it possible to fix the format string to remove the type ignore?

pymongo/topology.py
```diff
@@ -128,15 +145,17 @@ def __init__(self, topology_settings):
         self._opened = False
         self._closed = False
         self._lock = _create_lock()
+        assert self._settings.condition_class is not None
```
ShaneHarvey, line 148: condition_class should be changed to not optional.

pymongo/topology.py
```diff
@@ -128,15 +145,17 @@ def __init__(self, topology_settings):
...
         self._condition = self._settings.condition_class(self._lock)
-        self._servers = {}
-        self._pid = None
-        self._max_cluster_time = None
+        self._servers: Dict[_Address, Server] = {}
+        self._pid: Optional[int] = None
+        self._max_cluster_time: Optional[Mapping[str, Any]] = None
```
ShaneHarvey, line 152: Can we add a type alias for cluster time called ClusterTime?

pymongo/topology.py
```diff
@@ -128,15 +145,17 @@ def __init__(self, topology_settings):
...
         self._session_pool = _ServerSessionPool()
 
         if self._publish_server or self._publish_tp:
+            weak: weakref.ReferenceType[Optional[queue.Queue]]
```
ShaneHarvey, line 156: Should be `ReferenceType[queue.Queue]` (ditch the optional).

pymongo/topology.py
```diff
@@ -430,21 +475,22 @@ def _get_replica_set_members(self, selector):
             ):
                 return set()
 
-            return {sd.address for sd in selector(self._new_selection())}
+            # mypy doesn't recognize classes that implement a __getitem__ as something that is iterable
+            return {sd.address for sd in selector(self._new_selection())}  # type: ignore[attr-defined]
```
ShaneHarvey, line 479: Does calling iter() explicitly fix this: `iter(selector(...))`?

ShaneHarvey: https://github.com/python/mypy/issues/2220

pymongo/mongo_client.py
```diff
@@ -1096,7 +1096,7 @@ def address(self) -> Optional[Tuple[str, int]]:
         return self._server_property("address")
 
     @property
-    def primary(self) -> Optional[Tuple[str, int]]:
+    def primary(self) -> Optional[_Address]:
```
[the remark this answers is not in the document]
ShaneHarvey: If you want to ignore this rabbit hole I suggest undoing the _Address changes in this PR and opening a new jira ticket.

ShaneHarvey (changes requested):

pymongo/mongo_client.py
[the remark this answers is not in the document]
sleepyStick: hm I see, I open a new jira ticket for this then.

pymongo/mongo_client.py
sleepyStick: sounds good

pymongo/topology.py
sleepyStick: done!

pymongo/topology.py
sleepyStick: done

pymongo/topology.py
sleepyStick: I see now, fixed!

pymongo/topology.py
sleepyStick: yes! good idea!

test/test_read_preferences.py
sleepyStick: I believe the error is that round_trip_time is an `Optional[float]`, my first instinct was to add an `assert server.description.round_trip_time is not None` statement but that doesn't work well with the list comprehension...alternatively, I could cast it? What are your thoughts?

ShaneHarvey: Can we use `%s` instead of `%d`?

pymongo/mongo_client.py
```diff
@@ -1906,7 +1906,7 @@ def _send_cluster_time(
         session_time = session.cluster_time if session else None
         if topology_time and session_time:
             if topology_time["clusterTime"] > session_time["clusterTime"]:
-                cluster_time = topology_time
+                cluster_time: Optional[Mapping[str, Any]] = topology_time
```
ShaneHarvey, line 1909: Let's use ClusterTime here too.

pymongo/topology.py
```diff
@@ -430,21 +477,21 @@ def _get_replica_set_members(self, selector):
...
         """Return set of secondary addresses."""
         return self._get_replica_set_members(secondary_server_selector)
 
-    def get_arbiters(self):
+    def get_arbiters(self) -> Set[_Address]:
         """Return set of arbiter addresses."""
         return self._get_replica_set_members(arbiter_server_selector)
 
-    def max_cluster_time(self):
+    def max_cluster_time(self) -> Optional[Mapping[str, Any]]:
```
ShaneHarvey, line 490: Let's use ClusterTime here and anywhere else we use cluster_time or $clusterTime.

pymongo/topology.py
```diff
@@ -57,8 +70,15 @@
...
+if TYPE_CHECKING:
+    from bson import ObjectId
+    from pymongo.settings import TopologySettings
+    from pymongo.typings import _Address
+
+ClusterTime = Optional[Mapping[str, Any]]
 
-def process_events_queue(queue_ref):
+
+def process_events_queue(queue_ref: weakref.ReferenceType[Optional[queue.Queue]]) -> bool:
```
ShaneHarvey, line 81: Ditch the optional here as well.

ShaneHarvey (changes requested):

test/test_read_preferences.py
sleepyStick: yes! good idea!

pymongo/topology.py
sleepyStick: yes! done!

pymongo/topology.py
sleepyStick: right, sorry for missing it, thanks for catching that!

pymongo/typings.py
```diff
@@ -34,6 +34,7 @@
 _Address = Tuple[str, Optional[int]]
 _CollationIn = Union[Mapping[str, Any], "Collation"]
 _Pipeline = Sequence[Mapping[str, Any]]
+ClusterTime = Optional[Mapping[str, Any]]
```
ShaneHarvey, line 37: Personally I think we should nix the Optional here.

sleepyStick: I don't have a huge preference either way, so its changed :)

ShaneHarvey (approved):

PR #1347, opened 2023-08-09
PYTHON-3885 Use tox in remaining Evergreen Tests

blink1073:


.evergreen/run-mongodb-aws-test.sh
```diff
@@ -39,37 +39,12 @@ fi
...
-    fi
-    python -m pip install '.[aws]'
-    python test/auth_aws/test_auth_aws.py -v
-    deactivate
-    rm -rf venvaws
-}
-
-PYTHON=${PYTHON_BINARY:-}
-if [ -z "$PYTHON" ]; then
+if [ -z "$PYTHON_BINARY" ]; then
```
NoahStapp, line 42: Do we need this file at all? What if we put the new contents directly into `config.yml`?

.evergreen/run-mongodb-oidc-test.sh
```diff
@@ -45,41 +45,13 @@ if [ "$ASSERT_NO_URI_CREDS" = "true" ]; then
...
-    python -m pip install -U pip setuptools
-    python -m pip install '.[aws]'
-    python test/auth_aws/test_auth_oidc.py -v
-    deactivate
-    rm -rf venvoidc
-}
-
-PYTHON=${PYTHON_BINARY:-}
-if [ -z "$PYTHON" ]; then
+if [ -z "$PYTHON_BINARY" ]; then
```
NoahStapp, line 48: Same for this file

NoahStapp (changes requested):

.evergreen/run-mongodb-aws-test.sh
blink1073: We would still run these two files as entrypoints for local testing.  I wanted to keep the logic out of config.yml where possible, and there are parts of this that we don't want to echo and parts that we do.

NoahStapp: Makes sense!

NoahStapp (approved):

blink1073:
Note: the OIDC failure is tracked in [PYTHON-3909](https://jira.mongodb.org/browse/PYTHON-3909)

PR #1349, opened 2023-08-09
PYTHON-3855: Appends suggestion `Use CodecOptions` into errorhandler for datetime conversions

Jibola:


bson/datetime_ms.py
```diff
@@ -127,24 +129,33 @@ def _millis_to_datetime(millis: int, opts: CodecOptions) -> Union[datetime.datet
...
-
-        if opts.tz_aware:
-            dt = EPOCH_AWARE + datetime.timedelta(seconds=seconds, microseconds=micros)
-            if opts.tzinfo:
-                dt = dt.astimezone(tz)
-            return dt
-        else:
-            return EPOCH_NAIVE + datetime.timedelta(seconds=seconds, microseconds=micros)
+        try:
+            tz = opts.tzinfo or datetime.timezone.utc
```
blink1073, line 133: Perhaps put this part in a function to avoid such a large `try:` section.

bson/datetime_ms.py
```diff
@@ -127,24 +129,33 @@ def _millis_to_datetime(millis: int, opts: CodecOptions) -> Union[datetime.datet
...
+            micros = diff * 1000
+
+            if opts.tz_aware:
+                dt = EPOCH_AWARE + datetime.timedelta(seconds=seconds, microseconds=micros)
+                if opts.tzinfo:
+                    dt = dt.astimezone(tz)
+                return dt
+            else:
+                return EPOCH_NAIVE + datetime.timedelta(seconds=seconds, microseconds=micros)
+        except Exception:
```
blink1073, line 151: We should only catch `InvalidBSON`

bson/datetime_ms.py
```diff
@@ -127,24 +129,33 @@ def _millis_to_datetime(millis: int, opts: CodecOptions) -> Union[datetime.datet
...
+            _, exc_value, _ = sys.exc_info()
```
blink1073, line 152: I think it would be cleaner to make a new `InvalidBSON` error with the new error string than modify the `exc_value`.

blink1073 (changes requested):

bson/datetime_ms.py
ShaneHarvey: I think we can move the "try:" down to cover only the lines that might raise the OverflowError.

ShaneHarvey (commented):
Just a note, once we finalize the Python exception handling we'll need to make a similar change to the C bson module.

bson/datetime_ms.py
Jibola: Ah, so the thing here is this exception is actually an `OverflowError` that gets thrown at this point in the code, so catching only `InvalidBSON` will not trigger the exception capture. Further down the pipe it gets re-converted into `InvalidBSON`. We run the same/similar behavior in the calling function [`_bson_to_dict`](https://github.com/mongodb/mongo-python-driver/blob/master/bson/__init__.py#L594-L607). 
```
def _bson_to_dict(data: Any, opts: CodecOptions) -> Any:
    """Decode a BSON string to document_class."""
    data, view = get_data_and_view(data)
    try:
        if _raw_document_class(opts.document_class):
            return opts.document_class(data, opts)
        _, end = _get_object_size(data, 0, len(data))
        return _elements_to_dict(data, view, 4, end, opts)
    except InvalidBSON:
        raise
    except Exception:
        # Change exception type to InvalidBSON but preserve traceback.
        _, exc_value, exc_tb = sys.exc_info()
        raise InvalidBSON(str(exc_value)).with_traceback(exc_tb)
```

Based on the criteria we set, though, I can definitely remove using `Exception` and I think favor using something still fairly broad, but clearly more specific. 

e.g. `OverflowError`'s base class `ArithmeticError`.

bson/datetime_ms.py
Jibola: Yeah, I'll go ahead and move this down to 144. (Right before the first arithmetic operation occurs)

bson/datetime_ms.py
Jibola: Completely agree here! I think what I wanted to achieve is more cleanly done via 
`raise _ from` syntax.

blink1073:
The Python portion LGTM!

bson/datetime_ms.py
```diff
@@ -138,13 +141,17 @@ def _millis_to_datetime(millis: int, opts: CodecOptions) -> Union[datetime.datet
...
+        try:
+            if opts.tz_aware:
+                dt = EPOCH_AWARE + datetime.timedelta(seconds=seconds, microseconds=micros)
+                if opts.tzinfo:
+                    dt = dt.astimezone(tz)
+                return dt
+            else:
+                return EPOCH_NAIVE + datetime.timedelta(seconds=seconds, microseconds=micros)
+        except ArithmeticError as err:
+            raise InvalidBSON(f"{err.args[0]} {DATETIME_ERROR_SUGGESTION}") from err
```
ShaneHarvey, line 153: Could we add a test that asserts this behavior?

ShaneHarvey (changes requested):

bson/datetime_ms.py
Jibola: Added! :)

bson/_cbsonmodule.c
```diff
@@ -281,13 +283,40 @@ static PyObject* datetime_from_millis(long long millis) {
...
+        * Calling _error clears the error state, so fetch it first.
+        */
+        PyErr_Fetch(&etype, &evalue, &etrace);
+
+        /* Only add addition error message on ValueError exceptions. */
+        if (PyErr_GivenExceptionMatches(etype, PyExc_ValueError)) {
+            if (evalue) {
+                PyObject* err_msg = PyObject_Repr(evalue);
+                if(err_msg) {
+                    PyObject* appendage = PyUnicode_FromString(" (Use CodecOptions(datetime_conversion=DATETIME_AUTO) or MongoClient(datetime_conversion='DATETIME_AUTO'))");
```
ShaneHarvey, line 306: We need to error check the results of PyUnicode_FromString and PyUnicode_Concat.

bson/datetime_ms.py
```diff
@@ -138,13 +140,17 @@ def _millis_to_datetime(millis: int, opts: CodecOptions) -> Union[datetime.datet
...
```
ShaneHarvey, line 152: What do you think about `{err}` instead of `{err.args[0]}`? I think it would result in a friendlier error message.

bson/_cbsonmodule.c
```diff
@@ -281,13 +283,40 @@ static PyObject* datetime_from_millis(long long millis) {
...
+
+        /*
+        * Calling _error clears the error state, so fetch it first.
+        */
+        PyErr_Fetch(&etype, &evalue, &etrace);
+
+        /* Only add addition error message on ValueError exceptions. */
+        if (PyErr_GivenExceptionMatches(etype, PyExc_ValueError)) {
+            if (evalue) {
+                PyObject* err_msg = PyObject_Repr(evalue);
```
ShaneHarvey, line 304: Related to my other comment, I think this should be PyObject_Str for a friendlier error message.

bson/datetime_ms.py
```diff
@@ -23,10 +23,12 @@
 from typing import Any, Union, cast
 
 from bson.codec_options import DEFAULT_CODEC_OPTIONS, CodecOptions, DatetimeConversion
+from bson.errors import InvalidBSON
 from bson.tz_util import utc
 
 EPOCH_AWARE = datetime.datetime.fromtimestamp(0, utc)
 EPOCH_NAIVE = EPOCH_AWARE.replace(tzinfo=None)
+DATETIME_ERROR_SUGGESTION = "(Use CodecOptions(datetime_conversion=DATETIME_AUTO) or MongoClient(datetime_conversion='DATETIME_AUTO'))"
```
ShaneHarvey, line 31: I suggest "(Consider using Codec...)" because not all users will want/need to use this feature when this error happens. For example sometimes the user accidentally introduced out of range years and they might want to remove those dates instead of preserving them. 

Also it would be great to link to `https://pymongo.readthedocs.io/en/stable/examples/datetimes.html#handling-out-of-range-datetimes`.

One last thing is that `DATETIME_ERROR_SUGGESTION` should be `_DATETIME_ERROR_SUGGESTION` to make it private. Anything not beginning with `_` can be misinterpreted as public api.

ShaneHarvey (changes requested):

test/test_bson.py
```diff
@@ -1309,6 +1309,11 @@ def __int__(self):
         with self.assertRaises(TypeError):
             encode({"x": float_ms})
 
+        # Test InvalidBSON Errors on conversion include _DATETIME_ERROR_SUGGESTION
```
Jibola, line 1312:
```suggestion
        # Test InvalidBSON errors on conversion include _DATETIME_ERROR_SUGGESTION
```

bson/_cbsonmodule.c
```diff
@@ -275,19 +275,51 @@ static PyObject* datetime_from_millis(long long millis) {
...
+    PyObject* datetime;
     int diff = (int)(((millis % 1000) + 1000) % 1000);
     int microseconds = diff * 1000;
     Time64_T seconds = (millis - diff) / 1000;
     struct TM timeinfo;
     cbson_gmtime64_r(&seconds, &timeinfo);
 
-    return PyDateTime_FromDateAndTime(timeinfo.tm_year + 1900,
+    datetime = PyDateTime_FromDateAndTime(timeinfo.tm_year + 1900,
                                       timeinfo.tm_mon + 1,
```
ShaneHarvey, line 286: Could you fix the offset argument indentation here?

ShaneHarvey (changes requested):

blink1073 (approved):
LGTM!

PR #1351, opened 2023-08-10
PYTHON-3907 add --disallow-untyped-defs for mypy

sleepyStick:
Added the flag to the tox.ini file and fixed any untyped defs that came up as a result of this changed

https://jira.mongodb.org/browse/PYTHON-3907

### DIFF
--- a/pymongo/topology.py
+++ b/pymongo/topology.py
@@ -890,7 +890,7 @@ class Topology:
             msg = "CLOSED "
         return f"<{self.__class__.__name__} {msg}{self._description!r}>"
 
-    def eq_props(self) -> Tuple[Tuple[Tuple[str, int], ...], Optional[str], Optional[str], str]:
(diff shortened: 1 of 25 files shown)


pymongo/topology.py
```diff
@@ -890,7 +890,7 @@ def __repr__(self) -> str:
             msg = "CLOSED "
         return f"<{self.__class__.__name__} {msg}{self._description!r}>"
 
-    def eq_props(self):
+    def eq_props(self) -> Tuple[Tuple[Tuple[str, int], ...], Optional[str], Optional[str], str]:
```
ShaneHarvey, line 893: Probably can be `Tuple[Any]`.

sleepyStick: It's probably better to have the specifics; replaced the innermost `Tuple[str, int]` with `_Address`, which is probably more accurate.

PR #1352, opened 2023-08-10
PYTHON-3821 use overload pattern for _DocumentType

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3821

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -1143,6 +1143,11 @@ def decode_all(
        Replaced `as_class`, `tz_aware`, and `uuid_subtype` options with
        `codec_options`.
     """
+    # opts = codec_options or DEFAULT_CODEC_OPTIONS
+    # if not isinstance(opts, CodecOptions):
+    #     raise _CODEC_OPTIONS_TYPE_ERROR
+
+    # return _decode_all(data, opts)
     if codec_options is None:
         return _decode_all(data, DEFAULT_CODEC_OPTIONS)
 
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -1106,9 +1106,21 @@ if _USE_C:
     _decode_all = _cbson._decode_all  # noqa: F811
 
 
+@overload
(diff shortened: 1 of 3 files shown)


bson/__init__.py
```diff
@@ -1131,11 +1143,13 @@ def decode_all(
        Replaced `as_class`, `tz_aware`, and `uuid_subtype` options with
        `codec_options`.
     """
-    opts = codec_options or DEFAULT_CODEC_OPTIONS
```
ShaneHarvey, line 1134: What was wrong (according to mypy) with this code? I find the new code less readable since it calls _decode_all in multiple places.

ShaneHarvey (commented):
Is it possible to add tests for this change to show that it improves the behavior?

bson/__init__.py
sleepyStick: I was getting the following mypy error:
```
typecheck-mypy: commands[0]> mypy --install-types --non-interactive bson gridfs tools pymongo
bson/__init__.py: note: In function "decode_all":
bson/__init__.py:1150: error: Cannot infer type argument 1 of "_decode_all"  [misc]
        return _decode_all(data, opts)
               ^~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 85 source files)
typecheck-mypy: exit 1 (2.96 seconds) /Users/iris.ho/GitHub/mongo-python-driver> mypy --install-types --non-interactive bson gridfs tools pymongo pid=40888
.pkg: _exit> python /Users/iris.ho/Library/Python/3.9/lib/python/site-packages/pyproject_api/_backend.py True setuptools.build_meta
  typecheck-mypy: FAIL code 1 (10.28=setup[7.32]+cmd[2.96] seconds)
  evaluation failed :( (10.35 seconds)
 ```

sleepyStick:
Yes! Just wrote some tests! :)

ShaneHarvey (approved):
Nice!

PR #1360, opened 2023-08-21
PYTHON-3668 Use polling SDAM mode on FaaS and add sdamMode=auto/stream/poll

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3668

The implementation of https://github.com/mongodb/specifications/pull/1452.

- Introduces sdamMode URI option.
- Disable streaming on FaaS by default.
- Resync SDAM and uri-options spec tests.
- Update Lambda test to assert streaming is not used.


blink1073 (approved):
LGTM!

blink1073:
I'd say let's wait for consensus on https://github.com/mongodb/specifications/pull/1452, and defer this to 4.6.

pymongo/common.py
```diff
@@ -669,6 +672,16 @@ def validate_datetime_conversion(option: Any, value: Any) -> Optional[DatetimeCo
     raise TypeError(f"{option} must be a str or int representing DatetimeConversion")
 
 
+def validate_sdam_mode(option: Any, value: Any) -> str:
```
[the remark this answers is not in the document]
ShaneHarvey: Good call. Done.

pymongo/monitor.py
```diff
@@ -255,7 +261,17 @@ def _check_once(self) -> ServerDescription:
         address = self._server_description.address
         if self._publish:
             assert self._listeners is not None
-            self._listeners.publish_server_heartbeat_started(address)
+            sd = self._server_description
+            # TODO: this is not accurate in the rare case the pool has a
```
blink1073, line 265: Still TODO?   Needs a follow up ticket?

ShaneHarvey: I don't think fixing this is worth the effort, especially since it should be so rare (in streaming the socket is almost always checked out of the pool so the connection would need to become stale in the short window between the check in and the checkout) and the impact of this is trivial (ServerHeartbeatStartedEvent.awaited will be True instead of False).

blink1073:
I opened https://jira.mongodb.org/browse/PYTHON-3994 to track the docs failure.

blink1073 (approved):

PR #1361, opened 2023-08-23
[bug-fix] Return empty list if server_descriptions do not exist

piyiotisk:
This PR fixes the following error. This is a part of our CI pipeline (withplum.com)
Unfortunately, almost in every build we encounter this failure. Locally it's difficult to reproduce. 
Here's the Jira ticket from Mongodb https://jira.mongodb.org/browse/PYTHON-1271

```
.venv/lib/python3.9/site-packages/pymongo/database.py:959: in list_collection_names
    return [result["name"] for result in self.list_collections(session=session, **kwargs)]
.venv/lib/python3.9/site-packages/pymongo/database.py:911: in list_collections
    return self.__client._retryable_read(_cmd, read_pref, session)
.venv/lib/python3.9/site-packages/pymongo/_csot.py:105: in csot_wrapper
    return func(self, *args, **kwargs)
.venv/lib/python3.9/site-packages/pymongo/mongo_client.py:1441: in _retryable_read
    server = self._select_server(read_pref, session, address=address)
.venv/lib/python3.9/site-packages/pymongo/mongo_client.py:1257: in _select_server
    server = topology.select_server(server_selector)
.venv/lib/python3.9/site-packages/pymongo/topology.py:272: in select_server
    server = self._select_server(selector, server_selection_timeout, address)
.venv/lib/python3.9/site-packages/pymongo/topology.py:261: in _select_server
    servers = self.select_servers(selector, server_selection_timeout, address)
.venv/lib/python3.9/site-packages/pymongo/topology.py:223: in select_servers
    server_descriptions = self._select_servers_loop(selector, server_timeout, address)
.venv/lib/python3.9/site-packages/pymongo/topology.py:231: in _select_servers_loop
    server_descriptions = self._description.apply_selector(
.venv/lib/python3.9/site-packages/pymongo/topology_description.py:317: in apply_selector
    return self._apply_local_threshold(selection)
.venv/lib/python3.9/site-packages/pymongo/topology_description.py:261: in _apply_local_threshold
    return [
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

.0 = <list_iterator object at 0x7f562981f5b0>

    return [
>       s for s in selection.server_descriptions if (s.round_trip_time - fastest) <= threshold
    ]
E   TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'

.venv/lib/python3.9/site-packages/pymongo/topology_description.py:262: TypeError
```

### DIFF
--- a/pymongo/topology_description.py
+++ b/pymongo/topology_description.py
@@ -267,12 +267,18 @@ class TopologyDescription:
     def _apply_local_threshold(self, selection: Optional[Selection]) -> List[ServerDescription]:
         if not selection:
             return []
+    
+        server_descriptions = selection.server_descriptions
+        server_descriptions = [s for s in server_descriptions if s.round_trip_time != None]
+        if not server_descriptions:
+            return []
+        
         # Round trip time in seconds.
         fastest = min(cast(float, s.round_trip_time) for s in selection.server_descriptions)
         threshold = self._topology_settings.local_threshold_ms / 1000.0
         return [
             s
-            for s in selection.server_descriptions
+            for s in server_descriptions
             if (cast(float, s.round_trip_time) - fastest) <= threshold
         ]
 


pymongo/topology_description.py
```diff
@@ -267,12 +267,18 @@ def srv_max_hosts(self) -> int:
     def _apply_local_threshold(self, selection: Optional[Selection]) -> List[ServerDescription]:
         if not selection:
             return []
+    
+        server_descriptions = selection.server_descriptions
+        server_descriptions = [s for s in server_descriptions if s.round_trip_time != None]
```
KonstantinosTsitoumis, line 272: Do we need to re-assign here? Shouldn't we do it in a one-liner?

pymongo/topology_description.py
```diff
@@ -267,12 +267,18 @@ def srv_max_hosts(self) -> int:
...
+        if not server_descriptions:
+            return []
```
stavros47, line 274:
```suggestion
server_descriptions = [s for s in (selection.server_descriptions or []) if s.round_trip_time != None]

if not server_descriptions:
   return []
```

expobrain (changes requested):
We have already the patched version in the branch https://github.com/withplum/mongo-python-driver/tree/4.5.0-patched

ShaneHarvey:
Hello, thanks for the PR. As stated in [PYTHON-1271](https://jira.mongodb.org/browse/PYTHON-1271) it's unexpected for this case to happen since at this point all the servers _should_ be in a known state with a non-None value for round_trip_time. Are you able to reproduce this issue? If so, it would be better for us to debug this case to understand the root cause.

ShaneHarvey:
@piyiotisk, @KonstantinosTsitoumis, @stavros47, @expobrain Are you able to reproduce this issue? If so, it would be better for us to debug this case to understand the root cause.

blink1073:
We're going to close this pull request for now, please comment if you'd like us to open it back up.

Ale-Cas:
Hi, i'm still able to reproduce this issue with pymongo v4.6.1. 

The root cause of the error is probably due to our configuration of the topology / replica set but it would be helpful if this was handled in a more specific way, instead of getting a generic TypeError when computing this list comprehension. 

I think the solution proposed here was to basically return an empty list in such cases, but I'm not sure wether that's the best way to handle this, and I would prefer to have a ConfigurationError specifying what's wrong in the configuration in this case.

PR #1362, opened 2023-08-24
PYTHON-3932 Support mypy --strict testing in bson package

blink1073:


bson/codec_options.py
```diff
@@ -246,36 +246,36 @@ class CodecOptions(Tuple, Generic[_DocumentType]):
...
             tzinfo: Optional[datetime.tzinfo] = ...,
             type_registry: Optional[TypeRegistry] = ...,
             datetime_conversion: Optional[int] = ...,
-        ) -> "CodecOptions[_DocumentType]":
+        ) -> CodecOptions[_DocumentType]:
             ...
 
         # CodecOptions API
-        def with_options(self, **kwargs: Any) -> "CodecOptions[_DocumentType]":
+        def with_options(self, **kwargs: Any) -> CodecOptions[Any]:
```
Jibola, line 261: Wouldn't we still want this to type to `CodecOptions[_DocumentType]`

bson/son.py
```diff
@@ -170,15 +171,15 @@ def __eq__(self, other: Any) -> bool:
         """
         if isinstance(other, SON):
             return len(self) == len(other) and list(self.items()) == list(other.items())
-        return self.to_dict() == other
+        return cast(bool, self.to_dict() == other)
```
Jibola, line 174: Should this need an explicit cast? I assumed the `self.to_dict() == other` would automatically evaluate to a boolean

pymongo/message.py
```diff
@@ -1127,7 +1127,7 @@ class _EncryptedBulkWriteContext(_BulkWriteContext):
 
     def __batch_command(
         self, cmd: MutableMapping[str, Any], docs: List[Mapping[str, Any]]
-    ) -> Tuple[Dict[str, Any], List[Mapping[str, Any]]]:
+    ) -> Tuple[MutableMapping[str, Any], List[Mapping[str, Any]]]:
```
Jibola, line 1130: NIT: Not sure if you wanted to do this for all of the typing parts (mainly for consistency), but thoughts on converting all of the `typing.[built-ins] -> [built-ins]` in another task?

That also balloons scope here so I wouldn't want to bloat this PR. 
```suggestion
    ) -> tuple[MutableMapping[str, Any], list[Mapping[str, Any]]]:
```

bson/codec_options.py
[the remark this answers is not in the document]
blink1073: I don't think so, because one of the options can be "document_class", which would change the type.

bson/son.py
blink1073: I agree, but the type checker complained anyway.  ¯\_(ツ)_/¯

pymongo/message.py
blink1073: Ah, I had forgotten about `tuple`.  I agree this should be done as part of https://jira.mongodb.org/browse/PYTHON-3905, which I should have done first.

bson/codec_options.py
```diff
@@ -140,8 +140,8 @@ def __init__(
     ) -> None:
         self.__type_codecs = list(type_codecs or [])
         self._fallback_encoder = fallback_encoder
-        self._encoder_map: Dict[Any, Any] = {}
-        self._decoder_map: Dict[Any, Any] = {}
+        self._encoder_map: dict[Any, Any] = {}
```
NoahStapp, line 143: This should be kept as `Dict` to be consistent with the `pymongo` package typing.

bson/codec_options.py
```diff
@@ -246,36 +246,36 @@ class CodecOptions(Tuple, Generic[_DocumentType]):
...
         # CodecOptions API
-        def with_options(self, **kwargs: Any) -> "CodecOptions[_DocumentType]":
+        def with_options(self, **kwargs: Any) -> CodecOptions[Any]:
             ...
 
         def _arguments_repr(self) -> str:
             ...
 
-        def _options_dict(self) -> Dict[Any, Any]:
+        def _options_dict(self) -> dict[Any, Any]:
```
NoahStapp, line 267: This should be kept as `Dict` to be consistent with the `pymongo` package typing.

bson/codec_options.py
```diff
@@ -246,36 +246,36 @@ class CodecOptions(Tuple, Generic[_DocumentType]):
...
             ...
 
         # NamedTuple API
         @classmethod
-        def _make(cls, obj: Iterable) -> "CodecOptions[_DocumentType]":
+        def _make(cls, obj: Iterable[Any]) -> CodecOptions[_DocumentType]:
             ...
 
-        def _asdict(self) -> Dict[str, Any]:
+        def _asdict(self) -> dict[str, Any]:
```
NoahStapp, line 275: This should be kept as `Dict` to be consistent with the `pymongo` package typing.

bson/raw_bson.py
```diff
@@ -156,7 +160,9 @@ def __inflated(self) -> Mapping[str, Any]:
         return self.__inflated_doc
 
     @staticmethod
-    def _inflate_bson(bson_bytes: bytes, codec_options: CodecOptions) -> Mapping[Any, Any]:
+    def _inflate_bson(
+        bson_bytes: bytes, codec_options: CodecOptions[RawBSONDocument]
+    ) -> Mapping[str, Any]:
```
NoahStapp, line 165: Should return `MutableMapping[str, Any]` to match the signature of `_inflate_bson`.

pymongo/message.py
NoahStapp: Python 3.9 is the oldest version to support the use of the literal generics such as `dict` without the future annotations import. I recall that we decided against using the literals while we still supported older versions.

NoahStapp (changes requested):

pymongo/message.py
blink1073: Right, but we're going to use `from __future__ import annotations` in all files, and there is not harm in continuing to do so indefinitely.

bson/codec_options.py
blink1073: Those will be changed as part of [PYTHON-3905](https://jira.mongodb.org/browse/PYTHON-3905)

bson/raw_bson.py
blink1073: Actually I think it should be `Mapping` everywhere in this file, updated.

blink1073: Actually, this method is somewhat out of place, I think it really should have been in `bson/__init__.py`.  I think it is fine for this to return `MutableMapping` since it is used in other contexts, while the APIs of `RawBSONDocument` return `Mapping`, since it is meant to be a read-only object.

NoahStapp (approved):

PR #1367, opened 2023-08-30
PYTHON-3920 - Migrate AWS Auth Tests to use AWS Secrets

NoahStapp:


.evergreen/config.yml
```diff
@@ -553,39 +554,21 @@ functions:
...
 
-  "add aws auth variables to file":
+  "get aws auth secrets":
+    - command: ec2.assume_role
+      params:
+        role_arn: ${aws_test_secrets_role}
     - command: shell.exec
       type: test
       params:
+        add_expansions_to_env: true
```
blink1073, line 564: These can be simplified to `include_expansions_in_env: ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN"]`.  I tried it out in https://github.com/mongodb/mongo-go-driver/pull/1365

.evergreen/config.yml
```diff
@@ -664,29 +635,33 @@ functions:
...
-             exit 0
-          fi
-          # Write an empty prepare_mongodb_aws so no auth environment variables
-          # are set.
-          rm "${PROJECT_DIRECTORY}/prepare_mongodb_aws.sh" || true
-          PYTHON_BINARY=${PYTHON_BINARY} ASSERT_NO_URI_CREDS=true .evergreen/run-mongodb-aws-test.sh
+
+          cd -
+
+          # Write an empty prepare_aws_env so no auth environment variables are set.
```
blink1073, line 660: We shouldn't need to do this anymore (perhaps something still needs to change in D-E-T?

.evergreen/config.yml
```diff
@@ -695,60 +670,42 @@ functions:
...
         shell: "bash"
         script: |
-          ${PREPARE_SHELL}
           if [ "${skip_EC2_auth_test}" = "true" ]; then
              echo "This platform does not support the web identity auth test, skipping..."
              exit 0
           fi
-          set -ex
+
+          # Try to source exported AWS Secrets
```
blink1073, line 678: We can probably move this logic into `run-mongodb-aws-test.sh` and pass in the test to run (`web-identity` in this case).

.evergreen/run-tests.sh
```diff
@@ -69,6 +69,7 @@ if [ -n "$TEST_ENTERPRISE_AUTH" ]; then
         echo "Writing keytab"
         echo ${KEYTAB_BASE64} | base64 -d > ${PROJECT_DIRECTORY}/.evergreen/drivers.keytab
         echo "Running kinit"
+        python -c "import os;print(sorted([(k, v[:2]) for k, v in os.environ.items()]))"
```
blink1073, line 72: Remove?

blink1073 (changes requested):

.evergreen/config.yml
blink1073: Ah, I see how it is used now.  I think we can also absorb this logic into `run-mongodb-aws-test.sh`.  I do like the idea of adding the script in D-E-T as you've done.

blink1073:
Nice, I like the direction this is heading.  Looks like you still need to convert the last few tests?

NoahStapp:
> Nice, I like the direction this is heading. Looks like you still need to convert the last few tests?

Should all be converted and good to go now!

.evergreen/config.yml
blink1073: This particular instance still needs to be updated.  :)

NoahStapp: for `get_aws_secrets` it fails without `expansions_to_env`, `test-atlas` has been updated to use `include_expansions_in_env`

blink1073: Wait so there are things we're missing from the new vault?

NoahStapp: `get_aws_secrets` is what fetches the secrets from the vault.

blink1073: Right, okay, so now I'm confused, because that's what I'm doing in the go driver ticket.

NoahStapp:
> Beautiful!

Are these failures expected? 

Screen Shot 2023-09-05 at 11 20 28 AM

blink1073:
No, I something changed on those tests, I'll open a ticket.

NoahStapp:
> No, I something changed on those tests, I'll open a ticket.

Is this good to merge then?

blink1073:
> Is this good to merge then?

Yes, I opened https://jira.mongodb.org/browse/PYTHON-3951

PR #1368, opened 2023-08-31
PYTHON-3598 Add Command(Succeeded|Failed)Event.database_name property

Jibola:


pymongo/monitoring.py
```diff
@@ -570,12 +570,14 @@ def __init__(
         connection_id: _Address,
         operation_id: Optional[int],
         service_id: Optional[ObjectId] = None,
+        database_name: Optional[str] = None,
```
ShaneHarvey, line 573: Why is database_name optional?

test/command_monitoring/pre-42-server-connection-id.json
```diff
@@ -0,0 +1,101 @@
+{
+  "description": "pre-42-server-connection-id",
```
ShaneHarvey, line 2: We have not implemented the server connection id spec yet so let's remove these test files.

pymongo/monitoring.py
```diff
@@ -652,7 +666,9 @@ def command(self) -> _DocumentOut:
     @property
     def database_name(self) -> str:
         """The name of the database this command was run against."""
-        return self.__db
+        db = super().database_name
+        assert db is not None
```
ShaneHarvey, line 670: Why do we need the assert?

Jibola: Our typechecker had been raising issues because `self.__db` is actually typed as `Optional[str]`. Used an assert to mitigate that error. 

Alternatively, I could just `# type: ignore`

pymongo/monitoring.py
Jibola: I followed the conventions above `service_id`. I also assumed `Command(Succeeded|Failed)Event`s could be instantiated in some other codebase and fail because it's now a required argument. I figured keeping it optional doesn't introduce any backward breaking changes.

ShaneHarvey: Oh good point. I suppose one alternative would be to make it `database_name: str = "",`. What do you think about using `""` to mean no database was given (which could only happen if the user instantiated these classes on their own)? That would avoid the None/assert issue.

Jibola: Sounds good to me.

pymongo/monitoring.py
```diff
@@ -635,25 +645,24 @@ def __init__(
...
 
     @property
     def command(self) -> _DocumentOut:
         """The command document."""
         return self.__cmd
 
-    @property
-    def database_name(self) -> str:
-        """The name of the database this command was run against."""
-        return self.__db
```
ShaneHarvey, line 655: I think we still need this? Otherwise the docs will say that CommandStartedEvent.database_name was added in 4.6 when it really has existed the whole time.

pymongo/server.py
```diff
@@ -140,8 +140,8 @@ def run_operation(
             request_id, data, max_doc_size = self._split_message(message)
 
         if publish:
-            cmd, dbn = operation.as_command(conn)
             assert listeners is not None
+            cmd, dbn = operation.as_command(conn)
```
ShaneHarvey, line 144: I assume this change was accidental. Can we revert it?

ShaneHarvey (changes requested):

ShaneHarvey:
Since this is a public api change could you add a mention in the changelog?

pymongo/server.py
Jibola: Yeah, a vestige of an failed attempt to refactor from earlier.

doc/changelog.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Changes in Version 4.6
+----------------------
+
+PyMongo 4.6 brings a number of improvements including:
+- Added support for recording database name to command*Events (`PYTHON-3598`_)
```
ShaneHarvey, line 8: Let's be a little more direct here:
```
- Added the :attr:`pymongo.monitoring.CommandSucceededEvent.database_name` property.
- Added the :attr:`pymongo.monitoring.CommandFailedEvent.database_name` property.
```

pymongo/monitoring.py
```diff
@@ -651,8 +665,12 @@ def command(self) -> _DocumentOut:
 
     @property
     def database_name(self) -> str:
-        """The name of the database this command was run against."""
-        return self.__db
+        """The name of the database this command was run against.
+
+        .. versionchanged:: 4.6
+           Derived from base class's property.
```
ShaneHarvey, line 671: This versionchanged is not needed since nothing in the api has changed.

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM, one last thing, when you merge could you update the title/commit message to describe the api change too, like:
```
PYTHON-3598 Add Command(Succeeded|Failed)Event.database_name property
```

PR #1371, opened 2023-09-11
PYTHON-3698 Support mypy 1.2 --strict testing (part 1)

blink1073:
I broke this up into about half of the errors to save review.


mypy.ini
```diff
@@ -1,4 +1,5 @@
 [mypy]
+python_version = 3.7
```
blink1073, line 2: This setting enables us to test against Python 3.7 typing semantics directly, simplifying local running and CI.

ShaneHarvey: Does this mean that we now _never_ test against 3.11?

pymongo/encryption.py
```diff
@@ -923,7 +941,7 @@ def get_key(self, id: Binary) -> Optional[RawBSONDocument]:
         """
         self._check_closed()
         assert self._key_vault_coll is not None
-        return self._key_vault_coll.find_one({"_id": id})
+        return cast(RawBSONDocument, self._key_vault_coll.find_one({"_id": id}))
```
ShaneHarvey, line 944: Can we type _key_vault_coll as `Collection[RawBSONDocument]` rather than adding these casts?

pymongo/change_stream.py
```diff
@@ -422,17 +422,17 @@ def try_next(self) -> Optional[_DocumentType]:
         self._start_at_operation_time = None
 
         if self._decode_custom:
-            return _bson_to_dict(change.raw, self._orig_codec_options)
-        return change
+            return cast(_DocumentType, _bson_to_dict(change.raw, self._orig_codec_options))
+        return cast(_DocumentType, change)
```
ShaneHarvey, line 426: Can we avoid these casts by defining _orig_codec_options in terms of _DocumentType? casts are bad in my opinion because they can mask legitimate typing bugs so we should only use them as a last resort.

ShaneHarvey (changes requested):

mypy.ini
blink1073: It's in the example config from `mypy`, I think of it like the version we use in `black`.    Unless there is a backwards-incompatible change to typing, this should be fine.

pymongo/encryption.py
blink1073: Done

pymongo/change_stream.py
blink1073: Fixed.  However, I noticed that `with_options` should defer to the `CodecOptions` type, since that is often the option you are changing.  This requires casting the parent type.

mypy.ini
ShaneHarvey: Okay sounds good to me.

pymongo/collection.py
```diff
@@ -431,7 +431,7 @@ def with_options(
         read_preference: Optional[_ServerMode] = None,
         write_concern: Optional[WriteConcern] = None,
         read_concern: Optional[ReadConcern] = None,
-    ) -> Collection[_DocumentType]:
+    ) -> Collection[_DocumentTypeArg]:
```
ShaneHarvey, line 434: Could we revert this change and discuss it in a new jira ticket? I think the interaction of Collection and CodecOption's DocumentType is not well defined so I don't think we should make breaking changes like this without more consideration of the whole api.

ShaneHarvey (changes requested):

pymongo/collection.py
blink1073: Yeah, I agree there isn't a good way to handle this with the typing system.  I think we should stick with what we've previously shipped.

ShaneHarvey (approved):

PR #1373, opened 2023-09-13
PYTHON-3940 Pytest Config Improvements

NoahStapp:


blink1073 (approved):
LGTM, thanks!

.evergreen/run-tests.sh
```diff
@@ -262,7 +262,7 @@ if [ -z "$GREEN_FRAMEWORK" ]; then
         # causing this script to exit.
         python -c "from bson import _cbson; from pymongo import _cmessage"
     fi
-    python -m pytest -v $TEST_ARGS
+    python -m pytest -v --durations=5 --maxfail=10 $TEST_ARGS
```
ShaneHarvey, line 265: I can see the benefit of `--maxfail=10` for running tox locally but what's the rationale for using it in Evergreen? I'm worried it will make it harder to get test passing in some cases. For example let's say 10 tests fail on the first attempt, then you fix those locally, then run a new patch only to see other test failures.

NoahStapp: I've often wanted a test suite to fail quickly when debugging failures for faster iteration time. If a suite has 10 failures already, it's likely something is broken across all or most tests.

ShaneHarvey: Okay doesn't hurt to try it out!

PR #1374, opened 2023-09-13
PYTHON-3943 Serverless failure - TestUnifiedRunCursorCommand

NoahStapp:
This issue is caused by a side-effect of `pymongo.message._op_msg`. Because we pass `_GetMore._as_command` as the `command` parameter, this line directly adds the `$db` value to ` _GetMore._as_command`: https://github.com/mongodb/mongo-python-driver/blob/14640256cb0e4be21a94dbd3926b490ad8257bbc/pymongo/message.py#L746

As a result, when the final `getMore` of the test hits this line: https://github.com/mongodb/mongo-python-driver/blob/14640256cb0e4be21a94dbd3926b490ad8257bbc/pymongo/server.py#L127 
it does not get `$db` added to its command spec for the test listener, causing the test to fail.

### DIFF
--- a/pymongo/server.py
+++ b/pymongo/server.py
@@ -147,11 +147,6 @@ class Server:
                 conn.send_message(data, max_doc_size)
                 reply = conn.receive_message(request_id)
 
-            print(
-                f"\nOP_MSG for request {request_id}\nCHECKSUM_PRESENT: {reply.CHECKSUM_PRESENT}\nEXHAUST_ALLOWED: {reply.EXHAUST_ALLOWED}\nMORE_TO_COME: {reply.MORE_TO_COME}\nOP_CODE: {reply.OP_CODE}\nflags: {reply.flags}\nmore_to_come: {reply.more_to_come}\ndata: {reply.payload_document.obj}",
-                flush=True,
-            )
-
             # Unpack and check for command errors.
             if use_cmd:
                 user_fields = _CURSOR_DOC_FIELDS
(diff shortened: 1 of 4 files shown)


pymongo/server.py
```diff
@@ -132,6 +132,8 @@ def run_operation(
 
         if publish:
             cmd, dbn = operation.as_command(conn)
+            if "$db" not in cmd and dbn is not None:
```
ShaneHarvey, line 135: 2 questions: Is dbn ever None here? Could we move this logic to as_command?

ShaneHarvey (changes requested):
Is it possible to add a test for this that runs on regular servers?

pymongo/server.py
ShaneHarvey: I still have this question.

ShaneHarvey:
> Is it possible to add a test for this that runs on regular servers?

This seems possible. Just add a test that checks for $db on OP_MSG an exhaust cursor.

pymongo/server.py
NoahStapp: `dbn` does not appear to ever be `None`, but moving this logic to `as_command` does not work, the test fails with the same "missing $db" error message.

NoahStapp:
> > Is it possible to add a test for this that runs on regular servers?
> 
> This seems possible by adding a test that checks for $db on an OP_MSG exhaust cursor.

This test would verify that `$db` is set on exhaust cursors?

ShaneHarvey:
Yes, using the CommandListener to assert on the CommandStartedEvents for find/getMore commands.

ShaneHarvey (changes requested):
Nice test. I have some comments that would help make it more robust.

ShaneHarvey (approved):
LGTM!

PR #1375, opened 2023-09-14
PYTHON-3926 Add more information to connection errors and timeouts

NoahStapp:


pymongo/common.py
```diff
@@ -871,6 +870,37 @@ def _ecoc_coll_name(encrypted_fields: Mapping[str, Any], name: str) -> Any:
...
+
+
+def format_timeout_details(details: dict[str, Any]) -> str:
+    result = " (configured timeouts:"
+    if "socketTimeout" in details:
+        result += f" socketTimeout: {details['socketTimeout']}ms,"
+    if "timeout" in details:
+        result += f" timeout: {details['timeout']}ms,"
+    if "connectTimeout" in details:
+        result += f" connectTimeout: {details['connectTimeout']}ms,"
```
Jibola, line 897: NIT:
```suggestion
    for timeout in ["socketTimeout", "timeout", "connectTimeout"]:
      if timeout in details:
         result += f" {timeout}:  {details[timeout]}ms,"
```

Jibola (changes requested):
Overall, this looks great and I love how concise an improvement it is! 

Left some commentary, but only one blocking comment around `topology`

NoahStapp:
I just noticed that this doesn't cover the Connection socket timeouts like 

https://github.com/mongodb/mongo-python-driver/blob/db526f7d36d0db6c80780ec71c81ec1ef839c5e7/pymongo/pool.py#L956

and

https://github.com/mongodb/mongo-python-driver/blob/db526f7d36d0db6c80780ec71c81ec1ef839c5e7/pymongo/pool.py#L972

and

 https://github.com/mongodb/mongo-python-driver/blob/db526f7d36d0db6c80780ec71c81ec1ef839c5e7/pymongo/pool.py#L982

We already have `socket_timeout` and `connect_timeout` in the `Connection`'s `PoolOptions`, thoughts on adding `timeout` there as well? `toplogy` may require more extensive refactoring if we want to include that always as well.

pymongo/common.py
```diff
@@ -871,6 +870,37 @@ def _ecoc_coll_name(encrypted_fields: Mapping[str, Any], name: str) -> Any:
 WRITE_CONCERN_OPTIONS = frozenset(["w", "wtimeout", "wtimeoutms", "fsync", "j", "journal"])
 
 
+def _get_timeout_details(client: MongoClient) -> dict[str, Any]:
+    details = {}
+    timeout = client.options.timeout
+    socket_timeout = client.options.pool_options.socket_timeout
+    connect_timeout = client.options.pool_options.connect_timeout
+    topology = client.topology_description.server_descriptions()
```
ShaneHarvey, line 878: This should be `client.topology_description` which is what we include in ServerSelectionTimeoutError: https://github.com/mongodb/mongo-python-driver/blob/6f4e617e6db6323b1da389b94a8a598d85e7c233/pymongo/topology.py#L259

pymongo/common.py
```diff
@@ -871,6 +870,37 @@ def _ecoc_coll_name(encrypted_fields: Mapping[str, Any], name: str) -> Any:
...
+    if timeout:
+        details["timeout"] = timeout * 1000
```
ShaneHarvey, line 880: Would it make more sense to add _csot.get_timeout() here? That would reflect the last CSOT timeout that was set.

pymongo/pool.py
```diff
@@ -391,13 +396,14 @@ def _raise_connection_failure(
     if msg_prefix:
         msg = msg_prefix + msg
     if isinstance(error, socket.timeout):
-        raise NetworkTimeout(msg) from error
+        print(f"DETAILS: {details}")
```
ShaneHarvey, line 399: errant print here.

pymongo/pool.py
```diff
@@ -1549,7 +1559,10 @@ def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> Connect
                 )
 
             if isinstance(error, (IOError, OSError, SSLError)):
-                _raise_connection_failure(self.address, error)
+                timeout_details = None
+                if handler:
+                    timeout_details = _get_timeout_details(handler.client)
+                _raise_connection_failure(self.address, error, details=timeout_details)
```
ShaneHarvey, line 1565: Hmm is there a central place we could catch and re-raise these exceptions (rather than sprinkling format_timeout_details in various places)?

ShaneHarvey (changes requested):

pymongo/pool.py
NoahStapp: Ideally we'd be able to access the `_MongoClientErrorHandler` whenever we catch a timeout exception. Are there existing limitations to our internal network interfaces that prevent us from passing an instance of the handler around more widely?

NoahStapp: Following up @ShaneHarvey

ShaneHarvey: I believe we do already wrap all operations with _MongoClientErrorHandler. Does that answer your question?

NoahStapp: `_MongoClientErrorHandler` is only wrapped around `_checkout` and `_run_operation` in MongoClient, are those the only two operations relevant for this issue? If so, we could maybe move the logic in `format_timeout_details` into the error handler itself to centralize the processing.

ShaneHarvey: Answered in Zoom but summarizing here:
Yes that's true but it's better we add the info at the point the exception is raised to ensure that event publishing and (soon) command logging also include the same info. If we catch and reraise the exception in _MongoClientErrorHandler then the events/logs will see a different error message than the user.

pymongo/pool.py
```diff
@@ -407,6 +411,32 @@ def _cond_wait(condition: threading.Condition, deadline: Optional[float]) -> boo
     return condition.wait(timeout)
 
 
+def _get_timeout_details(options: PoolOptions) -> dict[str, float]:
+    details = {}
+    timeout = _csot.get_timeout()
+    socket_timeout = options.socket_timeout
+    connect_timeout = options.connect_timeout
+    if timeout:
+        details["timeout"] = timeout * 1000
```
ShaneHarvey, line 420: I suspect "timeout" will be confusing to users. Perhaps "operation timeout"? Also if _csot.get_timeout is enabled then should we avoid adding socketTimeoutMS here since CSOT overrides socketTimeoutMS?

ShaneHarvey (changes requested):
Could you add a few tests for this behavior (asserting the error message includes the expected timeout info)?

pymongo/pool.py
NoahStapp: I went with `operationTimeoutMS`. Agreed, if CSOT is present then socketTimeoutMS should not be displayed.

pymongo/pool.py
```diff
@@ -1549,7 +1586,8 @@ def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> Connect
                 )
 
             if isinstance(error, (IOError, OSError, SSLError)):
-                _raise_connection_failure(self.address, error)
+                details = _get_timeout_details(self.opts)
+                _raise_connection_failure(self.address, error, timeout_details=details)
```
ShaneHarvey, line 1590: I see _configured_socket() also calls _raise_connection_failure(). Will this then add timeout_details to the error twice?

test/test_pooling.py
```diff
@@ -411,6 +411,31 @@ def find_one():
         # maxConnecting = unbounded: 30+ connections in ~0.140+ seconds
         print(len(pool.conns))
 
+    def test_csot_timeout_message(self):
+        client = rs_or_single_client(timeoutMS=1)
+
+        with self.assertRaises(Exception) as error:
+            client.db.t.find_one({"$where": delay(2)})
+            self.assertTrue(
```
ShaneHarvey, line 419: This line needs to be de-indented otherwise it never runs.

test/test_pooling.py
```diff
@@ -411,6 +411,31 @@ def find_one():
...
+                "(configured timeouts: operationTimeoutMS: 1.0ms)" in str(error.exception)
+            )
+
+    def test_socket_timeout_message(self):
+        client = rs_or_single_client(socketTimeoutMS=1)
+
+        with self.assertRaises(Exception) as error:
+            client.db.t.find_one({"$where": delay(2)})
+            self.assertTrue("(configured timeouts: socketTimeoutMS: 1.0ms)" in str(error.exception))
```
ShaneHarvey, line 428: same here.

test/test_pooling.py
```diff
@@ -411,6 +411,31 @@ def find_one():
         # maxConnecting = unbounded: 30+ connections in ~0.140+ seconds
         print(len(pool.conns))
 
+    def test_csot_timeout_message(self):
+        client = rs_or_single_client(timeoutMS=1)
+
+        with self.assertRaises(Exception) as error:
+            client.db.t.find_one({"$where": delay(2)})
```
ShaneHarvey, line 418: We can use pymongo.timeout() where instead of creating a new client.

pymongo/pool.py
```diff
@@ -407,6 +411,32 @@ def _cond_wait(condition: threading.Condition, deadline: Optional[float]) -> boo
     return condition.wait(timeout)
 
 
+def _get_timeout_details(options: PoolOptions) -> dict[str, float]:
+    details = {}
+    timeout = _csot.get_timeout()
+    socket_timeout = options.socket_timeout
+    connect_timeout = options.connect_timeout
+    if timeout:
+        details["operationTimeoutMS"] = timeout * 1000
```
ShaneHarvey, line 420: Suggest remaining this to "timeoutMS" to match the URI option name. operationTimeoutMS doesn't appear anywhere else so this could be confusing.

test/test_pooling.py
```diff
@@ -411,6 +411,31 @@ def find_one():
...
+            self.assertTrue("(configured timeouts: socketTimeoutMS: 1.0ms)" in str(error.exception))
+
+    def test_connection_timeout_message(self):
+        client = rs_or_single_client("badhost", connectTimeoutMS=1, serverSelectionTimeoutMS=10)
+
+        with self.assertRaises(Exception) as error:
+            client.db.t.find_one({"$where": delay(2)})
+            self.assertTrue(
```
ShaneHarvey, line 435: Same here.

test/test_pooling.py
```diff
@@ -411,6 +411,31 @@ def find_one():
...
+
+    def test_socket_timeout_message(self):
+        client = rs_or_single_client(socketTimeoutMS=1)
+
+        with self.assertRaises(Exception) as error:
+            client.db.t.find_one({"$where": delay(2)})
+            self.assertTrue("(configured timeouts: socketTimeoutMS: 1.0ms)" in str(error.exception))
+
+    def test_connection_timeout_message(self):
+        client = rs_or_single_client("badhost", connectTimeoutMS=1, serverSelectionTimeoutMS=10)
```
ShaneHarvey, line 431: Might need to use a failpoint to fail connection creation, otherwise this will result in a ServerSelectionTimeoutError, not a pool checkout error.

ShaneHarvey (changes requested):

pymongo/pool.py
NoahStapp: Agreed it should match the URI option name. Is `timeoutMS` generally specific enough for users, or should we consider changing that option name to reflect its functionality better?

pymongo/pool.py
NoahStapp: Good catch. I'll add a check in `_raise_connection_failure` to ensure it doesn't add `timeout_details` if they're already in the message.

test/test_pooling.py
NoahStapp: Wouldn't failing connection creation result in a `ServerSelectionTimeoutError`? In this case we want to have the client successfully select the server and then timeout when it tries to execute a command, correct?

test/test_pooling.py
```diff
@@ -411,6 +411,41 @@ def find_one():
         # maxConnecting = unbounded: 30+ connections in ~0.140+ seconds
         print(len(pool.conns))
 
+    def test_csot_timeout_message(self):
+        with self.assertRaises(Exception) as error:
+            with timeout(0.001):
```
ShaneHarvey, line 416: We'll need to increase this to 500ms to make sure there's enough time to actually run the command (instead of timing out in server selection for example).

test/test_pooling.py
```diff
@@ -411,6 +411,41 @@ def find_one():
...
+                self.client.db.t.find_one({"$where": delay(1)})
+
+        self.assertTrue("(configured timeouts: timeoutMS: 1.0ms" in str(error.exception))
+
+    def test_socket_timeout_message(self):
+        client = rs_or_single_client(socketTimeoutMS=0.01)
```
ShaneHarvey, line 422:
```
___________________ TestPooling.test_socket_timeout_message ____________________

self = <test.test_pooling.TestPooling testMethod=test_socket_timeout_message>

    def test_socket_timeout_message(self):
        client = rs_or_single_client(socketTimeoutMS=0.01)
    
>       with self.assertRaises(Exception) as error:
E       AssertionError: Exception not ### raised
```

Not sure how to explain the above error but could you increase the timeout to 500ms to make sure there's enough time to actually run the command (instead of timing out in server selection for example).

test/test_pooling.py
```diff
@@ -411,6 +411,41 @@ def find_one():
...
+            client.db.t.find_one({"$where": delay(1)})
+
+        self.assertTrue("(configured timeouts: socketTimeoutMS: 0.01ms" in str(error.exception))
+
+    @client_context.require_failCommand_fail_point
+    def test_connection_timeout_message(self):
+        client = rs_or_single_client(connectTimeoutMS=1)
+
+        # Mock a connection failing due to timeout.
+        mock_connection_timeout = {
```
ShaneHarvey, line 434: This is failing on the find command, not the connection establishment. Is that expected?

ShaneHarvey (changes requested):

test/test_pooling.py
```diff
@@ -411,6 +412,48 @@ def find_one():
         # maxConnecting = unbounded: 30+ connections in ~0.140+ seconds
         print(len(pool.conns))
 
+    def test_csot_timeout_message(self):
+        with self.assertRaises(Exception) as error:
+            with timeout(0.5):
+                self.client.db.t.find_one({"$where": delay(2)})
```
ShaneHarvey, line 418: This is failing because the query ($where) will only be evaluated on non-empty collections. On missing/empty collections the server returns an empty result without executing the query. Same for test_socket_timeout_message below.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

test/test_pooling.py
ShaneHarvey: @NoahStapp ping.

ShaneHarvey: Oh I see test_connection_timeout_message does the right thing now. Resolving.

ShaneHarvey (approved):
Nice tests!

test/test_pooling.py
NoahStapp: That is outdated, the current code for `test_connection_timeout_message` correctly fails on the connection establishment.

PR #1376, opened 2023-09-15
PYTHON-3890 Data Lake Tests Were Failing Silently

NoahStapp:


.evergreen/config.yml
```diff
@@ -1687,6 +1687,7 @@ tasks:
         - func: "run tests"
           vars:
             TEST_DATA_LAKE: "true"
+            PYMONGO_MUST_CONNECT: "true"
```
ShaneHarvey, line 1690: This is an EVG var but does it get exported as a shell env var?

ShaneHarvey (changes requested):

.evergreen/config.yml
NoahStapp: Good point, I've made it so if PYMONGO_MUST_CONNECT is set in EG it will get pulled explicitly by test-eg.

ShaneHarvey:
Could you schedule the data lake tests? I suspect test.test_client_context.TestClientContext.test_must_connect won't run in that suite regardless of PYMONGO_MUST_CONNECT.

NoahStapp:
> Could you schedule the data lake tests? I suspect test.test_client_context.TestClientContext.test_must_connect won't run in that suite regardless of PYMONGO_MUST_CONNECT.

Correct, that test does not even with PYMONGO_MUST_CONNECT set.

ShaneHarvey:
So what we need to do here is not use PYMONGO_MUST_CONNECT at all and instead find out why this test was not protecting us like we'd expoect:
```python
class TestDataLakeMustConnect(IntegrationTest):
    def test_connected_to_data_lake(self):
        data_lake = os.environ.get("TEST_DATA_LAKE")
        if not data_lake:
            self.skipTest("TEST_DATA_LAKE is not set")

        self.assertTrue(
            client_context.is_data_lake,
            "client context.is_data_lake must be True when DATA_LAKE is set",
        )
```
https://github.com/mongodb/mongo-python-driver/blob/b8c13643689b7fd04eb249582e7f6bd72a04dd5f/test/test_data_lake.py#L35-L44

ShaneHarvey:
Perhaps we need to remove IntegrationTest there?

NoahStapp:
> Perhaps we need to remove IntegrationTest there?

Here's the patch with that change: https://spruce.mongodb.com/version/6504d71030661555703878ac/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I don't quite understand why we expect `test.test_client_context.TestClientContext.test_must_connect` to run in this suite, can you elaborate?

ShaneHarvey:
PYMONGO_MUST_CONNECT is only relevant for TestClientContext.test_must_connect but that test is never run when tests data lake. So setting PYMONGO_MUST_CONNECT does nothing for this PR. What we want to do is ensure TestDataLakeMustConnect fails when the driver cannot connect to data lake.

NoahStapp:
> PYMONGO_MUST_CONNECT



> PYMONGO_MUST_CONNECT is only relevant for TestClientContext.test_must_connect but that test is never run when tests data lake. So setting PYMONGO_MUST_CONNECT does nothing for this PR. What we want to do is ensure TestDataLakeMustConnect fails when the driver cannot connect to data lake.

Got it, added `client_context.connected` to `test_connected_to_data_lake`.

ShaneHarvey (approved):

PR #1379, opened 2023-09-19
PYTHON-3916 Deprecate Deprecated GridFS Fields, remove in 5.0

NoahStapp:


ShaneHarvey:
I think this needs more work. These deprecated properties should emit a DeprecationWarning on set/get.

PR #1380, opened 2023-09-25
PYTHON-3698 Support mypy 1.2 --strict testing (part 2)

blink1073:
The number of typing errors fluctuates as changes are made.  This gets us down to less than 100 for me locally.

### DIFF
--- a/pymongo/cursor.py
+++ b/pymongo/cursor.py
@@ -434,7 +434,6 @@ class Cursor(Generic[_DocumentType]):
             # Skip killCursors.
             cursor_id = 0
             address = None
-        assert self.__sock_mgr is not None
         self.__collection.database.client._cleanup_cursor(
             synchronous,
             cursor_id,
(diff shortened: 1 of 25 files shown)


pymongo/cursor.py
```diff
@@ -204,9 +204,9 @@ def __init__(
         # Initialize all attributes used in __del__ before possibly raising
         # an error to avoid attribute errors during garbage collection.
         self.__collection: Collection[_DocumentType] = collection
-        self.__id: Any = None
+        self.__id: int | None = None
```
NoahStapp, line 207: Use `Optional` instead?

pymongo/cursor.py
```diff
@@ -204,9 +204,9 @@ def __init__(
...
         self.__exhaust = False
-        self.__sock_mgr: Any = None
+        self.__sock_mgr: _ConnectionManager | None = None
```
NoahStapp, line 209: Use `Optional` instead?

NoahStapp (changes requested):
I can see the benefits of adding `--strict` support, but a lot of these changes negatively impact readability of the code. How much do we expect to gain from `--strict` support, and is it worth this trade-off?

PR #1382, opened 2023-10-02
PYTHON-3983 Clean up handling of PYTHON_BINARY

blink1073:


.evergreen/run-mongodb-oidc-test.sh
```diff
@@ -13,20 +13,21 @@ fi
 
 # Get the drivers secrets.  Use an existing secrets file first.
 if [ ! -f "./secrets-export.sh" ]; then
-    bash .evergreen/tox.sh -m aws-secrets -- drivers/oidc
```
NoahStapp, line 16: What's the rationale behind directly calling `setup_secrets.sh` instead of using tox?

NoahStapp (changes requested):

.evergreen/run-mongodb-oidc-test.sh
blink1073: Our tox env replicates the env requirements in DRIVERS_TOOLS, I think we should migrate away from it.

NoahStapp: If we migrate away from the tox env, what process do we envision for running these tests locally?

ShaneHarvey:
In the past we discovered that using venv was problematic on some platforms. Please run the *entire* test suite to make sure that's no longer the case.

.evergreen/run-mongodb-oidc-test.sh
NoahStapp: Good point! Can we add that step to the local testing documentation?

blink1073:
> In the past we discovered that using venv was problematic on some platforms. Please run the entire test suite to make sure that's no longer the case.

Good call out, so far macos does not [work](https://spruce.mongodb.com/task/mongo_python_driver_test_macos_arm64__platform~macos_1100_arm64_auth_ssl~noauth_nossl_test_latest_sharded_cluster_patch_0590ce49caef13ac50ed98707dbe834184dd4be9_651ab616a4cf4734cfa5a511_23_10_02_12_22_48/logs?execution=1&sortBy=STATUS&sortDir=ASC) yet with this approach.

blink1073:
MacOS 64 was the only newly failing platform, will investigate.

.evergreen/run-mongodb-oidc-test.sh
blink1073: I opened https://jira.mongodb.org/browse/PYTHON-3986

blink1073:
MacOS is [fixed](https://spruce.mongodb.com/version/651bef6657e85a9882f4efb9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), I'll run them all again once #1383 is merged to cut down on the number of failures.

NoahStapp (approved):

ShaneHarvey (approved):

PR #1383, opened 2023-10-03
PYTHON-3985 [Build Failure] TestCollectionChangeStream.test_split_large_change

blink1073:
I verified that the test passes in 6.0.x and 7.0.x and is skipped in 6.3.x.


blink1073:
The doctest was failing at teardown, so I removed gevent from the doctests, since it relies on being imported and patched first, which we can't guarantee.

Jibola (approved):

PR #1388, opened 2023-10-12
PYTHON-3991 Fix MockupDB tests when run through tox

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3991


blink1073:
I think the fix is to use `python -m pytest` in the `test` task.

NoahStapp (approved):
LGTM, good find!

NoahStapp:
> I think the fix is to use `python -m pytest` in the `test` task.

The MockupDB task needs the `test` extra to have pytest be present in the environment.

blink1073 (approved):

PR #1389, opened 2023-10-13
PYTHON-2878 Allow passing dict to sort/create_index/hint

NoahStapp:


blink1073:
Typing failures:

```
pymongo/helpers.py: note: In function "_index_document":
pymongo/helpers.py:142: error: No overload variant of "__getitem__" of
"Sequence" matches argument type
"Union[str, Tuple[str, Union[int, str, Mapping[str, Any]]]]"  [call-overload]
                value = index_list[key]
                        ^~~~~~~~~~~~~~~
pymongo/helpers.py:1[42](https://github.com/mongodb/mongo-python-driver/actions/runs/6511237574/job/17686536214?pr=1389#step:5:43): note: Possible overload variants:
pymongo/helpers.py:142: note:     def __getitem__(self, int, /) -> Union[str, Tuple[str, Union[int, str, Mapping[str, Any]]]]
pymongo/helpers.py:142: note:     def __getitem__(self, slice, /) -> Sequence[Union[str, Tuple[str, Union[int, str, Mapping[str, Any]]]]]
pymongo/helpers.py: note: In function "_validate_index_key_pair":
pymongo/helpers.py:155: error: Implicit return in function which does not
return  [misc]
    def _validate_index_key_pair(key: any, value: any) -> NoReturn:
    ^
pymongo/helpers.py:1[55](https://github.com/mongodb/mongo-python-driver/actions/runs/6511237574/job/17686536214?pr=1389#step:5:56): error: Function "builtins.any" is not valid as a type 
[valid-type]
    def _validate_index_key_pair(key: any, value: any) -> NoReturn:
                                      ^
pymongo/helpers.py:155: note: Perhaps you meant "typing.Any" instead of "any"?
```

blink1073:
Can you also please add a test that passes in a dictionary?

pymongo/helpers.py
```diff
@@ -111,7 +111,7 @@ def _index_list(
         if isinstance(key_or_list, str):
             return [(key_or_list, ASCENDING)]
         if isinstance(key_or_list, abc.ItemsView):
-            return list(key_or_list)
+            return list(key_or_list)  # type: ignore
```
blink1073, line 114: Can you please include the specific error code for these ignores?

test/test_client.py
```diff
@@ -1836,6 +1836,13 @@ def test_handshake_08_invalid_aws_ec2(self):
             None,
         )
 
+    def test_dict_hints(self):
+        c = rs_or_single_client()
+        try:
+            c.t.t.find(hint={"x": 1})
```
blink1073, line 1842: Can we also test `sort` and `create_index`?

blink1073 (changes requested):

blink1073 (approved):
Thank you!

test/test_client.py
```diff
@@ -1836,6 +1836,28 @@ def test_handshake_08_invalid_aws_ec2(self):
             None,
         )
 
+    def test_dict_hints(self):
+        c = rs_or_single_client()
```
ShaneHarvey, line 1840: These tests should use the global client instead of creating a new one since creating clients is expensive and slows down the test suite.

NoahStapp: Good catch, opened https://github.com/mongodb/mongo-python-driver/pull/1396 to fix.

PR #1395, opened 2023-10-16
PYTHON-3988 Python Test Suite still uses legacy isMaster Command when…

NoahStapp:
… loadbalanced=true

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -307,7 +307,10 @@ class ClientContext:
     @property
     def hello(self):
         if not self._hello:
-            self._hello = self.client.admin.command(HelloCompat.LEGACY_CMD)
+            if self.serverless:
+                self._hello = self.client.admin.command(HelloCompat.CMD)
+            else:
+                self._hello = self.client.admin.command(HelloCompat.LEGACY_CMD)
         return self._hello
 
     def _connect(self, host, port, **kwargs):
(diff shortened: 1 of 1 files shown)


test/__init__.py
```diff
@@ -317,7 +320,10 @@ def _connect(self, host, port, **kwargs):
         )
         try:
             try:
-                client.admin.command(HelloCompat.LEGACY_CMD)  # Can we connect?
```
ShaneHarvey, line 320: I believe we can use the "ping" command here instead.

test/__init__.py
```diff
@@ -307,7 +307,10 @@ def uri(self):
     @property
     def hello(self):
         if not self._hello:
-            self._hello = self.client.admin.command(HelloCompat.LEGACY_CMD)
+            if self.serverless:
```
ShaneHarvey, line 310: Suggest `if self.serverless or self.load_balancer:`

ShaneHarvey (changes requested):

test/__init__.py
```diff
@@ -317,7 +320,10 @@ def _connect(self, host, port, **kwargs):
         )
         try:
             try:
-                client.admin.command(HelloCompat.LEGACY_CMD)  # Can we connect?
+                if self.serverless:
+                    client.admin.command(HelloCompat.CMD)  # Can we connect?
+                else:
+                    client.admin.command("ping")  # Can we connect?
```
ShaneHarvey, line 326: We don't need to use the legacy command on serverless. We can use ping in all cases.

ShaneHarvey (changes requested):

PR #1396, opened 2023-10-16
PYTHON-2878 Allow passing dict to sort/create_index/hint performance …

NoahStapp:
…improvement

### DIFF
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -1837,26 +1837,16 @@ class TestClient(IntegrationTest):
         )
 
     def test_dict_hints(self):
-        c = self.client
-        try:
-            c.t.t.find(hint={"x": 1})
-        except Exception:
-            self.fail("passing a dictionary hint to find failed!")
+        self.db.t.find(hint={"x": 1})
 
     def test_dict_hints_sort(self):
-        c = self.client
-        try:
-            result = c.t.t.find()
-            result.sort({"x": 1})
-        except Exception:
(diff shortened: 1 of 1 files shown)


test/test_client.py
```diff
@@ -1837,22 +1837,22 @@ def test_handshake_08_invalid_aws_ec2(self):
         )
 
     def test_dict_hints(self):
-        c = rs_or_single_client()
+        c = self.client
```
ShaneHarvey, line 1840: Since we're at it can we clean these tests up a bit more? The try/except is redundant since any exception will case the test to fail anyway. I'd suggest:
```python
def test_dict_hints(self):
    self.db.t.find(hint={"x": 1})
...
```

test/test_client.py
```diff
@@ -1837,22 +1837,22 @@ def test_handshake_08_invalid_aws_ec2(self):
...
         try:
             c.t.t.find(hint={"x": 1})
         except Exception:
             self.fail("passing a dictionary hint to find failed!")
 
     def test_dict_hints_sort(self):
-        c = rs_or_single_client()
+        c = self.client
         try:
             result = c.t.t.find()
```
ShaneHarvey, line 1849: Can we always ensure it's possible to pass a dict sort to the find() call itself: `find(sort=...)`?

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM!

PR #1398, opened 2023-10-17
PYTHON-4002 No module named coverage.__main__

blink1073:
I also added coverage to the tasks that are run on PR commits


.evergreen/run-tests.sh
```diff
@@ -258,7 +258,8 @@ python -c 'import sys; print(sys.version)'
 # Only cover CPython. PyPy reports suspiciously low coverage.
 PYTHON_IMPL=$($PYTHON -c "import platform; print(platform.python_implementation())")
 if [ -n "$COVERAGE" ] && [ "$PYTHON_IMPL" = "CPython" ]; then
-    python -m pip install pytest-cov "coverage"
+    # coverage 7.3 dropped support for Python 3.7, keep in sync with combine-coverage.sh.
+    python -m pip install pytest-cov "coverage<7.3"
```
Jibola, line 262: Could we see a world where we have a version of python or dependent libraries that's incompatible with coverage<7.3? I don't imagine that's likely considering the python release schedules, and our support timelines but just curious.

.evergreen/combine-coverage.sh
```diff
@@ -0,0 +1,23 @@
...
+if [ -z "$PYTHON_BINARY" ]; then
+    PYTHON_BINARY=$(find_python3)
+fi
+
+createvirtualenv "$PYTHON_BINARY" covenv
+# coverage 7.3 dropped support for Python 3.7, keep in sync with run-tests.sh
+pip install -q "coverage<7.3"
+
+pip list
+ls -la coverage/
```
Jibola, line 20: Are these two just for capturing the information in our logs?

.evergreen/run-tests.sh
blink1073: Eventually, yes.   We'll do a `grep` for Python 3.7 when we drop support for it, and should update this min version.

PR #1399, opened 2023-10-18
PYTHON-4005 Replace flake8 and isort with ruff

blink1073:
I disabled many of the rules that would have caused extra work, especially in `tests`.  We can consider re-enabling them over time.  I used https://learn.scientific-python.org/development/guides/style/#ruff as the basis for the configuration.

It automatically added `from __future__ import annotations`, which accounts for the high file changed count.


pymongo/operations.py
```diff
@@ -547,7 +547,7 @@ def document(self) -> dict[str, Any]:
 class SearchIndexModel:
     """Represents a search index to create."""
 
-    __slots__ = "__document"
+    __slots__ = ("__document",)
```
blink1073, line 550: Found a bug!

pymongo/monitoring.py
```diff
@@ -1216,7 +1212,7 @@ class ServerClosedEvent(_ServerEvent):
 class TopologyEvent:
     """Base class for topology description events."""
 
-    __slots__ = "__topology_id"
+    __slots__ = ("__topology_id",)
```
blink1073, line 1215: Found a bug!

test/__init__.py
```diff
@@ -795,11 +796,12 @@ def is_topology_type(self, topologies):
             return True
         return False
 
-    def require_cluster_type(self, topologies=[]):  # noqa
+    def require_cluster_type(self, topologies=None):
```
blink1073, line 799: Found a bug!

doc/docs-requirements.txt
```diff
@@ -1,5 +1,5 @@
-sphinx>=5.3,<7
-sphinx_rtd_theme~=0.5
-readthedocs-sphinx-search~=0.1
+sphinx>=7,<8
```
ShaneHarvey, line 1: Is this related to ruff?

gridfs/grid_file.py
```diff
@@ -357,12 +357,14 @@ def write(self, data: Any) -> None:
         except AttributeError:
             # string
             if not isinstance(data, (str, bytes)):
-                raise TypeError("can only write strings or file-like objects")
+                raise TypeError("can only write strings or file-like objects") from None
```
ShaneHarvey, line 360: What are the pros of adding `from None`? Are there any cons?

test/__init__.py
ShaneHarvey: Was there a bug here? I don't see it. `topologies` is read only IIUC so the mutable default arg doesn't have and effect.

pymongo/monitoring.py
ShaneHarvey: Also not a bug.

pymongo/operations.py
ShaneHarvey: Not a bug :). `__slots__` can be assigned a single attr name or a tuple of names: https://docs.python.org/3/reference/datamodel.html#object.__slots__

bson/__init__.py
```diff
@@ -244,7 +244,7 @@ def _raise_unknown_type(element_type: int, element_name: str) -> NoReturn:
 
 
 def _get_int(
-    data: Any, view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
+    data: Any, _view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
```
ShaneHarvey, line 247: Why did ruff rename "view" -> "_view"? It seems out of place.

blink1073: I did that manually, `dummy` is explicitly part of `dummy-variable-rgx`.

pymongo/uri_parser.py
```diff
@@ -598,14 +598,14 @@ def _parse_kms_tls_options(kms_tls_options: Optional[Mapping[str, Any]]) -> dict
     if not isinstance(kms_tls_options, dict):
         raise TypeError("kms_tls_options must be a dict")
     contexts = {}
-    for provider, opts in kms_tls_options.items():
-        if not isinstance(opts, dict):
+    for provider, options in kms_tls_options.items():
```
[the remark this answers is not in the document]
blink1073: It was shadowing the function parameter

gridfs/grid_file.py
blink1073: It affects whether the previous error shows up in the traceback or not.

pymongo/operations.py
blink1073: Ah, shucks.

test/__init__.py
blink1073: Yeah fair, but nice to see that it catches this case.

ShaneHarvey (approved):
LGTM!

blink1073:
cc @charliermarsh 🚀

charliermarsh:
Awesome to see this, thanks @blink1073!

PR #1400, opened 2023-10-19
PYTHON-1424 Add repr to write result classes

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-1424

Examples:
```python
>>> client.t.t.insert_one({})
InsertOneResult(ObjectId('65319acdd55bb3a27ab5502b'), acknowledged=True)
>>> client.t.t.insert_many([{} for _ in range(3)])
InsertManyResult([ObjectId('65319af2d55bb3a27ab5502c'), ObjectId('65319af2d55bb3a27ab5502d'), ObjectId('65319af2d55bb3a27ab5502e')], acknowledged=True)
>>> client.t.t.delete_one({})
DeleteResult({'n': 1, 'ok': 1.0, '$clusterTime': {'clusterTime': Timestamp(1697749775, 3), 'signature': {'hash': b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 'keyId': 0}}, 'operationTime': Timestamp(1697749775, 3)}, acknowledged=True)
>>> client.t.t.update_one({}, {'$set': {'a': 1}})
UpdateResult({'nModified': 1, 'n': 1, 'ok': 1.0, '$clusterTime': {'clusterTime': Timestamp(1697749806, 2), 'signature': {'hash': b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 'keyId': 0}}, 'operationTime': Timestamp(1697749806, 2), 'updatedExisting': True}, acknowledged=True)
>>> client.t.t.bulk_write([InsertOne({})])
BulkWriteResult({'writeErrors': [], 'writeConcernErrors': [], 'nInserted': 1, 'nUpserted': 0, 'nMatched': 0, 'nModified': 0, 'nRemoved': 0, 'upserted': []}, acknowledged=True)
```

The only part that sticks out as a little odd is that UpdateResult and DeleteResult include the full command response with operationTime, etc... but I'm not sure we can change that due to backwards compatibility.

### DIFF
--- a/doc/examples/custom_type.rst
+++ b/doc/examples/custom_type.rst
@@ -135,7 +135,7 @@ Now, we can seamlessly encode and decode instances of
 .. doctest::
 
   >>> collection.insert_one({"num": Decimal("45.321")})
-  InsertOneResult(ObjectId('...'), True)
+  InsertOneResult(ObjectId('...'), acknowledged=True)
   >>> mydoc = collection.find_one()
   >>> import pprint
   >>> pprint.pprint(mydoc)
@@ -217,7 +217,7 @@ object, we can seamlessly encode instances of ``DecimalInt``:
   >>> collection = db.get_collection("test", codec_options=codec_options)
   >>> collection.drop()
   >>> collection.insert_one({"num": DecimalInt("45.321")})
-  InsertOneResult(ObjectId('...'), True)
+  InsertOneResult(ObjectId('...'), acknowledged=True)
   >>> mydoc = collection.find_one()
   >>> pprint.pprint(mydoc)
   {'_id': ObjectId('...'), 'num': Decimal('45.321')}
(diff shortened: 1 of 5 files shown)


blink1073:
> but I'm not sure we can change that due to backwards compatibility.

Why not remove them only for repr purposes?

Also, it might be good to show `acknowledged=True` instead of just `True`.

ShaneHarvey:
> Why not remove them only for repr purposes?

The idea behind repr is that it can be eval’d to get an identical object. Omitting fields would break that contract. 

> show acknowledged=True instead of just True.

Good idea!

blink1073 (approved):
LGTM!

PR #1402, opened 2023-10-20
PYTHON-3757 Bulk write operations don't consider hint field in equality checks

NoahStapp:


ShaneHarvey (commented):
Is this only a bug with DeleteOne or are there other operations we need to change?

test/test_crud_v1.py
```diff
@@ -185,9 +185,21 @@ def test_InsertOneNotEquals(self):
 
     def test_DeleteOneEquals(self):
         self.assertEqual(DeleteOne({"foo": 42}), DeleteOne({"foo": 42}))
+        self.assertEqual(DeleteOne({"foo": 42}, {"bar": 43}), DeleteOne({"foo": 42}, {"bar": 43}))
```
ShaneHarvey, line 188: Could you use a minimal collation like `{"locale": "en_US"}` instead of "bar"?

ShaneHarvey (changes requested):

NoahStapp:
> Is this only a bug with DeleteOne or are there other operations we need to change?

Good catch, I fixed DeleteMany as well!

ShaneHarvey (approved):
LGTM!

PR #1403, opened 2023-10-20
PYTHON-3747 Fix flaky test_list_databases

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3747

See also https://jira.mongodb.org/browse/PYTHON-3529 and https://jira.mongodb.org/browse/PYTHON-1532. Comparing the full document is flaky, instead let's just make sure they have the expected fields.


NoahStapp (approved):
LGTM!

PR #1404, opened 2023-10-20
PYTHON-3962 Make delimiting slash between hosts and options optional

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3962

Also added some changelog entries that were missing.

### DIFF
--- a/pymongo/uri_parser.py
+++ b/pymongo/uri_parser.py
@@ -513,7 +513,7 @@ def parse_uri(
     if path_part:
         dbase, _, opts = path_part.partition("?")
     else:
-        # No slash scheme_free, check for sole "?".
+        # There was no slash in scheme_free, check for a sole "?".
         host_part, _, opts = host_part.partition("?")
 
     if dbase:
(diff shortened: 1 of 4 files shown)


pymongo/uri_parser.py
```diff
@@ -506,22 +510,23 @@ def parse_uri(
...
-                dbase, collection = dbase.split(".", 1)
-            if _BAD_DB_CHARS.search(dbase):
-                raise InvalidURI('Bad database name "%s"' % dbase)
-        else:
-            dbase = None
+    else:
+        # There was no slash in scheme_free, check for a sole "?".
+        host_part, _, opts = host_part.partition("?")
+
+    if dbase:
```
NoahStapp, line 519: `dbase` will always be `None` for slashless URIs. Is that expected?

NoahStapp (changes requested):

pymongo/uri_parser.py
ShaneHarvey: Yes you can't specify a db without a slash. EG:
```
mongodb://example.com/admin?tls=True
```
without the slash is invalid:
```
mongodb://example.comadmin?tls=True
```

NoahStapp: Got it, looks good then!

NoahStapp (approved):

PR #1409, opened 2023-10-27
PYTHON-3968 Allow pymongo to be installed in tox when C extension fails to build

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-3968


NoahStapp (approved):
LGTM!

PR #1418, opened 2023-11-01
PYTHON-3837 Driver Container and Kubernetes Awareness

NoahStapp:
Based on the Go Driver implementation: https://github.com/mongodb/mongo-go-driver/commit/d8f19ab9cc105a06edcf186acb85f4e5656754a3

### DIFF
--- a/pymongo/pool.py
+++ b/pymongo/pool.py
@@ -25,6 +25,8 @@ import sys
 import threading
 import time
 import weakref
+from dataclasses import dataclass
+from pathlib import Path
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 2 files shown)


pymongo/pool.py
```diff
@@ -268,6 +270,34 @@ def _set_keepalive_times(sock: socket.socket) -> None:
...
+    runtime: Optional[str]
+    orchestrator: Optional[str]
+
+
+def get_container_env_info() -> Optional[ContainerInfo]:
+    """Returns the runtime and orchestrator of a container.
+    If neither value is present, the metadata client.env.container field will be omitted."""
+    runtime, orchestrator = None, None
+
+    if Path.exists(Path(DOCKER_ENV_PATH)):
```
ShaneHarvey, line 291: Suggest `if Path(DOCKER_ENV_PATH).exists():`

pymongo/pool.py
```diff
@@ -307,6 +337,13 @@ def _getenv_int(key: str) -> Optional[int]:
 
 def _metadata_env() -> dict[str, Any]:
     env: dict[str, Any] = {}
+    container = get_container_env_info()
+    if container is not None:
+        env["container"] = {}
+        if container.runtime:
+            env["container"]["runtime"] = container.runtime
+        if container.orchestrator:
+            env["container"]["orchestrator"] = container.orchestrator
```
ShaneHarvey, line 346: What do you think about inlining the get_container_env_info() logic and removing the ContainerInfo class? I believe it will make this code more straightforward to read.

test/test_client.py
```diff
@@ -347,6 +347,16 @@ def test_metadata(self):
         options = client._MongoClient__options
         self.assertEqual(options.pool_options.metadata, metadata)
 
+    def test_container_metadata(self):
+        os.environ[ENV_VAR_K8S] = "1"
```
ShaneHarvey, line 351: Let's use mock.patch.dict to edit the environ that way it gets correctly reset to the previous value.

ShaneHarvey (changes requested):

pymongo/pool.py
NoahStapp: I like the extensibility of having a separate `get_container_env_info` function + `ContainerInfo` class. If we add similar detection for other platforms, or want more verbose metadata for K8s or Docker, inlining will no longer be very readable.

ShaneHarvey: I think it's the opposite. The inlined version removes 20~ lines of code and is much easier to read:
```python
    container = {}
    if Path(DOCKER_ENV_PATH).exists():
        container["runtime"] = RUNTIME_NAME_DOCKER
    if os.getenv(ENV_VAR_K8S):
        container["orchestrator"] = ORCHESTRATOR_NAME_K8S
    if container:
        env["container"] = container
```

NoahStapp: It does now, but if we add further metadata like this inline we'll want to move it out for readability. I'm fine going with the inline version now and coming back to it if we add more in the future.

ShaneHarvey: What about a compromise where we just remove the dataclass and return the "container" dict itself?:
```python
container = get_container_env_info()
if container:
    env["container"] = container
```

NoahStapp: I like that suggestion, let's go with that.

ShaneHarvey (approved):

PR #1421, opened 2023-11-06
PYTHON-2834 Direct read/write retries to another mongos if possible

NoahStapp:


pymongo/mongo_client.py
```diff
@@ -2334,6 +2338,8 @@ def run(self) -> T:
                             raise
                         self._retrying = True
                         self._last_error = exc
+                        if self._client.topology_description.topology_type == TOPOLOGY_TYPE.Sharded:
```
blink1073, line 2341: I think we need to clear the deprioritized servers after all retry attempts (pass or fail).

NoahStapp: Good catch, thanks!

ShaneHarvey: I would also expect multiple failures to append to the deprioritized server list. What does the spec say?

NoahStapp: The spec seems to suggest that only a single retry will occur, but using a list of deprioritized servers suggests to me that multiple failures would result in multiple servers being deprioritized.

blink1073: Yes, I meant after the retry completes (whether it succeeds or not).

ShaneHarvey: _ClientConnectionRetryable is a temp object so I don't see a reason to clear the list.

blink1073: Ah, very nice!

test/utils.py
```diff
@@ -1153,3 +1153,9 @@ def prepare_spec_arguments(spec, arguments, opname, entity_map, with_txn_callbac
                 raise AssertionError(f"Unsupported cursorType: {cursor_type}")
         else:
             arguments[c2s] = arguments.pop(arg_name)
+
+
+def set_fail_point(client, command_args):
```
ShaneHarvey, line 1158: Can you use the [with self.fail_point()](https://github.com/mongodb/mongo-python-driver/blob/7936708d97b6bb77a0d6fe6c378f81594c34a7c5/test/__init__.py#L988) helper instead?

test/test_retryable_writes.py
```diff
@@ -469,6 +471,78 @@ def test_batch_splitting_retry_fails(self):
...
+            "configureFailPoint": "failCommand",
+            "mode": {"times": 1},
+            "data": {
+                "failCommands": ["insert"],
+                "closeConnection": True,
+                "appName": "retryableWriteTest",
+            },
+        }
+
+        for mongos in client_context.mongos_seeds().split(","):
```
ShaneHarvey, line 487: `for mongos in client_context.mongoses:`

test/test_retryable_writes.py
```diff
@@ -469,6 +471,78 @@ def test_batch_splitting_retry_fails(self):
...
+            client = rs_or_single_client(mongos, directConnection=True)
```
ShaneHarvey, line 488: directConnection is not needed

test/test_retryable_writes.py
```diff
@@ -469,6 +471,78 @@ def test_batch_splitting_retry_fails(self):
...
+            appName="retryableWriteTest",
+            event_listeners=[listener],
+            retryWrites=True,
+        )
+
+        with self.assertRaises(AutoReconnect):
+            client.t.t.insert_one({"x": 1})
+
+        # Disable failpoints on each mongos
+        for mongos in client_context.mongos_seeds().split(","):
```
ShaneHarvey, line 503: Rather than opening new clients to disable the failpoint, let's save the clients that were opened above and reuse them here. Finally we should close them by registering an addCleanup(client.close) when creating them.

test/test_retryable_writes.py
```diff
@@ -469,6 +471,78 @@ def test_batch_splitting_retry_fails(self):
...
+
+            fail_command["mode"] = "off"
+            set_fail_point(client, fail_command)
+
+        self.assertEqual(len(listener.failed_events), 2)
+        self.assertEqual(len(listener.succeeded_events), 0)
+
+    @client_context.require_multiple_mongoses
+    @client_context.require_failCommand_fail_point
+    def test_retryable_writes_in_sharded_cluster_one_available(self):
```
ShaneHarvey, line 514: This test is not needed. We already run this scenario 100's of times because this is this default test suite behavior.

ShaneHarvey (changes requested):

test/test_retryable_writes.py
NoahStapp: That change causes the following error: 

```
h = ('localhost', 27017)

    def _connection_string(h):
>       if h.startswith(("mongodb://", "mongodb+srv://")):
E       AttributeError: 'tuple' object has no attribute 'startswith'

test/utils.py:544: AttributeError
```

test/utils.py
NoahStapp: We need to set the fail point on each individual mongos to ensure the failure occurs on every node. `with self.fail_point()` does not set the fail point correctly.

test/test_retryable_writes.py
NoahStapp: We run the scenario of a mongos cluster with a single available node retrying in other tests?

ShaneHarvey: Yes, all of other sharded tests use a single mongos by default.

NoahStapp: Got it, sounds good!

pymongo/mongo_client.py
```diff
@@ -2334,6 +2338,8 @@ def run(self) -> T:
...
+                            self._deprioritized_servers.append(self._server)
```
ShaneHarvey, line 2342: Since this logic is the same for reads and writes and we do it in one place?

ShaneHarvey (changes requested):

pymongo/mongo_client.py
NoahStapp: Good catch, thanks!

pymongo/topology.py
```diff
@@ -282,11 +282,13 @@ def _select_server(
         selector: Callable[[Selection], Selection],
         server_selection_timeout: Optional[float] = None,
         address: Optional[_Address] = None,
+        deprioritized_servers: Optional[list[Server]] = None,
     ) -> Server:
         servers = self.select_servers(selector, server_selection_timeout, address)
-        if len(servers) == 1:
-            return servers[0]
-        server1, server2 = random.sample(servers, 2)
+        filtered_servers = _filter_servers(servers, deprioritized_servers)
```
ShaneHarvey, line 288: Can we rename filtered_servers->servers  so the rest of the code doesn't have to change?

ShaneHarvey (commented):
One more minor thing, otherwise LGTM

blink1073 (approved):
LGTM!

ShaneHarvey (approved):

PR #1425, opened 2023-11-08
PYTHON-4038: Ensure retryable read `OperationFailure`s re-raise exception when 0 or NoneType error code is provided. 

Jibola:
TICKET: https://jira.mongodb.org/browse/PYTHON-4038

Pymongo is retrying a failed find connection ONCE that has a 0 or NoneType error code. 

This is undesired behavior because if, theoretically, we introduce an error type that does not have an error code but is an instance of an `OperationFailure`, we will mistakenly retry leading to unexpected consequences.

Added 3 tests: 
- Test nonetype case
- Test 0 case 
- Test 1 case

### DIFF
--- a/pymongo/mongo_client.py
+++ b/pymongo/mongo_client.py
@@ -2329,7 +2329,8 @@ class _ClientConnectionRetryable(Generic[T]):
                         # ConnectionFailures do not supply a code property
                         exc_code = getattr(exc, "code", None)
                         if self._is_not_eligible_for_retry() or (
-                            exc_code and exc_code not in helpers._RETRYABLE_ERROR_CODES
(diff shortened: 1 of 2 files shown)


Jibola:
I need to add a test case into PyMongo to ensure this doesn't recur

NoahStapp (approved):

ShaneHarvey (changes requested):
Still needs a mockupdb test where the server returns a missing or 0 error code. 

Also could you update the PR title and changelog to describe the bug that's being fixed?

ShaneHarvey (approved):

NoahStapp (approved):

PR #1426, opened 2023-11-08
PYTHON-2884: Replaced SON usage in all internal classes and commands

Jibola:
## Callouts 
* Pool.metadata is a _somewhat_ public method that is now returning `SON` instead of `dict`. I would love perspective on if this could be considered a breaking conversion.
* There are two test cases that now needed changing. Commentary needed on whether or not these are breaking changes.


blink1073:
> I would love perspective on if this could be considered a breaking conversion.

Yeah I think to be safe we should leave it as SON until 5.0.  It looks like some tests need to be updated as well.

ShaneHarvey:
Pool is a private API so there's no compat concerns there,

blink1073:
> Pool is a private API so there are no compat concerns.

Ah, true, I missed that `MongoClient.__options` is private, I saw `options.pool_options.metadata` in a grep.

ShaneHarvey:
Oh I didn't realize it was PoolOptions. That is indeed public:
```python
>>> client.options.pool_options.metadata
SON([('driver', SON([('name', 'PyMongo'), ('version', '4.7.0.dev0')])), ('os', SON([('type', 'Darwin'), ('name', 'Darwin'), ('architecture', 'arm64'), ('version', '13.6')])), ('platform', 'CPython 3.11.4.final.0')])
```

blink1073:
Looks like two more tests need to be addressed:

```
FAILED test/test_collection.py::TestCollection::test_find_command_generation
FAILED test/test_cursor.py::TestCursor::test_clone - AssertionError: False is...
```

blink1073 (approved):
LGTM!

ShaneHarvey:
Please wait for my review before merging.

pymongo/message.py
```diff
@@ -399,7 +394,7 @@ def as_command(
         # Support auto encryption
         client = self.client
         if client._encrypter and not client._encrypter._bypass_auto_encryption:
-            cmd = cast(SON[str, Any], client._encrypter.encrypt(self.db, cmd, self.codec_options))
+            cmd = cast(dict[str, Any], client._encrypter.encrypt(self.db, cmd, self.codec_options))
```
ShaneHarvey, line 397: Let's change _encrypter.encrypt's return type to dict so we can remove the redundant casts here.

pymongo/pool.py
```diff
@@ -268,6 +269,25 @@ def _set_keepalive_times(sock: socket.socket) -> None:
...
+    """Returns the runtime and orchestrator of a container.
+    If neither value is present, the metadata client.env.container field will be omitted."""
+    container = {}
+
+    if Path(DOCKER_ENV_PATH).exists():
+        container["runtime"] = RUNTIME_NAME_DOCKER
+    if os.getenv(ENV_VAR_K8S):
+        container["orchestrator"] = ORCHESTRATOR_NAME_K8S
+
+    return container
```
ShaneHarvey, line 289: Odd that this shows up in this PR. Did you `git merge master`?

test/test_collection.py
```diff
@@ -2125,7 +2125,7 @@ def test_find_command_generation(self):
             None,
         )
         self.assertEqual(
-            cmd.to_dict(), SON([("find", "coll"), ("$dumb", 2), ("filter", {"foo": 1})]).to_dict()
+            cmd, SON([("find", "coll"), ("$dumb", 2), ("filter", {"foo": 1})]).to_dict()
```
ShaneHarvey, line 2128: Hmm, this highlights how this is a subtle breaking change since the command payloads published in CommandStartedEvent will change from SON to dict. Code relying on SON will break when it changes to dict.

We should at least document in the changelog and mark the jira ticket as a minor breaking change.

ShaneHarvey (changes requested):

test/test_collection.py
ShaneHarvey: Also since this is the case I revise my earlier opinion. I think it's fine to change PoolOptions.metadata type.

pymongo/encryption.py
```diff
@@ -23,6 +23,7 @@
 from typing import (
     TYPE_CHECKING,
     Any,
+    Dict,
```
NoahStapp, line 26: Should this be lowercase `dict` for consistency across type hints?

test/test_collection.py
NoahStapp: If we're removing SON entirely, this test case should compare against an actual `dict`, not a converted `SON`.

NoahStapp (changes requested):
A few minor comments, otherwise looks great, thanks Jib! Also, if we're removing `SON` entirely, can we also delete the `bson.son` file?

Jibola:
> A few minor comments, otherwise looks great, thanks Jib! Also, if we're removing `SON` entirely, can we also delete the `bson.son` file?

Ah, I don't think we're removing`SON` entirely, we're just deprecating our internal use of it. IIUC, SON is still perfectly viable, but outright removal would be a major version breaking change.

pymongo/encryption.py
Jibola: This was the error I ran into, which left me just to leave it. 
```
typecheck: commands[1]> mypy --install-types --non-interactive bson gridfs tools pymongo
pymongo/encryption.py:86: error: "dict" is not subscriptable  [misc]
        document_class=dict[str, Any], uuid_representation=STANDARD
                       ^
Found 1 error in 1 file (checked 86 source files)
```

NoahStapp: We're importing future `__annotations__`, so I would expect that to work, hmm.

NoahStapp (approved):
Looks good! I'd prefer to figure out the `Dict` vs `dict` issue in `encryption.py`, but otherwise LGTM!

pymongo/encryption.py
```diff
@@ -83,9 +83,8 @@
 _KMS_CONNECT_TIMEOUT = CONNECT_TIMEOUT  # CDRIVER-3262 redefined this value to CONNECT_TIMEOUT
 _MONGOCRYPTD_TIMEOUT_MS = 10000
 
-
-_DATA_KEY_OPTS: CodecOptions[SON[str, Any]] = CodecOptions(
-    document_class=SON[str, Any], uuid_representation=STANDARD
+_DATA_KEY_OPTS: CodecOptions[dict[str, Any]] = CodecOptions(
```
blink1073, line 86: This `Dict` is a runtime object, that's what the annotations import doesn't help.

NoahStapp: Ahhh right, we're passing it directly as a keyword argument.

doc/changelog.rst
```diff
@@ -11,6 +11,8 @@ PyMongo 4.7 brings a number of improvements including:
   :attr:`pymongo.monitoring.CommandSucceededEvent.server_connection_id`, and
   :attr:`pymongo.monitoring.CommandFailedEvent.server_connection_id` properties.
 - Fixed a bug where inflating a :class:`~bson.raw_bson.RawBSONDocument` containing a :class:`~bson.code.Code` would cause an error.
+- Replaced usage of :class:`bson.son.SON` on all internal classes and commands to dict,
```
ShaneHarvey, line 14: Let's call this out more prominently under an "Unavoidable breaking changes" section like we do here: https://github.com/mongodb/mongo-python-driver/blob/3d63eedb/doc/changelog.rst?plain=1#L291

It's also best practice to show how this breaking change could show up and what to do about it. Including example code that expects `CommandStartedEvent.command` to be SON would be good.

ShaneHarvey: https://github.com/mongodb/mongo-python-driver/pull/1498

PR #1430, opened 2023-11-15
PYTHON-3457 & PYTHON-3458 - Standardized debugging logging for command monitoring 

NoahStapp:


NoahStapp:
The `test_Unacknowledged` tests for the `unified` `crud` tests required the change to `publish_command_start`.

ShaneHarvey:
Could you link to the test you're referring to? I did't find it.

NoahStapp:
> Could you link to the test you're referring to? I did't find it.

Here's one: https://github.com/mongodb/mongo-python-driver/blob/134b7efaec4cf2b613d3cec44d541d88635a33fa/test/crud/unified/bulkWrite-updateMany-hint-unacknowledged.json#L179

ShaneHarvey:
That test already exists and is already passing though.

NoahStapp:
> That test already exists and is already passing though.

You're right, I just tested it in master and it passes. Must be a bug in the logging code or something, I'll investigate.

pymongo/network.py
```diff
@@ -162,15 +164,34 @@ def command(
 
     if max_bson_size is not None and size > max_bson_size + message._COMMAND_OVERHEAD:
         message._raise_document_too_large(name, size, max_bson_size + message._COMMAND_OVERHEAD)
-
+    encoding_duration = datetime.datetime.now() - start
+    command_logger = logging.getLogger("pymongo.command")
+    # TODO: add serverConnectionId
+    if name == "insert":
+        assert True
```
Jibola, line 171:
```suggestion
```

pymongo/network.py
```diff
@@ -162,15 +164,34 @@ def command(
 
     if max_bson_size is not None and size > max_bson_size + message._COMMAND_OVERHEAD:
         message._raise_document_too_large(name, size, max_bson_size + message._COMMAND_OVERHEAD)
-
+    encoding_duration = datetime.datetime.now() - start
+    command_logger = logging.getLogger("pymongo.command")
+    # TODO: add serverConnectionId
```
Jibola, line 169: Is this a TODO for this PR or a TODO to be deferred for another? If it's for another, we can make a JIRA ticket to track and place it here.

Jibola (changes requested):
Great code! I've added some nits and other comments. I've pinched it off now as I see there's still some changes you're making, but great work thus far! 🥳

pymongo/network.py
NoahStapp: Whoops missed these comments: I'll be opening https://jira.mongodb.org/browse/PYTHON-2822 shortly to resolve this.

NoahStapp: Opened: https://github.com/mongodb/mongo-python-driver/pull/1438

pymongo/logger.py
```diff
@@ -0,0 +1,64 @@
...
+        for doc in _documents:
+            if doc in kwargs:
+                if doc == "failure" and kwargs.pop("isServerSideError", False):
+                    kwargs[doc] = {
+                        k: v for k, v in kwargs[doc].items() if k in REDACTED_FAILURE_FIELDS
+                    }
+                is_speculative_authenticate = (
+                    kwargs.pop("speculative_authenticate", False)
+                    or "speculativeAuthenticate" in kwargs[doc]
+                )
```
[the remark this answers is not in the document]
NoahStapp: `isServerSideError` and `speculative_authenticate` are not required by the spec to be present in the log message, and are only used for defining conditional logging behavior. We could leave them in, but it's cleaner to remove all unnecessary fields from the log messages.

pymongo/logger.py
```diff
@@ -0,0 +1,64 @@
...
+                    and ("commandName" in kwargs and kwargs["commandName"] in SENSITIVE_COMMANDS)
+                    or (kwargs["commandName"] in HELLO_COMMANDS and is_speculative_authenticate)
+                ):
+                    kwargs[doc] = json_util.dumps({})
+                else:
+                    kwargs[doc] = json_util.dumps(
+                        kwargs[doc], json_options=json_options, default=lambda o: o.__repr__()
+                    )
+                if len(kwargs[doc]) > document_length:
+                    kwargs[doc] = kwargs[doc][:document_length] + "..."
```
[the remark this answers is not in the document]
NoahStapp: The spec specifies that the `...` does not count towards the max document length: https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#configurable-max-document-length.

pymongo/message.py
```diff
@@ -1007,33 +1009,83 @@ def unack_write(
         msg: bytes,
         max_doc_size: int,
         docs: list[Mapping[str, Any]],
+        client: MongoClient,
     ) -> Optional[Mapping[str, Any]]:
         """A proxy for Connection.unack_write that handles event publishing."""
+        command_logger = logging.getLogger("pymongo.command")
```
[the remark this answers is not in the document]
NoahStapp: Can you expand on this?

pymongo/logger.py
```diff
@@ -0,0 +1,64 @@
...
+                    }
+                is_speculative_authenticate = (
+                    kwargs.pop("speculative_authenticate", False)
+                    or "speculativeAuthenticate" in kwargs[doc]
+                )
+                if (
+                    doc != "failure"
+                    and ("commandName" in kwargs and kwargs["commandName"] in SENSITIVE_COMMANDS)
+                    or (kwargs["commandName"] in HELLO_COMMANDS and is_speculative_authenticate)
+                ):
```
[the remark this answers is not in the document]
NoahStapp: Good idea!

pymongo/message.py
```diff
@@ -1007,33 +1009,83 @@ def unack_write(
...
+        duration = datetime.datetime.now() - self.start_time
```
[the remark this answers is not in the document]
NoahStapp: Yeah, I think we can remove this and just have duration set once as `datetime.datetime.now() - self.start_time` in the try/except block below.

pymongo/message.py
```diff
@@ -1007,33 +1009,83 @@ def unack_write(
...
+                driverConnectionId=self.conn.id,
+                serverHost=self.conn.address[0],
+                serverPort=self.conn.address[1],
+                serviceId=self.conn.service_id,
+            )
+        )
         if self.publish:
-            assert self.start_time is not None
-            duration = datetime.datetime.now() - self.start_time
             cmd = self._start(cmd, request_id, docs)
```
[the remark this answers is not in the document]
NoahStapp: The spec requires that `durationMS` include the time spent sending the message to the server and reading the reply. Since we only need to report `durationMS` in  `Command Succeeded` logs, I don't think we need this.

pymongo/logger.py
```diff
@@ -0,0 +1,64 @@
...
+HELLO_COMMANDS = ["hello", "ismaster", "isMaster"]
+
+REDACTED_FAILURE_FIELDS = ["code", "codeName", "errorLabels"]
+
+
+class StructuredMessage:
+    def __init__(self, **kwargs: Any):
+        document_length = int(os.getenv("MONGOB_LOG_MAX_DOCUMENT_LENGTH", DEFAULT_DOCUMENT_LENGTH))
+        if document_length < 0:
+            document_length = DEFAULT_DOCUMENT_LENGTH
```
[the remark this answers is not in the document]
NoahStapp: `test_configured_truncation_limit` caught this: since the max document length is configurable, and could be configured between different logging calls, we need to get `document_length` on every call of `LogMessage`.

pymongo/message.py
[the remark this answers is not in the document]
Jibola: Sure: 
```
def __init__(self, *args, **kwargs):
    ...
    `self._logger = logging.getLogger("pymongo.command")`
...
def _some_fn(self):
    self._logger.debug({...})
```

test/unified_format.py
```diff
@@ -1024,6 +1032,9 @@ def maybe_skip_test(self, spec):
             if "timeoutMode" in op.get("arguments", {}):
                 self.skipTest("PyMongo does not support timeoutMode")
 
+        # if spec["description"] != "A failed command":
+        #     self.skipTest("for now...")
+
```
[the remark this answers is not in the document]
Jibola: We can make a TODO for this just to eliminate lingering code comments.

test/test_logger.py
```diff
@@ -0,0 +1,54 @@
...
+
+                cmd_succeeded_log = json_util.loads(cm.records[1].message)
+                self.assertLessEqual(len(cmd_succeeded_log["reply"]), 5 + 3)
+                with self.assertRaises(OperationFailure):
+                    db.command({"notARealCommand": True})
+                cmd_failed_log = json_util.loads(cm.records[-1].message)
+                self.assertEqual(len(cmd_failed_log["failure"]), 5 + 3)
+
+    # def test_truncation_multi_byte_codepoints(self):
+    #     ...
```
Jibola, line 50: Similarly, we can make a TODO: for this so that it doesn't exist as a lingering code comment.

test/unified_format.py
```diff
@@ -1650,6 +1661,45 @@ def check_events(self, spec):
             else:
                 assert server_connection_id is None
 
+    def check_log_messages(self, operations, spec):
+        def format_logs(log_list):
+            client_to_log = defaultdict(list)
+            for log in log_list:
+                data = json_util.loads(log.message)
+                client = data["clientId"]
+                data.pop("clientId")
```
Jibola, line 1670:
```suggestion
                client = data.pop("clientId")
```

test/unified_format.py
```diff
@@ -1650,6 +1661,45 @@ def check_events(self, spec):
...
+                client_to_log[client].append(
+                    {
+                        "level": log.levelname.lower(),
+                        "component": log.name.strip("pymongo."),
```
Jibola, line 1674: TIL strip doesn't do _exactly_ what I thought it does. Let's opt for using something more explicit like `replace` since strip can cause unexpected behavior.
See: https://docs.astral.sh/ruff/rules/strip-with-multi-characters/

Here's a "real world" example: 
```
>>> "pymongo.hello".strip("pymongo.")
'hell'
```


```suggestion
                        "component": log.name.replace("pymongo.", "", 1),
```

pyproject.toml
```diff
@@ -185,6 +185,7 @@ ignore = [
   "EM103",  # Exception must not use a `.format()` string directly, assign to variable first
   "C408",   # Unnecessary `dict` call (rewrite as a literal)
   "SIM117", # Use a single `with` statement with multiple contexts instead of nested `with` statements
+  "B005",   # Using `.strip()` with multi-character strings is misleading
```
Jibola, line 188: How do you feel if we kept this error as one that gets raised and made the change in the test provided above?

```suggestion
```

Jibola (approved):
Great Work and thanks for taking this on! Really looking forward to having nice and standardized logs coming through.

Jibola (changes requested):
Whoops. Applied the wrong review status 😅

ShaneHarvey:
Could you schedule the perf task to see if the logging lines add any noticeable overhead?

pymongo/message.py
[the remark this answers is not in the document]
Jibola: NIT: Not sure if you still wanted to go with adding this.

Jibola (approved):
LGTM!

.evergreen/run-tests.sh
```diff
@@ -30,7 +30,7 @@ set -o xtrace
 
 AUTH=${AUTH:-noauth}
 SSL=${SSL:-nossl}
-TEST_ARGS="$1"
+TEST_ARGS=$*
```
Jibola, line 33: Is this an intended inclusion in your change?

NoahStapp:
> NIT: Not sure if you still wanted to go with adding this.

Sorry, what was the comment here? Github won't let me see what you're referring to.

pymongo/logger.py
```diff
@@ -0,0 +1,91 @@
+from __future__ import annotations
```
ShaneHarvey, line 1: Missing the standard LICENSE comment we add to all files.

test/test_logger.py
```diff
@@ -0,0 +1,71 @@
+from __future__ import annotations
```
ShaneHarvey, line 1: Missing the standard LICENSE comment we add to all files.

test/unified_format.py
```diff
@@ -992,6 +998,8 @@ def maybe_skip_test(self, spec):
             self.skipTest("Implement PYTHON-1894")
         if "timeoutMS applied to entire download" in spec["description"]:
             self.skipTest("PyMongo's open_download_stream does not cap the stream's lifetime")
+        if "command log messages include server connection id" == spec["description"]:
+            self.skipTest("TODO: Implement https://jira.mongodb.org/browse/PYTHON-2822")
```
ShaneHarvey, line 1002: This is fixed now. Could you open a new ticket unless you plan to address in this PR?

pymongo/logger.py
```diff
@@ -0,0 +1,91 @@
...
+
+import enum
+import os
+from typing import Any
+
+from bson import UuidRepresentation, json_util
+from bson.json_util import JSONOptions
+
+
+class LogMessageStatus(str, enum.Enum):
```
ShaneHarvey, line 11: Let's prefix all globals in this file with `_` to avoid people mistakenly thinking this is public.

pymongo/logger.py
```diff
@@ -0,0 +1,91 @@
...
+    def _process_documents(self) -> None:
+        document_length = int(os.getenv("MONGOB_LOG_MAX_DOCUMENT_LENGTH", DEFAULT_DOCUMENT_LENGTH))
+        if document_length < 0:
+            document_length = DEFAULT_DOCUMENT_LENGTH
+        is_server_side_error = self.kwargs.pop("isServerSideError", False)
+
+        for doc in _DOCUMENTS:
+            if doc in self.kwargs:
+                if doc == "failure" and is_server_side_error:
+                    self.kwargs[doc] = {
```
ShaneHarvey, line 75: suggest refactoring `self.kwargs[doc]` to avoid repeated lookups and renaming _DOCUMENTS to _REDACTED_KEYS (or something else). "doc" usually refers to a document, not a field name:
```pyhton
for key in _REDACTED_KEYS:
    doc = self.kwargs.get(key)
    if doc:...
```

pymongo/logger.py
```diff
@@ -0,0 +1,91 @@
...
+_REDACTED_FAILURE_FIELDS = ["code", "codeName", "errorLabels"]
+_DOCUMENTS = ["command", "reply", "failure"]
+_JSON_OPTIONS = JSONOptions(uuid_representation=UuidRepresentation.STANDARD)
+
+
+class LogMessage:
+    def __init__(self, **kwargs: Any):
+        self.kwargs = kwargs
+
+        self._process_documents()
```
ShaneHarvey, line 39: This should be delayed to __str__ for performance reasons. No need to go through all the effort of redacting info that will never be serialized.

pymongo/logger.py
```diff
@@ -0,0 +1,91 @@
...
+            self.kwargs["durationMS"] = self.kwargs["durationMS"].total_seconds() * 1000
+        if "serviceId" in self.kwargs and self.kwargs["serviceId"] is None:
+            del self.kwargs["serviceId"]
+
+    def __str__(self) -> str:
+        return "%s" % (
+            json_util.dumps(self.kwargs, json_options=_JSON_OPTIONS, default=lambda o: o.__repr__())
+        )
+
+    def _is_sensitive(self, doc: str) -> bool:
```
ShaneHarvey, line 51: same comment about "doc" here.

pymongo/logger.py
```diff
@@ -0,0 +1,91 @@
...
+            "commandName" in self.kwargs and self.kwargs["commandName"] in _SENSITIVE_COMMANDS
+        )
+
+        is_sensitive_hello = (
+            self.kwargs["commandName"] in _HELLO_COMMANDS and is_speculative_authenticate
+        )
+
+        return is_sensitive_command or is_sensitive_hello
+
+    def _process_documents(self) -> None:
```
ShaneHarvey, line 66: suggest renaming to "redact"

pymongo/logger.py
```diff
@@ -0,0 +1,91 @@
...
+    "copydbsaslstart",
+    "copydb",
+]
+_HELLO_COMMANDS = ["hello", "ismaster", "isMaster"]
+_REDACTED_FAILURE_FIELDS = ["code", "codeName", "errorLabels"]
+_DOCUMENTS = ["command", "reply", "failure"]
+_JSON_OPTIONS = JSONOptions(uuid_representation=UuidRepresentation.STANDARD)
+
+
+class LogMessage:
```
ShaneHarvey, line 35: Suggest adding `__slots__` for better perf.

ShaneHarvey (changes requested):
> Could you schedule the perf task to see if the logging lines add any noticeable overhead?

Still interested in this piece.

pymongo/logger.py
NoahStapp: Is `__slots__` still useful if there are no public-facing methods? The intended use of `LogMessage` is to have it call `__str__` only.

ShaneHarvey: Yes it will reduce the memory footprint of instantiating the LogMessage class.

NoahStapp: Got it, so just make all the class methods slots?

ShaneHarvey: Not quite. slots is for the properties (self.kwargs in this case), not methods. Check out how we use slots on other classes like DatetimeMS: https://github.com/mongodb/mongo-python-driver/blob/6537415da70c2a15cad52ebe7669673947ace65a/bson/datetime_ms.py#L42

NoahStapp:
@ShaneHarvey perf task logs: https://evergreen.mongodb.com/task_log_raw/mongo_python_driver_perf_tests_perf_6.0_standalone_patch_9d32a09e30aec795ffb381f4fc3f827f780e662c_6578ca2132f4177fd5fa9cc3_23_12_12_21_01_22/0?type=T.

ShaneHarvey:
Did you see the 50-99% perf regression across the board? https://spruce.mongodb.com/task/mongo_python_driver_perf_tests_perf_6.0_standalone_patch_9d32a09e30aec795ffb381f4fc3f827f780e662c_6578ca2132f4177fd5fa9cc3_23_12_12_21_01_22/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC

Screenshot 2023-12-13 at 3 23 50 PM

We can't merge until those issues are fixed.

PR #1434, opened 2023-11-16
PYTHON-3745 Add support for multiphase initialization

blink1073:


NoahStapp (approved):

ShaneHarvey (changes requested):
Should we set https://docs.python.org/3/c-api/module.html#c.Py_MOD_MULTIPLE_INTERPRETERS_SUPPORTED on 3.12+? IIUC that's what we support. The new work would be to set Py_MOD_PER_INTERPRETER_GIL_SUPPORTED.

blink1073:
> Without Py_MOD_MULTIPLE_INTERPRETERS_SUPPORTED I think this change could break some users.

Fixed

blink1073:
mod_wsgi 4.x Python 3.7 Ubuntu 22.04 is failing intermittently, both with and without the Py_MOD_MULTIPLE_INTERPRETERS_SUPPORTED block.   `mod_wsgi` is a bit of a black box, should we skip that particular configuration and wait until we can more directly test multiple interpreters?

ShaneHarvey:
> mod_wsgi 4.x Python 3.7 Ubuntu 22.04 is failing intermittently

Is it also failing on main or only after the multiphase changes?

blink1073:
After the multiphase changes

ShaneHarvey:
I suggest we open a mod_wsgi issue to ask if multiphase init modules are supported: https://github.com/GrahamDumpleton/mod_wsgi

blink1073:
It's the interpreter itself that defines compat as I understand it, the feature should be usable from Python 3.5.  I propose we close this issue and re-open when we can test directly against subinterpreters to see if there is an actual bug.

ShaneHarvey:
I think it could be a bug in mod_wsgi because it has custom code to create subinterpreters. We'll also still need to keep the mod_wsgi tests even once we can test subinterpreters directly so I think we should try investigating the mod_wsgi bug now to avoid delaying this change.

blink1073:
I'm going to have to backlog this for now until I finish OIDC work.

ShaneHarvey:
Closing in favor of https://github.com/mongodb/mongo-python-driver/pull/1499

PR #1436, opened 2023-11-16
PYTHON-3823 Merge perf task into main project

ShaneHarvey:
[PYTHON-3823](https://jira.mongodb.org/browse/PYTHON-3823). This PR merges the  perf task into main project. I've disabled the mongo-python-driver-perf project. This makes it easier to test and track perf changes. For example, we can now schedule the perf job on the EVG patch build in a pull request rather than manually creating another patch on a separate project. Most other drivers use this model.

This change also updates to clone the test data from the spec repo and changes the benchmark to report megabytes per second instead of bytes per second to comply with the spec: 
> In addition to timing data, all micro-benchmark tasks will be measured in terms of "megabytes/second" (MB/s) of documents processed, with higher scores being better. (In this document, "megabyte" refers to the SI decimal unit, i.e. 1,000,000 bytes.) This makes cross-benchmark comparisons easier.

https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.rst#measurement

This the PR task here: https://spruce.mongodb.com/task/mongo_python_driver_perf_tests_perf_6.0_standalone_patch_2ac7f0d88d6856f6fc2ba88567627851e528956d_6556880932f417c7eef470d2_23_11_16_21_22_19/logs?execution=0


blink1073 (approved):
LGTM!

ShaneHarvey:
Didn't see your LGTM before I pushed other changes. I reverted the unreviewed changes and I'll on a new PR.

PR #1437, opened 2023-11-17
PYTHON-4049 Use MyST Parser for Narrative Docs

blink1073:
Notes:
- I converted our changelog and the encryption example page, with no hand edits.
- We need to use explicit ref targets instead of implicit refs based on header names, which is more robust anyway.  See change in `doc/installation.rst`.
- Links to API docs are working as expected.  When copying from docstrings, we would replace the `:class:` with `{class}` (or other directive), and update the link styles from RST to Markdown as appropriate, or we could have a tox job that takes an input file and uses RST-to-Myst to convert it for us.


ShaneHarvey:
I still don't think we should do this:
- we'll no longer be able to copy/paste between our docs and the docstrings
- we'll need to context switch between two formats instead of one
- myst style markdown render just as poorly in github as rst so there's no benefit there
- we need a new dependency

PR #1438, opened 2023-11-17
PYTHON-2822 Add server connectionId to command monitoring events

NoahStapp:


pymongo/hello.py
```diff
@@ -218,3 +218,7 @@ def service_id(self) -> Optional[ObjectId]:
     @property
     def hello_ok(self) -> bool:
         return self._doc.get("helloOk", False)
+
+    @property
+    def connection_id(self) -> Optional[ObjectId]:
```
ShaneHarvey, line 223: We already have a [public `connection_id` concept](https://pymongo.readthedocs.io/en/stable/api/pymongo/monitoring.html#pymongo.monitoring.ConnectionCheckedOutEvent.connection_id) so unfortunately I think we need to rename this. Perhaps `server_connection_id`?

pymongo/monitoring.py
```diff
@@ -615,6 +625,14 @@ def database_name(self) -> str:
         """
         return self.__db
 
+    @property
+    def server_connection_id(self) -> Optional[ObjectId]:
+        """The server-side connection id for the connection this command was sent to, or ``None``.
```
ShaneHarvey, line 630: sent to -> sent on

pymongo/monitoring.py
```diff
@@ -615,6 +625,14 @@ def database_name(self) -> str:
         """
         return self.__db
 
+    @property
+    def server_connection_id(self) -> Optional[ObjectId]:
```
ShaneHarvey, line 629: ObjectId? I thought connectionId was an integer?

ShaneHarvey (changes requested):

pymongo/hello.py
ShaneHarvey: Oh I didn't realize this was on Hello. Ignore this comment, connection_id is fine here.

pymongo/monitoring.py
NoahStapp: Good catch. The typecheck must have missed it because the tests use `assertIsInstance` instead of a `mypy`-friendly typecheck.

Jibola (approved):
LGTM

pymongo/monitoring.py
```diff
@@ -1494,6 +1523,7 @@ def publish_command_start(
         connection_id: _Address,
         op_id: Optional[int] = None,
         service_id: Optional[ObjectId] = None,
+        server_connection_id: Optional[int] = None,
```
ShaneHarvey, line 1526: If the server connection id is always passed here then perhaps we should make it a required argument? Same for the succeeded/failed methods.

NoahStapp: `server_connection_id` was added in MongoDB 4.2. Do we still need to support versions older than that? If so, we need to keep this optional.

ShaneHarvey: I meant that we are still always passing it to these methods even when server_connection_id is None. So we should consider removing the default value since the default is never used. Now that I'm looking at it the same is true for service_id and database_name so I'll open a new issue for this.

ShaneHarvey (commented):
Could you update the changelog to mention all the new apis?

NoahStapp:
> Could you update the changelog to mention all the new apis?

Do we mention internal code changes in the changelog? None of the changes will effect user behavior/APIs.

ShaneHarvey:
No we don't mention internal api changes but Hello and the command events are public, not internal.

ShaneHarvey (changes requested):

doc/changelog.rst
```diff
@@ -4,6 +4,16 @@ Changelog
 Changes in Version 4.6.1
```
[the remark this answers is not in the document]
NoahStapp: Oh we remove the in-between releases in the changelog?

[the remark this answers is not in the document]
ShaneHarvey: Nope, look at the file again. There are two 4.6.1 headers.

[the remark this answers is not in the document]
NoahStapp: Oh I see lol thanks

PR #1440, opened 2023-11-20
[PYTHON-3036] Improve error message for unknown MongoClient options

caseyclements:
https://jira.mongodb.org/browse/PYTHON-3036

When an unknown kwarg was passed to MongoClient, the ConfigurationError wasn't helpful. I use the standard lib package `difflib` to look for similar known options. It appears to work pretty well. 

First contribution to this mongodb. This PR / ticket doubles as SDLC tutorial. I've got a feature branch with a couple tests that I'll now use.

It will now look something like this:

```
ConfigurationError('Unknown option: auth. Did you mean: authsource, authmechanism, authoidcallowedhosts?')
```

### DIFF
--- a/pymongo/common.py
+++ b/pymongo/common.py
@@ -820,7 +820,7 @@ def validate_auth_option(option: str, value: Any) -> tuple[str, Any]:
 
 def _get_validator(
     key: str, validators: dict[str, Callable[[Any, Any], Any]], normed_key: Optional[str] = None
-):
+) -> Callable:
     normed_key = normed_key or key
     try:
         return validators[normed_key]


test/test_client.py
```diff
@@ -535,6 +535,18 @@ def test_client_options(self):
...
         self.assertIsInstance(c.options.retry_reads, bool)
 
+    def test_validate_suggestion(self):
+        """Validate kwargs in constructor."""
+        self.assertRaises(ConfigurationError, MongoClient, auth='standard')
+
+        try:
+            MongoClient(auth='standard')
+        except ConfigurationError as exc:
+            expected = "Unknown option: auth. Did you mean: authsource, authmechanism, authoidcallowedhosts?"
```
ShaneHarvey, line 545: Would it be possible to change this to:
```
Did you mean: authSource, authMechanism, ...?"
```

ShaneHarvey (changes requested):

test/test_client.py
caseyclements: Upon investigation, updating all the cases of the piles and piles of keywords in these dicts, finding the matching ones, etc etc isn't worth it. It's a lot of work, and it's brittle. 
I have a much simpler proposal. I've added a note to my error message about camelCase. What do you think?
ConfigurationError('Unknown option: auth. Did you mean one of (authsource, authmechanism, authoidcallowedhosts) **or maybe a camelCase version of one?'**)

test/test_client.py
```diff
@@ -535,6 +535,16 @@ def test_client_options(self):
         self.assertIsInstance(c.options.retry_writes, bool)
         self.assertIsInstance(c.options.retry_reads, bool)
 
+    def test_validate_suggestion(self):
+        """Validate kwargs in constructor."""
+        self.assertRaises(ConfigurationError, MongoClient, auth="standard")
```
ShaneHarvey, line 540: We can replace this two asserts with one that uses `with self.assertRaisesRegex(...)`, for example: https://github.com/mongodb/mongo-python-driver/blob/6c88c732194335c8ec8e9c2c6fbd8c81d34151fc/test/test_client.py#L237-L238

test/test_client.py
ShaneHarvey: Okay that's fair. Although I suspect there is a non-brittle way to do it but it would require more work than needed for this PR. Could you open this idea as a new PYTHON ticket so we can reevaluate later on?

pymongo/common.py
```diff
@@ -817,9 +821,13 @@ def validate_auth_option(option: str, value: Any) -> tuple[str, Any]:
...
-    validator = VALIDATORS.get(lower, raise_config_error)
-    value = validator(option, value)
-    return option, value
+    try:
+        validator = VALIDATORS[lower]
+        value = validator(option, value)
+        return option, value
+    except KeyError:
+        suggestions = get_close_matches(lower, VALIDATORS, cutoff=0.2)
+        raise_config_error(option, suggestions)
```
ShaneHarvey, line 830: What do you think about refactoring this with the code block below? It seems like they are essentially identical besides the VALIDATORS vs URI_OPTIONS_VALIDATOR_MAP.

pymongo/common.py
```diff
@@ -817,9 +821,13 @@ def validate_auth_option(option: str, value: Any) -> tuple[str, Any]:
 def validate(option: str, value: Any) -> tuple[str, Any]:
     """Generic validation function."""
     lower = option.lower()
-    validator = VALIDATORS.get(lower, raise_config_error)
-    value = validator(option, value)
-    return option, value
+    try:
+        validator = VALIDATORS[lower]
+        value = validator(option, value)
```
ShaneHarvey, line 826: Let's move the validator() call to after the try/except so that we only ever handle KeyErrors from the dict look up.

test/test_client.py
```diff
@@ -535,6 +535,16 @@ def test_client_options(self):
...
+
+        try:
+            MongoClient(auth="standard")
```
ShaneHarvey, line 543: Could you also add a test for a camelCase option to ensure we report the original non-lowercased name?

ShaneHarvey (changes requested):

pymongo/common.py
```diff
@@ -817,9 +821,13 @@ def validate_auth_option(option: str, value: Any) -> tuple[str, Any]:
...
+        return option, value
+    except KeyError:
+        suggestions = get_close_matches(lower, VALIDATORS, cutoff=0.2)
```
ShaneHarvey, line 829: Also this is neat! I hadn't seen difflib or get_close_matches used before. Did you try various values for cutoff? How did you end up with 0.2?

pymongo/common.py
caseyclements: I've added a note /hint on camelCase and suggested referring to docstring.
I've also opened a ticket to explore rationalizing the keys to kwargs: https://jira.mongodb.org/browse/PYTHON-4056

caseyclements (commented):
@ShaneHarvey When you've got a moment, let's review my review :)

test/test_client.py
ShaneHarvey: Still todo.

test/test_client.py
ShaneHarvey: still todo.

pymongo/common.py
ShaneHarvey: Still todo.

pymongo/common.py
ShaneHarvey: Sounds good.

test/test_uri_parser.py
```diff
@@ -144,6 +144,13 @@ def test_split_options(self):
         self.assertEqual({"authsource": "foobar"}, split_options("authSource=foobar"))
         self.assertEqual({"maxpoolsize": 50}, split_options("maxpoolsize=50"))
 
+        # Test suggestions given when invalid kwarg passed
+        try:
```
ShaneHarvey, line 148: Should use assertRaisesRegex here.

ShaneHarvey (changes requested):

pymongo/common.py
caseyclements: Hi. If the original intent hasn't been resolved, would you please elaborate?

ShaneHarvey: Can you refactor this to remove the duplicated code so that there's only one call to try/except + get_close_matches/raise_config_error, eg:
```python
def _get_validator(name, validators):
    lower = name.lower()
    try:
        return validators[lower]
    except KeyError:
        suggestions = get_close_matches(lower, validators, cutoff=0.2)
        raise_config_error(name, suggestions)
```

pymongo/common.py
```diff
@@ -856,15 +864,14 @@ def get_setter_key(x: str) -> str:
     for opt, value in options.items():
         normed_key = get_normed_key(opt)
         try:
-            validator = URI_OPTIONS_VALIDATOR_MAP.get(normed_key, raise_config_error)
```
ShaneHarvey, line 859: This code should still be using URI_OPTIONS_VALIDATOR_MAP but by calling validate() it now uses VALIDATORS. This is why I suggested the `_get_validator` function above.

ShaneHarvey (changes requested):

pymongo/common.py
```diff
@@ -810,14 +814,24 @@ def validate_auth_option(option: str, value: Any) -> tuple[str, Any]:
...
     lower, value = validate(option, value)
     if lower not in _AUTH_OPTIONS:
-        raise ConfigurationError(f"Unknown authentication option: {option}")
+        raise ConfigurationError(f"Unknown option: {option}. Must be in {_AUTH_OPTIONS}")
     return option, value
 
 
+def _get_validator(
+    key: str, validators: dict[str, Callable[[Any, Any], Any]], normed_key: Optional[str] = None
+):
```
ShaneHarvey, line 823: Typecheck is failing due to the missing return type here.

ShaneHarvey (approved):

PR #1441, opened 2023-11-21
PYTHON-3605 Move type annotations to parameter list in rendered docs

blink1073:
Leaving as draft until https://github.com/mongodb/mongo-python-driver/pull/1432 is merged so I can update `CONTRIBUTING.md` to reference https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html.

I used https://gist.github.com/blink1073/c5c174f64eaec3f118c4829cef05884a to reformat the docstrings.


ShaneHarvey:
Nice! Does this give us a path to removing the inline argument types in the docs?

like this one:
```
decode(codec_options: CodecOptions[Any] = CodecOptions(document_class=dict, tz_aware=False, uuid_representation=UuidRepresentation.UNSPECIFIED, unicode_decode_error_handler='strict', tzinfo=None, type_registry=TypeRegistry(type_codecs=[], fallback_encoder=None), datetime_conversion=DatetimeConversion.DATETIME)) → dict[str, Any]
```

blink1073:
Already part of this ticket!  See https://pymongo--1441.org.readthedocs.build/en/1441/api/bson/index.html#bson.decode

blink1073:
Hmm, some of the APIs are doing that in this PR, but not that one, I'll investigate.  The original ticket had that feature as its main goal.

blink1073:
Ah, it might be the presence of `@overload`.

ShaneHarvey:
Ah yes, the overload on codec_options=None piece, I like it.

blink1073:
Because https://pymongo--1441.org.readthedocs.build/en/1441/api/pymongo/database.html#pymongo.database.Database.command has the same issue, but not https://pymongo--1441.org.readthedocs.build/en/1441/api/pymongo/database.html#pymongo.database.Database.cursor_command

ShaneHarvey:
I think it would be good to keep this PR as just changing the params and open a new PR if we want to add more overloads.

blink1073:
Note that some of the overloads will go away with [PYTHON-3992](https://jira.mongodb.org/browse/PYTHON-3992)

blink1073:
>  if we want to add more overloads.

This PR doesn't add new `@overloads`, it just looks like `autodoc_typehints` doesn't work with `@overload`

PR #1445, opened 2023-11-30
PYTHON-4055 Add xunit-results to .gitignore

caseyclements:
One line change to `.gitignore`. xunit-results/ and test/xunit-results/ can be safely ignored. I imagine everyone has some hack to ignore them already. It's nice to be able to `git add .` :)


ShaneHarvey:
Could you add the other files mentioned in the ticket?
> Additional ones appearing are: 
> - .eggs/
> - .vscode/
> - server.log

PR #1446, opened 2023-12-01
PYTHON-3963 Migrate CSFLE Tests to AWS Secrets Vault

NoahStapp:


blink1073:
Some ideas:

I don't think we want to tie all these things together into a single script/tox env.   We'd want to leave the server running and libmongocrypt in place between runs.

What I'd recommend is:

- Rename `run-encryption-tests.sh` to `start-kmip-server.sh` and have it write out a `.pid` file, and call that script from EVG to remove the redundancy.
- Move the logic for handling the secrets file and AWS env variables into `run-tests.sh`.
- Remove the local script and `encryption-tests` tox entry.
- Add tox entries for `setup-encryption` and `teardown-encryption`.
  - Setup encryption starts the kmip server and fetches the secrets
  - Teardown encryption stops the kmip server and removes libmongocrypt dir
- Add an entry in `CONTRIBUTING.md` for encryption testing telling you to:
  - Run `AWS_PROFILE=<foo> tox -e setup-encryption`
  - Run your tests as `TEST_ENCRYPTION=1 tox -e test-eg` as desired.
  - Run `tox -e teardown-encryption` when done.

NoahStapp:
I like the approach of separating out the different components, just a few clarifying questions.

> * Rename `run-encryption-tests.sh` to `start-kmip-server.sh` and have it write out a `.pid` file, and call that script from EVG to remove the redundancy.

The existing redundancy of starting the `kmip` servers is due to what seems to be an EG bug that prevents the `kmip` server startup from working if `run-encryption-tests-sh.` is used. Ideally this redundancy is eliminated, but I've been unable to determine the cause.

> * Move the logic for handling the secrets file and AWS env variables into `run-tests.sh`.

Is our eventual goal to have `run-tests.sh` contain the bulk of the existing EG `config.yml` logic?

> * Add an entry in `CONTRIBUTING.md` for encryption testing telling you to:
>   * Run `AWS_PROFILE=<foo> tox -e setup-encryption`
>   * Run your tests as `TEST_ENCRYPTION=1 tox -e test-eg` as desired.
>   * Run `tox -e teardown-encryption` when done.

I like all of these, but I think there's still a lot of value to be had in a single command that does all the setup, tests, and teardown without requiring manual user intervention.

.evergreen/start-kmip-server.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# The -u options forces the stdout and stderr streams to be unbuffered.
+# TMPDIR is required to avoid "AF_UNIX path too long" errors.
+./activate-kmstlsvenv.sh
+TMPDIR="$(dirname "$DRIVERS_TOOLS")" python -u kms_kmip_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/server.pem  --port 5698 &
+python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/expired.pem --port 8000 &
+python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/wrong-host.pem --port 8001 &
+python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/server.pem --port 8002 --require_client_cert &
+
+pgrep -f "kms_...._server.py" > kmip_pids.pid
```
ShaneHarvey, line 16: Does this pgrep command work on all the hosts we test on?

tox.ini
```diff
@@ -175,3 +175,43 @@ deps =
...
+    {[testenv:test-eg]deps}
+    {[testenv:setup-encryption]deps}
+extras =
+    {[testenv:test-eg]extras}
+allowlist_externals =
+    {[testenv:test-eg]allowlist_externals}
+    {[testenv:teardown-encryption]allowlist_externals}
+passenv = *
+setenv =
+    LIBMONGOCRYPT_URL = https://s3.amazonaws.com/mciuploads/libmongocrypt/macos/master/latest/libmongocrypt.tar.gz
```
ShaneHarvey, line 210: This would be the first time we hard code macos into tox. Can we avoid this?

.evergreen/run-tests.sh
```diff
@@ -172,6 +172,7 @@ if [ -n "$TEST_ENCRYPTION" ]; then
 
     # Get access to the AWS temporary credentials:
     # CSFLE_AWS_TEMP_ACCESS_KEY_ID, CSFLE_AWS_TEMP_SECRET_ACCESS_KEY, CSFLE_AWS_TEMP_SESSION_TOKEN
+    export AWS_SESSION_TOKEN=
```
ShaneHarvey, line 175: Is this a bug fix?

.evergreen/run-encryption-tests-local.sh
```diff
@@ -0,0 +1,16 @@
...
+export AWS_ACCESS_KEY_ID=$FLE_AWS_KEY
+export AWS_SECRET_ACCESS_KEY=$FLE_AWS_SECRET
+export AWS_DEFAULT_REGION=us-east-1
+export AWS_SESSION_TOKEN=
+
+tox -e encryption-tests
+
+rm -rf libmongocrypt*
+# shellcheck disable=SC2046
+kill -9 $(pgrep -f 'python')
```
ShaneHarvey, line 16: We can't kill all python processes since that would disrupt other things running on the system.

.evergreen/cleanup-tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Clean up CSFLE kmip servers
+cd "$DRIVERS_TOOLS"/.evergreen/csfle || exit
+./activate-kmstlsvenv.sh
```
ShaneHarvey, line 5: Do we need to activate the venv here?

.evergreen/cleanup-tests.sh
```diff
@@ -0,0 +1,9 @@
...
+
+if [ -f "kmip_pids.pid" ]; then
+  < kmip_pids.pid xargs kill -9
```
ShaneHarvey, line 8: Should we delete the kmip_pids.pid file here too?

.evergreen/cleanup-tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
ShaneHarvey, line 1: Suggest renaming this file to "stop-kmip-server.sh" to mirror "start-kmip-server.sh".

ShaneHarvey (changes requested):

.evergreen/start-kmip-server.sh
NoahStapp: I don't see any errors in my latest patches, so it seems so.

tox.ini
NoahStapp: I think the best move is to just remove this tox env entirely, as we need a different `LIBMONGOCRYPT_URL` for each platform.

.evergreen/run-tests.sh
NoahStapp: Previously we did this as part of the encryption env setup block. I moved it here for clarity since it's required before the next line.

.evergreen/run-encryption-tests-local.sh
NoahStapp: Whoops. this script was for testing, included it by mistake.

.evergreen/cleanup-tests.sh
NoahStapp: No, good catch thanks.

NoahStapp:
Doc failure is for `http://api.mongodb.com/python/`, is that a known failure?

.evergreen/start-kmip-server.sh
```diff
@@ -0,0 +1,23 @@
...
+. ./activate-kmstlsvenv.sh
+
+# The -u options forces the stdout and stderr streams to be unbuffered.
+# TMPDIR is required to avoid "AF_UNIX path too long" errors.
+TMPDIR="$(dirname "$DRIVERS_TOOLS")" python -u kms_kmip_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/server.pem  --port 5698 &
+python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/expired.pem --port 8000 &
+python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/wrong-host.pem --port 8001 &
+python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/server.pem --port 8002 --require_client_cert &
+
+pgrep -f "kms_...._server.py" > kmip_pids.pid
```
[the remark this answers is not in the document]
blink1073: Ah, good!

blink1073 (changes requested):

PR #1454, opened 2023-12-07
PYTHON-4079 Generate ManyLinux2014 and ManyLinux1 Wheels

blink1073:


blink1073:
The docs failure is unrelated, due to a DNS outage.

ShaneHarvey:
In EVG we also build the manylinux1 wheels, can we add those as well?

blink1073:
> If manylinux1 wheels are no longer needed then could you update the changelog to mention we stopped releasing them?

I'd prefer that route since we don't support Python 3.6 anymore, should I also remove them from the EVG script?

ShaneHarvey:
Sounds good to me. This comment from Oct 2021 shows 3.7 has 8% manylinux1 usage and 2% manylinux2010 usage so it seems fine to drop them now: https://github.com/pypa/manylinux/issues/994#issuecomment-983058889

doc/changelog.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Changes in Version 4.8
```
ShaneHarvey, line 4: The next version is 4.7 which already has an entry below.

ShaneHarvey (changes requested):

doc/changelog.rst
blink1073: Lol, fixed.

.evergreen/build-manylinux.sh
```diff
@@ -6,24 +6,20 @@ docker version
...
 # the last releases to generate pip < 20.3 compatible wheels. After that
 # auditwheel was upgraded to v4 which produces PEP 600 manylinux_x_y wheels
 # which requires pip >= 20.3. We use the older docker image to support older
 # pip versions.
 BUILD_WITH_TAG="$1"
 if [ -n "$BUILD_WITH_TAG" ]; then
-  images=(quay.io/pypa/manylinux1_x86_64:2021-05-05-b64d921 \
-          quay.io/pypa/manylinux1_i686:2021-05-05-b64d921 \
-          quay.io/pypa/manylinux2014_x86_64:2021-05-05-1ac6ef3 \
+  images=(quay.io/pypa/manylinux2014_x86_64:2021-05-05-1ac6ef3 \
```
caseyclements, line 16: Is the removal of i686 unrelated?

ShaneHarvey (approved):

ShaneHarvey:
Should we rename the PR to "Remove manylinux1 wheels"?

.evergreen/build-manylinux.sh
caseyclements: If I understand correctly, manylinux1 wheels are Centos5. Is that what we were using before, and this upgrades to Centos7? We can't move to manylinux_x_y because it requires the new pip solver. Please respond if I've gotten any of this wrong.

I think I understand what's going on and why. Looks good!

caseyclements (approved):

ShaneHarvey:
I guess I also have one question: why not just keep manylinux1 until it becomes a maintenance burden? It seems like it doesn't cost us anything to keep building and publishing them. And it would benefit users on older distros.

.evergreen/build-manylinux.sh
blink1073: Both x86 and i686 were manylinux1, but we're going to try and bring both back.  Yes, this is related so supporting older versions of pip that come preinstalled on stock Centos 7.  See the compat table here: https://github.com/pypa/manylinux.  If we bump to manylinux2014, then we'd still support CentOS 7 rh-python38, but not stock CentOS 7.

blink1073:
> why not just keep manylinux1 until it becomes a maintenance burden

Looks like that can work:

- pymongo-4.7.0.dev0-cp39-cp39-manylinux_2_5_x86_64.manylinux1_x86_64.whl
- pymongo-4.7.0.dev0-cp39-cp39-manylinux_2_5_i686.manylinux1_i686.whl

caseyclements:
@blink1073 All looks good. Two questions.

1.  Do wheels names like this mean that they're cross-compiled: pymongo-4.7.0.dev0-cp39-cp39-manylinux_2_5_x86_64.manylinux1_x86_64.whl

2.  test-skip = "*-macosx_universal2:arm64" Does this mean that we build the wheel for apple silicon, but don't test it?

ShaneHarvey (approved):

blink1073:
> Do wheels names like this mean that they're cross-compiled: pymongo-4.7.0.dev0-cp39-cp39-manylinux_2_5_x86_64.manylinux1_x86_64.whl

It means the symbols used support both of those manylinux targets.

> Does this mean that we build the wheel for apple silicon, but don't test it?

Yes, apple silicon is not yet available on GitHub Actions

PR #1455, opened 2023-12-07
PYTHON-1366 About DocumentTooLarge - insert_one case

caseyclements:
This PR handles the first case of https://jira.mongodb.org/browse/PYTHON-1366 - insert_one(large_doc)
We no longer raise an Operation error if the document is larger than the document size limit but under the command size limit.

### DIFF
--- a/test/test_encryption.py
+++ b/test/test_encryption.py
@@ -74,6 +74,7 @@ from pymongo.errors import (
     AutoReconnect,
     BulkWriteError,
     ConfigurationError,
+    DocumentTooLarge,
     DuplicateKeyError,
     EncryptedCollectionError,
     EncryptionError,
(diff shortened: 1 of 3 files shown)


test/test_encryption.py
```diff
@@ -1232,16 +1233,13 @@ def test_06_insert_fails_over_16MiB(self):
...
 
-        with self.assertRaisesRegex(WriteError, "object to insert too large"):
+        err_msg = "the connected server supports BSON document sizes up to 16777216 bytes."
+        with self.assertRaisesRegex(DocumentTooLarge, expected_regex=err_msg):
             self.coll_encrypted.insert_one(doc)
 
         # Same with bulk_write.
-        doc["_id"] = "encryption_exceeds_16mib_bulk"
-        with self.assertRaises(BulkWriteError) as ctx:
+        with self.assertRaisesRegex(DocumentTooLarge, expected_regex=err_msg):
```
ShaneHarvey, line 1241: Hmm this is surprising. Why does encrypted bulk_write raise DocumentTooLarge now but not regular bulk_write?

ShaneHarvey: That doesn’t answer my question though.

ShaneHarvey:
The link check failure is unrelated, the docs team is looking into it.

test/test_encryption.py
caseyclements: Where is the test for the regular bulk_write?

caseyclements: Or somewhere in `test_bulk.py`?

PR #1457, opened 2023-12-10
PYTHON-4085 - First commit. Adds insert as special case, alongside un…

caseyclements:
checks individual docs are less than max_bson_size, simply by adding insert as special case, alongside unacknowledged writes, that checks individual docs in addition to command size.

This means all of the following sizes of "large" now throw DocumentTooLarge(BSONError). Previously 3 different sizes produced 3 different exceptions. Exception handling for clients will be significantly easier. These now all raise DocumentTooLarge.

```
    large = {"large": "1" * 1024 * 1024 * 16}  # max_bson_size < doc < max_command_size
    xlarge = {"xlarge": "1" * 1024 * 1024 * 18}  #  max_command_size < doc
    xxl = {"xxl": "1" * 1024 * 1024 * 48}  # max_message_size < doc
    
    CASE 2B: coll.insert_many([large] where large in [L, XL, XXL])
insert_many([large])type(eDocDict16)=<class 'pymongo.errors.BulkWriteError'>. errmsg: object to insert too large. size in bytes: 16777250, max size: 16777216
insert_many([xlarge]): type(eDocDict18)=<class 'pymongo.errors.OperationFailure'>. errmsg: BSONObj size: 18874403 (0x1200023) is invalid. Size must be between 0 and 16793600(16MB) First element: _id: ObjectId('657546bce7630c6acf9cdd53')
insert_many([xxl]): type(eDocDict48)=<class 'pymongo.errors.DocumentTooLarge'>. errmsg: BSON document too large (50331680 bytes) - the connected server supports BSON document sizes up to 16777216 bytes.
```

### DIFF
--- a/pymongo/message.py
+++ b/pymongo/message.py
@@ -1174,7 +1174,7 @@ def _raise_document_too_large(operation: str, doc_size: int, max_size: int) -> N
     """Internal helper for raising DocumentTooLarge."""
     if operation == "insert":
         raise DocumentTooLarge(
-            "BSON document too large (%d bytes)"
+            "BSON document too large (%d bytes) "
             " - the connected server supports"
             " BSON document sizes up to %d"
             " bytes." % (doc_size, max_size)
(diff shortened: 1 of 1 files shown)


caseyclements:
Closing to wait for 5.0

PR #1460, opened 2023-12-14
PYTHON-1374 Optimize json_util performance

NoahStapp:
Summary of changes:

- Use `_type_marker` to detect BSON types and move into separate helper function.
- Remove SON from `json_util`, use plain `dict` instead.
- Reorder cases in `default` to optimize for more common/faster cases.


bson/json_util.py
```diff
@@ -869,9 +845,31 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
...
+        return {"$maxKey": 1}
+    elif type_marker == 17:  # Timestamp
+        return {"$timestamp": {"t": obj.time, "i": obj.inc}}
+    elif type_marker == 13:  # Code
+        if obj.scope is None:
+            return {"$code": str(obj)}
+        return {"$code": str(obj), "$scope": _json_convert(obj.scope, json_options)}
+    elif type_marker == 5:  # Binary
+        return _encode_binary(obj, obj.subtype, json_options)
+    elif type_marker == 19:  # Decimal128
```
ShaneHarvey, line 862: Nice work! I suspect this could be made even faster by switching to a lookup table, something like:
```python
_encoders = {
    19: lambda obj: {"$numberDecimal": str(obj)},
    ...
}
...

    return _encoders.get(type_marker, raise_type_error)(obj)
```

I believe this is part of what Bernie was suggesting in the ticket.

ShaneHarvey (commented):
Looking good so far! Does the existing JSON benchmark cover these improvements? We may want to add some fine grained benchmarks to track these improvements more closely.

NoahStapp:
Latest perf run is still not great: https://spruce.mongodb.com/task/mongo_python_driver_perf_tests_perf_6.0_standalone_patch_9d32a09e30aec795ffb381f4fc3f827f780e662c_657ca9ec0ae60651e7c3a4b6_23_12_15_19_33_09/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC

NoahStapp:
> Before merging this it would be great to pull in the JSON micro benchmarks added in [PYTHON-4125](https://jira.mongodb.org/browse/PYTHON-4125) to see the improvement.

Pulled in and ran a benchmark: https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz?comparison_id=821845e0-c46b-4521-8c7d-82e187a6757c&selected_tab=data-table&percent_filter=0%7C%7C100&z_filter=0%7C%7C10.

Most of the JSON benchmarks don't have a data point to compare to in the mainline commits, but I manually compared them and this branch is even for decoding tests and around twice as fast for encoding tests.

ShaneHarvey:
> Most of the JSON benchmarks don't have a data point to compare to in the mainline commits

There should be a data point in mainline from this commit: https://spruce.mongodb.com/task/mongo_python_driver_perf_tests_perf_6.0_standalone_70bd1632cf94c66f7661e5c234418421f664c4ed_24_01_09_18_17_33/tests?execution=0&page=0&sortBy=DURATION&sortDir=DESC

I guess the issue must be that this work is in the standardized-logging branch, not master.

ShaneHarvey (changes requested):
Could these changes be merged into master instead? That way we can merge the general JSON improvements in one commit. I would prefer to see the logging specific max_length feature added separately.

NoahStapp:
> Could we create a new PR with only the general JSON improvements? That way we can merge the general JSON improvements in one commit and split out the logging specific max_length feature as a separate PR. Currently it's tough to review this PR as it has a lot of unrelated changes.

To clarify, split this PR into the JSON improvements and a separate PR for the logging-specific changes?

ShaneHarvey:
Yes exactly.

pymongo/logger.py
```diff
@@ -50,14 +50,15 @@ class LogMessage:
 
     def __init__(self, **kwargs: Any):
         self._kwargs = kwargs
+        self._redact()
```
ShaneHarvey, line 53: Can you change the base branch to master?

bson/json_util.py
```diff
@@ -448,6 +448,20 @@ def with_options(self, **kwargs: Any) -> JSONOptions:
 .. versionadded:: 3.4
 """
 
+# Encoders for BSON types
+_encoders = {
+    5: lambda obj, json_options: _encode_binary(obj, obj.subtype, json_options),  # Binary
+    7: lambda obj, json_options: {"$oid": str(obj)},  # noqa: ARG005 ObjectId
+    9: lambda obj, json_options: _encode_datetimems(obj, json_options),  # DatetimeMS
```
ShaneHarvey, line 455: A number of these can be simplified to remove a redundant intermediate call:
```
 9: _encode_datetimems,
 13: _encode_code,
 18: _encode_int64,
```

bson/json_util.py
```diff
@@ -824,22 +838,74 @@ def _parse_canonical_maxkey(doc: Any) -> MaxKey:
     return MaxKey()
 
 
+def _encode_bson(obj: Any, json_options: JSONOptions) -> Any:
```
ShaneHarvey, line 841: Can you inline this method? It's only called once and it should improve performance to remove a method call (per bson value).

bson/json_util.py
```diff
@@ -824,22 +838,74 @@ def _parse_canonical_maxkey(doc: Any) -> MaxKey:
...
+    type_marker = obj._type_marker
+    try:
+        return _encoders.get(type_marker)(obj, json_options)  # type: ignore
```
ShaneHarvey, line 844: Why the `# type: ignore`?

ShaneHarvey (changes requested):

bson/json_util.py
NoahStapp: `_encode_code` and `_encode_int64` are fine to inline, `_encode_datetimems` would be significantly less readable with its three distinct return paths.

bson/json_util.py
NoahStapp: The typechecker sees that the `get` call can return `None`, so it complains even with the try block around it.

ShaneHarvey: Oh but that's a bug that we need to guard against. It should be similar to our BSON logic here: https://github.com/mongodb/mongo-python-driver/blob/6537415da70c2a15cad52ebe7669673947ace65a/bson/__init__.py#L934-L935

bson/json_util.py
ShaneHarvey: Uh? I'm saying the code can literally be:
```python
 9: _encode_datetimems,
 13: _encode_code,
 18: _encode_int64,
```

bson/json_util.py
NoahStapp: Whoops, it's meant to look like this: 

```
_encoders[type_marker](obj, json_options)  # type: ignore[no-untyped-call]
```

That way if `type_marker` isn't a valid type we support, it will raise a `KeyError` and be caught in the try block.

bson/json_util.py
NoahStapp: Oh sorry, I thought you meant inline those functions into the lambda itself.

NoahStapp: That works, just have to move the definition of `_encoders` down to after all of the encoding methods are defined.

ShaneHarvey: Oh I see now thanks. I also agree that embedding complex logic into a lambda would not be ideal.

ShaneHarvey (approved):
Nice work! Great to see such a big impact on JSON encoding performance. Feel free to mention this improvement in the changelog.
Screenshot 2024-01-11 at 4 08 28 PM

ShaneHarvey (approved):

PR #1461, opened 2023-12-15
PYTHON-4103 Add badges to README

blink1073:
Also updates the Project [URLs](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#urls) that will be rendered on PyPI since we're removing a couple links from the readme.


ShaneHarvey (approved):
LGTM!

PR #1463, opened 2023-12-23
PYTHON-4015 Add test that ExceededTimeLimit is a retryable exception for reads

caseyclements:
…for reads

ExceededTimeLimit, Code 262 had already been added to pymongo.helpers._RETRYABLE_ERROR_CODES.

This simply adds the test. Exact same unified json as used in the C driver used for reference: 

https://github.com/mongodb/mongo-c-driver/pull/1482/files#diff-d96d8295b32780116338ab4eacb607f962ac00cbb900b17667493e8ec2072eac


blink1073 (approved):
LGTM!

PR #1464, opened 2024-01-05
PYTHON-4087 Flush writes before testing validate with background=True

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4087

### DIFF
--- a/test/test_database.py
+++ b/test/test_database.py
@@ -377,12 +377,13 @@ class TestDatabase(IntegrationTest):
     @client_context.require_version_min(4, 3, 3)
     @client_context.require_no_standalone
     def test_validate_collection_background(self):
-        db = self.client.pymongo_test
+        db = self.client.pymongo_test.with_options(write_concern=WriteConcern(w="majority"))
         db.test.insert_one({"dummy": "object"})
         coll = db.test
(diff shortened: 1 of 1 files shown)


Jibola (approved):

test/test_database.py
```diff
@@ -377,12 +377,14 @@ def test_validate_collection(self):
...
     @client_context.require_no_standalone
     def test_validate_collection_background(self):
-        db = self.client.pymongo_test
+        db = self.client.pymongo_test.with_options(write_concern=WriteConcern(w="majority"))
         db.test.insert_one({"dummy": "object"})
         coll = db.test
         self.assertTrue(db.validate_collection(coll, background=False))
         # The inMemory storage engine does not support background=True.
         if client_context.storage_engine != "inMemory":
+            # background=True requires the collection exist in a checkpoint.
```
Jibola, line 386: Should we also capture how fsync helps enforce the collection exists? Maybe just adding in a link to fsync? https://www.mongodb.com/docs/manual/reference/command/fsync/#fsync-lock

ShaneHarvey: I don't think more explanation on how fsync works is needed here. I prefer comments to cover the "why" more so than the "how".

PR #1465, opened 2024-01-05
PYTHON-3175 Preemptively cancel in progress operations when SDAM heartbeats timeout

NoahStapp:


pymongo/network.py
```diff
@@ -297,7 +297,6 @@ def receive_message(
 def wait_for_read(conn: Connection, deadline: Optional[float]) -> None:
     """Block until at least one byte is read, or a timeout, or a cancel."""
     context = conn.cancel_context
-    # Only Monitor connections can be cancelled.
     if context:
```
caseyclements, line 300: Now that conn.cancel_context is never None, I might consider updating this code block. 

I'm also not crazy about context = conn.cancel_context. It draws attention to a class which appears to be just a boolean. Maybe we could add a docstring to _CancellableContext to explain why it exists.

pymongo/pool.py
```diff
@@ -1377,7 +1373,8 @@ def __init__(self, address: _Address, options: PoolOptions, handshake: bool = Tr
         # LIFO pool. Sockets are ordered on idle time. Sockets claimed
         # and returned to pool from the left side. Stale sockets removed
         # from the right side.
-        self.conns: collections.deque = collections.deque()
+        self.available_conns: collections.deque = collections.deque()
+        self.active_conns: list[Connection] = []
         self.lock = _create_lock()
         self.active_sockets = 0
```
caseyclements, line 1379: There's now an ambiguity between `active_sockets` and active_conns`. Is it the case that active_sockets == len(self.active_conns)? Even if there is a reason to keep both,  it would be a good exercise to review where each is used and make sure they behave as expected.

pymongo/pool.py
```diff
@@ -1557,7 +1570,7 @@ def remove_stale_sockets(self, reference_generation: int) -> None:
                     if self.gen.get_overall() != reference_generation:
                         conn.close_conn(ConnectionClosedReason.STALE)
                         return
-                    self.conns.appendleft(conn)
```
caseyclements, line 1560: Here in `remove_stale_sockets` is the 2nd usage of Pool.connect(). Inside the latter we add the new Connection to active_conns. Is that the intended behavior in this scenario. I am unfamiliar with `self.gen` but I notice that the connection also gets added to self.available_conns.

pymongo/network.py
NoahStapp: @ShaneHarvey is there a reason `_CancellableContext` can't be removed in favor of a simple boolean?

ShaneHarvey: _CancellationContext is implemented via a boolean right now but it's possible to implement it in other ways. For example, at one point it was going to be implemented via a pipe so that the socket read/write could be interrupted more quickly. The other motivation for it being a class was that way a single instance could be shared between multiple connections. I agree the if-statement can be removed if cancel_context is never None. 

I'll leave it up to you if want to remove _CancellableContext and embed the bool within the Connection itself or not.

NoahStapp: There is a slight complication with removing `_CancellableContext`: `Monitor` takes advantage of `conn.cancel_context` being a reference to an object rather than a bool in order for its `_cancel_context` to affect the underlying connection. If `cancel_context` becomes a bool, `Monitor` will have to hold a reference to the last used connection instead in order to properly cancel the connection when needed.

pymongo/pool.py
```diff
@@ -1602,6 +1615,7 @@ def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> Connect
             raise
 
         conn = Connection(sock, self, self.address, conn_id)  # type: ignore[arg-type]
+        self.active_conns.append(conn)
```
ShaneHarvey, line 1618: I'm a little concerned about the impact of adding another place where Connections are stored. In particular around potential bugs where `active_conns` becomes out of sync with the pool, or a connection is kept alive because of this reference. It could be safer to instead store the CancellationContexts since those have no resources and are take up less memory. What do you think?

ShaneHarvey (changes requested):

pymongo/network.py
```diff
@@ -296,35 +296,32 @@ def receive_message(
...
-            if context.cancelled:
-                raise _OperationCancelled("hello cancelled")
-            if readable:
-                return
-            if timed_out:
-                raise socket.timeout("timed out")
+                timeout = _POLL_TIMEOUT
+            readable = conn.socket_checker.select(sock, read=True, timeout=timeout)
+        if conn.cancel_context.cancelled:
+            raise _OperationCancelled("operation cancelled")
```
ShaneHarvey, line 320: This block is under indented one-level now.

pymongo/pool.py
```diff
@@ -1794,6 +1803,9 @@ def checkin(self, conn: Connection) -> None:
         conn.pinned_cursor = False
         self.__pinned_sockets.discard(conn)
         listeners = self.opts._event_listeners
+        with self.lock:
+            if conn in self.active_conns:
+                self.active_conns.remove(conn)
```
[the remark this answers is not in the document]
ShaneHarvey: I retract this. It's not safe.

pymongo/pool.py
```diff
@@ -739,10 +739,7 @@ def __init__(
         self.pool_gen = pool.gen
         self.generation = self.pool_gen.get_overall()
         self.ready = False
-        self.cancel_context: Optional[_CancellationContext] = None
-        if not pool.handshake:
-            # This is a Monitor connection.
-            self.cancel_context = _CancellationContext()
+        self.cancel_context: _CancellationContext = _CancellationContext()
```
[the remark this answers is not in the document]
NoahStapp: Yes, it fails typechecking without it.

pymongo/pool.py
```diff
@@ -1377,9 +1373,9 @@ def __init__(self, address: _Address, options: PoolOptions, handshake: bool = Tr
         # LIFO pool. Sockets are ordered on idle time. Sockets claimed
         # and returned to pool from the left side. Stale sockets removed
         # from the right side.
-        self.conns: collections.deque = collections.deque()
+        self.available_conns: collections.deque = collections.deque()
```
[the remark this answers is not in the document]
NoahStapp: It's confusing to have a variable named `conns` represent the connections available for checkout from the pool if we also have a variable called `active_conns` that represents a different set of connections. Changing `conns` to `available_conns` makes the distinction and function of each more clear.

pymongo/pool.py
NoahStapp: I like this idea, we only need to track contexts, not connections themselves.

pymongo/pool.py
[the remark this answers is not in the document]
ShaneHarvey: I disagree because the same confusion exists between "available_conns" and "active_conns" but regardless now that we have active_conns_context can we rename this back to "conns"?

pymongo/pool.py
```diff
@@ -1602,6 +1618,7 @@ def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> Connect
             raise
 
         conn = Connection(sock, self, self.address, conn_id)  # type: ignore[arg-type]
+        self.active_conns_context.add(conn.cancel_context)
```
ShaneHarvey, line 1621: Can't modify this without holding the lock.

pymongo/pool.py
```diff
@@ -1644,6 +1661,8 @@ def checkout(self, handler: Optional[_MongoClientErrorHandler] = None) -> Iterat
             assert listeners is not None
             listeners.publish_connection_checked_out(self.address, conn.id)
         try:
+            if conn.cancel_context not in self.active_conns_context:
+                self.active_conns_context.add(conn.cancel_context)
```
ShaneHarvey, line 1665: Can't modify this without holding the lock.

pymongo/pool.py
```diff
@@ -1794,6 +1806,9 @@ def checkin(self, conn: Connection) -> None:
         conn.pinned_cursor = False
         self.__pinned_sockets.discard(conn)
         listeners = self.opts._event_listeners
+        with self.lock:
+            if conn.cancel_context in self.active_conns_context:
```
ShaneHarvey, line 1810: set.remove already handles this case so the if statement can be removed.

pymongo/pool.py
```diff
@@ -1557,7 +1571,9 @@ def remove_stale_sockets(self, reference_generation: int) -> None:
                     if self.gen.get_overall() != reference_generation:
                         conn.close_conn(ConnectionClosedReason.STALE)
                         return
-                    self.conns.appendleft(conn)
+                    self.available_conns.appendleft(conn)
+                    if conn.cancel_context in self.active_conns_context:
+                        self.active_conns_context.remove(conn.cancel_context)
```
ShaneHarvey, line 1576: set.remove already handles this case so the if statement can be removed.

ShaneHarvey (changes requested):

pymongo/pool.py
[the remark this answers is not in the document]
NoahStapp: Sure, fair enough.

pymongo/pool.py
NoahStapp: I've added removing the conn's cancellation context from the set of active contexts here, as suggested.

pymongo/pool.py
NoahStapp: I believe both are required, as `active_sockets` tracks socket creation and deletion, which is not coupled exactly with connection check-in and check-out, which `active_conns` tracks.

pymongo/pool.py
NoahStapp: Turns out `set.discard` is the method we're looking for, good catch!

pymongo/pool.py
NoahStapp: Turns out `set.discard` is the method we're looking for, good catch!

pymongo/pool.py
ShaneHarvey: Ah that's the one!

pymongo/pool.py
```diff
@@ -1378,6 +1374,7 @@ def __init__(self, address: _Address, options: PoolOptions, handshake: bool = Tr
         # and returned to pool from the left side. Stale sockets removed
         # from the right side.
         self.conns: collections.deque = collections.deque()
+        self.active_conns_context: set[_CancellationContext] = set()
```
ShaneHarvey, line 1377: suggest renaming to `active_contexts` for brevity.

ShaneHarvey (changes requested):
Could you run a benchmark with:
- 100 threads all running short running op (eg ping)
- 100 threads all running a long running op (eg find one with a sleep)

As well as the EVG benchmark? I'm wondering if there's a cost to adding this polling loop to every connection.

pymongo/network.py
NoahStapp: The original code has the following if statements all at the same level:

```
if context.cancelled:
	...
if readable:
	...
if timed_out:
	...
```

This change keeps that consistency, but removes the initial `if context:` check above, which might be causing the confusion.

pymongo/monitoring.py
```diff
@@ -874,15 +874,23 @@ class PoolClearedEvent(_PoolEvent):
     :param address: The address (host, port) pair of the server this Pool is
        attempting to connect to.
     :param service_id: The service_id this command was sent to, or ``None``.
+     - `service_id`: The service_id this command was sent to, or ``None``.
+     - '__interrupt_in_use_connections': True if all active connections were interrupted by the Pool during clearing.
```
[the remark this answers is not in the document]
NoahStapp: Not sure if we follow the spec this closely, but the spec calls out this flag as explicitly named `interruptInUseConnections`.

NoahStapp:
> Could you run a benchmark with:
> 
> * 100 threads all running short running op (eg ping)
> * 100 threads all running a long running op (eg find one with a sleep)
> 
> As well as the EVG benchmark? I'm wondering if there's a cost to adding this polling loop to every connection.

EVG benchmark results: https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz?comparison_id=9802c9f9-659e-46b3-97ac-c268f34a581c&selected_tab=data-table&percent_filter=0%7C%7C100&z_filter=0%7C%7C10 

A slight decrease for some operations, a slight improvement or no change for others. 

I ran a very simple local benchmark that created 100 threads and executed either a `ping` or a `find_one` with a sleep on each, timing the runtime and averaging out the result over all the threads. The runtime over 100 iterations of this benchmark is essentially unchanged when compared to master.

caseyclements (approved):
Good to go. Thank you. I learned a lot doing this review!

pymongo/monitoring.py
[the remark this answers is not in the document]
ShaneHarvey: Let's rename it `interrupt_connections` throughout, even on the monitoring events.

ShaneHarvey (changes requested):

ShaneHarvey:
Looking at the perf results it does appear this results in a ~5% decrease. Locally with 100 threads I see a decrease of somewhere between 5-10%:
```
$ git:(master) time python bench-100-threads.py 
python bench-100-threads.py  6.61s user 8.06s system 129% cpu 11.320 total
$ git:(master) time python bench-100-threads.py 
python bench-100-threads.py  6.65s user 8.13s system 130% cpu 11.297 total
$ git:(master) time python bench-100-threads.py 
python bench-100-threads.py  6.86s user 8.52s system 129% cpu 11.842 total
```
vs
```
$ git:(NoahStapp-PYTHON-3175) time python bench-100-threads.py 
python bench-100-threads.py  7.18s user 8.52s system 127% cpu 12.323 total
$ git:(NoahStapp-PYTHON-3175) time python bench-100-threads.py 
python bench-100-threads.py  7.26s user 8.88s system 130% cpu 12.320 total
$ git:(NoahStapp-PYTHON-3175) time python bench-100-threads.py 
python bench-100-threads.py  7.34s user 9.02s system 132% cpu 12.335 total
```

I don't think this should hold up this PR but it would be good to circle back and see how we can improve. Could you open a new ticket?

pymongo/monitoring.py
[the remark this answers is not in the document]
NoahStapp: The spec tests actually explicitly check for the existence of `interruptInUseConnections` in the event, so renaming this breaks the tests.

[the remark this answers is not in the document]
ShaneHarvey: What we can do in that situation is update the test runner to map the field name. It looks like the test runner ignores the interruptInUseConnections field and only checks for hasServiceId: https://github.com/mongodb/mongo-python-driver/blob/b1939e147073ad601666820a1b94af45949a8a27/test/unified_format.py#L836-L838

ShaneHarvey (changes requested):
Still need to update the unified runner to check interruptInUseConnections. It looks like the test runner ignores the interruptInUseConnections field and only checks for hasServiceId: https://github.com/mongodb/mongo-python-driver/blob/b1939e147073ad601666820a1b94af45949a8a27/test/unified_format.py#L836-L838

ShaneHarvey (approved):
LGTM!

PR #1472, opened 2024-01-12
PYTHON-1374 Optimize json_util performance for log truncation

NoahStapp:


bson/json_util.py
```diff
@@ -860,67 +936,19 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
...
             return _encode_binary(binval, binval.subtype, json_options)
         else:
             return {"$uuid": obj.hex}
-    if isinstance(obj, Decimal128):
-        return {"$numberDecimal": str(obj)}
-    if isinstance(obj, bool):
-        return obj
-    if json_options.json_mode == JSONMode.CANONICAL and isinstance(obj, int):
+    elif json_options.json_mode == JSONMode.CANONICAL and isinstance(obj, int):
         if -(2**31) <= obj < 2**31:
```
behackett, line 948: Make 2**31 a constant?

bson/json_util.py
```diff
@@ -235,6 +235,8 @@ class JSONMode:
 else:
     _BASE_CLASS = CodecOptions
 
+INT32_MAX = 2**31
```
behackett, line 238: You probably don't want people using this in their own code, so prepend it with an underscore.

bson/json_util.py
```diff
@@ -946,3 +961,87 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
             # original value, when float() is called on it.
             return {"$numberDouble": str(repr(obj))}
     raise TypeError("%r is not JSON serializable" % obj)
+
+
+def get_size(obj: Any, max_size: int, current_size: int = 0) -> int:
+    """Recursively finds size of objects"""
+    if current_size >= max_size:
+        return current_size
+    if isinstance(obj, ObjectId):
```
blink1073, line 970: This seems like another good place to use a lut based on object type, and store the lengths of these strings as constants.

pymongo/logger.py
```diff
@@ -50,6 +50,7 @@ class LogMessage:
 
     def __init__(self, **kwargs: Any):
         self._kwargs = kwargs
+        # self._redact()
```
blink1073, line 53: Let's not forget to restore this

NoahStapp: We actually want this to be removed entirely, the redaction should only take place when the log is going to be output.

bson/json_util.py
```diff
@@ -946,3 +961,87 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
...
+                return current_size
+    elif hasattr(obj, "__iter__"):
+        for i in obj:
+            current_size += get_size(i, max_size, current_size)
+            if current_size >= max_size:
+                return current_size
+    return current_size
+
+
+def truncate_documents(obj: Any, max_length: int) -> Tuple[Any, int]:
```
ShaneHarvey, line 1010: Let's make this private.

bson/json_util.py
```diff
@@ -946,3 +961,87 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
...
+            truncated_v, remaining = truncate_documents(v, remaining)
+            if truncated_v:
+                truncated[k] = truncated_v
+            if remaining <= 0:
+                break
+        return truncated, remaining
+    elif hasattr(obj, "__iter__") and not isinstance(obj, (str, bytes)):
+        truncated: Any = []  # type:ignore[no-redef]
+        for v in obj:
+            truncated_v, remaining = truncate_documents(v, remaining)
```
ShaneHarvey, line 1027: I suspect this whole method would be must faster if it was converted from recursion to iteration. Recursive calls can be expensive on deeply nested data.

bson/json_util.py
```diff
@@ -824,6 +831,14 @@ def _parse_canonical_maxkey(doc: Any) -> MaxKey:
     return MaxKey()
 
 
+def _encode_bson(obj: Any, json_options: JSONOptions) -> Any:
```
ShaneHarvey, line 834: Can this be removed? Or is it just an artifact from rebasing?

bson/json_util.py
```diff
@@ -467,6 +469,11 @@ def dumps(obj: Any, *args: Any, **kwargs: Any) -> str:
        Accepts optional parameter `json_options`. See :class:`JSONOptions`.
     """
     json_options = kwargs.pop("json_options", DEFAULT_JSON_OPTIONS)
+    if "max_length" in kwargs:
```
ShaneHarvey, line 472: What do think about calling `truncate_documents` explicitly in the logging code instead of here? That way we don't need to add or document a new "max_length" feature which would only be used internally.

ShaneHarvey (changes requested):

bson/json_util.py
NoahStapp: Good idea, thanks!

bson/json_util.py
NoahStapp: Artifact from rebasing, removed.

bson/json_util.py
NoahStapp: Perf result for an iterative implementation: https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz?comparison_id=27101777-b813-40aa-ba21-75a1d0ded08a&selected_tab=data-table&percent_filter=0%7C%7C100&z_filter=0%7C%7C10. Doesn't seem much better.

ShaneHarvey: Thanks for investigating that. Let's go back to recursive in that case. The code is much easier to follow and since there's no perf benefit it's not worth the complexity. Also there's the issue that the list and dict elements will be processed in reverse order so the behavior is subtly different.

ShaneHarvey (approved):
LGTM

PR #1474, opened 2024-01-12
PYTHON-4143 Optimize JSON encoding of int, float, str, and None

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4143

Explanation:
Previously, when our default() method would encounter an object of type int, float, str, or None it would go through every isintance if/else check, reach the bottom, raise a TypeError, then `_json_convert` would catch the TypeError and return the object. This is extremely wasteful and it's much faster to simply return the object when we know it's one of these types.

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -920,6 +920,8 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
                 # original value, when float() is called on it.
                 return {"$numberDouble": str(repr(obj))}
         return obj
+    elif isinstance(obj, str):
+        return obj
     elif isinstance(obj, datetime.datetime):
         if json_options.datetime_representation == DatetimeRepresentation.ISO8601:
             if not obj.tzinfo:
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -902,6 +902,24 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
             return _encoders[type_marker](obj, json_options)  # type: ignore[no-untyped-call]
         except KeyError:
             raise TypeError("%r is not JSON serializable" % obj) from None
+    elif isinstance(obj, int):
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
BEFORE:
```
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(None for _ in range(100000))}' 'dumps(doc)'
2 loops, best of 5: 160 msec per loop
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(str("1") for _ in range(100000))}' 'dumps(doc)'
2 loops, best of 5: 165 msec per loop
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(range(100000))}' 'dumps(doc)'
2 loops, best of 5: 163 msec per loop
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(float(i) for i in range(100000))}' 'dumps(doc)'
2 loops, best of 5: 182 msec per loop
```

AFTER:
```
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(None for _ in range(100000))}' 'dumps(doc)'
20 loops, best of 5: 11.4 msec per loop
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(str("1") for _ in range(100000))}' 'dumps(doc)'
10 loops, best of 5: 29.1 msec per loop
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(range(100000))}' 'dumps(doc)'
10 loops, best of 5: 25.8 msec per loop
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(float(i) for i in range(100000))}' 'dumps(doc)'
10 loops, best of 5: 39.4 msec per loop
```

Improvement:
- None 14x
- str 5.7x
- int 6.3x
- float 4.6x

NoahStapp (approved):
Nice catch, thanks!

PR #1475, opened 2024-01-12
PYTHON-4144 Optimize json_util performance using single dispatch table

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4144

Beanchmark results here: https://spruce.mongodb.com/task/mongo_python_driver_perf_tests_perf_6.0_standalone_patch_cd3f502de6405657c1e825978ec13b295b0d76ef_65a70b9257e85a54e3540958_24_01_16_23_05_08/trend-charts?execution=0

Screenshot 2024-01-16 at 4 19 18 PM

Results in a 15% improvement to the JSON benchmarks.

Before:
```
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(float(i) for i in range(100000))}' 'dumps(doc)'
10 loops, best of 5: 39 msec per loop
```
After:
```
$ python -m timeit -s 'from bson.json_util import dumps,loads;doc={"a": list(float(i) for i in range(100000))}' 'dumps(doc)'
10 loops, best of 5: 31.1 msec per loop
```

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -110,6 +110,7 @@ import uuid
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Mapping,
     MutableMapping,
     Optional,
@@ -971,7 +972,7 @@ def _encode_maxkey(dummy0: Any, dummy1: Any) -> dict:
 # Each encoder function's signature is:
 #   - obj: a Python data type, e.g. a Python int for _encode_int
 #   - json_options: a JSONOptions
-_ENCODERS: dict[Any, Any] = {
+_ENCODERS: dict[Any, Callable[[Any, JSONOptions], Any]] = {
     bool: _encode_noop,
     bytes: _encode_bytes,
     datetime.datetime: _encode_datetime,
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -904,7 +904,7 @@ def _encode_float(obj: float, json_options: JSONOptions) -> Any:
     return obj
 
 
-def _encode_datetime(obj: datetime, json_options: JSONOptions) -> dict:
+def _encode_datetime(obj: datetime.datetime, json_options: JSONOptions) -> dict:
     if json_options.datetime_representation == DatetimeRepresentation.ISO8601:
         if not obj.tzinfo:
             obj = obj.replace(tzinfo=utc)
@@ -935,7 +935,7 @@ def _encode_binary_obj(obj: Binary, json_options: JSONOptions) -> dict:
     return _encode_binary(obj, obj.subtype, json_options)
 
 
-def _encode_uuid(obj: bytes, json_options: JSONOptions) -> dict:
+def _encode_uuid(obj: uuid.UUID, json_options: JSONOptions) -> dict:
     if json_options.strict_uuid:
         binval = Binary.from_uuid(obj, uuid_representation=json_options.uuid_representation)
         return _encode_binary(binval, binval.subtype, json_options)
@@ -971,7 +971,7 @@ def _encode_maxkey(dummy0: Any, dummy1: Any) -> dict:
 # Each encoder function's signature is:
 #   - obj: a Python data type, e.g. a Python int for _encode_int
 #   - json_options: a JSONOptions
-_ENCODERS = {
+_ENCODERS: dict[Any, Any] = {
     bool: _encode_noop,
     bytes: _encode_bytes,
     datetime.datetime: _encode_datetime,
@@ -994,7 +994,7 @@ _ENCODERS = {
     Decimal128: _encode_decimal128,
 }
 
-_MARKERS = {
+_MARKERS: dict[Any, Any] = {
     5: _encode_binary_obj,
     7: _encode_objectid,
     9: _encode_datetimems,
@@ -1016,7 +1016,7 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
     # happen once per subtype.
     obj_type = type(obj)
     try:
-        return _ENCODERS[obj_type](obj, json_options)  # type: ignore[no-untyped-call]
+        return _ENCODERS[obj_type](obj, json_options)
     except KeyError:
         pass
 
@@ -1029,7 +1029,7 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
             func = _MARKERS[marker]
             # Cache this type for faster subsequent lookup.
             _ENCODERS[obj_type] = func
-            return func(obj, json_options)  # type: ignore[no-untyped-call]
+            return func(obj, json_options)
 
     # Third, test each base type. This will only happen once for
     # a subtype of a supported base type.
@@ -1038,6 +1038,6 @@ def default(obj: Any, json_options: JSONOptions = DEFAULT_JSON_OPTIONS) -> Any:
             func = _ENCODERS[base]
             # Cache this type for faster subsequent lookup.
             _ENCODERS[obj_type] = func
-            return func(obj, json_options)  # type: ignore[no-untyped-call]
+            return func(obj, json_options)
 
     raise TypeError("%r is not JSON serializable" % obj)
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -1004,8 +1004,8 @@ _MARKERS: dict[Any, Any] = {
     18: _encode_int64,
     19: _encode_decimal128,
     100: _encode_dbref,
-    127: _encode_minkey,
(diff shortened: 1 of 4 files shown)


ShaneHarvey:
I decided to track the perf delta for each BSON type:
<google-sheets-html-origin><style type="text/css"></style>
type | before (secs) | after (secs) | diff %
-- | -- | -- | --
bool | 0.125 | 0.143 | -12.58741259
NoneType | 1.648 | 0.144 | 1044.444444
int | 0.246 | 0.189 | 30.15873016
float | 0.368 | 0.294 | 25.17006803
Int64 | 0.287 | 0.2 | 43.5
str | 0.287 | 0.19 | 51.05263158
bytes | 1.112 | 0.984 | 13.00813008
datetime | 2.816 | 2.699 | 4.334938866
DatetimeMS | 3.878 | 3.696 | 4.924242424
UUID | 1.678 | 1.523 | 10.17728168
Binary | 1.096 | 1.014 | 8.08678501
Code | 0.581 | 0.494 | 17.61133603
DBRef | 3.964 | 3.652 | 8.543263965
ObjectId | 0.592 | 0.506 | 16.99604743
MaxKey | 0.427 | 0.348 | 22.70114943
MinKey | 0.432 | 0.342 | 26.31578947
Regex | 2.935 | 2.947 | -0.4071937564
Pattern | 3.009 | 3.024 | -0.496031746
Timestamp | 0.802 | 0.717 | 11.85495119
Decimal128 | 2.994 | 2.938 | 1.906058543

You can see that this is pretty much an easy win across the board except for `bool` which dropped 12%. The explanation is that in the old code `bool` was the first type we check for so it was being encoded as fast as possible. This advantage (bias) for `bool` disappears now that all type encoding functions are looked up in the same table. Personally I think a small drop for bools is fine given the overall benefit.

ShaneHarvey:
Note this was the benchmark:
```python
import uuid
import re
import timeit
from datetime import datetime
from bson import *
from bson import json_util

objects = [
    True,
    None,
    1,
    1.1,
    Int64(64),
    'str',
    b'bytes',
    datetime.now(),
    DatetimeMS(1),
    uuid.uuid4(),
    Binary(b'1'),
    Code("code"),
    DBRef("coll", ObjectId()),
    ObjectId(),
    MaxKey(),
    MinKey(),
    Regex("pat"),
    re.compile("pat"),
    Timestamp(1, 1),
    Decimal128("0.5"),
]


for o in objects:
    doc = {"a": [o for _ in range(10000)]}
    opts = json_util.RELAXED_JSON_OPTIONS.with_options(uuid_representation=STANDARD)
    tot = timeit.timeit("json_util.dumps(doc, json_options=opts)", globals=globals(), number=100)
    print(f"{type(o).__name__}, {tot:.3f}")
```

bson/json_util.py
```diff
@@ -855,100 +855,189 @@ def _encode_int64(obj: Int64, json_options: JSONOptions) -> Any:
...
+    obj_type = type(obj)
+    try:
+        return _ENCODERS[obj_type](obj, json_options)
+    except KeyError:
+        pass
+
+    # Second, fall back to trying _type_marker. This has to be done
+    # before the loop below since users could subclass one of our
+    # custom types that subclasses a python built-in (e.g. Binary)
+    if hasattr(obj, "_type_marker"):
```
NoahStapp, line 1026: Is this check explicitly for subclassed BSON types that users define? All of our own BSON types should already be cached in `_ENCODERS`, correct?

bson/json_util.py
```diff
@@ -855,100 +855,189 @@ def _encode_int64(obj: Int64, json_options: JSONOptions) -> Any:
...
+        marker = obj._type_marker
+        if marker in _MARKERS:
+            func = _MARKERS[marker]
+            # Cache this type for faster subsequent lookup.
+            _ENCODERS[obj_type] = func
+            return func(obj, json_options)
+
+    # Third, test each base type. This will only happen once for
+    # a subtype of a supported base type.
+    for base in _BUILT_IN_TYPES:
```
NoahStapp, line 1036: And this is the same for non-BSON types?

bson/json_util.py
```diff
@@ -855,100 +855,189 @@ def _encode_int64(obj: Int64, json_options: JSONOptions) -> Any:
...
-    17: lambda obj, json_options: {"$timestamp": {"t": obj.time, "i": obj.inc}},  # noqa: ARG005 Timestamp
-    18: _encode_int64,  # Int64
-    19: lambda obj, json_options: {"$numberDecimal": str(obj)},  # noqa: ARG005 Decimal128
-    100: lambda obj, json_options: _json_convert(obj.as_doc(), json_options=json_options),  # DBRef
-    127: lambda obj, json_options: {"$maxKey": 1},  # noqa: ARG005 MaxKey
-    255: lambda obj, json_options: {"$minKey": 1},  # noqa: ARG005 MinKey
+# Each encoder function's signature is:
+#   - obj: a Python data type, e.g. a Python int for _encode_int
+#   - json_options: a JSONOptions
+_ENCODERS: dict[Any, Any] = {
```
NoahStapp, line 974: Is it worth type-hinting the value type here since we know all the function signatures?

NoahStapp (changes requested):
A few clarifying questions, the code itself looks good!

bson/json_util.py
ShaneHarvey: Correct, I'll add some test cases for this (if they don't already exist).

ShaneHarvey: Wow there was no tests for this in BSON, I added one. The use-case is a little convoluted. Not sure if this is something we'll support but at least we have more consistency between bson and json encoding now.

bson/json_util.py
ShaneHarvey: This is to handle cases like:
```python
        class myint(int):
            pass
```

bson/json_util.py
ShaneHarvey: Done.

ShaneHarvey:
Ready for another look.

bson/json_util.py
NoahStapp: Makes sense, I think the comment would be clearer if it explicitly said why this is needed.

bson/json_util.py
NoahStapp: Makes sense, I think the comment would be clearer if it explicitly said why this is needed.

bson/json_util.py
ShaneHarvey: Oh this comment it clear to me now. It's saying we need to check type_marker first because otherwise a class like this might be encoded as a `bytes` instead of a `Binary` because both `isinstance(MyBin, Binary)` and `isinstance(MyBin, bytes)` would be true and we don't control which check happens first:
```python
class MyBin(Binary):  # Will this be encoded as bytes or as Binary?
    pass
```

I'll keep the comments as they are (to mirror the bson comments) and add this case to the tests.

bson/json_util.py
```diff
@@ -855,100 +856,188 @@ def _encode_int64(obj: Int64, json_options: JSONOptions) -> Any:
...
+    MaxKey: _encode_maxkey,
+    MinKey: _encode_minkey,
+    ObjectId: _encode_objectid,
+    Regex: _encode_regex,
+    RE_TYPE: _encode_regex,
+    Timestamp: _encode_timestamp,
+    Decimal128: _encode_decimal128,
+}
+
+_MARKERS: dict[Any, Any] = {
```
Jibola, line 998: NIT: 
```suggestion
_MARKERS: dict[int, Callable[[Any, JSONOptions], Any]] = {
```

bson/json_util.py
```diff
@@ -855,100 +856,188 @@ def _encode_int64(obj: Int64, json_options: JSONOptions) -> Any:
...
+    7: _encode_objectid,
+    9: _encode_datetimems,
+    11: _encode_regex,
+    13: _encode_code,
+    17: _encode_timestamp,
+    18: _encode_int64,
+    19: _encode_decimal128,
+    100: _encode_dbref,
+    127: _encode_maxkey,
+    255: _encode_minkey,
```
Jibola, line 1009: Out of scope for this PR, but do we have some constants/enum type object we could have these numbers reference for readability in the future? I imagine we need to handle these numerical types elsewhere, and I'm curious thoughts on that.

Jibola (approved):
LGTM!

bson/json_util.py
ShaneHarvey: Done.

bson/json_util.py
ShaneHarvey: Removed these magic constants by refactoring it to use `_type_marker`.

NoahStapp (approved):
LGTM!

PR #1477, opened 2024-01-17
PYTHON-4151 Use Centralized Scripts for CSFLE Tests

blink1073:
Requires https://github.com/mongodb-labs/drivers-evergreen-tools/pull/390

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -136,12 +136,6 @@ if [ -n "$TEST_ENCRYPTION" ] || [ -n "$TEST_FLE_AZURE_AUTO" ] || [ -n "$TEST_FLE
         exit 1
     fi
     export PYMONGOCRYPT_LIB
-
-    # TODO: Test with 'pip install pymongocrypt'
-    if [ ! -d "libmongocrypt_git" ]; then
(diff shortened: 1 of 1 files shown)


.evergreen/run-tests.sh
```diff
@@ -157,8 +138,9 @@ if [ -n "$TEST_ENCRYPTION" ] || [ -n "$TEST_FLE_AZURE_AUTO" ] || [ -n "$TEST_FLE
     export PYMONGOCRYPT_LIB
 
     # TODO: Test with 'pip install pymongocrypt'
-    git clone https://github.com/mongodb/libmongocrypt.git libmongocrypt_git
-    python -m pip install --prefer-binary -r .evergreen/test-encryption-requirements.txt
+    if [ ! -d "libmongocrypt_git" ]; then
+        git clone https://github.com/mongodb/libmongocrypt.git libmongocrypt_git
```
NoahStapp, line 142: Do we need to do this if we already install libmongocrypt above?

NoahStapp (changes requested):

.evergreen/run-tests.sh
blink1073: It shouldn't be

NoahStapp (approved):

PR #1478, opened 2024-01-18
PYTHON-4146 Improve GridFS upload performance by batch writing chunks with insert_many

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4146

### DIFF
--- a/gridfs/grid_file.py
+++ b/gridfs/grid_file.py
@@ -28,14 +28,17 @@ from gridfs.errors import CorruptGridFile, FileExists, NoFile
 from pymongo import ASCENDING
 from pymongo.client_session import ClientSession
 from pymongo.collection import Collection
+from pymongo.common import MAX_MESSAGE_SIZE
 from pymongo.cursor import Cursor
 from pymongo.errors import (
+    BulkWriteError,
     ConfigurationError,
     CursorNotFound,
     DuplicateKeyError,
     InvalidOperation,
     OperationFailure,
 )
+from pymongo.helpers import _check_write_command_response
 from pymongo.read_preferences import ReadPreference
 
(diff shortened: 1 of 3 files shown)


gridfs/grid_file.py
```diff
@@ -268,32 +277,47 @@ def __setattr__(self, name: str, value: Any) -> None:
...
-
-        try:
-            self._chunks.insert_one(chunk, session=self._session)
-        except DuplicateKeyError:
-            self._raise_file_exists(self._file["_id"])
+        if data:
+            self._buffered_docs.append(
+                {"files_id": self._file["_id"], "n": self._chunk_number, "data": Binary(data)}
+            )
+            self._buffered_docs_size += len(data)
```
blink1073, line 288: Should this include rest of the document size as well?

ShaneHarvey: Hmm good point. This won't matter much in practice because the default chunk_size (256KB) dominates the document overhead but I could see some pathological cases where the app uses chunk_size=1 or a very large "files_id" where this could behave poorly. In those cases we'd batch up far more documents in memory than we'd need.

I see two fixes that are simpler than guestimating the document size:
- encode to RawBSONDocument so we know the real size
- limit the batch to at most 100,000 documents.

ShaneHarvey: I decided not to use RawBSONDocument as is was tricky to get right (which codec options to use), is less readable code-wise, and is a little less efficient (due to extra memory copies shuffling around the raw documents). 

I did add the 100,000 limit (OP_MSG's maxWriteBatchSize) and added some extra overhead to account for the size of the chunk doc encoded as bson.

blink1073 (approved):
LGTM!

PR #1483, opened 2024-01-23
PYTHON-4158 Fix typo in create_index docstring

jvanmelckebeke:
change a minuscule typo which formats a code part in the docs wrong

image

basically adds one space. I know it is a small change but it bothered me ;)


blink1073 (approved):
Thank you!

PR #1486, opened 2024-01-25
PYTHON-4147: RuntimeError fix: Check for explicit runtime error message

Jibola:


pymongo/periodic_executor.py
```diff
@@ -91,7 +91,14 @@ def open(self) -> None:
             thread.daemon = True
             self._thread = weakref.proxy(thread)
             _register_executor(self)
-            thread.start()
+            try:
+                thread.start()
+            except RuntimeError as e:
+                if str(e) == "can't create new thread at interpreter shutdown":
```
NoahStapp, line 97: We're assuming this error message will remain constant,  right?

NoahStapp (changes requested):
One clarification for my own understanding, otherwise good!

pymongo/periodic_executor.py
Jibola: Yep

NoahStapp (approved):

pymongo/periodic_executor.py
```diff
@@ -91,7 +91,14 @@ def open(self) -> None:
...
+                    # Result of change
```
ShaneHarvey, line 98: This comment can be removed.

pymongo/periodic_executor.py
```diff
@@ -91,7 +91,14 @@ def open(self) -> None:
             thread.daemon = True
             self._thread = weakref.proxy(thread)
             _register_executor(self)
-            thread.start()
+            try:
```
ShaneHarvey, line 94: It's worth adding a comment explaining when and why this code hits this error.

ShaneHarvey (changes requested):
Could you update the PR title to explain the higher level of what's being changed? Also we should add this to the changelog.

ShaneHarvey (commented):
Is it possible to add a test for this?

pymongo/periodic_executor.py
Jibola: Added the new commentary. Let me know what you think.

test/test_monitor.py
```diff
@@ -79,6 +80,17 @@ def test_cleanup_executors_on_client_close(self):
         for executor in executors:
             wait_until(lambda: executor._stopped, f"closed executor: {executor._name}", timeout=5)
 
+    @unittest.skipIf(sys.version_info[:2] >= (3, 12), reason="Python version must be (>=3.12)")
```
ShaneHarvey, line 83: Let's run this on every python version.

test/test_monitor.py
```diff
@@ -79,6 +80,17 @@ def test_cleanup_executors_on_client_close(self):
...
+    def test_no_thread_start_runtime_err_on_shutdown(self):
+        """Test we silence noisy runtime errors fired when the MongoClient spawns a new thread
+        on process shutdown."""
+        command = ["python", "-c", "'from pymongo import MongoClient; c = MongoClient()'"]
```
ShaneHarvey, line 87: "python" could be any python version here. Instead let's use `sys.executable`.

test/test_monitor.py
```diff
@@ -79,6 +80,17 @@ def test_cleanup_executors_on_client_close(self):
...
+        completed_process: subprocess.CompletedProcess = subprocess.run(
+            " ".join(command), shell=True, capture_output=True
```
ShaneHarvey, line 89: No need to `join` on command since run() accepts a list.

test/test_monitor.py
```diff
@@ -79,6 +80,17 @@ def test_cleanup_executors_on_client_close(self):
...
+        )
+
+        assert not completed_process.stderr
```
ShaneHarvey, line 92: Let's assert stdout is empty too? Also should this be `self.assertEmpty` or `self.assertFalse`?

pymongo/periodic_executor.py
```diff
@@ -91,7 +93,15 @@ def open(self) -> None:
             thread.daemon = True
             self._thread = weakref.proxy(thread)
             _register_executor(self)
-            thread.start()
+            # Mitigation to RuntimeError firing when thread starts on shutdown
+            # https://github.com/python/cpython/issues/114570
+            try:
+                thread.start()
+            except RuntimeError as e:
+                if str(e) == _THREAD_START_ON_SHUTDOWN_ERR:
```
ShaneHarvey, line 101: What do you think of `if "interpreter shutdown" in str(e) or sys.is_finalizing():`? This could be more future proof.

ShaneHarvey (changes requested):

test/test_monitor.py
Jibola: I ended up using `self.assertFalse`

test/test_monitor.py
Jibola: I've been attempting to capture the output, and `shell=True` has been the main way for me to capture that output. It seems to not capture all of the shell output when done without the `shell=True`. 

I'm not sure why this is the case and haven't found much good material explaining why, but since these tests are scoped and explicitly defined, I'd argue it's fine to keep this way. The only difference I know is that `shell=True` uses whatever native `sh` the underlying system uses, environment variables and all.

If anyone does know why it's not capturing the output in either stderr/stdout when using `shell=False` that would be great.

pymongo/periodic_executor.py
Jibola: Wasn't sure if you meant explicitly "interpreter shutdown", so I've left it with the current, longer string.

ShaneHarvey: Yes I meant "interpreter shutdown" since it's a more broad but still safe.

test/test_monitor.py
ShaneHarvey: Oh I see. Thanks for explaining.

ShaneHarvey (commented):
One last thing is to update the changelog to describe this bugfix.

doc/changelog.rst
```diff
@@ -28,6 +28,8 @@ PyMongo 4.7 brings a number of improvements including:
     >>> orjson.dumps({'a': Int64(1)}, default=json_util.default, option=orjson.OPT_PASSTHROUGH_SUBCLASS)
 
 .. _orjson: https://github.com/ijl/orjson
+- Fixed a bug appearing in Python 3.12 where :attr:`RuntimeError: can't create new thread at interpreter shutdown`
+  fires when a MongoClient's :class:`PeriodicExecutor` thread starts as the python interpreter is shutting down.
```
ShaneHarvey, line 32: `fires` -> `could be written to stderr`

Jibola:
```suggestion
  could be written to stderr when a MongoClient's :class:`PeriodicExecutor` thread starts as the python interpreter is shutting down.
```

ShaneHarvey (approved):

ShaneHarvey:
Whoops, lint is failing:
> doc/changelog.rst:31: D000 Explicit markup ends without a blank line; unexpected unindent.

doc/changelog.rst
```diff
@@ -28,6 +28,8 @@ PyMongo 4.7 brings a number of improvements including:
     >>> orjson.dumps({'a': Int64(1)}, default=json_util.default, option=orjson.OPT_PASSTHROUGH_SUBCLASS)
 
 .. _orjson: https://github.com/ijl/orjson
+- Fixed a bug appearing in Python 3.12 where :attr:`RuntimeError: can't create new thread at interpreter shutdown`
```
ShaneHarvey, line 31: remove `:attr` and replace the backticks with "?

doc/changelog.rst
```diff
@@ -28,6 +28,8 @@ PyMongo 4.7 brings a number of improvements including:
...
+  could be written to stderr when a MongoClient's :class:`PeriodicExecutor` thread starts as the python interpreter is shutting down.
```
ShaneHarvey, line 32: We can remove PeriodicExecutor here since it's private.

ShaneHarvey (approved):
LGTM

PR #1489, opened 2024-01-29
PYTHON-4018 Clarify exactly what code/label fields drivers should inspect to determine retryability

blink1073:


NoahStapp (approved):
LGTM!

PR #1490, opened 2024-01-29
PYTHON-3473 Add log messages to CMAP spec

NoahStapp:


pymongo/server.py
```diff
@@ -89,7 +89,7 @@ def close(self) -> None:
                 )
             )
         self._monitor.close()
-        self._pool.reset_without_pause()
+        self._pool.close()
```
Jibola, line 92: why do we opt to close here rather than the old behavior?

NoahStapp: See https://jira.mongodb.org/browse/PYTHON-4166.

NoahStapp:
Perf results: (https://spruce.mongodb.com/task/mongo_python_driver_perf_tests_perf_6.0_standalone_patch_41a131ea1c15ffa969a14ce1334ce19837dc226b_65b80026d6d80a6e319730d1_24_01_29_19_44_47/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC)
image

pymongo/logger.py
```diff
@@ -45,18 +62,33 @@ class _CommandStatusMessage(str, enum.Enum):
...
+def _verbose_connection_error_reason(reason: str) -> str:
+    return _VERBOSE_CONNECTION_ERROR_REASONS.get(reason, reason)
+
+
 class LogMessage:
     __slots__ = ["_kwargs"]
 
     def __init__(self, **kwargs: Any):
         self._kwargs = kwargs
+        self._kwargs = {k: v for k, v in self._kwargs.items() if v is not None}
```
ShaneHarvey, line 91: suggest `k: v for k, v in kwargs.items()` so we can remove `self._kwargs = kwargs`.

pymongo/pool.py
```diff
@@ -1606,7 +1675,16 @@ def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> Connect
...
                 )
-
+                _debug_log(
+                    _CONNECTION_LOGGER,
+                    clientId=self._client_id,
+                    message=_ConnectionStatusMessage.CONN_CLOSED,
+                    serverHost=self.address[0],
+                    serverPort=self.address[1],
+                    driverConnectionId=conn_id,
+                    reason="An error occurred while using the connection",
```
ShaneHarvey, line 1685: Should this use _verbose_connection_error_reason?

test/unified_format.py
```diff
@@ -1704,7 +1704,9 @@ def format_logs(log_list):
                 )
             return client_to_log
 
-        with self.assertLogs("pymongo.command", level="DEBUG") as cm:
+        # Assumes that tests only verify one logging component at a time
```
ShaneHarvey, line 1707: Does this assumption hold? What happens if a test specifies multiple components?

NoahStapp: So far this assumption holds, but if a test specifies multiple components only the first component's logs will be checked. Our current testing implementation would require a significant overhaul if a multi-component test was written, as we'd have to figure out a way to distinguish which operations belonged to which logging components.

NoahStapp: Good point! I didn't realize that `assertLogs` would catch all of the logging components if we just passed it the top-level `pymongo` logger. Since it does, supporting multiple components is easy, done.

test/unified_format.py
```diff
@@ -1704,13 +1704,18 @@ def format_logs(log_list):
                 )
             return client_to_log
 
-        with self.assertLogs("pymongo.command", level="DEBUG") as cm:
+        with self.assertLogs("pymongo", level="DEBUG") as cm:
             self.run_operations(operations)
             formatted_logs = format_logs(cm.records)
             # FIXME: currently I assume all msgs are coming from client
```
ShaneHarvey, line 1710: Can we remove this fixme now?

ShaneHarvey (approved):
LGTM!

PR #1491, opened 2024-01-30
PYTHON-3587 Do not perform server selection to determine sessions support

NoahStapp:


pymongo/client_session.py
```diff
@@ -1097,19 +1101,12 @@ def pop_all(self) -> list[_ServerSession]:
...
-        # clear stale sessions there. In case many sessions were returned via
-        # __del__, check for stale sessions here too.
-        self._clear_stale(session_timeout_minutes)
-
+    def get_server_session(self) -> _ServerSession:
         # The most recently used sessions are on the left.
         while self:
-            s = self.popleft()
-            if not s.timed_out(session_timeout_minutes):
-                return s
```
NoahStapp, line 1112: This change is necessary because we no longer have access to `session_timeout_minutes` at the time of session creation. Is it feasible to move this timeout check to the moment of session application instead, where we have the underlying connection and access to the timeout?

ShaneHarvey: Let's keep this method the same but delay calling it until ClientSession._materialize() is called.

pymongo/mongo_client.py
```diff
@@ -1751,9 +1751,7 @@ def _process_periodic_tasks(self) -> None:
                 helpers._handle_exception()
 
     def __start_session(self, implicit: bool, **kwargs: Any) -> ClientSession:
-        # Raises ConfigurationError if sessions are not supported.
         if implicit:
-            self._topology._check_implicit_session_support()
             server_session: Union[_EmptyServerSession, _ServerSession] = _EmptyServerSession()
         else:
             server_session = self._get_server_session()
```
ShaneHarvey, line 1757: Let's have both implicit and explicit sessions start with _EmptyServerSession(). That way we can check logicalSessionTimeoutMinutes on first use in _materialize().

pymongo/server_description.py
```diff
@@ -243,8 +243,7 @@ def is_server_type_known(self) -> bool:
     def retryable_writes_supported(self) -> bool:
         """Checks if this server supports retryable writes."""
         return (
-            self._ls_timeout_minutes is not None
```
ShaneHarvey, line 246: Should we remove _ls_timeout_minutes and all code for it?

pymongo/topology.py
```diff
@@ -609,8 +609,7 @@ def _check_session_support(self) -> float:
     def get_server_session(self) -> _ServerSession:
         """Start or resume a server session, or raise ConfigurationError."""
         with self._lock:
-            session_timeout = self._check_session_support()
```
ShaneHarvey, line 612: Can we remove _check_session_support?

test/test_encryption.py
```diff
@@ -2997,5 +2999,58 @@ def test_collection_name_collision(self):
...
+    args = ["mongocryptd", f"--port={port}", "--idleShutdownTimeoutSecs=60"]
+    _spawn_daemon(args)
+
+
+class TestNoSessionsSupport(EncryptionIntegrationTest):
+    mongocryptd_client: MongoClient
+    MONGOCRYPTD_PORT = 27020
+
+    @classmethod
+    @client_context.require_sessions
```
ShaneHarvey, line 3012: This test is for servers that _don't_ support sessions, but require_sessions does the opposite. Can we remove this?

test/test_encryption.py
```diff
@@ -2997,5 +2999,58 @@ def test_collection_name_collision(self):
...
+        super().setUpClass()
+        start_mongocryptd(cls.MONGOCRYPTD_PORT)
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+
+    def setUp(self) -> None:
+        self.listener = OvertCommandListener()
+        self.listener.reset()
```
ShaneHarvey, line 3024: Let's remove this reset() call since the listener was just created.

test/test_encryption.py
```diff
@@ -2997,5 +2999,58 @@ def test_collection_name_collision(self):
...
+        with self.assertRaises(OperationFailure):
+            self.mongocryptd_client.db.test.insert_one({"x": 1})
+
+        self.assertNotIn("lsid", self.listener.started_events[1].command)
+        self.mongocryptd_client.close()
+
+    def test_explicit_session_errors_when_unsupported(self):
+        self.listener.reset()
+        with self.mongocryptd_client.start_session() as s:
+            with self.assertRaises(ConfigurationError):
```
ShaneHarvey, line 3048: Let's use assertRaisesRegex to assert the error message too.

test/test_encryption.py
```diff
@@ -2997,5 +2999,58 @@ def test_collection_name_collision(self):
...
+                self.mongocryptd_client.db.test.find_one(session=s)
+            with self.assertRaises(ConfigurationError):
+                self.mongocryptd_client.db.test.insert_one({"x": 1}, session=s)
+        self.mongocryptd_client.close()
```
ShaneHarvey, line 3052: instead of closing at the end of each test I suggest using `self.addCleanup(self.mongocryptd_client.close)` right after creating the client in setup.

test/test_encryption.py
```diff
@@ -22,6 +22,7 @@
 import socket
 import socketserver
 import ssl
+import subprocess
```
ShaneHarvey, line 25: unused.

ShaneHarvey (changes requested):

pymongo/server_description.py
NoahStapp: Actually, according to the spec https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst, we still need to store the lowest `logicalSessionTimeoutMinutes` value in the `ToplogyDescription`.

ShaneHarvey: I see, so we still need to track it so we can use the lowest value for the server session cache expiration.

pymongo/client_session.py
```diff
@@ -965,10 +965,12 @@ def _txn_read_preference(self) -> Optional[_ServerMode]:
             return self._transaction.opts.read_preference
         return None
 
-    def _materialize(self) -> None:
+    def _materialize(self, logical_session_timeout_minutes: float) -> None:
```
ShaneHarvey, line 968: float -> int.

pymongo/client_session.py
```diff
@@ -1097,7 +1104,9 @@ def pop_all(self) -> list[_ServerSession]:
             ids.append(self.pop().session_id)
         return ids
 
-    def get_server_session(self, session_timeout_minutes: float) -> _ServerSession:
+    def get_server_session(
+        self, session_timeout_minutes: float, old: _EmptyServerSession
```
ShaneHarvey, line 1108: float -> int.

pymongo/server_description.py
NoahStapp: Bingo.

pymongo/client_session.py
```diff
@@ -1111,7 +1120,7 @@ def get_server_session(self, session_timeout_minutes: float) -> _ServerSession:
             if not s.timed_out(session_timeout_minutes):
                 return s
 
-        return _ServerSession(self.generation)
+        return _ServerSession(self.generation, old.session_id)
```
ShaneHarvey, line 1123: This is odd. If the cache is empty we'll create a new session with the same id but if the cache isn't empty we'll return an existing session with a different id. What problem are we trying to solve here?

ShaneHarvey (changes requested):

pymongo/pool.py
```diff
@@ -870,6 +871,7 @@ def _hello(
         self.max_message_size = hello.max_message_size
         self.max_write_batch_size = hello.max_write_batch_size
         self.supports_sessions = hello.logical_session_timeout_minutes is not None
+        self.logical_session_timeout_minutes = hello.logical_session_timeout_minutes or 0.0
```
ShaneHarvey, line 874: `or 0` since logical_session_timeout_minutes is an int.

pymongo/client_session.py
NoahStapp: Hmm, good point if the cache is not empty. Our unified test runner expects that sessions have an id on creation, with several tests relying on that behavior. Adding a simple ping command after creation to populate the session id emits unexpected events that cause those same tests to fail.

ShaneHarvey: Oh I see. I think the only approach is that accessing the session_id property needs to materialize the session too.

pymongo/client_session.py
ShaneHarvey: Yeah that was a mistake. It's int everywhere else.

pymongo/client_session.py
NoahStapp: Sorry, I don't quite understand what you mean. You're saying we need to have the session materialize when `session_id` is accessed, but without actually checking if sessions are supported?

ShaneHarvey: Yes. The session support check would still happen later in _apply_to.

pymongo/mongo_client.py
```diff
@@ -2393,12 +2384,15 @@ def _write(self) -> T:
...
             self._server = self._get_server()
-            supports_session = (
-                self._session is not None and self._server.description.retryable_writes_supported
-            )
             with self._client._checkout(self._server, self._session) as conn:
                 max_wire_version = conn.max_wire_version
-                if self._retryable and not supports_session:
+                sessions_supported = (
+                    self._retryable
+                    and self._server.description.retryable_writes_supported
```
ShaneHarvey, line 2391: Can we remove this line?

ShaneHarvey (changes requested):

pymongo/client_session.py
```diff
@@ -979,16 +988,20 @@ def _apply_to(
...
+        self._materialize(conn.logical_session_timeout_minutes)
         if self.options.snapshot:
             self._update_read_concern(command, conn)
 
         self._server_session.last_use = time.monotonic()
         command["lsid"] = self._server_session.session_id
 
         if is_retryable:
-            command["txnNumber"] = self._server_session.transaction_id
+            command["txnNumber"] = self._transaction_id
```
ShaneHarvey, line 1004: I'd say change these back to _server_session.transaction_id since we've already called materialize().

pymongo/client_session.py
```diff
@@ -1097,24 +1110,25 @@ def pop_all(self) -> list[_ServerSession]:
...
 
-    def get_server_session(self, session_timeout_minutes: float) -> _ServerSession:
+    def get_server_session(self, session_timeout_minutes: Optional[int]) -> _ServerSession:
         # Although the Driver Sessions Spec says we only clear stale sessions
         # in return_server_session, PyMongo can't take a lock when returning
         # sessions from a __del__ method (like in Cursor.__die), so it can't
         # clear stale sessions there. In case many sessions were returned via
         # __del__, check for stale sessions here too.
-        self._clear_stale(session_timeout_minutes)
+        if session_timeout_minutes:
```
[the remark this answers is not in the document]
NoahStapp: Good point, I think it's safer to check explicitly for `None`.

pymongo/mongo_client.py
NoahStapp: And just check `conn.supports_sessions` instead?

test/test_session.py
```diff
@@ -116,16 +116,12 @@ def _test_ops(self, client, *ops):
 
         for f, args, kw in ops:
             with client.start_session() as s:
-                last_use = s._server_session.last_use
-                start = time.monotonic()
-                self.assertLessEqual(last_use, start)
```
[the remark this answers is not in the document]
NoahStapp: No, we can fix the issue by materializing the session before accessing last_use.

pymongo/mongo_client.py
NoahStapp: We can remove two of the variables in the check, the remaining two are required for tests to pass.

pymongo/mongo_client.py
```diff
@@ -2393,12 +2384,12 @@ def _write(self) -> T:
...
             max_wire_version = 0
             self._server = self._get_server()
-            supports_session = (
-                self._session is not None and self._server.description.retryable_writes_supported
-            )
             with self._client._checkout(self._server, self._session) as conn:
                 max_wire_version = conn.max_wire_version
-                if self._retryable and not supports_session:
+                sessions_supported = (
+                    self._server.description.retryable_writes_supported and conn.supports_sessions
```
ShaneHarvey, line 2390: Is `self._session` never `None` anymore?

test/test_session.py
```diff
@@ -245,10 +245,12 @@ def test_pool_lifo(self):
         b.end_session()
 
         s = self.client.start_session()
+        self.client.admin.command("ping", session=s)
```
ShaneHarvey, line 248: Can we remove these pings now that `s.session_id` materializes the session?

test/test_session.py
```diff
@@ -274,6 +276,8 @@ def test_end_sessions(self):
         client = rs_or_single_client(event_listeners=[listener])
         # Start many sessions.
         sessions = [client.start_session() for _ in range(_MAX_END_SESSIONS + 1)]
+        for s in sessions:
+            client.admin.command("ping", session=s)
```
ShaneHarvey, line 280: This is gonna make the test take a long time. Can we instead just do `s.session_id`?

test/test_session.py
NoahStapp: No, this test requires the sessions actually enter the pool, which materializing itself does not do. The test fails without these pings, sadly.

pymongo/mongo_client.py
NoahStapp: Added it back for safety.

test/test_session.py
ShaneHarvey: I mean replace this ping with `s.session_id`.

ShaneHarvey: Can you explain? It appears these tests are only checking `session_id`.

NoahStapp: The test checks that session `a` and `b` are both in the session pool after they are ended, with `b` at the top due to LIFO ordering. We expect the next session created to be `b`, but since session creation no longer interacts with the pool unless a timeout is present, materializing the new session instead of using a `ping` results in a new session entirely.

pymongo/client_session.py
```diff
@@ -1097,24 +1110,25 @@ def pop_all(self) -> list[_ServerSession]:
...
-        # The most recently used sessions are on the left.
-        while self:
-            s = self.popleft()
-            if not s.timed_out(session_timeout_minutes):
-                return s
+            # The most recently used sessions are on the left.
+            while self:
+                s = self.popleft()
+                if not s.timed_out(session_timeout_minutes):
+                    return s
```
ShaneHarvey, line 1126: We still want to acquire a session from the pool here even if the timeout is None.

test/test_session.py
ShaneHarvey: Oh that's a bug in get_server_session. After it's fixed we should remove the pings here.

ShaneHarvey (changes requested):

pymongo/client_session.py
```diff
@@ -1062,11 +1075,14 @@ def mark_dirty(self) -> None:
         """
         self.dirty = True
 
-    def timed_out(self, session_timeout_minutes: float) -> bool:
-        idle_seconds = time.monotonic() - self.last_use
+    def timed_out(self, session_timeout_minutes: Optional[int]) -> bool:
+        if session_timeout_minutes is None:
+            return False
+        else:
+            idle_seconds = time.monotonic() - self.last_use
```
ShaneHarvey, line 1082: nit: since the if-statement returns we don't need the `else`. The benefit is that:
1) code is easier to read with less indentation
2) there will be less code churn in the PR and in the git history.

What do you think?

ShaneHarvey (changes requested):
Could you schedule the mockupdb tests in EVG? It would also be good to run the motor test suite after merging this to see if anything breaks there.

NoahStapp:
> Could you schedule the mockupdb tests in EVG? It would also be good to run the motor test suite after merging this to see if anything breaks there.

mockupdb tests timed out on `test_rsghost`: https://spruce.mongodb.com/version/65bd38bbc9ec443a931ddce7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

ShaneHarvey:
Looks like the test gets stuck attempting to run endSessions when closing the client in test
```
 [2024/02/02 11:19:49.085] Thread 0x00007f64de62f740 (most recent call first):
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/socket_checker.py", line 66 in select
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/network.py", line 321 in wait_for_read
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/network.py", line 340 in _receive_data_on_socket
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/network.py", line 263 in receive_message
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/network.py", line 186 in command
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/pool.py", line 997 in command
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/helpers.py", line 322 in inner
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/mongo_client.py", line 1193 in _end_sessions
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/mongo_client.py", line 1216 in close
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/mongo_client.py", line 2112 in __exit__
 [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/test/mockupdb/test_rsghost.py", line 56 in test_rsghost
```

NoahStapp:
> Looks like the test gets stuck attempting to run endSessions when closing the client in test_rsghost:
> 
> ```
>  [2024/02/02 11:19:49.085] Thread 0x00007f64de62f740 (most recent call first):
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/socket_checker.py", line 66 in select
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/network.py", line 321 in wait_for_read
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/network.py", line 340 in _receive_data_on_socket
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/network.py", line 263 in receive_message
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/network.py", line 186 in command
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/pool.py", line 997 in command
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/helpers.py", line 322 in inner
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/mongo_client.py", line 1193 in _end_sessions
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/mongo_client.py", line 1216 in close
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/pymongo/mongo_client.py", line 2112 in __exit__
>  [2024/02/02 11:19:49.085]   File "/data/mci/1adae7ce73c33f49ebb23345c8812e53/src/test/mockupdb/test_rsghost.py", line 56 in test_rsghost
> ```

Yeah, while sending the `endSessions` commands back to the server.

ShaneHarvey:
test_rsghost needs to set up an auto responder for endSessions. Although I wonder why the endSession was not needed before.

ShaneHarvey:
Previously we never sent sessions to RSGhost servers but now we do. That could be a problem.

NoahStapp:
> Previously we never sent sessions to RSGhost servers but now we do. That could be a problem.

We do now because of the sessions changes? That shouldn't happen, should it?

NoahStapp:
Passing mockupdb tests: https://spruce.mongodb.com/version/65c13ac13e8e86a4392b3080/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

pymongo/pool.py
```diff
@@ -752,6 +752,7 @@ def __init__(
         self.active = False
         self.last_timeout = self.opts.socket_timeout
         self.connect_rtt = 0.0
+        self.logical_session_timeout_minutes: int = 0
```
ShaneHarvey, line 755: Let's default this to None or avoid confusing the 0 and None cases. (It's technically possible to set logical_session_timeout_minutes=0.)

pymongo/pool.py
```diff
@@ -869,7 +870,10 @@ def _hello(
         self.max_bson_size = hello.max_bson_size
         self.max_message_size = hello.max_message_size
         self.max_write_batch_size = hello.max_write_batch_size
-        self.supports_sessions = hello.logical_session_timeout_minutes is not None
+        self.supports_sessions = (
+            hello.logical_session_timeout_minutes is not None and hello.is_readable
+        )
+        self.logical_session_timeout_minutes = hello.logical_session_timeout_minutes or 0
```
ShaneHarvey, line 876: Let's remove the `or 0`.

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM!

PR #1503, opened 2024-02-02
PYTHON-4187 Ensure secrets are not logged in Evergreen

blink1073:


.evergreen/config.yml
```diff
@@ -380,6 +380,7 @@ functions:
       params:
         working_dir: "src"
         shell: bash
+        include_expansions_in_env: true
```
ShaneHarvey, line 383: IIUC this should be a list of vars to avoid pulling in vars we don't need.

ShaneHarvey (changes requested):

.evergreen/config.yml
blink1073: According to the mitigation document, this is an appropriate solution for now.

blink1073: Updating to use add_expansions_to_env, which we have used in several places.  I'll open up a follow-up ticket to make them all explicit.

ShaneHarvey: Wasn't there an issue where add_expansions_to_env caused AWS env vars to be injected thus causing issues in the MONGODB-AWS tests? If not I'm fine with your approach.

Jibola (approved):
LGTM

PR #1504, opened 2024-02-05
PYTHON-4196 Fix import in OIDC Test

blink1073:


caseyclements (approved):
one day i'll get used to starting function names with _ :)

PR #1508, opened 2024-02-05
PYTHON-4182 Skip test_unpin_after_TransientTransactionError_error_on_abort on latest

blink1073:


test/unified_format.py
```diff
@@ -1026,6 +1026,10 @@ def maybe_skip_test(self, spec):
         if "timeoutMS applied to entire download" in spec["description"]:
             self.skipTest("PyMongo's open_download_stream does not cap the stream's lifetime")
 
+        if "unpin after TransientTransactionError error on abort" in spec["description"]:
+            if os.environ.get("MONGODB_VERSION", "") == "latest":
```
ShaneHarvey, line 1030: Can we use client_context.version instead? I don't set MONGODB_VERSION locally when testing.

ShaneHarvey (changes requested):

test/unified_format.py
blink1073: Done

ShaneHarvey (approved):

PR #1511, opened 2024-02-06
PYTHON-3459 Add log messages to Server selection spec

NoahStapp:
For additional context, the [spec](https://github.com/mongodb/specifications/blob/3b8e7e4135f2615fedef1ea6cd2046f5be5a1725/source/server-selection/server-selection.rst#L1239) for server selection logging requires that every time we select a server, we log the operation we plan to send to the server. Since we do not have a unified operations layer (see [PYTHON-1833](https://jira.mongodb.org/browse/PYTHON-1833)), this requires a large number of scattered changes to pass the operation name all the way down to server selection.

The spec tests are quite minimal, presumably because they assume drivers have a unified operations layer that makes logging operation names easy. I've added additional unified format tests for most of the operations we have separate handlers for. Depending on other driver implementations, these tests may be useful--I'll open a DRIVERS ticket to start the discussion for possible inclusion in the spec tests.


pymongo/operations.py
```diff
@@ -44,6 +45,35 @@
 _IndexKeyHint = Union[str, _IndexList]
 
 
+class _Operations(str, enum.Enum):
+    ABORT_TRANSACTION_OP = "abortTransaction"
```
ShaneHarvey, line 49: I'd optimzie for shorter names here. We can ditch the `_OP` suffix since these are all nested a named enum. We could also shorten `_Operations` to `_Op` so these become: `_Op.ABORT`

pymongo/collection.py
```diff
@@ -970,6 +979,7 @@ def replace_one(
                 session=session,
                 let=let,
                 comment=comment,
+                operation="replace",
```
ShaneHarvey, line 982: Does this spec require this to be "replace" or should it instead be "update" (the command name).

pymongo/mongo_client.py
```diff
@@ -1680,10 +1713,10 @@ def _kill_cursors(
         if address:
             # address could be a tuple or _CursorAddress, but
             # select_server_by_address needs (host, port).
-            server = topology.select_server_by_address(tuple(address))  # type: ignore[arg-type]
+            server = topology.select_server_by_address(tuple(address), operation="killCursors")  # type: ignore[arg-type]
         else:
             # Application called close_cursor() with no address.
-            server = topology.select_server(writable_server_selector)
+            server = topology.select_server(writable_server_selector, operation="killCursors")
```
ShaneHarvey, line 1719: Should killCursors be in the Enum?

pymongo/collection.py
NoahStapp: It should be `update`, good catch.

pymongo/bulk.py
```diff
@@ -408,6 +408,7 @@ def execute_command(
         generator: Iterator[Any],
         write_concern: WriteConcern,
         session: Optional[ClientSession],
+        operation: Optional[str] = "TEST_OPERATION",
```
Jibola, line 411: Can we just have this be `_Op.TEST`?

Additionally, for typing, is there a case where we would pass `operation=None` because if not, we could also type this to just be of type `str`.

Jibola: Found a case reading through, so ignore the second part :)

pymongo/database.py
```diff
@@ -786,6 +790,7 @@ def command(
         codec_options: Optional[bson.codec_options.CodecOptions[_CodecDocumentType]] = None,
         session: Optional[ClientSession] = None,
         comment: Optional[Any] = None,
+        operation: Optional[str] = None,
```
Jibola, line 793: We'll have to update the docstring here.

pymongo/database.py
```diff
@@ -881,9 +886,14 @@ def command(
...
 
+        if operation is None and isinstance(command, str):
+            command_name = command
+        else:
+            command_name = next(iter(command))
+
         if read_preference is None:
             read_preference = (session and session._txn_read_preference()) or ReadPreference.PRIMARY
-        with self.__client._conn_for_reads(read_preference, session) as (
+        with self.__client._conn_for_reads(read_preference, session, operation=command_name) as (
```
Jibola, line 896: why do we pass `command_name` here instead of `operation`?

pymongo/bulk.py
ShaneHarvey: Good catch. Agree we should remove "TEST_OPERATION" and make this a required arg.

test/unified_format.py
```diff
@@ -395,13 +401,18 @@ def failed(self, event):
...
+        self, event: ServerDescriptionChangedEvent | TopologyDescriptionChangedEvent
+    ) -> None:
         self.add_event(event)
 
-    def description_changed(self, event: ServerDescriptionChangedEvent) -> None:
+    def topology_changed(self, event: TopologyDescriptionChangedEvent) -> None:
         self.add_event(event)
 
-    def closed(self, event: ServerClosedEvent) -> None:
+    def closed(self, event: ServerClosedEvent | TopologyClosedEvent) -> None:
```
Jibola, line 415: This will all need to use `Union`. iirc, supporting this operator in typing is only [supported in python 3.10+](https://docs.python.org/3/library/typing.html#typing.Union)

Jibola (changes requested):
Great job! Left a few comments

pymongo/bulk.py
NoahStapp: I've made `operation` required everywhere it appears for consistency with the spec and for ease of future refactors/development.

pymongo/collection.py
```diff
@@ -253,15 +253,10 @@ def __init__(
             else:
                 self.__create(name, kwargs, collation, session)
 
-    def _conn_for_reads(
-        self, session: ClientSession
-    ) -> ContextManager[tuple[Connection, _ServerMode]]:
-        return self.__database.client._conn_for_reads(self._read_preference_for(session), session)
-
```
Jibola, line 259: Was this an intentional removal?

Jibola (changes requested):
Great Job. I've got one question on a deleted method, but other than that, LGTM.

pymongo/collection.py
NoahStapp: Yes, that method is unused.

Jibola (approved):

PR #1514, opened 2024-02-08
PYTHON-4018 Clarify exactly what code/label fields drivers should inspect to determine retryability

blink1073:
I had missed updating the legacy test specs in https://github.com/mongodb/mongo-python-driver/pull/1489


blink1073:
Closing since PYTHON-4117 must be done first.

PR #1520, opened 2024-02-13
PYTHON-4167 - Create documentation for standardized logging

NoahStapp:


ShaneHarvey (changes requested):
This is looking good so far! You'll need to add this to the doc tree (in one of the index.rst files) so that it actually gets included in the docs. Should this go under the examples folder?

Also I suggest adding an example of the logging output for a simple app (eg create client, insert_one, find_one).

ShaneHarvey (approved):
LGTM!

PR #1522, opened 2024-02-14
Support tagSets in tests

NoahStapp:


pymongo/logger.py
```diff
@@ -83,8 +83,7 @@ class _ConnectionStatusMessage(str, enum.Enum):
 
 
 def _debug_log(logger: logging.Logger, **fields: Any) -> None:
-    if logger.isEnabledFor(logging.DEBUG):
-        logger.debug(LogMessage(**fields))
```
ShaneHarvey, line 87: Was this change included in your benchmark results from yesterday?

NoahStapp: Yes: https://evergreen.mongodb.com/filediff/65caab760305b9deed94222f/?file_name=pymongo%2Flogger.py&patch_number=0&commit_number=23

PR #1531, opened 2024-02-20
PYTHON-3951: Remove SkipTest on ListSearchIndex tests

Jibola:


test/test_index_management.py
```diff
@@ -218,23 +218,12 @@ def test_case_5(self):
         coll0.drop_search_index("foo")
 
 
-if os.environ.get("TEST_INDEX_MANAGEMENT"):
```
ShaneHarvey, line 221: Can we remove all uses of TEST_INDEX_MANAGEMENT now?

ShaneHarvey (changes requested):

test/test_index_management.py
blink1073: No, because we need the unified tests to run on a localhost, and the prose tests to run on atlas.

ShaneHarvey: I see, thanks.

ShaneHarvey (approved):

PR #1537, opened 2024-02-27
PYTHON-3835 Log informational message client-side based on detected environment

NoahStapp:


NoahStapp:
Both EG test failures are CSOT.

pymongo/uri_parser.py
```diff
@@ -618,6 +621,28 @@ def _parse_kms_tls_options(kms_tls_options: Optional[Mapping[str, Any]]) -> dict
     return contexts
 
 
+def _detect_external_db(entity: str) -> bool:
```
caseyclements, line 624: I know it's not required, but would you mind adding a short docstring for this? Why do we have a fixed set of external db's? Will this keep expanding?

test/test_client.py
```diff
@@ -546,6 +546,28 @@ def test_validate_suggestion(self):
...
                 MongoClient(**{typo: "standard"})  # type: ignore[arg-type]
 
+    def test_dectected_environment_logging(self):
+        normal_hosts = [
+            "normalhost.com",
+            "host.cosmos.azure.com",
+            "host.docdb.amazonaws.com",
+            "host.docdb-elastic.amazonaws.com",
+        ]
+        srv_hosts = [
```
caseyclements, line 556: `srv_hosts = ["mongodb+srv://<test>:<test>@" + s for s in normal_hosts]`

test/test_uri_parser.py
```diff
@@ -521,6 +527,34 @@ def test_special_chars(self):
...
         self.assertEqual(pwd, res["password"])
 
+    def test_detect_external_db(self):
+        hosts = [
+            "normalhost.com",
+            "host.cosmos.AZURE.com",
+            "host.docdb.amazonaws.com",
+            "host.docdb-elastic.amazonaws.com",
+        ]
+        with self.assertLogs("pymongo", level="INFO") as cm:
```
caseyclements, line 537: How does `self.assertLogs("pymongo", level="INFO") `? Seems super useful!

caseyclements (commented):
Looks good. A couple comments.

pymongo/uri_parser.py
```diff
@@ -539,7 +540,9 @@ def parse_uri(
         fqdn, port = nodes[0]
         if port is not None:
             raise InvalidURI(f"{SRV_SCHEME} URIs must not include a port number")
-
+        for host in [node[0] for node in nodes]:
+            if _detect_external_db(host):
```
ShaneHarvey, line 544: Can we remove this and have _detect_external_db only be called once in MongoClient?

pymongo/mongo_client.py
```diff
@@ -785,7 +785,11 @@ def __init__(
                 opts = res["options"]
                 fqdn = res["fqdn"]
             else:
-                seeds.update(uri_parser.split_hosts(entity, port))
+                nodes = uri_parser.split_hosts(entity, port)
+                for host in [node[0] for node in nodes]:
+                    if uri_parser._detect_external_db(host):
+                        break
```
ShaneHarvey, line 791: Can you add a test case for the "break" logic? ie that we only log once even when multiple hosts are given?

pymongo/uri_parser.py
NoahStapp: The current set of external DBs is specified in the ticket, but I expect that list to expand with time.

test/test_uri_parser.py
NoahStapp: Here's the official docs: https://docs.python.org/3.11/library/unittest.html#unittest.TestCase.assertLogs. It's really helpful for testing our logging behavior, especially now that we have several components that log.

ShaneHarvey (changes requested):

pymongo/mongo_client.py
NoahStapp: Good catch, will do!

pymongo/uri_parser.py
NoahStapp: Unless we want to refactor our SRV/URI logic to always pull out hosts as a separate step in MongoClient.

NoahStapp: The ticket specifies that we do so: https://jira.mongodb.org/browse/PYTHON-3835.

ShaneHarvey: I'd prefer to ignore that guidance. The practical issue is that for MongoClient() we want the warning.warn to have the correct stacklevel so that the warning is reported for the line that creates the client.

caseyclements (approved):
With comments replied to, I'm happy to approve.

pymongo/uri_parser.py
NoahStapp: I don't quite understand how the stacklevel for `warning.warn` relates to the SRV lookup timing, can you elaborate?

ShaneHarvey: The stacklevel for the uri parser will need to be different than the stack level for the client. So the warning will show up differently if you pass a mongodb:// uri or a mongodb+srv:// uri which seems odd. Ideally the warning is always in once place and the stack level make it so that it gets reported on the line that created the MongoClient().

ShaneHarvey: For example play around with a program like this:
```
client1 = MongoClient("mongodb://...")
client2 = MongoClient("mongodb+srv://...")
```

And see what warnings are reported.

NoahStapp: This does make testing more difficult, as now we need an actual live host for the resolver to hit. That doesn't seem like a solvable issue for a unit or even an integration test.

pymongo/mongo_client.py
```diff
@@ -785,10 +786,15 @@ def __init__(
                 opts = res["options"]
                 fqdn = res["fqdn"]
             else:
-                seeds.update(uri_parser.split_hosts(entity, port))
+                nodes = uri_parser.split_hosts(entity, port)
+                seeds.update(nodes)
```
ShaneHarvey, line 790: Can this change be reverted unless it was intentional?

pymongo/uri_parser.py
```diff
@@ -539,7 +539,6 @@ def parse_uri(
         fqdn, port = nodes[0]
         if port is not None:
             raise InvalidURI(f"{SRV_SCHEME} URIs must not include a port number")
-
```
ShaneHarvey, line 542: Unneeded change here.

ShaneHarvey (changes requested):
Could you add a test for the warning.warn behavior too? Also can you show an example of the UserWarning behavior? I want to ensure the stack level corresponds to the MongoClient line.

NoahStapp:
> Could you add a test for the warning.warn behavior too? Also can you show an example of the UserWarning behavior? I want to ensure the stack level corresponds to the MongoClient line.

This is what it looks like:

```
/Users/nstapp/Github/mongo-python-driver/pymongo/mongo_client.py:2503: UserWarning: You appear to be connected to a CosmosDB cluster. For more information regarding feature compatibility and support please visit https://www.mongodb.com/supportability/cosmosdb
  _log_or_warn(
```

ShaneHarvey (changes requested):
```

Could you add a test for the warning.warn behavior too? Also can you show an example of the UserWarning behavior? I want to ensure the stack level corresponds to the MongoClient line.

This is what it looks like:

/Users/nstapp/Github/mongo-python-driver/pymongo/mongo_client.py:2503: UserWarning: You appear to be connected to a CosmosDB cluster. For more information regarding feature compatibility and support please visit https://www.mongodb.com/supportability/cosmosdb
  _log_or_warn(
```
This is problematic, the stack level needs to be increased so that the warning is raised on the line of code that creates the MongoClient, not on pymongo's internal code.

NoahStapp:
> ```
> /Users/nstapp/Github/mongo-python-driver/pymongo/mongo_client.py:2503: UserWarning: You appear to be connected to a CosmosDB cluster. For more information regarding feature compatibility and support please visit https://www.mongodb.com/supportability/cosmosdb
>   _log_or_warn(
> ```
> 
> This is problematic, the stack level needs to be increased so that the warning is raised on the line of code that creates the MongoClient, not on pymongo's internal code.

Fixed, `stacklevel=4` ensures we get:

```
/Users/nstapp/Github/mongo-python-driver/dns_test.py:34: UserWarning: You appear to be connected to a DocumentDB cluster. For more information regarding feature compatibility and support please visit https://www.mongodb.com/supportability/documentdb
  c = MongoClient("host.docdb-elastic.amazonaws.com")
```

ShaneHarvey (commented):
One last piece is to describe this new behavior in the changelog.

doc/changelog.rst
```diff
@@ -32,6 +32,7 @@ PyMongo 4.7 brings a number of improvements including:
 
 - Fixed a bug appearing in Python 3.12 where "RuntimeError: can't create new thread at interpreter shutdown"
   could be written to stderr when a MongoClient's thread starts as the python interpreter is shutting down.
+- Added a warning when connecting to DocumentDB and CosmosDB clusters.
```
ShaneHarvey, line 35: Can you add this "For more information regarding feature compatibility and support please visit https://www.mongodb.com/supportability/documentdb and https://www.mongodb.com/supportability/cosmosdb."

ShaneHarvey (approved):
LGTM

PR #1548, opened 2024-03-12
PYTHON-4270 Run typing tests on Python 3.7

blink1073:


.github/workflows/test-python.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.7"
```
Jibola, line 91: should we test on both 3.7 and 3.11?

Jibola (approved):
LGTM

.github/workflows/test-python.yml
blink1073: Done

NoahStapp (approved):

PR #1554, opened 2024-03-18
PYTHON-4286 Remove duplicate test_detect_external_db

NoahStapp:


ShaneHarvey (approved):

ShaneHarvey (changes requested):
We need to remove the _detect_external_db import as well.

ShaneHarvey (approved):

PR #1555, opened 2024-03-18
PYTHON-4271 Skip failing serverless proxy tests

NoahStapp:


test/load_balancer/sdam-error-handling.json
```diff
@@ -261,6 +261,7 @@
     },
     {
       "description": "errors during the initial connection hello are ignored",
+      "skipReason": "waiting on CLOUDP-202309",
```
ShaneHarvey, line 264: Can we skip these tests only on the Serverless Proxy? They are still useful to run on all other variants.

ShaneHarvey (changes requested):

test/load_balancer/sdam-error-handling.json
NoahStapp: Is it possible to identify when a client context is for a serverless proxy and not just a serverless instance?

blink1073: We'd have to feed it in as an environment variable.   Here's where we feed it through: https://github.com/mongodb/mongo-python-driver/blob/a46331d478d68fca68bb9f85302fe411373f75b6/.evergreen/config.yml#L905

NoahStapp: Got it, there's no way to distinguish at the client level between proxy and non-proxy.

ShaneHarvey: In that case it would still be good to keep running these tests on all non-serverless clusters.

blink1073: Can't we skip in in the unified test runner, where we can inspect an environment variable?

NoahStapp: Sorry let me clarify--yes we can skip these using an environment variable, I was commenting that we don't have a cleaner way of doing so.

ShaneHarvey (approved):

PR #1558, opened 2024-03-22
PYTHON-4294 Add Note on Dict Ordering

zakarybk:


blink1073 (approved):
Thank you!

ShaneHarvey (commented):
We should remove this note altogether since we only support 3.7+ now.

zakarybk:
> We should remove this note altogether since we only support 3.7+ now.

Will this affect the documentation when selecting older Python versions?

PR #1560, opened 2024-03-23
PYTHON-4285 More consistent PyModule_GetState checks

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4285

### DIFF
--- a/pymongo/_cmessagemodule.c
+++ b/pymongo/_cmessagemodule.c
@@ -68,8 +68,6 @@ static int buffer_write_bytes_ssize_t(buffer_t buffer, const char* data, Py_ssiz
 
 static PyObject* _cbson_query_message(PyObject* self, PyObject* args) {
     /* NOTE just using a random number as the request_id */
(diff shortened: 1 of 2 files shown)


pymongo/_cmessagemodule.c
```diff
@@ -68,8 +68,6 @@ static int buffer_write_bytes_ssize_t(buffer_t buffer, const char* data, Py_ssiz
 
 static PyObject* _cbson_query_message(PyObject* self, PyObject* args) {
     /* NOTE just using a random number as the request_id */
-    struct module_state *state = GETSTATE(self);
-
```
Jibola, line 72: why do we move this down?

Jibola (approved):
LGTM

pymongo/_cmessagemodule.c
ShaneHarvey: So that the error check can happen right next to the definition. We're still using C89 I think which requires all variable declarations at the top of the function before any code.

PR #1561, opened 2024-03-25
PYTHON-4297 Allow passing arbitrary index options to create_search_index

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4297

### DIFF
--- a/pymongo/operations.py
+++ b/pymongo/operations.py
@@ -593,7 +593,7 @@ class SearchIndexModel:
         definition: Mapping[str, Any],
         name: Optional[str] = None,
         type: Optional[str] = "search",
-        **kwargs,
+        **kwargs: Any,
     ) -> None:
         """Create a Search Index instance.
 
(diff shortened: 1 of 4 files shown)


Jibola (approved):

pymongo/operations.py
```diff
@@ -593,23 +593,28 @@ def __init__(
         definition: Mapping[str, Any],
         name: Optional[str] = None,
         type: Optional[str] = "search",
```
NoahStapp, line 595: Do we want to remove this default?

ShaneHarvey: I'll ask that question in https://github.com/mongodb/specifications/pull/1541.

NoahStapp (approved):

ShaneHarvey:
Updated to fix these typing failures:
```
pymongo/operations.py:591: error: Function is missing a type annotation for one
or more arguments  [no-untyped-def]
        def __init__(
        ^
pymongo/operations.py:615: error: Incompatible types in assignment (expression
has type "Mapping[str, Any]", target has type "str")  [assignment]
            self.__document["definition"] = definition
                                            ^
pymongo/operations.py:616: error: Incompatible types in assignment (expression
has type "Optional[str]", target has type "str")  [assignment]
            self.__document["type"] = type
                                      ^
```

PR #1565, opened 2024-03-28
PYTHON-4285 Fix PyModule_GetState check

ShaneHarvey:
Fixes a mistake in PYTHON-4285. Note there's still no bug here since `state` will never be null. We are adding the check because it's harmless and Coverity flagged it (even though it's a false positive).


blink1073 (approved):
LGTM!

PR #1570, opened 2024-04-02
PYTHON-4037 Avoid Appending Write/Read Concern in Atlas Search Index Helper Commands

blink1073:


pymongo/collection.py
```diff
@@ -2364,7 +2364,10 @@ def list_search_indexes(
             pipeline = [{"$listSearchIndexes": {"name": name}}]
 
         coll = self.with_options(
-            codec_options=DEFAULT_CODEC_OPTIONS, read_preference=ReadPreference.PRIMARY
+            codec_options=DEFAULT_CODEC_OPTIONS,
+            read_preference=ReadPreference.PRIMARY,
+            write_concern=WriteConcern(),
+            read_concern=ReadConcern(),
```
ShaneHarvey, line 2370: Let's use the DEFAULT_READ_CONCERN/DEFAULT_WRITE_CONCERN objects.

ShaneHarvey (changes requested):
Let's add a changelog for this bugfix.

pymongo/collection.py
blink1073: Done

ShaneHarvey (approved):

PR #1579, opened 2024-04-11
PYTHON-4349 Fix snappy support on PyPy 3.8

blink1073:


blink1073:
Passing build here: https://spruce.mongodb.com/version/6617fc94a5cc2000073d2344/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC%3BNAME%3AASC

pyproject.toml
```diff
@@ -71,7 +71,8 @@ ocsp = [
     "service_identity>=18.1.0",
 ]
 snappy = [
-    "python-snappy",
+    "python-snappy;implementation_name!='pypy' or implementation_version>'7.3.11'",
+    "python-snappy<0.7.0;implementation_name=='pypy' and implementation_version<='7.3.11'"
```
ShaneHarvey, line 75: We should make this change only in the test suite. That's what we usually do for issues like this. That way if python-snappy fixes this bug then we won't need to update pymongo.

ShaneHarvey (changes requested):

pyproject.toml
blink1073: Updated

ShaneHarvey (approved):

PR #1584, opened 2024-04-12
PYTHON-4354 & PYTHON-4355 - Increase logging limit for tests

NoahStapp:
The green framework failures were actually due to `ocsp_support` logging non-JSON messages. Our test framework expects all logs to be in JSON format, so I've added a check to exclude `ocsp_support` logs from test verification.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -129,6 +129,10 @@ GCP_CREDS = {
 }
 KMIP_CREDS = {"endpoint": os.environ.get("FLE_KMIP_ENDPOINT", "localhost:5698")}
 
+os.environ[
+    "MONGOB_LOG_MAX_DOCUMENT_LENGTH"
+] = "2000"  # Ensure Evergreen metadata doesn't result in truncation
+
 
 def is_server_resolvable():
     """Returns True if 'server' is resolvable."""
--- a/test/unified_format.py
+++ b/test/unified_format.py
@@ -1770,15 +1770,16 @@ class UnifiedSpecTestMixinV1(IntegrationTest):
         def format_logs(log_list):
             client_to_log = defaultdict(list)
             for log in log_list:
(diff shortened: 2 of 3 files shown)


blink1073 (approved):
LGTM!

test/unified_format.py
```diff
@@ -1770,15 +1770,16 @@ def check_log_messages(self, operations, spec):
...
-                data = json_util.loads(log.message)
-                client = data.pop("clientId")
-                client_to_log[client].append(
-                    {
-                        "level": log.levelname.lower(),
-                        "component": log.name.replace("pymongo.", "", 1),
-                        "data": data,
-                    }
-                )
+                if log.module != "ocsp_support":
```
ShaneHarvey, line 1773: Suggest writing like this to avoid code churn:
```python
if log.module == "ocsp_support":
    continue
```

test/__init__.py
```diff
@@ -129,6 +129,10 @@
 }
 KMIP_CREDS = {"endpoint": os.environ.get("FLE_KMIP_ENDPOINT", "localhost:5698")}
 
+os.environ[
+    "MONGOB_LOG_MAX_DOCUMENT_LENGTH"
+] = "2000"  # Ensure Evergreen metadata doesn't result in truncation
```
ShaneHarvey, line 134: Suggest using setdefault so that we can still set this manually. For example I recently found it useful to set MONGOB_LOG_MAX_DOCUMENT_LENGTH when running the perf tests.

ShaneHarvey (changes requested):

test/unified_format.py
ShaneHarvey: Eh too late I guess.

NoahStapp: I can open another, I agree your suggestions make for cleaner code.

PR #1591, opened 2024-04-16
PYTHON-3984 Deprecate WriteConcern.wtimeout in favor of pymongo.timeout()

blink1073:


pymongo/write_concern.py
```diff
@@ -38,9 +38,9 @@ class WriteConcern:
           replicated to **two** secondaries). **w=0 disables acknowledgement
           of write operations and can not be used with other write concern
           options.**
-    :param wtimeout: (integer) Used in conjunction with `w`. Specify a value
-          in milliseconds to control how long to wait for write propagation
-          to complete. If replication does not complete in the given
+    :param wtimeout: (integer) **DEPRECATED** Used in conjunction with `w`.
```
ShaneHarvey, line 41: Let's also say MongoClient's wTimeoutMS option is deprecated too.

ShaneHarvey (changes requested):

pymongo/write_concern.py
blink1073: Done

ShaneHarvey (approved):

PR #1593, opened 2024-04-16
PYTHON-4057 Emit DeprecationWarning for deprecated GridFS apis

blink1073:


blink1073:
As expected, tests fail with messages like: `DeprecationWarning: GridIn property 'md5' is deprecated and will be removed in PyMongo 5.0`.  I'll add these to the known warnings.

NoahStapp (approved):

PR #1600, opened 2024-04-17
PYTHON-2048 Improve error message for bulk_write failures due to unsupported storage enginge

blink1073:
```python
>>> client.test.foo.insert_many(data)
  File "/Users/steve.silvester/workspace/mongo-python-driver/pymongo/_csot.py", line 108, in csot_wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/steve.silvester/workspace/mongo-python-driver/pymongo/collection.py", line 742, in insert_many
    blk.execute(write_concern, session, _Op.INSERT)
  File "/Users/steve.silvester/workspace/mongo-python-driver/pymongo/bulk.py", line 595, in execute
    return self.execute_command(generator, write_concern, session, operation)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/steve.silvester/workspace/mongo-python-driver/pymongo/bulk.py", line 462, in execute_command
    _raise_bulk_write_error(full_result)
  File "/Users/steve.silvester/workspace/mongo-python-driver/pymongo/bulk.py", line 164, in _raise_bulk_write_error
    raise OperationFailure(errmsg, code, full_result)
pymongo.errors.OperationFailure: This MongoDB deployment does not support retryable writes. Please add retryWrites=false to your connection string., full error: {'writeErrors': [{'index': 0, 'code': 20, 'codeName': 'IllegalOperation', 'errmsg': 'Transaction numbers are only allowed on storage engines that support document-level locking', 'op': {'a': 1, '_id': ObjectId('661f32dd2507a26812abf937')}}], 'writeConcernErrors': [], 'nInserted': 0, 'nUpserted': 0, 'nMatched': 0, 'nModified': 0, 'nRemoved': 0, 'upserted': []}
```

### DIFF
--- a/pymongo/bulk.py
+++ b/pymongo/bulk.py
@@ -149,8 +149,19 @@ def _merge_command(
 
 def _raise_bulk_write_error(full_result: _DocumentOut) -> NoReturn:
     """Raise a BulkWriteError from the full bulk api result."""
+    # retryWrites on MMAPv1 should raise an actionable error.
     if full_result["writeErrors"]:
         full_result["writeErrors"].sort(key=lambda error: error["index"])
+        err = full_result["writeErrors"][0]
+        code = err["code"]
+        msg = err["errmsg"]
+        if code == 20 and msg.startswith("Transaction numbers"):
+            errmsg = (
+                "This MongoDB deployment does not support "
+                "retryable writes. Please add retryWrites=false "
+                "to your connection string."
+            )
+            raise OperationFailure(errmsg, code, full_result)
     raise BulkWriteError(full_result)
 
 


pymongo/bulk.py
```diff
@@ -149,8 +149,19 @@ def _merge_command(
...
+        err = full_result["writeErrors"][0]
+        code = err["code"]
+        msg = err["errmsg"]
+        if code == 20 and msg.startswith("Transaction numbers"):
+            errmsg = (
+                "This MongoDB deployment does not support "
+                "retryable writes. Please add retryWrites=false "
+                "to your connection string."
+            )
+            raise OperationFailure(errmsg, code, full_result)
```
ShaneHarvey, line 164: We should raise BulkWriteError here instead of OperationFailure. Each writeError with "Transaction numbers" + code 20 we can replace the errmsg  be replaced with "This MongoDB deployment does support...".

ShaneHarvey: Actually I was incorrect. I was testing the behavior of mmapv1 but this new code handles a different case, when a shard is a standalone mongod. I added a test case here: https://github.com/mongodb/mongo-python-driver/pull/1606

PR #1603, opened 2024-04-17
PYTHON-4227 Unified tests: Advance cluster_time of ClientSessions after initialData creation

caseyclements:
## Changes

After initialData is added in test.unified_format's run_scenario, we get the $clusterTime from the server, and add this to each of the ClientSession entities before further operations.

## Motivation for the change.

[DRIVERS-2816](https://jira.mongodb.org/browse/DRIVERS-2816)

[Here is the pertinent section of the specifications](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md?plain=1#L2987-L2998)

### DIFF
--- a/test/unified_format.py
+++ b/test/unified_format.py
@@ -1883,7 +1883,8 @@ class UnifiedSpecTestMixinV1(IntegrationTest):
         self.entity_map.create_entities_from_spec(self.TEST_SPEC.get("createEntities", []), uri=uri)
         # process initialData
         if "initialData" in self.TEST_SPEC:
-            self.insert_initial_data(self.TEST_SPEC.get("initialData", []))
+            self.insert_initial_data(self.TEST_SPEC["initialData"])
+            self._cluster_time = self.client.admin.command("ping").get("$clusterTime")
             self.entity_map.advance_cluster_times()
 
         if "expectLogMessages" in spec:
--- a/test/unified_format.py
+++ b/test/unified_format.py
@@ -428,11 +428,11 @@ class EntityMapUtil:
     """
 
     def __init__(self, test_class):
-        self._cluster_time = None
         self._entities: Dict[str, Any] = {}
         self._listeners: Dict[str, EventListenerUtil] = {}
         self._session_lsids: Dict[str, Mapping[str, Any]] = {}
         self.test: UnifiedSpecTestMixinV1 = test_class
+        self._cluster_time: Optional[ClusterTime] = None
 
     def __contains__(self, item):
         return item in self._entities
@@ -625,10 +625,9 @@ class EntityMapUtil:
             # session has been closed.
             return self._session_lsids[session_name]
 
-    def advance_cluster_times(self, cluster_time: Optional[ClusterTime] = None):
-        if cluster_time is not None:
-            self._cluster_time = cluster_time
-        elif getattr(self, "_cluster_time", None) is None:
+    def advance_cluster_times(self):
+        """Manually synchronize entities when desired"""
+        if self._cluster_time is None:
             self._cluster_time = self.test.client.admin.command("ping").get("$clusterTime")
         for entity in self._entities:
             if isinstance(entity, ClientSession):
--- a/test/unified_format.py
+++ b/test/unified_format.py
@@ -629,7 +629,7 @@ class EntityMapUtil:
         """Manually synchronize entities when desired"""
         if not self._cluster_time:
             self._cluster_time = self.test.client.admin.command("ping").get("$clusterTime")
-        for entity in self._entities:
+        for entity in self._entities.values():
             if isinstance(entity, ClientSession):
                 entity.advance_cluster_time(self._cluster_time)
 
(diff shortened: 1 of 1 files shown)


caseyclements:
I'm investigating the new failure that popped up. I can reproduce locally.

```
FAILURE: pymongo.errors.OperationFailure: PlanExecutor error during aggregation :: caused by :: indexes of target collection db0.coll0 changed during processing., 

self = <test.test_crud_unified.TestUnifiedDbAggregateWriteReadPreference testMethod=test_Database-level_aggregate_with_$out_includes_read_preference_for_5_0+_server>
```

test/unified_format.py
```diff
@@ -1880,7 +1872,15 @@ def _run_scenario(self, spec, uri=None):
...
-        self.insert_initial_data(self.TEST_SPEC.get("initialData", []))
+        if "initialData" in self.TEST_SPEC:
+            self.insert_initial_data(self.TEST_SPEC.get("initialData", []))
+            # advance cluster times of session entities,
+            # to ensure consistency in transactions against a sharded deployment,
+            cluster_time = self.client.admin.command("ping").get("$clusterTime")
+            if cluster_time:
+                for entity in self.entity_map._entities.values():
+                    if isinstance(entity, ClientSession):
+                        entity.advance_cluster_time(cluster_time)
```
ShaneHarvey, line 1883: We need to do this for sessions created later on in `_testOperation_createEntities` too.

blink1073 (dismissed):
LGTM!

test/unified_format.py
```diff
@@ -1880,7 +1872,15 @@ def _run_scenario(self, spec, uri=None):
...
         self.entity_map.create_entities_from_spec(self.TEST_SPEC.get("createEntities", []), uri=uri)
         # process initialData
-        self.insert_initial_data(self.TEST_SPEC.get("initialData", []))
+        if "initialData" in self.TEST_SPEC:
+            self.insert_initial_data(self.TEST_SPEC.get("initialData", []))
+            # advance cluster times of session entities,
+            # to ensure consistency in transactions against a sharded deployment,
+            cluster_time = self.client.admin.command("ping").get("$clusterTime")
+            if cluster_time:
+                for entity in self.entity_map._entities.values():
```
ShaneHarvey, line 1881: Suggest adding a EntityMapUtil.values() method that we can call here so we don't need to look at the internals.

ShaneHarvey (dismissed):

test/unified_format.py
caseyclements: Where is `_testOperation_createEntities` called from? Would we want to ping again before updating?

ShaneHarvey: We should only run ping once, store the cluster_time, then use it later when _testOperation_createEntities is called.

ShaneHarvey (commented):
LGTM once the typing issue is fixed.

caseyclements:
@ShaneHarvey, @blink1073 `test.test_transactions_unified.TestUnifiedMongosUnpin.test_unpin_after_TransientTransactionError_error_on_abort` is still failing intermittently. This PR had removed the skip. Should I put it back on?

blink1073:
> Should I put it back on?

Yeah, apparently https://jira.mongodb.org/browse/PYTHON-4182 was unrelated, I unblocked it and put it in scheduled.

blink1073 (dismissed):
LGTM!

test/unified_format.py
```diff
@@ -624,6 +625,14 @@ def get_lsid_for_session(self, session_name):
             # session has been closed.
             return self._session_lsids[session_name]
 
+    def advance_cluster_times(self) -> None:
+        """Manually synchronize entities when desired"""
+        if not self._cluster_time:
+            self._cluster_time = self.test.client.admin.command("ping").get("$clusterTime")
+        for entity in self._entities:
```
NoahStapp, line 632:
```suggestion
        for entity in self._entities.values():
```

The current version iterates over the names instead of the values and never passes the `isinstance` check.

test/unified_format.py
```diff
@@ -1050,17 +1059,11 @@ def maybe_skip_test(self, spec):
...
-            self.skipTest("Skipping TransientTransactionError pending PYTHON-4227")
-        if "withTransaction commits after callback returns" in spec["description"]:
-            self.skipTest("Skipping TransientTransactionError pending PYTHON-4303")
-        if "unpin on successful abort" in spec["description"]:
-            self.skipTest("Skipping TransientTransactionError pending PYTHON-4227")
-
         if "unpin after non-transient error on abort" in spec["description"]:
             if client_context.version[0] == 8:
                 self.skipTest("Skipping TransientTransactionError pending PYTHON-4182")
+        if "unpin after TransientTransactionError error on" in spec["description"]:
```
NoahStapp, line 1065: Is this intended? The failure is the same for [PYTHON-4182](https://jira.mongodb.org/browse/PYTHON-4182), so I would expect this PR to fix it as well.

NoahStapp (changes requested):

test/unified_format.py
caseyclements: I can try it out. The failure was popping back up earlier but now that everything is cleaner, and I've synced everything up to master, it is worth trying again.

blink1073:
Merging so we can check the waterfall afterward.

PR #1604, opened 2024-04-17
Revert "PYTHON-2048 Improve error message for bulk_write failures due to unsupported storage enginge"

blink1073:
Reverts mongodb/mongo-python-driver#1600


ShaneHarvey:
Okay it turns out that I was mistaken. Looks like we need this code to handle the standalone shard case described in SERVER-44591. I'll add a mockupdb test for it.

ShaneHarvey:
Opened: https://github.com/mongodb/mongo-python-driver/pull/1606

PR #1613, opened 2024-04-22
PYTHON-4363 Add dependabot config for Python

blink1073:
Dependabot will check all of our requirements*.txt files once a week and raise PRs if it sees any CVEs that need to be addressed.  The PR description will show the linked vulnerability for authenticated, and just the version bump publicly.


caseyclements (approved):
LGTM

PR #1618, opened 2024-04-24
PYTHON-4401 Fix make_sdist job

blink1073:


blink1073:
We need to run these tests on PRs so we don't get blindsided during a release again.

NoahStapp (approved):

PR #1619, opened 2024-04-24
PYTHON-4402 Remove dead links

blink1073:


caseyclements (changes requested):
Still needs formatting and such, right?

caseyclements (approved):
Assuming test_encryption failures arr unrelated, good to go.

PR #1621, opened 2024-04-26
PYTHON-3601 OIDC: Clarify TOKEN_RESOURCE and client_id usage

blink1073:


doc/examples/authentication.rst
```diff
@@ -408,8 +408,10 @@ Azure IMDS
 ^^^^^^^^^^
 
 For an application running on an Azure VM or otherwise using the `Azure Internal Metadata Service`_,
-you can use the built-in support for Azure, where "<client_id>" below is the client id of the Azure
-managed identity, and ``<audience>`` is the url-encoded ``audience`` `configured on your MongoDB deployment`_.
+you can use the built-in support for Azure. If using an Azure Managed Identity, the "<client_id>" is
+the Client ID.  If using a Service Principal (such as an Enterprise Application), the "<client_id>" is
+Application Id of the Enterprise Application.  The ``<audience>`` value is the ``audience``
+`configured on your MongoDB deployment`_.
```
mongoKart, line 414: I got the sense from this that managed identities and service principals are mutually exclusive, but [Microsoft classifies managed identity as a type of service principal.](https://learn.microsoft.com/en-us/entra/identity-platform/app-objects-and-service-principals?tabs=browser)

Something like this might be clearer (also removed caps to follow MS conventions):

```suggestion
you can use the built-in support for Azure. If using an Azure managed identity, the "<client_id>" is
the client ID.  If using a service principal to represent an enterprise application, the "<client_id>" is
the application ID of the service principal. The ``<audience>`` value is the ``audience``
`configured on your MongoDB deployment`_.
```

PR #1624, opened 2024-04-29
PYTHON-4409 Streamline Wheel Building

blink1073:


.github/workflows/release-python.yml
```diff
@@ -29,43 +29,25 @@ jobs:
...
-            - buildplat: [macos-11, macosx_*]
-              python: "cp38"
-            - buildplat: [macos-11, macosx_*]
-              python: "cp39"
-            - buildplat: [macos-14, macosx_*]
-              python: "cp310"
-            - buildplat: [macos-14, macosx_*]
-              python: "cp311"
-            - buildplat: [macos-14, macosx_*]
-              python: "cp312"
```
Jibola, line 52: Why can we get away with not specifying the python version anymore Is It that the way we build wheels now will automatically create wheels packages for 3.7 - 3.12?

.github/workflows/release-python.yml
```diff
@@ -79,75 +61,37 @@ jobs:
...
-          python -m pip install "cibuildwheel>=2.4,<3"
+          python -m pip install "cibuildwheel>=2.17,<3"
 
-      - name: Build wheel
-        if: ${{ matrix.buildplat[0] != 'macos-14' }}
+      - name: Build wheels
         env:
-          CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
-          MACOS_TEST_SKIP: "*arm64"
+          CIBW_BUILD: ${{ matrix.buildplat[2] }}
```
Jibola, line 68: Should this step exclude manylinux1?

Jibola (approved):
LGTM. 
Left questions just to improve my understanding.

.github/workflows/release-python.yml
blink1073: Yes, and we verify it later. :)

.github/workflows/release-python.yml
blink1073: No need to exclude it.  We have a separate manylinux1 to ensure that we have wheels that include that arch, but it is fine if we have duplicate wheels.

PR #1629, opened 2024-05-02
PYTHON-4264 Async PyMongo Beta

NoahStapp:
### Remaining work to be done before beta release (in no particular order):

- [x] Fix `no module named pymongo.asynchronous/synchronous` errors in release scripts
- [x] Fix `slotscheck` `ERROR: Module 'grid_file' not found` error
- [x] Automate running import sorting on synchronous files after generation
- [x] Fix doc links (either change all import paths or export sync/async modules at the top level of the package)
- [x] Automate `synchro.py` running as part of pre-commit hook
- [x] Add async test suite (likely a mix of manual and automated at first until we can write a whole async test runner)
- [x] Add docstring translation from async to sync

Since there are over 30k lines changed, reviewing the changes line-by-line is not feasible. I've added several comments to files in this PR to explain choices or the important parts to look over closely, but we'll have to rely on our test suite to do the bulk of ensuring correctness. 

Additionally, as shown by the list above, there is still remaining work to be done before the beta is fully ready for release. I've elected to mark this PR as ready for review now to give people plenty of time now that the tests are fully passing. There aren't any planned fundamental changes remaining, only fixing the last few parts of our infrastructure and related tooling.

I plan to start writing up the design document this week, which should help answer any additional questions raised as part of the review process. 

### For reviewers
The most important code to look at closely is `pymongo/network.py`, `pymongo/asynchronous/mongo_client.py`, `pymongo/asynchronous/cursor.py`, and `pymongo/asynchronous/command_cursor.py`. The first is one of the only low-level components that contains both async and sync code, while the others all use `IS_SYNC` for behavior unsupported by magic methods like `__init__`. Reading those files will give a good understanding of how the rest of the async code is structured. 

The goal of the `asynchronous` sub-module is to be at feature parity with Motor, while the `synchronous` sub-module should be at parity with the existing PyMongo. Our test suite should cover almost all of the work of ensuring parity, so my biggest concern is code readability and maintainability. 

All of our work going forward will have to be asynchronous, adding another layer of complexity and verbosity. Minimizing this additional overhead is a priority whenever possible, especially since it can be unclear when a method does or does not need to be asynchronous.


pymongo/network.py
NoahStapp, line 1: `pymongo/network.py` has two separate sets of socket + network IO implementations: one for async, one for sync. All code above this layer is effectively just colored async or sync, with no difference in code structure or flow.

As a result, ensuring that `pymongo/network.py` is maintainable and understandable is the most critical component of the async migration. A line-by-line review of this file specifically is worthwhile.

pymongo/asynchronous/aggregation.py
NoahStapp, line 1: Every file inside the `asynchronous` sub-module has a `IS_SYNC = False` definition near the top. This is mirrored by `IS_SYNC = True` definitions in the `synchronous` sub-module. This allows us to check for allowing different behavior between async and sync versions of the same code.

These values should only be used when the async and sync codebases require different code or control flow besides the addition/removal of `async/await` keywords.

pymongo/__init__.py
NoahStapp, line 1: We currently expose `AsyncMongoClient` as the only public API for the async codebase. If we expect users to use any other async APIs, we will need to expose them here as well.

pymongo/periodic_executor.py
NoahStapp, line 1: `PeriodicExecutor` can take either a `coroutine` or a normal function as its `target`. Coroutines are run using `asyncio.run` to avoid halting on any unintended blocking of the existing asyncio loop.

pymongo/pyopenssl_context.py
NoahStapp, line 1: `PyOpenSSL` does not have an async API, forcing us to run synchronous code inside an asyncio executor loop.

test/__init__.py
NoahStapp, line 1: We currently test only the synchronous code for parity with the existing synchronous PyMongo release. We'll need to also test the asynchronous code going forward. Our current plan is to add this test suite during the beta testing period. One proposal is to rewrite the test suite as async and translate it using unasync to sync, mirroring the actual PyMongo codebase.

mypy_test.ini
```diff
@@ -6,3 +6,10 @@ exclude = (?x)(
     ^test/mypy_fails/*.*$
     | ^test/conftest.py$
   )
+
+[mypy-pymongo.synchronous.*,gridfs.synchronous.*]
+warn_unused_ignores = false
+disable_error_code = unused-coroutine
```
NoahStapp, line 12: Several places use `IS_SYNC` to call either the sync or async version of a network-level operation. This causes linting errors in the synchronous codebase due to the async coroutine for the `IS_SYNC = False` path missing an await.

pymongo/asynchronous/helpers.py
NoahStapp, line 1: This file and several others don't contain any `async` code, but still import files that do for either type hints or helper methods/constants. Moving them outside of the `asynchronous` sub-module causes typing issues and test failures due to `pymongo.asynchronous.ClassName` not being identical to `pymongo.synchronous.ClassName`.

synchro.py
NoahStapp, line 1: `unasync` is a very small and simple library that effectively just does token replacement for `async/await` keywords and known equivalents like `__anext__ -> __next__`. The names inside `replacements` here are mapped from async to sync on a token-by-token basis. 

This relatively limited functionality means any asynchronous code that uses async-only features that do not have a synchronous parallel will require an `IS_ASYNC` block to implement the synchronous version.

pymongo/asynchronous/helpers.py
```diff
@@ -0,0 +1,317 @@
...
+                    await conn.authenticate(reauthenticate=True)
+                else:
+                    raise
+                return func(*args, **kwargs)
+            raise
+
+    return cast(F, inner)
+
+
+async def anext(cls: Any) -> Any:
```
NoahStapp, line 315: `anext` was added in Python 3.10, so we need to use our own version here for compatibility with Python 3.8 and 3.9. `anext` is one of the special names that `unasync` automatically converts to the sync version, which is `next` in this case.

pymongo/pyopenssl_context.py
blink1073: I think it should be possible to use `PyOpenSSL` without using an explicit loop, can you add this as a follow-on item?

pymongo/periodic_executor.py
blink1073: Since this is running in a dedicated thread, there will be no existing asyncio loop, but `asyncio.run` is still appropriate.

pymongo/__init__.py
blink1073: If we're targeting 4.x, we'll need to preserve the existing modules that are documented in our API docs.  We'll need to discuss this in our design doc.

pymongo/network.py
blink1073: It is hard to tell what parts are duplicated in the top level functions.  I think we should move this logic to the asynchronous folder and make use of IS_SYNC to choose which function to call.

blink1073:
We need to add the subpackages to `[tool.setuptools.packages.find]` in `pyproject.toml`.

blink1073:
Can you please move `synchro.py` under `tools`?

pymongo/pyopenssl_context.py
NoahStapp: We can remove the explicit loop if blocking the asyncio loop is acceptable for wrap_context.

pymongo/network.py
NoahStapp: My only concern with moving `network.py` to the `asynchronous` sub-module is that `async_send` and `async_sendall` are entirely different than their synchronous counterparts. The other functions should be compatible with `unasync` translation.

What if we kept `async_send` and `async_sendall` at the top level and moved the other functions into `asynchronous`?

pymongo/__init__.py
NoahStapp: Re-exporting all the documented synchronous modules from the top-level `__init__.py` should work in that case.

pymongo/network.py
blink1073: That sounds reasonable

pymongo/__init__.py
blink1073: Yeah, let's do that in this PR at least.  It will address `ModuleNotFoundError: No module named 'pymongo.mongo_client'` in the docs build.

pymongo/pyopenssl_context.py
blink1073: Looking at the implementation, `request_ocsp` is non-blocking, it sets a flag.  We can use the pattern from `aioopenssl` to perform the handshake on the [loop](https://github.com/horazont/aioopenssl/blob/8f250c0d4a772ed9e58a5668bdf9872fa689df5d/aioopenssl/__init__.py#L348) rather than using an exector.

NoahStapp: Sure I'll open a ticket for that change? Unless we want to do it as part of this initial PR.

blink1073: As a ticket please

blink1073:
The release scripts work now, I updated the PR checklist.

test/utils.py
```diff
@@ -823,6 +896,32 @@ def wait_until(predicate, success_description, timeout=10):
...
+    interval = min(float(timeout) / 100, 0.1)
+    while True:
+        retval = await predicate()
+        if retval:
+            return retval
+
+        if time.time() - start > timeout:
+            raise AssertionError("Didn't ever %s" % success_description)
+
+        time.sleep(interval)
```
blink1073, line 922: Should this be an `asyncio.sleep`?

NoahStapp: Yes, good catch!

blink1073:
We should add the [ASYNCIO](https://docs.astral.sh/ruff/rules/#flake8-async-async) ruff check, I just learned it existed.

blink1073:
> PyMongo import got 65 percent worse

Can you investigate this with https://pypi.org/project/tuna/?  I wouldn't expect it to get that much worse.

pymongo/asynchronous/helpers.py
ShaneHarvey: I'm a little concerned about duplicating code that could actually be shared by both the sync/async codepaths. Can we revisit this? message.py for example duplicates like 1500 lines of sans-I/O code just because a few classes call blocking APIs in a few places.

pymongo/asynchronous/helpers.py
ShaneHarvey: I suggest using the builtin method on 3.10+ and only using the fallback on <3.9.

test/__init__.py
```diff
@@ -45,14 +45,14 @@
...
 from pymongo.hello import HelloCompat
-from pymongo.mongo_client import MongoClient
 from pymongo.server_api import ServerApi
 from pymongo.ssl_support import HAVE_SSL, _ssl  # type:ignore[attr-defined]
-from pymongo.uri_parser import parse_uri
+from pymongo.synchronous import common, message
+from pymongo.synchronous.common import partition_node
+from pymongo.synchronous.database import Database
+from pymongo.synchronous.mongo_client import MongoClient
+from pymongo.synchronous.uri_parser import parse_uri
```
ShaneHarvey, line 55: I think we should revert all the test import changes because we want to be sure the old imports still work.

pymongo/network_layer.py
```diff
@@ -0,0 +1,60 @@
...
+    from pymongo.pyopenssl_context import _sslConn
+
+_UNPACK_HEADER = struct.Struct("<iiii").unpack
+_UNPACK_COMPRESSION_HEADER = struct.Struct("<iiB").unpack
+_POLL_TIMEOUT = 0.5
+# Errors raised by sockets (and TLS sockets) when in non-blocking mode.
+BLOCKING_IO_ERRORS = (BlockingIOError, *ssl_support.BLOCKING_IO_ERRORS)
+
+
+async def async_send(socket: Union[socket.socket, _sslConn], buf: bytes, flags: int = 0) -> int:
```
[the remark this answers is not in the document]
NoahStapp: Good point, it's completely unused. Removed!

pymongo/network_layer.py
```diff
@@ -0,0 +1,60 @@
...
+import struct
+from typing import (
+    TYPE_CHECKING,
+    Union,
+)
+
+from pymongo import ssl_support
+
+if TYPE_CHECKING:
+    from pymongo.pyopenssl_context import _sslConn
```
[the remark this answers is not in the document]
NoahStapp: That fails the typechecker: 

```
pymongo/asynchronous/pool.py:1034: error: Argument 1 to "async_sendall" has incompatible type "Union[socket, _sslConn]"; expected
"Union[socket, SSLSocket]"  [arg-type]
                await async_sendall(self.conn, message)
```

gridfs/__init__.py
```diff
@@ -21,980 +21,46 @@
...
+
+# Export synchronous modules as top-level gridfs modules for compatibility
+_sync_modules = {}
+for _name in sys.modules:
+    if _name.startswith("gridfs.synchronous."):
+        _full_name = "{}.{}".format("gridfs", _name.rsplit(".")[-1])
+        _sync_modules[_full_name] = _name
+
+for _module in _sync_modules:
+    sys.modules[_module] = sys.modules[_sync_modules[_module]]
```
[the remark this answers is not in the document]
ShaneHarvey: Sounds good but I want to clarify that this issue doesn't need to block this PR. It would be fine to open a ticket to change this later if that's easier.

pymongo/network_layer.py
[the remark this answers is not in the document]
ShaneHarvey: Interesting, maybe this is because we always install pyopenssl when type checking. I'll open a new ticket since I just realized this is an existing problem in the types.

pymongo/lock.py
```diff
@@ -38,3 +41,150 @@ def _release_locks() -> None:
...
+        return self
+
+    async def __aexit__(self, exc_type: Any, exc: Any, tb: Any) -> None:
+        self.release()
+
+    def __exit__(self, exc_type: Any, exc: Any, tb: Any) -> None:
+        self.release()
+
+
+class _Lock:
```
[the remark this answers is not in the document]
NoahStapp: I can just do it in this PR it's a simple change.

test/__init__.py
NoahStapp: We have a lot of tests that import private modules, so we'd either have to re-export everything at the top level or have a mix of `pymongo.module` and `pymongo.synchronous.module` imports in the tests.

ShaneHarvey: Right, anything private is free to change but the public imports should remain the same.

blink1073: Of note, we will most likely be using unasync for the tests as well, so we'll need to have the qualified imports.

gridfs/__init__.py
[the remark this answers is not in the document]
NoahStapp: It's easy enough to do now and get the structure of the repo into a correct state.

test/__init__.py
ShaneHarvey: Ah good point! Relying on test_import_compatibility SGTM.

test/test_import_compatibility.py
```diff
@@ -0,0 +1,49 @@
...
+        operations,
+        pool,
+        read_concern,
+        read_preferences,
+        results,
+        server_api,
+        server_description,
+        topology_description,
+        uri_parser,
+        write_concern,
```
ShaneHarvey, line 48: I think we need to actually import everything from each module:
```
from pymongo.collation import Collation, CollationStrength, ...
...
```

ShaneHarvey (changes requested):

test/test_import_compatibility.py
NoahStapp: Everything public? Or is `from pymongo.collation import *` sufficient?

ShaneHarvey: Yeah everything, that way we know every object we expect to be there is actually importable.

pymongo/pyopenssl_context.py
```diff
@@ -367,6 +368,56 @@ def set_default_verify_paths(self) -> None:
...
+                # Request a stapled OCSP response.
+                await loop.run_in_executor(None, ssl_conn.request_ocsp)
+            ssl_conn.set_connect_state()
+        # If this wasn't true the caller of wrap_socket would call
+        # do_handshake()
+        if do_handshake_on_connect:
+            # XXX: If we do hostname checking in a callback we can get rid
+            # of this call to do_handshake() since the handshake
+            # will happen automatically later.
+            await loop.run_in_executor(None, ssl_conn.do_handshake)
```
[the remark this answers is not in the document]
NoahStapp: By avoiding using `loop.run_in_executor`? Or creating a single background thread we keep alive the whole life of the program for this purpose?

test/test_import_compatibility.py
NoahStapp: Wouldn't using `from x import *` be sufficient then? Since that imports everything without a `_` in front?

blink1073: We also have `test_default_exports.py`, perhaps that just needs to be updated?

NoahStapp: Good point: `test_default_exports` checks that all of our top-level modules exist, but doesn't check that importing them exposes the correct public APIs. Adding this new import test to that file makes sense to me.

NoahStapp:
> @NoahStapp we've picked up conflicts from the delayed import PR

Resynced and fixed.

pymongo/asynchronous/auth.py
```diff
@@ -0,0 +1,663 @@
...
+    if not isinstance(password, str):
+        raise TypeError("password must be an instance of str")
+    if len(password) == 0:
+        raise ValueError("password can't be empty")
+    if not isinstance(username, str):
+        raise TypeError("username must be an instance of str")
+
+    md5hash = hashlib.md5()  # noqa: S324
+    data = f"{username}:mongo:{password}"
+    md5hash.update(data.encode("utf-8"))
```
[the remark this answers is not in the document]
NoahStapp: Is this a false positive?

pymongo/asynchronous/auth.py
```diff
@@ -0,0 +1,663 @@
...
+    return md5hash.hexdigest()
+
+
+def _auth_key(nonce: str, username: str, password: str) -> str:
+    """Get an auth key to use for authentication."""
+    digest = _password_digest(username, password)
+    md5hash = hashlib.md5()  # noqa: S324
+    data = f"{nonce}{username}{digest}"
+    md5hash.update(data.encode("utf-8"))
```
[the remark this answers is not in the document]
NoahStapp: Is this a false positive? @blink1073

pymongo/asynchronous/auth.py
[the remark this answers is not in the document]
NoahStapp: Is there any reason for us to not switch to a more secure hash algorithm like sha256?

pymongo/asynchronous/auth.py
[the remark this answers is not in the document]
blink1073: No, I had to dismiss these alerts previously, pointing out that it is used in a deprecated API.

pymongo/asynchronous/auth.py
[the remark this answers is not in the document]
blink1073: It is required by the spec: https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#mongodb-cr

pymongo/asynchronous/auth.py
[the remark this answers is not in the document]
blink1073: I dismissed it last time with "This code is used in a deprecated feature."

pymongo/pyopenssl_context.py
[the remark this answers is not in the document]
ShaneHarvey: By avoiding run_in_executor and avoiding threads altogether. Could you open a ticket for this problem? As you explained in person, pyopenssl doesn't yet support an async compatible OCSP handshake.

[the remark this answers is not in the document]
NoahStapp: Done

pymongo/__init__.py
```diff
@@ -87,11 +88,13 @@
...
 from pymongo import _csot
 from pymongo._version import __version__, get_version_string, version_tuple
-from pymongo.collection import ReturnDocument
-from pymongo.common import MAX_SUPPORTED_WIRE_VERSION, MIN_SUPPORTED_WIRE_VERSION
-from pymongo.cursor import CursorType
-from pymongo.mongo_client import MongoClient
-from pymongo.operations import (
+from pymongo.cursor_shared import CursorType
+
+# isort: off
```
ShaneHarvey, line 93: Why the isort disable? Does it cause import cycles? Could you add a comment explaining?

pymongo/__init__.py
```diff
@@ -87,11 +88,13 @@
 
 from pymongo import _csot
 from pymongo._version import __version__, get_version_string, version_tuple
-from pymongo.collection import ReturnDocument
-from pymongo.common import MAX_SUPPORTED_WIRE_VERSION, MIN_SUPPORTED_WIRE_VERSION
-from pymongo.cursor import CursorType
-from pymongo.mongo_client import MongoClient
-from pymongo.operations import (
+from pymongo.cursor_shared import CursorType
```
ShaneHarvey, line 91: CursorType should still be imported from `pymongo.cursor`.

pymongo/__init__.py
```diff
@@ -87,11 +88,13 @@
...
+
+# isort: off
+from pymongo.synchronous.collection import ReturnDocument
+from pymongo.synchronous.common import MAX_SUPPORTED_WIRE_VERSION, MIN_SUPPORTED_WIRE_VERSION
+from pymongo.synchronous.mongo_client import MongoClient
+from pymongo.synchronous.operations import (
```
ShaneHarvey, line 97: Why do we have multiple operations classes? IMO there should only be one set of InsertOne, etc... operations which can be passed to either sync or async apis.

test/auth_aws/test_auth_aws.py
```diff
@@ -26,7 +26,7 @@
 
 from pymongo import MongoClient
 from pymongo.errors import OperationFailure
-from pymongo.uri_parser import parse_uri
+from pymongo.synchronous.uri_parser import parse_uri
```
ShaneHarvey, line 29: I still think these unneeded test import changes should be reverted, even if only to make this change easier to review.

test/test_collection.py
```diff
@@ -192,17 +192,15 @@ def test_create(self):
             lambda: "create_test_no_wc" not in db.list_collection_names(),
             "drop create_test_no_wc collection",
         )
-        Collection(db, name="create_test_no_wc", create=True)
+        db.create_collection("create_test_no_wc")
```
ShaneHarvey, line 195: Did we remove support for `Collection(db, name="create_test_no_wc", create=True)`?

ShaneHarvey (changes requested):

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=63.0"]
+requires = ["setuptools>=63.0", "unasync"]
```
blink1073, line 2: We shouldn't need unasync to build a wheel, right?

pymongo/__init__.py
NoahStapp: Yeah it causes import cycles, will add a comment.

pymongo/__init__.py
NoahStapp: Good catch, fixed.

pymongo/__init__.py
NoahStapp: This is a consequence of the mixing of data and IO operations in much of our codebase.  I completely agree we should only have a single set of these operations classes.

NoahStapp: We could decide to do that refactoring and further separation now, but that would take a significant amount of time and delay this beta release. We'd also continue to have the constant merge conflicts and rebasing issues on this PR as more commits are merged to master.

test/auth_aws/test_auth_aws.py
NoahStapp: I'd argue they're necessary, as we'll need these changes once we rewrite our test suite to be async and generate the synchronous tests with `unasync`.

test/test_collection.py
NoahStapp: Yes, we decided to remove that option entirely from the driver.

pyproject.toml
NoahStapp: Now that we're using pre-commit for the synchro script, yup! Good catch.

pymongo/__init__.py
NoahStapp: Turns out I fixed that issue with the refactoring, so we can remove this entirely.

pymongo/__init__.py
blink1073: I agree, [PYTHON-4476](https://jira.mongodb.org/browse/PYTHON-4476) is tracking this.

test/auth_aws/test_auth_aws.py
blink1073: I second this, the work is already done.

test/test_collection.py
ShaneHarvey: Then this can't be merged in the 4.x branch since it's a major breaking change. So we either need to bring it back or change the plan on when we can release this.

NoahStapp: What if we make the full release of the async API the 5.0 release?

pymongo/asynchronous/database.py
```diff
@@ -0,0 +1,1425 @@
...
+            coll = AsyncCollection(
+                self,
+                name,
+                codec_options,
+                read_preference,
+                write_concern,
+                read_concern,
+                **kwargs,
+            )
+            await coll._create(kwargs, s, **kwargs)
```
ShaneHarvey, line 604: kwargs should only be passed once here.

test/test_collection.py
ShaneHarvey: It sounds simpler to add this option back in for the sync api. Isn't that what _IS_SYNC is for?:
```python
def __init__(...):
    ...
    if _IS_SYNC and <should auto create>:
        await self._create()
```

pymongo/__init__.py
ShaneHarvey: Okay as long as [PYTHON-4476](https://jira.mongodb.org/browse/PYTHON-4476) is implemented before we release anything non-alpha (betas aren't intended to have breaking changes either).

test/test_collection.py
ShaneHarvey: Do you have a list of other known breaking changes or is this the only one?

NoahStapp: I'm not a huge fan of having two arguments (`create` and `session`) for `AsyncCollection` that aren't valid. Doing so requires either keyword arguments everywhere we construct an `AsyncCollection` to keep the existing argument order or just passing `create=False` everywhere. 

My recollection of the original reasoning behind this change is that users shouldn't be using `create=True` in the first place. If that's accurate, I'd much rather remove it and have a breaking change than continue to support something we don't think should be supported.

NoahStapp: I don't recall any other breaking changes. I would expect our tests to catch all breaking changes for the synchronous API. Is that a fair assumption?

ShaneHarvey (approved):

ShaneHarvey (changes requested):
There are still a few test regressions: 
```
 [2024/06/05 16:22:30.833] =================================== FAILURES ===================================
 [2024/06/05 16:22:30.833] _____________________ TestCollation.test_create_collection _____________________
 [2024/06/05 16:22:30.833] self = <test.test_collation.TestCollation testMethod=test_create_collection>
 [2024/06/05 16:22:30.833]     def test_create_collection(self):
 [2024/06/05 16:22:30.833]         self.db.test.drop()
 [2024/06/05 16:22:30.833] >       self.db.create_collection("test", collation=self.collation)
 [2024/06/05 16:22:30.833] test/test_collation.py:128:
 [2024/06/05 16:22:30.833] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
 [2024/06/05 16:22:30.833] pymongo/_csot.py:120: in csot_wrapper
 [2024/06/05 16:22:30.833]     return func(self, *args, **kwargs)
 [2024/06/05 16:22:30.833] pymongo/synchronous/database.py:606: in create_collection
 [2024/06/05 16:22:30.833]     coll._create(kwargs, s)
 [2024/06/05 16:22:30.833] pymongo/synchronous/collection.py:664: in _create
 [2024/06/05 16:22:30.833]     self._create_helper(self._name, options, collation, session)
 [2024/06/05 16:22:30.833] pymongo/synchronous/collection.py:632: in _create_helper
 [2024/06/05 16:22:30.833]     self._command(
 [2024/06/05 16:22:30.833] pymongo/synchronous/collection.py:589: in _command
 [2024/06/05 16:22:30.833]     return conn.command(
 [2024/06/05 16:22:30.833] pymongo/synchronous/helpers.py:290: in inner
 [2024/06/05 16:22:30.833]     return func(*args, **kwargs)
 [2024/06/05 16:22:30.833] pymongo/synchronous/pool.py:992: in command
 [2024/06/05 16:22:30.833]     return command(
 [2024/06/05 16:22:30.833] pymongo/synchronous/network.py:222: in command
 [2024/06/05 16:22:30.833]     _async_helpers._check_command_response(
 [2024/06/05 16:22:30.833] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
 [2024/06/05 16:22:30.833] response = {'code': 48, 'codeName': 'NamespaceExists', 'errmsg': 'Collection pymongo_test.test already exists.', 'ok': 0.0}
 [2024/06/05 16:22:30.833] max_wire_version = 17, allowable_errors = None, parse_write_concern_error = True
 [2024/06/05 16:22:30.833]     def _check_command_response(
 [2024/06/05 16:22:30.833]         response: _DocumentOut,
 [2024/06/05 16:22:30.833]         max_wire_version: Optional[int],
 [2024/06/05 16:22:30.833]         allowable_errors: Optional[Container[Union[int, str]]] = None,
 [2024/06/05 16:22:30.833]         parse_write_concern_error: bool = False,
 [2024/06/05 16:22:30.833]     ) -> None:
 [2024/06/05 16:22:30.833]         """Check the response to a command for errors."""
 [2024/06/05 16:22:30.833]         if "ok" not in response:
 [2024/06/05 16:22:30.833]             # Server didn't recognize our message as a command.
 [2024/06/05 16:22:30.833]             raise OperationFailure(
 [2024/06/05 16:22:30.833]                 response.get("$err"),  # type: ignore[arg-type]
 [2024/06/05 16:22:30.833]                 response.get("code"),
 [2024/06/05 16:22:30.833]                 response,
 [2024/06/05 16:22:30.833]                 max_wire_version,
 [2024/06/05 16:22:30.833]             )
 [2024/06/05 16:22:30.833]         if parse_write_concern_error and "writeConcernError" in response:
 [2024/06/05 16:22:30.833]             _error = response["writeConcernError"]
 [2024/06/05 16:22:30.833]             _labels = response.get("errorLabels")
 [2024/06/05 16:22:30.833]             if _labels:
 [2024/06/05 16:22:30.833]                 _error.update({"errorLabels": _labels})
 [2024/06/05 16:22:30.833]             _raise_write_concern_error(_error)
 [2024/06/05 16:22:30.833]         if response["ok"]:
 [2024/06/05 16:22:30.833]             return
 [2024/06/05 16:22:30.833]         details = response
 [2024/06/05 16:22:30.833]         # Mongos returns the error details in a 'raw' object
 [2024/06/05 16:22:30.833]         # for some errors.
 [2024/06/05 16:22:30.834]         if "raw" in response:
 [2024/06/05 16:22:30.834]             for shard in response["raw"].values():
 [2024/06/05 16:22:30.834]                 # Grab the first non-empty raw error from a shard.
 [2024/06/05 16:22:30.834]                 if shard.get("errmsg") and not shard.get("ok"):
 [2024/06/05 16:22:30.834]                     details = shard
 [2024/06/05 16:22:30.834]                     break
 [2024/06/05 16:22:30.834]         errmsg = details["errmsg"]
 [2024/06/05 16:22:30.834]         code = details.get("code")
 [2024/06/05 16:22:30.834]         # For allowable errors, only check for error messages when the code is not
 [2024/06/05 16:22:30.834]         # included.
 [2024/06/05 16:22:30.834]         if allowable_errors:
 [2024/06/05 16:22:30.834]             if code is not None:
 [2024/06/05 16:22:30.834]                 if code in allowable_errors:
 [2024/06/05 16:22:30.834]                     return
 [2024/06/05 16:22:30.834]             elif errmsg in allowable_errors:
 [2024/06/05 16:22:30.834]                 return
 [2024/06/05 16:22:30.834]         # Server is "not primary" or "recovering"
 [2024/06/05 16:22:30.834]         if code is not None:
 [2024/06/05 16:22:30.834]             if code in _NOT_PRIMARY_CODES:
 [2024/06/05 16:22:30.834]                 raise NotPrimaryError(errmsg, response)
 [2024/06/05 16:22:30.834]         elif HelloCompat.LEGACY_ERROR in errmsg or "node is recovering" in errmsg:
 [2024/06/05 16:22:30.834]             raise NotPrimaryError(errmsg, response)
 [2024/06/05 16:22:30.834]         # Other errors
 [2024/06/05 16:22:30.834]         # findAndModify with upsert can raise duplicate key error
 [2024/06/05 16:22:30.834]         if code in (11000, 11001, 12582):
 [2024/06/05 16:22:30.834]             raise DuplicateKeyError(errmsg, code, response, max_wire_version)
 [2024/06/05 16:22:30.834]         elif code == 50:
 [2024/06/05 16:22:30.834]             raise ExecutionTimeout(errmsg, code, response, max_wire_version)
 [2024/06/05 16:22:30.834]         elif code == 43:
 [2024/06/05 16:22:30.834]             raise CursorNotFound(errmsg, code, response, max_wire_version)
 [2024/06/05 16:22:30.834] >       raise OperationFailure(errmsg, code, response, max_wire_version)
 [2024/06/05 16:22:30.834] E       pymongo.errors.OperationFailure: Collection pymongo_test.test already exists., full error: {'ok': 0.0, 'errmsg': 'Collection pymongo_test.test already exists.', 'code': 48, 'codeName': 'NamespaceExists'}
 [2024/06/05 16:22:30.834] pymongo/synchronous/helpers.py:196: OperationFailure
 [2024/06/05 16:22:30.834] __________ TestCollection.test_bypass_document_validation_bulk_write ___________
 [2024/06/05 16:22:30.834] self = <test.test_collection.TestCollection testMethod=test_bypass_document_validation_bulk_write>
...
```
https://spruce.mongodb.com/task/mongo_python_driver_tests_python_version_rhel8_test_ssl__platform~rhel8_auth_ssl~auth_ssl_python_version~3.10_coverage~coverage_test_6.0_standalone_patch_e9c86f4c00d793ea771e51e19478efcd21487a3d_6660efb33d72350007e0d11d_24_06_05_23_07_37?execution=0&sortBy=STATUS&sortDir=ASC

NoahStapp:
> There are still a few test regressions:
> 
> ```
>  [2024/06/05 16:22:30.833] =================================== FAILURES ===================================
>  [2024/06/05 16:22:30.833] _____________________ TestCollation.test_create_collection _____________________
>  [2024/06/05 16:22:30.833] self = <test.test_collation.TestCollation testMethod=test_create_collection>
>  [2024/06/05 16:22:30.833]     def test_create_collection(self):
>  [2024/06/05 16:22:30.833]         self.db.test.drop()
>  [2024/06/05 16:22:30.833] >       self.db.create_collection("test", collation=self.collation)
>  [2024/06/05 16:22:30.833] test/test_collation.py:128:
>  [2024/06/05 16:22:30.833] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
>  [2024/06/05 16:22:30.833] pymongo/_csot.py:120: in csot_wrapper
>  [2024/06/05 16:22:30.833]     return func(self, *args, **kwargs)
>  [2024/06/05 16:22:30.833] pymongo/synchronous/database.py:606: in create_collection
>  [2024/06/05 16:22:30.833]     coll._create(kwargs, s)
>  [2024/06/05 16:22:30.833] pymongo/synchronous/collection.py:664: in _create
>  [2024/06/05 16:22:30.833]     self._create_helper(self._name, options, collation, session)
>  [2024/06/05 16:22:30.833] pymongo/synchronous/collection.py:632: in _create_helper
>  [2024/06/05 16:22:30.833]     self._command(
>  [2024/06/05 16:22:30.833] pymongo/synchronous/collection.py:589: in _command
>  [2024/06/05 16:22:30.833]     return conn.command(
>  [2024/06/05 16:22:30.833] pymongo/synchronous/helpers.py:290: in inner
>  [2024/06/05 16:22:30.833]     return func(*args, **kwargs)
>  [2024/06/05 16:22:30.833] pymongo/synchronous/pool.py:992: in command
>  [2024/06/05 16:22:30.833]     return command(
>  [2024/06/05 16:22:30.833] pymongo/synchronous/network.py:222: in command
>  [2024/06/05 16:22:30.833]     _async_helpers._check_command_response(
>  [2024/06/05 16:22:30.833] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
>  [2024/06/05 16:22:30.833] response = {'code': 48, 'codeName': 'NamespaceExists', 'errmsg': 'Collection pymongo_test.test already exists.', 'ok': 0.0}
>  [2024/06/05 16:22:30.833] max_wire_version = 17, allowable_errors = None, parse_write_concern_error = True
>  [2024/06/05 16:22:30.833]     def _check_command_response(
>  [2024/06/05 16:22:30.833]         response: _DocumentOut,
>  [2024/06/05 16:22:30.833]         max_wire_version: Optional[int],
>  [2024/06/05 16:22:30.833]         allowable_errors: Optional[Container[Union[int, str]]] = None,
>  [2024/06/05 16:22:30.833]         parse_write_concern_error: bool = False,
>  [2024/06/05 16:22:30.833]     ) -> None:
>  [2024/06/05 16:22:30.833]         """Check the response to a command for errors."""
>  [2024/06/05 16:22:30.833]         if "ok" not in response:
>  [2024/06/05 16:22:30.833]             # Server didn't recognize our message as a command.
>  [2024/06/05 16:22:30.833]             raise OperationFailure(
>  [2024/06/05 16:22:30.833]                 response.get("$err"),  # type: ignore[arg-type]
>  [2024/06/05 16:22:30.833]                 response.get("code"),
>  [2024/06/05 16:22:30.833]                 response,
>  [2024/06/05 16:22:30.833]                 max_wire_version,
>  [2024/06/05 16:22:30.833]             )
>  [2024/06/05 16:22:30.833]         if parse_write_concern_error and "writeConcernError" in response:
>  [2024/06/05 16:22:30.833]             _error = response["writeConcernError"]
>  [2024/06/05 16:22:30.833]             _labels = response.get("errorLabels")
>  [2024/06/05 16:22:30.833]             if _labels:
>  [2024/06/05 16:22:30.833]                 _error.update({"errorLabels": _labels})
>  [2024/06/05 16:22:30.833]             _raise_write_concern_error(_error)
>  [2024/06/05 16:22:30.833]         if response["ok"]:
>  [2024/06/05 16:22:30.833]             return
>  [2024/06/05 16:22:30.833]         details = response
>  [2024/06/05 16:22:30.833]         # Mongos returns the error details in a 'raw' object
>  [2024/06/05 16:22:30.833]         # for some errors.
>  [2024/06/05 16:22:30.834]         if "raw" in response:
>  [2024/06/05 16:22:30.834]             for shard in response["raw"].values():
>  [2024/06/05 16:22:30.834]                 # Grab the first non-empty raw error from a shard.
>  [2024/06/05 16:22:30.834]                 if shard.get("errmsg") and not shard.get("ok"):
>  [2024/06/05 16:22:30.834]                     details = shard
>  [2024/06/05 16:22:30.834]                     break
>  [2024/06/05 16:22:30.834]         errmsg = details["errmsg"]
>  [2024/06/05 16:22:30.834]         code = details.get("code")
>  [2024/06/05 16:22:30.834]         # For allowable errors, only check for error messages when the code is not
>  [2024/06/05 16:22:30.834]         # included.
>  [2024/06/05 16:22:30.834]         if allowable_errors:
>  [2024/06/05 16:22:30.834]             if code is not None:
>  [2024/06/05 16:22:30.834]                 if code in allowable_errors:
>  [2024/06/05 16:22:30.834]                     return
>  [2024/06/05 16:22:30.834]             elif errmsg in allowable_errors:
>  [2024/06/05 16:22:30.834]                 return
>  [2024/06/05 16:22:30.834]         # Server is "not primary" or "recovering"
>  [2024/06/05 16:22:30.834]         if code is not None:
>  [2024/06/05 16:22:30.834]             if code in _NOT_PRIMARY_CODES:
>  [2024/06/05 16:22:30.834]                 raise NotPrimaryError(errmsg, response)
>  [2024/06/05 16:22:30.834]         elif HelloCompat.LEGACY_ERROR in errmsg or "node is recovering" in errmsg:
>  [2024/06/05 16:22:30.834]             raise NotPrimaryError(errmsg, response)
>  [2024/06/05 16:22:30.834]         # Other errors
>  [2024/06/05 16:22:30.834]         # findAndModify with upsert can raise duplicate key error
>  [2024/06/05 16:22:30.834]         if code in (11000, 11001, 12582):
>  [2024/06/05 16:22:30.834]             raise DuplicateKeyError(errmsg, code, response, max_wire_version)
>  [2024/06/05 16:22:30.834]         elif code == 50:
>  [2024/06/05 16:22:30.834]             raise ExecutionTimeout(errmsg, code, response, max_wire_version)
>  [2024/06/05 16:22:30.834]         elif code == 43:
>  [2024/06/05 16:22:30.834]             raise CursorNotFound(errmsg, code, response, max_wire_version)
>  [2024/06/05 16:22:30.834] >       raise OperationFailure(errmsg, code, response, max_wire_version)
>  [2024/06/05 16:22:30.834] E       pymongo.errors.OperationFailure: Collection pymongo_test.test already exists., full error: {'ok': 0.0, 'errmsg': 'Collection pymongo_test.test already exists.', 'code': 48, 'codeName': 'NamespaceExists'}
>  [2024/06/05 16:22:30.834] pymongo/synchronous/helpers.py:196: OperationFailure
>  [2024/06/05 16:22:30.834] __________ TestCollection.test_bypass_document_validation_bulk_write ___________
>  [2024/06/05 16:22:30.834] self = <test.test_collection.TestCollection testMethod=test_bypass_document_validation_bulk_write>
> ...
> ```
> 
> [spruce.mongodb.com/task/mongo_python_driver_tests_python_version_rhel8_test_ssl__platform~rhel8_auth_ssl~auth_ssl_python_version~3.10_coverage~coverage_test_6.0_standalone_patch_e9c86f4c00d793ea771e51e19478efcd21487a3d_6660efb33d72350007e0d11d_24_06_05_23_07_37?execution=0&sortBy=STATUS&sortDir=ASC](https://spruce.mongodb.com/task/mongo_python_driver_tests_python_version_rhel8_test_ssl__platform~rhel8_auth_ssl~auth_ssl_python_version~3.10_coverage~coverage_test_6.0_standalone_patch_e9c86f4c00d793ea771e51e19478efcd21487a3d_6660efb33d72350007e0d11d_24_06_05_23_07_37?execution=0&sortBy=STATUS&sortDir=ASC)

Broke it with the `kwargs` changes, looking into it now.

blink1073 (approved):
:shipit:

ShaneHarvey (approved):

PR #1635, opened 2024-05-08
PYTHON-4429 Add Changelog for 4.7.2

blink1073:
(cherry picked from commit 29f4d5cf8952b1abe0a73b11631ff84818b7f9ed)

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -6,6 +6,22 @@ Changes in Version 4.8.0
 
 .. warning:: PyMongo 4.8 drops support for Python 3.7 and PyPy 3.8: Python 3.8+ or PyPy 3.9+ is now required.
 
+Changes in Version 4.7.2
+-------------------------
+
+Version 4.7.2 fixes a bug introduced in 4.7.0:
+
+- Fixed a bug where PyMongo could not be used with the Nuitka compiler.
+
+Issues Resolved
+...............
+
+See the `PyMongo 4.7.2 release notes in JIRA`_ for the list of resolved issues
+in this release.
+
+.. _PyMongo 4.7.2 release notes in JIRA: https://jira.mongodb.org/secure/ReleaseNote.jspa?projectId=10004&version=39710
+
+
 Changes in Version 4.7.1
 -------------------------
 


doc/changelog.rst
```diff
@@ -6,6 +6,22 @@ Changes in Version 4.8.0
 
 .. warning:: PyMongo 4.8 drops support for Python 3.7 and PyPy 3.8: Python 3.8+ or PyPy 3.9+ is now required.
 
+Changes in Version 4.7.2
+-------------------------
+
+Version 4.7.2 fixes a bug introduced in 4.7.0:
```
caseyclements, line 12: This is ambiguous. Is there just the Nuitka compiler bug?

doc/changelog.rst
```diff
@@ -6,6 +6,22 @@ Changes in Version 4.8.0
...
+
+- Fixed a bug where PyMongo could not be used with the Nuitka compiler.
+
+Issues Resolved
+...............
+
+See the `PyMongo 4.7.2 release notes in JIRA`_ for the list of resolved issues
```
caseyclements, line 19: There's something funny with these links. For one, they not automatically take you to them by clicking in github. But more importantly, when I paste the link into my browser, I don't see things rendered as html. I see raw tags, like <h2>.

caseyclements (changes requested):

doc/changelog.rst
blink1073: We're moving away from this changelog toward the "What's New" page on the internal docs site.

doc/changelog.rst
blink1073: Yes

caseyclements: How about just one line:

Fixes a bug where PyMongo could not be used with the Nuitka compiler.

doc/changelog.rst
caseyclements: So this is how it's supposed to look?

blink1073: 🤷🏼‍♂️

doc/changelog.rst
blink1073: I want to convey that the bug was introduced in 4.7.0.  It worked in prior versions.

caseyclements (approved):
Git-R-Done!

PR #1639, opened 2024-05-15
PYTHON-4447 Use 8.0 for local OIDC server

blink1073:
Also clean up the OIDC tasks a bit - rename, and mark as `test`.


caseyclements (commented):
Good to go assuming the monitoring tests are unrelated.

If you pin the MONGODB_VERSION  as env variable, what happens to all the build variants? Do they run the same thing?

blink1073:
It applies to all of the variants that use this task, yes.

PR #1644, opened 2024-05-20
PYTHON-4451 Use Hatch as Build Backend

blink1073:


blink1073:
I filed [PYTHON-4452](https://jira.mongodb.org/browse/PYTHON-4452) for the libmongocrypt failure.

blink1073:
evergreen retry

setup.py
```diff
@@ -136,32 +136,8 @@ def build_extension(self, ext):
     )
     ext_modules = []
 
-
```
caseyclements, line 139: You were able to get rid of all this because of the way that requirements.txt is handled? That's awesome!

.github/workflows/test-python.yml
```diff
@@ -31,12 +31,10 @@ jobs:
       - name: Run linters
         run: |
           tox -m lint-manual
-      - name: Check Manifest
```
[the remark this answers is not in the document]
caseyclements: I read the [Why Hatch?](https://hatch.pypa.io/latest/why/) page. I'm starting to get it. Very nice.

caseyclements (approved):
Looks good. It would be good to have a little demo sometime.

pymongo/_version.py
```diff
@@ -15,16 +15,22 @@
...
 
-version_tuple: Tuple[Union[int, str], ...] = (4, 8, 0, ".dev0")
+__version__ = "4.8.0.dev1"
 
-
-def get_version_string() -> str:
-    if isinstance(version_tuple[-1], str):
-        return ".".join(map(str, version_tuple[:-1])) + version_tuple[-1]
-    return ".".join(map(str, version_tuple))
+pattern = r"(?P<major>\d+).(?P<minor>\d+).(?P<patch>\d+)(?P<rest>.*)"
```
ShaneHarvey, line 23: Can we make this a function and add a few unit tests to make sure it works with a variety of versions? Eg "4.8.0.dev1", "4.8.0", "5.0", etc...

blink1073: Done.  We don't support "5.0", it has to have the patch version.

ShaneHarvey: I think not supporting "5.0" is a mistake. Our 4.0 release used "4.0" for example, same for "3.0". If we don't support "5.0" then get_version_tuple will silently return empty which is a bit odd.

blink1073: Fixed

ShaneHarvey (dismissed):

PR #1645, opened 2024-05-28
PYTHON-4455 Improve import time on Windows

blink1073:


caseyclements (dismissed):
Assuming CSOT failures are CSOT failures, LGTM. We can always change again.

ShaneHarvey (dismissed):
Could you update the CHANGELOG log mention we no longer include the windows version in the "name" field?


Can you give an example of the metadata before/after this change?

pymongo/pool.py
```diff
@@ -211,13 +211,14 @@ def _set_keepalive_times(sock: socket.socket) -> None:
...
-        # "Windows XP", "Windows 7", "Windows 10", etc.
-        "name": " ".join((platform.system(), platform.release())),
-        "architecture": platform.machine(),
-        # Windows patch level (e.g. 5.1.2600-SP3)
-        "version": "-".join(platform.win32_ver()[1:3]),
+        "type": "Windows",
+        "name": "Windows",
+        # Avoid using platform calls, see PYTHON-4455.
+        "architecture": os.environ.get("PROCESSOR_ARCHITECTURE") or platform.machine(),
+        # Windows patch level (e.g. 10-0-17763).
```
ShaneHarvey, line 220: is `10-0-17763` accurate? I would expect something like `10.0.17763-XXX`

blink1073: I need to update the comment, I just pushed a change to make it format the same as the old way.

ShaneHarvey (approved):

PR #1646, opened 2024-05-29
PYTHON-4463 Disallow comma character in authMechanismProperties connection string value

blink1073:


test/connection_string/test/valid-options.json
```diff
@@ -37,6 +37,25 @@
...
+          "host": "example.com",
+          "port": null
+        }
+      ],
+      "auth": null,
+      "options": {
+        "authmechanismProperties": {
+          "TOKEN_RESOURCE": "mongodb://test-cluster"
+        }
+      }
```
Jibola, line 58: Why did a new a colon test get added?

Jibola: do we now need an XFAIL test with a comma in it?

Jibola (changes requested):
Asked some clarifying questions, but overall LGTM!

test/connection_string/test/valid-options.json
blink1073: This came up because `TOKEN_RESOURCE` is the first auth mech property that will commonly have a colon in it, and that part of the spec was ambiguous.   I added a test for the case with a comma.

Jibola (approved):
LGTM

PR #1648, opened 2024-05-30
PYTHON-4441 Use deferred imports instead of lazy module loading

blink1073:
Before:
image

After:
image


pymongo/auth_aws.py
```diff
@@ -35,6 +27,8 @@
     from pymongo.auth import MongoCredential
     from pymongo.pool import Connection
 
+_HAVE_MONGODB_AWS = importlib.util.find_spec("pymongo_auth_aws") is not None
```
ShaneHarvey, line 30: Given the problems we've seen, I'm a little wary of using importlib at all. How about we remove _HAVE_MONGODB_AWS instead? Looks like it's only used where we do the import anyway.

pymongo/compression_support.py
```diff
@@ -13,33 +13,16 @@
...
-    _HAVE_ZLIB = False
-
-try:
-    zstandard = lazy_import("zstandard")
-    _HAVE_ZSTD = True
-except ImportError:
-    _HAVE_ZSTD = False
+_HAVE_SNAPPY = importlib.util.find_spec("snappy") is not None
+_HAVE_ZLIB = importlib.util.find_spec("zlib") is not None
+_HAVE_ZSTD = importlib.util.find_spec("zstandard") is not None
```
ShaneHarvey, line 25: Let's avoid using importlib. How about something like this instead?:
```python
def _have_snappy():
    try:
        import snappy
        return True
    except ImportError:
        return False
```

pymongo/srv_resolver.py
```diff
@@ -15,19 +15,18 @@
...
 from pymongo.errors import ConfigurationError
 
-try:
+if TYPE_CHECKING:
     from dns import resolver
 
-    _HAVE_DNSPYTHON = True
-except ImportError:
-    _HAVE_DNSPYTHON = False
+_HAVE_DNSPYTHON = importlib.util.find_spec("dns") is not None
```
ShaneHarvey, line 29: Similar comment about making this a function and avoiding importlib.

pymongo/pyopenssl_context.py
```diff
@@ -37,10 +36,6 @@
 from pymongo.socket_checker import _errno_from_exception
 from pymongo.write_concern import validate_boolean
 
-_x509 = lazy_import("cryptography.x509")
-_service_identity = lazy_import("service_identity")
-_service_identity_pyopenssl = lazy_import("service_identity.pyopenssl")
```
ShaneHarvey, line 42: Does this new version work correctly when pyopenssl is installed but service_identity or cryptography is not? What should happen in that case is the driver falls back to using stdlib ssl but I suspect this new code will attempt to use pyopenssl but fail with an import error at connection establishment time.

ShaneHarvey: The most straightforward way to fix this in my eyes would be to keep these cryptograph+service_identity imports at the top-level.

ShaneHarvey (changes requested):

pymongo/auth_aws.py
blink1073: Done

pymongo/compression_support.py
blink1073: Done

pymongo/srv_resolver.py
blink1073: Done

pymongo/pyopenssl_context.py
blink1073: Done

ShaneHarvey: Still need to do the same with the service_identity imports.

ShaneHarvey (changes requested):

pymongo/auth_aws.py
```diff
@@ -38,11 +29,13 @@
...
     """Authenticate using MONGODB-AWS."""
-    if not _HAVE_MONGODB_AWS:
+    try:
+        import pymongo_auth_aws  # type:ignore[import]
+    except ImportError:
         raise ConfigurationError(
             "MONGODB-AWS authentication requires pymongo-auth-aws: "
             "install with: python -m pip install 'pymongo[aws]'"
-        )
+        ) from None
```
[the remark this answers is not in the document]
blink1073: Done

pymongo/pyopenssl_context.py
blink1073: Done

ShaneHarvey (approved):

PR #1649, opened 2024-05-30
PYTHON-4468 Hide the value of sensitive subtype binary objects

TerryRPatterson:
[PYTHON-4468](https://jira.mongodb.org/browse/PYTHON-4468)

# Summary

I noticed the `bson.binary.Binary` type include it's
value even when the subtype is sensitive (8)

# Changes in this PR
This PR changes the Binary type's repr to be `<Binary ### {self.__subtype}>` when the subtype is 8.
The `< >` representation follow [general guideline](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_repr) that when a repr can't be roundtripped by `eval` it should use `< >` instead of a constructor like syntax.

### DIFF
--- a/bson/binary.py
+++ b/bson/binary.py
@@ -364,4 +364,7 @@ class Binary(bytes):
         return not self == other
 
     def __repr__(self) -> str:
-        return f"Binary({bytes.__repr__(self)}, {self.__subtype})"
+        if self.__subtype == SENSITIVE_SUBTYPE:
+            return f"<Binary(REDACTED, {self.__subtype})>"
+        else:
+            return f"Binary({bytes.__repr__(self)}, {self.__subtype})"
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -20,7 +20,7 @@ Issues Resolved
 See the `PyMongo 4.7.3 release notes in JIRA`_ for the list of resolved issues
 in this release.
 
-.. _PyMongo 4.7.3 release notes in JIRA:
+.. _PyMongo 4.7.3 release notes in JIRA: https://jira.mongodb.org/secure/ReleaseNote.jspa?projectId=10004&version=39865
 
 Changes in Version 4.7.2
 -------------------------
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -6,21 +6,9 @@ Changes in Version 4.8.0
 
 The handshake metadata for "os.name" on Windows has been simplified to "Windows" to improve import time.
 
+The repr of ``bson.binary.Binary`` is now redacted when the subtype is SENSITIVE_SUBTYPE(8).
+
 .. warning:: PyMongo 4.8 drops support for Python 3.7 and PyPy 3.8: Python 3.8+ or PyPy 3.9+ is now required.
-
-Changes in Version 4.7.3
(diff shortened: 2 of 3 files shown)


bson/binary.py
```diff
@@ -364,4 +364,7 @@ def __ne__(self, other: Any) -> bool:
         return not self == other
 
     def __repr__(self) -> str:
-        return f"Binary({bytes.__repr__(self)}, {self.__subtype})"
+        if self.__subtype == SENSITIVE_SUBTYPE:
+            return f"<Binary ### {self.__subtype}>"
```
blink1073, line 368: For consistency, can we format it like `f"Binary('<redacted>', {self._subtype))`>

TerryRPatterson: Okay will do.

blink1073 (dismissed):
Looks good, thank you!  I'll wait for our Evergreen tests to finish before merging.

bson/binary.py
```diff
@@ -364,4 +364,7 @@ def __ne__(self, other: Any) -> bool:
         return not self == other
 
     def __repr__(self) -> str:
-        return f"Binary({bytes.__repr__(self)}, {self.__subtype})"
+        if self.__subtype == SENSITIVE_SUBTYPE:
+            return f"Binary('<redacted>', {self.__subtype})"
```
ShaneHarvey, line 368: Since this is no longer an accurate representation of the object we need to surround it with <> so that repr+eval() will raise an error instead of silently changing the data, like this:
```
return f"<Binary(REDACTED, {self.__subtype})>"
```

Reference: https://docs.python.org/3/library/functions.html#repr

ShaneHarvey (changes requested):
Thanks @TerryRPatterson, do you mind adding a mention of this change in the doc/changelog.rst?

bson/binary.py
TerryRPatterson: I added an entry to the changelog. Since the the jira is tagged `Fixes 4.7.3` I assumed that is the release. I don't know how to get a link to the release notes from jira so I didn't include the link.

doc/changelog.rst
```diff
@@ -8,6 +8,20 @@ The handshake metadata for "os.name" on Windows has been simplified to "Windows"
...
+The repr of ``bson.binary.Binary`` is now redacted when the subtype
+is SENSITIVE_SUBTYPE(8).
+
+Issues Resolved
+...............
+
+See the `PyMongo 4.7.3 release notes in JIRA`_ for the list of resolved issues
+in this release.
+
+.. _PyMongo 4.7.3 release notes in JIRA:
```
blink1073, line 23:
```suggestion
.. _PyMongo 4.7.3 release notes in JIRA: https://jira.mongodb.org/secure/ReleaseNote.jspa?projectId=10004&version=39865
```

bson/binary.py
blink1073: Thanks!  I added the link.

blink1073:
> FYI since this is a minor breaking behavior change it can't go into 4.7.3 and has to wait for 4.8.

I updated the changelog.

ShaneHarvey (approved):

PR #1652, opened 2024-05-31
PYTHON-4473 Optimize find/aggregate/command by avoiding duplicate calls to _get_topology

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4473


ShaneHarvey:
The benchmarks don't show much improvement but I believe that's because they don't yet cover the case where there's many threads contending for the topology lock. 

Locally I see a ~10% improvement on find_one() and ~15% improvement on db.command() on a benchmark with 30-100 threads.

blink1073 (approved):
LGTM!

PR #1659, opened 2024-06-06
PYTHON-4480 Deprecate create=True for Collection

NoahStapp:


test/test_collection.py
```diff
@@ -197,12 +197,6 @@ def test_create(self):
             lambda: "create_test_no_wc" in db.list_collection_names(),
             "create create_test_no_wc collection",
         )
-        db.create_test_no_wc.drop()
```
blink1073, line 200: We should keep this and catch the warning.

pymongo/asynchronous/collection.py
```diff
@@ -248,6 +249,11 @@ def __init__(
 
         if create or kwargs:
             if _IS_SYNC:
+                warnings.warn(
+                    "The `create` and `kwargs` arguments are deprecated and will be removed in PyMongo 5.0",
```
ShaneHarvey, line 253: I would add "arguments to Collection are..."

test/test_collection.py
ShaneHarvey: Correct, when we deprecate something we keep the tests and also add an extra assertion to ensure the warning is actually raised.

blink1073 (approved):
LGTM!

PR #1667, opened 2024-06-10
PYTHON-4489 Make setup.py private

blink1073:


doc/changelog.rst
```diff
@@ -3,13 +3,21 @@ Changelog
...
+
 - The handshake metadata for "os.name" on Windows has been simplified to "Windows" to improve import time.
 - The repr of ``bson.binary.Binary`` is now redacted when the subtype is SENSITIVE_SUBTYPE(8).
 - A new asynchronous API with full asyncio support.
 
+Unavoidable breaking changes
+............................
+
+- Since we are now using ``hatch`` as our build backend, we no longer have a ``setup.py`` file
+  and require installation using ``pip``.
```
ShaneHarvey, line 19: This should be in 4.9 not 4.8, unless we are planning to backport PYTHON-4451.

ShaneHarvey (commented):
What do you think about adding a setup.py that prints a message and exits with an error to help out users? Something like "PyMongo>=4.9 no longer supports building via setup.py, use `python -m pip install path/to/pymongo` instead."

doc/changelog.rst
blink1073: It will be 4.8, PR is already open: https://github.com/mongodb/mongo-python-driver/pull/1663

blink1073:
> What do you think about adding a setup.py that prints a message and exits with an error to help out users? 

Done

ShaneHarvey (approved):

PR #1669, opened 2024-06-11
Handle incompatible PyOpenSSL & cryptography versions

oh2fih:
Catch `AttributeError` caused by incompatibility between PyOpenSSL < 23.2.0 & cryptography >= 42.0.0.

In a situation where incompatible versions are already installed on the system (regardless `pip install "pymongo[ocsp]"` as noticed in https://jira.mongodb.org/browse/PYTHON-4491), the user now gets an unhandled exception with a stack trace:

```
  File "/home/user/.local/lib/python3.10/site-packages/pymongo/ssl_support.py", line 24, in <module>
    import pymongo.pyopenssl_context as _ssl
  File "/home/user/.local/lib/python3.10/site-packages/pymongo/pyopenssl_context.py", line 29, in <module>
    from OpenSSL import SSL as _SSL
  File "/usr/lib/python3/dist-packages/OpenSSL/__init__.py", line 8, in <module>
    from OpenSSL import crypto, SSL
  File "/usr/lib/python3/dist-packages/OpenSSL/crypto.py", line 1579, in <module>
    class X509StoreFlags(object):
  File "/usr/lib/python3/dist-packages/OpenSSL/crypto.py", line 1598, in X509StoreFlags
    NOTIFY_POLICY = _lib.X509_V_FLAG_NOTIFY_POLICY
AttributeError: module 'lib' has no attribute 'X509_V_FLAG_NOTIFY_POLICY'. Did you mean: 'X509_V_FLAG_EXPLICIT_POLICY'?
```

The added exception handling tries to fix this problem by falling back to stdlib SSL as would happen if the PyOpenSSL was not installed at all. Additionally, the stack trace is replaced with a user friendly warning suggesting a possible solution in case the OCSP support is desired.

Ref. https://github.com/cve-search/cve-search/issues/1099#issuecomment-2144528101, https://github.com/conda/conda/issues/13619#issuecomment-2075825121 & https://github.com/mongodb/mongo-python-driver/pull/1666.


pymongo/ssl_support.py
```diff
@@ -28,6 +30,25 @@
         import pymongo.ssl_context as _ssl  # type: ignore[no-redef]
     except ImportError:
         HAVE_SSL = False
+except AttributeError:
```
ShaneHarvey, line 33: Could you refactor this into the `except ImportError:` block so we don't have to duplicate the stdlib ssl fallback?

```python
except (ImportError, AttributeError) as exc:
    if isinstance(exc, AttributeError):
        warnings.warn(...)
```

ShaneHarvey (changes requested):

pymongo/ssl_support.py
oh2fih: Excellent idea. I still kept the fallback warning as it seems appropriate for the other exception, too.

ShaneHarvey (changes requested):

pymongo/ssl_support.py
```diff
@@ -19,14 +19,31 @@
...
+            stacklevel=2,
+        )
     try:
         import pymongo.ssl_context as _ssl  # type: ignore[no-redef]
     except ImportError:
+        warnings.warn(
+            "Fallback to stdlib ssl failed, disabling SSL support.",
+            UserWarning,
+            stacklevel=2,
+        )
```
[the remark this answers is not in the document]
oh2fih: Thanks, removed.

ShaneHarvey:
Thanks @oh2fih! I scheduled the CI checks now.

pymongo/ssl_support.py
```diff
@@ -19,11 +19,23 @@
 
 from pymongo.errors import ConfigurationError
 
+import warnings
+
```
ShaneHarvey, line 23: This import needs to go above `from typing import Optional` in order to pass our pre-commit checks: https://github.com/mongodb/mongo-python-driver/blob/master/CONTRIBUTING.md#running-linters

```
import warnings
from typing import Optional
```

ShaneHarvey (changes requested):

pymongo/ssl_support.py
oh2fih: But of course. Could be more efficient to let CI run early to save manual review time. It seems contributors aren't permitted to use the checks themselves, which could help improving the coding style without manual intervention. I would suggest a bit more relaxed policy to allow contributors use CI after they have added something sane i.e. just rule out vandalism.

However, the mentioned pre-commit lint warning should now be solved.

blink1073: Hi @oh2fih, the GitHub options are limited, we're using a balanced approach:

image

oh2fih: Thanks @blink1073 ; GitHub recognizes me as a first-time contributor since this is my first PR and not yet merged.

ShaneHarvey (approved):

PR #1670, opened 2024-06-11
PYTHON-4206 - QE Range Protocol V2

NoahStapp:


pymongo/asynchronous/encryption.py
```diff
@@ -836,7 +831,7 @@ async def encrypt(
             when the algorithm is :attr:`Algorithm.INDEXED`.  An integer value
             *must* be given when the :attr:`Algorithm.INDEXED` algorithm is
             used.
-        :param range_opts: Experimental only, not intended for public use.
+        :param range_opts: Index options for `range` queries.
```
blink1073, line 834: We need notes in the docstrings for the new public parameter being added.

NoahStapp: For `trimFactor`? Or for all of the `rangeOpts` themselves?

blink1073: `rangeOpts`

NoahStapp: Done.

blink1073 (approved):
LGTM!

Jibola (approved):

pymongo/asynchronous/encryption_options.py
```diff
@@ -231,20 +231,20 @@ def __init__(
 
 
 class RangeOpts:
-    """Options to configure encrypted queries using the rangePreview algorithm."""
+    """Options to configure encrypted queries using the range algorithm."""
 
     def __init__(
         self,
         sparsity: int,
+        trim_factor: int,
```
ShaneHarvey, line 239: Making this a required argument is a breaking change. Is that intended? Are users expected to always specifiy trim_factor?

NoahStapp: Yeah, as per the [spec](https://github.com/mongodb/specifications/blob/59844221ade468044ec7b1e86d9db81eb7c6afad/source/client-side-encryption/client-side-encryption.md?plain=1#L1170), `trim_factor` is a required argument.

ShaneHarvey: I'm not sure that is the case. Adrian says:

> My understanding is that [trimFactor is] a knob for our tech support to debug performance issues and that average users should not touch it.

PR #1672, opened 2024-06-11
BUMP 4.9.0.dev0

ShaneHarvey:


caseyclements (approved):
LGTM. 

Very cool that we link to JIRA's auto-generated list of tickets for upcoming releases, knowing that if we don't finish them, we push them back. Proper software engineering process init.

PR #1677, opened 2024-06-12
PYTHON-4497 Fix handling of Python executable in run-tests

blink1073:


.evergreen/run-tests.sh
```diff
@@ -31,9 +31,6 @@ set -o xtrace
 AUTH=${AUTH:-noauth}
 SSL=${SSL:-nossl}
 TEST_ARGS="${*:1}"
-PYTHON=$(which python)
-# TODO: Remove when we drop PyPy 3.8 support.
-OLD_PYPY=$(python -c "import sys; print(sys.implementation.name.lower() == 'pypy' and sys.implementation.version < (7, 3, 12))")
```
ShaneHarvey, line 36: We can remove the uses of OLD_PYPY now too.

ShaneHarvey (changes requested):
To clarify, is this fixing a bug or just cleaning up old code?

blink1073:
It what trigged by work in https://github.com/mongodb/winkerberos/pull/92

.evergreen/run-tests.sh
blink1073: Done

ShaneHarvey (approved):

PR #1678, opened 2024-06-12
PYTHON-4476 Separate data and IO classes more effectively

NoahStapp:
This PR refactors a significant amount of our codebase to reduce code duplication and more clearly separate network IO from data handling and business logic. There's more to be done, but breaking up future changes into smaller, more manageable PRs is a much better and more maintainable strategy. The changes here are largely circular in nature with import and dependency graphs, making it difficult to separate them out into multiple PRs.

Core changes:

- Moved `client_options.py, collation.py, common.py, compression_support.py, encryption_options.py, event_loggers.py, hello.py, logger.py, max_staleness_selectors.py, message.py, monitoring.py, operations.py, read_preferences.py, response.py, server_description.py, server_selectors.py, srv_resolver.py, topology_description.py, uri_parser.py` out of `asynchronous` and into the top-level package. This significantly reduces the amount of code duplication done by synchronization, and should result in only files with actual `async def` methods staying inside `asynchronous`.
- Separated `PoolOptions` and most of the helpers and constants out from `pool.py` into their own shared `pool_options.py` file.
- Moved many of the helpers, constants, and data classes of the remaining `asynchronous` files into `NAME_shared.py` files inside the top-level package.
- Removed all network IO from `message.py`. `_Query, _GetMore, _BulkWriteContext, _EncryptedBulkWriteContext` are all now pure data classes that rely on their users (`bulk.py` and `server.py`) to perform network IO.

The coloring nature of asynchronous code encourages very clear separation of network IO from the rest of the codebase. Going forward, I recommend we try to avoid bundling IO inside classes in order to reduce code duplication in the synchronization process.


gridfs/synchronous/grid_file.py
```diff
@@ -163,7 +164,7 @@ def get(self, file_id: Any, session: Optional[ClientSession] = None) -> GridOut:
 
         :param file_id: ``"_id"`` of the file to get
         :param session: a
-            :class:`~pymongo.client_session.ClientSession`
+            :class:`~pymongo.client_session.AsyncClientSession`
```
ShaneHarvey, line 167: Shouldn't this still be ClientSession not AsyncClientSession?

ShaneHarvey (changes requested):
> +8,955 −19,746

nice. Just one quick comment. I'll need to spend more time to  review this in the next day or two.

test/utils_selection_tests.py
```diff
@@ -19,7 +19,7 @@
 import os
 import sys
 
-from pymongo.synchronous.operations import _Op
+from pymongo.operations import _Op
```
ShaneHarvey, line 22: This import should be later in the file, after `sys.path[0:0] = [""]`.

test/unified_format.py
```diff
@@ -616,7 +616,7 @@ def get_lsid_for_session(self, session_name):
         session = self[session_name]
         if not isinstance(session, ClientSession):
             self.test.fail(
-                f"Expected entity {session_name} to be of type ClientSession, got {type(session)}"
+                f"Expected entity {session_name} to be of type AsyncClientSession, got {type(session)}"
```
ShaneHarvey, line 619: This one should be ClientSession, right?

test/synchronous/test_collection.py
```diff
@@ -39,13 +39,13 @@
     wait_until,
 )
 
-from bson import encode
+from bson import RawBSONDocument, encode
```
ShaneHarvey, line 42: Does this file ever get run? Seems like a duplicate of test/test_collection.py

test/asynchronous/test_collection.py
```diff
@@ -40,21 +40,17 @@
     wait_until,
 )
 
-from bson import encode
+from bson import RawBSONDocument, encode
```
ShaneHarvey, line 43: Does test/async ever get run in CI?

pymongo/message.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2009-present MongoDB, Inc.
+# Copyright 2024-present MongoDB, Inc.
```
ShaneHarvey, line 1: Let's make this 2009 since that's more accurate.

pymongo/message.py
```diff
@@ -694,7 +884,7 @@ def _op_msg_no_header(
     max_doc_size = 0
     if identifier and docs is not None:
         type_one = _pack_byte(1)
-        cstring = _make_c_string(identifier)
+        cstring = bson._make_c_string(identifier)
```
ShaneHarvey, line 887: These import changes seem unneeded.

pymongo/message.py
```diff
@@ -1555,206 +1531,3 @@ def _batched_write_command_impl(
     buf.write(_pack_int(length - command_start))
 
     return to_send, length
-
-
-class _OpReply:
```
ShaneHarvey, line 1560: Can you move these classes back down here to minimize the code churn?

pymongo/message.py
```diff
@@ -1243,59 +1259,19 @@ def __batch_command(
...
-        return to_send
+        return -1, outgoing, to_send
 
     @property
     def max_split_size(self) -> int:
         """Reduce the batch splitting size."""
         return _MAX_SPLIT_SIZE_ENC
 
 
-def _raise_document_too_large(operation: str, doc_size: int, max_size: int) -> NoReturn:
```
ShaneHarvey, line 1276: Can you move the new _raise_document_too_large back here to minimize code churn?

pymongo/synchronous/monitoring.py
```diff
@@ -1,1903 +0,0 @@
-# Copyright 2015-present MongoDB, Inc.
```
ShaneHarvey, line 1: How come this file doesn't show up as a rename like message.py does? Any way we can have git recognize it to minimize code churn?

I have the same question for many of the other files as well.

pymongo/synchronous/server.py
```diff
@@ -121,26 +135,26 @@ def run_operation(
         cursors.
         Can raise ConnectionFailure, OperationFailure, etc.
 
-        :param conn: A Connection instance.
+        :param conn: An AsyncConnection instance.
```
ShaneHarvey, line 138: "An AsyncConnection" is incorrect here.

pymongo/synchronous/server.py
```diff
@@ -121,26 +135,26 @@ def run_operation(
...
         :param operation: A _Query or _GetMore object.
         :param read_preference: The read preference to use.
         :param listeners: Instance of _EventListeners or None.
         :param unpack_res: A callable that decodes the wire protocol response.
+        :param client: An AsyncMongoClient instance.
```
ShaneHarvey, line 143: AsyncMongoClient is incorrect here.

ShaneHarvey (changes requested):

test/synchronous/test_collection.py
NoahStapp: This is generated from `test/asynchronous/test_collection.py`, left as an example result for our future test asynchronization work/

test/asynchronous/test_collection.py
NoahStapp: No, left in as an example for our future test asynchronization work

pymongo/synchronous/monitoring.py
NoahStapp: Seems like it's just git being confused. Let me do some commit reworking, I think I can significantly improve it.

NoahStapp: I was wrong sadly, there isn't a good way to do this. Because we are essentially removing a lot of the `asynchronous` and `synchronous` files and putting a single version into the top-level `pymongo` package, it complicates the history a lot for git. Git doesn't track renames/moves really, just the changes, so a "rename" just means it had few enough changes along with the location move to qualify.

pymongo/synchronous/bulk.py
```diff
@@ -180,7 +91,7 @@ def __init__(
         comment: Optional[str] = None,
         let: Optional[Any] = None,
     ) -> None:
-        """Initialize a _Bulk instance."""
+        """Initialize a _AsyncBulk instance."""
```
ShaneHarvey, line 94: _Bulk -> _AsyncBulk

pymongo/asynchronous/bulk.py
```diff
@@ -407,7 +509,23 @@ async def _execute_command(
                     if self.ordered and "writeErrors" in result:
                         break
                 else:
-                    to_send = await bwc.execute_unack(cmd, ops, client)
+                    if self.is_encrypted:
```
ShaneHarvey, line 512: Can this be refactored into a helper method? Something like:
```python
                    result, to_send = await self._execute_batch_unack(cmd, ops, client)
```

We can also use the new helper in execute_op_msg_no_results down below.

pymongo/asynchronous/bulk.py
```diff
@@ -387,7 +475,21 @@ async def _execute_command(
 
                 # Run as many ops as possible in one command.
                 if write_concern.acknowledged:
-                    result, to_send = await bwc.execute(cmd, ops, client)
+                    if self.is_encrypted:
```
ShaneHarvey, line 478: Can this be refactored into a helper method? Something like:
```python
                    result, to_send = await self._execute_batch(cmd, ops, client)
```

pymongo/asynchronous/bulk.py
```diff
@@ -387,7 +475,21 @@ async def _execute_command(
...
+                            codec_options=bwc.codec,
+                            session=bwc.session,
+                            client=client,
+                        )
+                    else:
+                        request_id, msg, to_send = bwc.batch_command(cmd, ops)
+                        result = await self.write_command(
+                            bwc, cmd, request_id, msg, to_send, client
+                        )
+                        await client._process_response(result, bwc.session)
```
ShaneHarvey, line 492: Shouldn't client._process_response be called in the encrypted case as well?

ShaneHarvey (changes requested):

pymongo/uri_parser.py
```diff
@@ -1,21 +1,624 @@
...
+    get_validated_options,
+)
+from pymongo.errors import ConfigurationError, InvalidURI
+from pymongo.srv_resolver import _have_dnspython, _SrvResolver
+from pymongo.typings import _Address
+
+if TYPE_CHECKING:
+    from pymongo.pyopenssl_context import SSLContext
+
+_IS_SYNC = False
```
Jibola, line 49: is _IS_SYNC needed for a file like this?

Also, I think the file is synchronous.

pymongo/topology_description.py
```diff
@@ -1,21 +1,678 @@
...
+                self._incompatible_err = (
+                    "Server at %s:%d requires wire version %d, but this "  # type: ignore
+                    "version of PyMongo only supports up to %d."
+                    % (
+                        s.address[0],
+                        s.address[1] or 0,
+                        s.min_wire_version,
+                        common.MAX_SUPPORTED_WIRE_VERSION,
+                    )
+                )
```
Jibola, line 143: Would you know if this if statements also needs a break?

Jibola (changes requested):
Great so far! Got as far as all of the non-test files. 

I decided I'd break it off and take the tests in a separate review!

pymongo/message.py
```diff
@@ -1758,3 +1178,356 @@ def unpack(cls, msg: bytes) -> _OpMsg:
...
+
+class _CursorAddress(tuple):
+    """The server address (host, port) of a cursor, with namespace property."""
+
+    __namespace: Any
+
+    def __new__(cls, address: _Address, namespace: str) -> _CursorAddress:
+        self = tuple.__new__(cls, address)
+        self.__namespace = namespace
+        return self
```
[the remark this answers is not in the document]
NoahStapp: I don't know, maybe something to do with `tuple`?

pymongo/uri_parser.py
NoahStapp: Good catch! Those are all artifacts from moving these files out of `pymongo/asynchronous`, fixed.

pymongo/message.py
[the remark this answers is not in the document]
ShaneHarvey: Yeah it's a tuple thing.

pymongo/topology_description.py
NoahStapp: I don't think it makes a difference in behavior, as if the `if` block is executed the `elif` block won't be.

Jibola (changes requested):
Last comment!

pymongo/collation.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2024-present MongoDB, Inc.
+# Copyright 2016-present MongoDB, Inc.
```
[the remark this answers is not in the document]
NoahStapp: They should reflect the year the original, pre-async file was committed.

Jibola (approved):
LGTM

PR #1681, opened 2024-06-13
PYTHON-4347 Ensure client can be opened after fork()

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4347

This regression should have been caught in the original PR but some of the fork() tests were accidentally disabled in PYTHON-4264:
```python
    @unittest.skip("testing")
    def test_topology_reset(self):
```

We'll open another PR to re-enable those tests.

### DIFF
--- a/pymongo/asynchronous/mongo_client.py
+++ b/pymongo/asynchronous/mongo_client.py
@@ -896,6 +896,7 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
         # this closure. When the client is freed, stop the executor soon.
         self_ref: Any = weakref.ref(self, executor.close)
         self._kill_cursors_executor = executor
+        self._opened = False
 
     def _should_pin_cursor(self, session: Optional[ClientSession]) -> Optional[bool]:
         return self._options.load_balanced and not (session and session.in_transaction)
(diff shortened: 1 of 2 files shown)


pymongo/asynchronous/mongo_client.py
```diff
@@ -896,6 +896,7 @@ async def target() -> bool:
         # this closure. When the client is freed, stop the executor soon.
         self_ref: Any = weakref.ref(self, executor.close)
         self._kill_cursors_executor = executor
+        self._opened = False
```
NoahStapp, line 899: Is this a new change? Before the async beta merge we didn't set `self._opened` here, only before calling `self._init_background()`.

NoahStapp (changes requested):

pymongo/asynchronous/mongo_client.py
ShaneHarvey: Yeah _opened was added in [PYTHON-4347](https://jira.mongodb.org/browse/PYTHON-4347).

NoahStapp: Can we remove the `self._opened = False` on line 857 then? Otherwise we assign it False twice for no reason.

ShaneHarvey: Best practice is to define instance variable only in `__init__`.

NoahStapp: Oh right I forgot that's the first definition of it.

NoahStapp (approved):

PR #1684, opened 2024-06-14
PYTHON-4504 Update mod_wsgi test to not call setup.py

blink1073:


NoahStapp (approved):

.evergreen/run-mod-wsgi-tests.sh
```diff
@@ -19,7 +19,7 @@ fi
 PYTHON_VERSION=$(${PYTHON_BINARY} -c "import sys; sys.stdout.write('.'.join(str(val) for val in sys.version_info[:2]))")
 
 # Ensure the C extensions are installed.
-${PYTHON_BINARY} setup.py build_ext -i
+${PYTHON_BINARY} -m pip install -e .
```
ShaneHarvey, line 22: We can't do pip install here because it will install on the evergreen host itself. We need to create a venv first.

PR #1686, opened 2024-06-14
PYTHON-4504 [v4.8] Update mod_wsgi test to not call setup.py (#1684)

blink1073:
(cherry picked from commit fe0faf74a6a408b492e0e7f77de2f985c5d68f63)

### DIFF
--- a/.evergreen/run-mod-wsgi-tests.sh
+++ b/.evergreen/run-mod-wsgi-tests.sh
@@ -19,7 +19,10 @@ fi
 PYTHON_VERSION=$(${PYTHON_BINARY} -c "import sys; sys.stdout.write('.'.join(str(val) for val in sys.version_info[:2]))")
 
 # Ensure the C extensions are installed.
-${PYTHON_BINARY} -m pip install -e .
+${PYTHON_BINARY} -m venv --system-site-packages .venv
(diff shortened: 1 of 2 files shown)


.evergreen/run-mod-wsgi-tests.sh
```diff
@@ -19,7 +19,7 @@ fi
 PYTHON_VERSION=$(${PYTHON_BINARY} -c "import sys; sys.stdout.write('.'.join(str(val) for val in sys.version_info[:2]))")
 
 # Ensure the C extensions are installed.
-${PYTHON_BINARY} setup.py build_ext -i
+${PYTHON_BINARY} -m pip install -e .
```
ShaneHarvey, line 22: We can't do pip install here because it will install on the evergreen host itself. We need to create a venv first.

ShaneHarvey (changes requested):
We can't do pip install here because it will install on the evergreen host itself. We need to create a venv first.

.evergreen/run-mod-wsgi-tests.sh
blink1073: Updated from master

ShaneHarvey (approved):

PR #1696, opened 2024-06-21
PYTHON-2052 Truncate client metadata document to 512 bytes

shruti-sridhar:
This PR adds to the existing pymongo client metadata truncation logic by truncating the `DriverInfo` name and version fields. It also introduces tests to validate this metadata truncation.

### DIFF
--- a/pymongo/asynchronous/pool.py
+++ b/pymongo/asynchronous/pool.py
@@ -394,6 +394,26 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
         metadata["platform"] = plat
     else:
         metadata.pop("platform", None)
+    encoded_size = len(bson.encode(metadata))
+    if encoded_size <= _MAX_METADATA_SIZE:
+        return
+    # 5. Truncate driver info.
+    overflow = encoded_size - _MAX_METADATA_SIZE
+    driver = metadata.get("driver", {})
+    if driver:
+        # Truncate driver version.
+        driver_version = driver.get("version")[:-overflow]
+        if driver_version:
+            metadata["driver"]["version"] = driver_version
+        else:
+            metadata["driver"]["version"] = _METADATA["driver"]["version"]
+        encoded_size = len(bson.encode(metadata))
+        if encoded_size <= _MAX_METADATA_SIZE:
+            return
+        # Truncate driver name.
+        overflow = encoded_size - _MAX_METADATA_SIZE
+        driver_name = driver.get("name")[:-overflow]
+        metadata["driver"]["name"] = driver_name
 
 
 # If the first getaddrinfo call of this interpreter's life is on a thread,
--- a/pymongo/asynchronous/pool.py
+++ b/pymongo/asynchronous/pool.py
@@ -403,7 +403,7 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
     if driver:
         # Truncate driver version.
         driver_version = driver.get("version")[:-overflow]
-        if driver_version:
+        if len(driver_version) >= len(_METADATA["driver"]["version"]):
             metadata["driver"]["version"] = driver_version
         else:
             metadata["driver"]["version"] = _METADATA["driver"]["version"]
@@ -413,7 +413,10 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
         # Truncate driver name.
         overflow = encoded_size - _MAX_METADATA_SIZE
         driver_name = driver.get("name")[:-overflow]
-        metadata["driver"]["name"] = driver_name
+        if len(driver_name) >= len(_METADATA["driver"]["name"]):
+            metadata["driver"]["name"] = driver_name
+        else:
+            metadata["driver"]["name"] = _METADATA["driver"]["name"]
 
 
 # If the first getaddrinfo call of this interpreter's life is on a thread,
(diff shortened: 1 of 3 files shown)


test/test_client.py
```diff
@@ -361,6 +361,25 @@ def test_metadata(self):
         )
         options = client.options
         self.assertEqual(options.pool_options.metadata, metadata)
+        # Test truncating driver info metadata.
+        client = MongoClient(driver=DriverInfo(name="s" * 512))
```
NoahStapp, line 365: Instead of using magic number like `512`, use pre-defined constants. In this case, we can use `_MAX_METADATA_SIZE` to ensure we're testing against the same limit as the code is using.

test/test_client.py
```diff
@@ -361,6 +361,25 @@ def test_metadata(self):
...
+        options = client.options
+        self.assertLess(
+            len(options.pool_options.metadata["driver"]["name"]),
```
NoahStapp, line 368: We want to make sure that the entire metadata object fits within `_MAX_METADATA_SIZE` instead of verifying the specific length of a field. Try something like:

```
        self.assertLessEqual(
            len(bson.encode(options.pool_options.metadata)),
            _MAX_METADATA_SIZE,
        )
```

This also removes the need for the magic number `514`.

test/test_client.py
```diff
@@ -361,6 +361,25 @@ def test_metadata(self):
...
+            len(_METADATA["driver"]["name"]) + 514,
+        )
+        client.admin.command("isMaster")
+        client = MongoClient(driver=DriverInfo(name="s" * 512, version="s" * 512))
+        options = client.options
+        self.assertLess(
```
NoahStapp, line 374: Same here as above, we want to verify that the entire metadata fits within the limit rather than verify a specific field's length.

test/test_client.py
```diff
@@ -361,6 +361,25 @@ def test_metadata(self):
...
         options = client.options
         self.assertEqual(options.pool_options.metadata, metadata)
+        # Test truncating driver info metadata.
+        client = MongoClient(driver=DriverInfo(name="s" * 512))
+        options = client.options
+        self.assertLess(
+            len(options.pool_options.metadata["driver"]["name"]),
+            len(_METADATA["driver"]["name"]) + 514,
+        )
+        client.admin.command("isMaster")
```
NoahStapp, line 371: We can remove these commands if the tests are updated to verify the metadata length fits within the limit.

NoahStapp (changes requested):
Great start! Just a few comments on the tests.

pymongo/asynchronous/pool.py
```diff
@@ -394,6 +394,26 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
+        if driver_version:
+            metadata["driver"]["version"] = driver_version
+        else:
+            metadata["driver"]["version"] = _METADATA["driver"]["version"]
+        encoded_size = len(bson.encode(metadata))
+        if encoded_size <= _MAX_METADATA_SIZE:
+            return
+        # Truncate driver name.
+        overflow = encoded_size - _MAX_METADATA_SIZE
+        driver_name = driver.get("name")[:-overflow]
```
NoahStapp, line 415: Need to check here that `driver_name` exists after truncation, similar to `driver_version` and `platform`.

NoahStapp (changes requested):
One more minor comment.

pymongo/asynchronous/pool.py
shruti-sridhar: Would it be fine to just pop the name field if `driver_name` doesn't exist after truncation?

ShaneHarvey: I don't think we should ever intentionally drop the driver name. At least we need to send the original `_METADATA["driver"]["name"]` which is `PyMongo`.

pymongo/asynchronous/pool.py
```diff
@@ -394,6 +394,26 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
         metadata.pop("platform", None)
+    encoded_size = len(bson.encode(metadata))
+    if encoded_size <= _MAX_METADATA_SIZE:
+        return
+    # 5. Truncate driver info.
+    overflow = encoded_size - _MAX_METADATA_SIZE
+    driver = metadata.get("driver", {})
+    if driver:
+        # Truncate driver version.
+        driver_version = driver.get("version")[:-overflow]
```
ShaneHarvey, line 405: As currently written we could truncate the version from `'4.2.0|1.2.3'` to something like `'4.'` which is a bit odd. Can we at least keep the pymongo version part in full?

NoahStapp (dismissed):
Great work, looks good!

test/test_client.py
```diff
@@ -361,6 +363,21 @@ def test_metadata(self):
         )
         options = client.options
         self.assertEqual(options.pool_options.metadata, metadata)
+        # Test truncating driver info metadata.
+        client = MongoClient(driver=DriverInfo(name="s" * _MAX_METADATA_SIZE))
```
ShaneHarvey, line 367: We need to pass connect=False to MongoClient here and below since this is part of ClientUnitTest: 
```
class ClientUnitTest(unittest.TestCase):
    """MongoClient tests that don't require a server."""
```

ShaneHarvey (changes requested):
Nice work!

ShaneHarvey (approved):

NoahStapp (approved):
👍

PR #1699, opened 2024-06-24
PYTHON-4520 Add test-async tox command and automated async testing

NoahStapp:
The failing test `test/asynchronous/test_collection.py::AsyncTestCollection::test_find_one_and_write_concern` will be fixed by https://github.com/mongodb/mongo-python-driver/pull/1678.


blink1073 (approved):
LGTM!

blink1073:
`TypeError: Listeners for event_listeners must be either a CommandListener, ServerHeartbeatListener, ServerListener, TopologyListener, or ConnectionPoolListener.`

NoahStapp:
> `TypeError: Listeners for event_listeners must be either a CommandListener, ServerHeartbeatListener, ServerListener, TopologyListener, or ConnectionPoolListener.`

That error will be fixed by https://github.com/mongodb/mongo-python-driver/pull/1678, which pulls all of the `Listener` types out of the `asynchronous/synchronous` directories.

tools/synchro.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-python ./tools/synchro.py
+python ./tools/synchro.py || exit 1
```
blink1073, line 3: We should use `set -eu` instead.

NoahStapp: Good call, done.

blink1073 (dismissed):
LGTM!

tools/synchro.py
```diff
@@ -89,6 +90,8 @@
 
 type_replacements = {"_Condition": "threading.Condition"}
 
+import_replacements = {"test.synchronous": "test"}
```
NoahStapp, line 93: This is for the conversion process, due to the bulk of our tests still using the original `test/__init__.py` and associated helpers. Once we've fully migrated all tests into the `test/asynchronous` and `test/synchronous` directories, we will switch to using `test.synchronous.<module>` again.

blink1073 (dismissed):

.evergreen/run-tests.sh
```diff
@@ -258,7 +258,9 @@ if [ -z "$GREEN_FRAMEWORK" ]; then
     # Use --capture=tee-sys so pytest prints test output inline:
     # https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html
     python -m pytest -v --capture=tee-sys --durations=5 --maxfail=10 $TEST_ARGS
-    python -m pytest -v --capture=tee-sys --durations=5 --maxfail=10 test/synchronous/ $TEST_ARGS
+    if [ -z "$TEST_ARGS" ]; then # TODO: remove this once test/__init__.py doesn't share with test/synchronous/__init__.py
```
blink1073, line 261: Is there a ticket for this?

NoahStapp: Good catch, created [PYTHON-4528](https://jira.mongodb.org/browse/PYTHON-4528) and replaced the TODO with the ticket name.

blink1073:
Merge conflict, ☹️

blink1073 (approved):
LGTM!

tox.ini
```diff
@@ -63,6 +66,14 @@ extras =
     test
 commands =
     pytest -v --durations=5 --maxfail=10 {posargs}
+    pytest -v --durations=5 --maxfail=10 test/synchronous/ {posargs}
```
ShaneHarvey, line 69: Does running pytest twice generate two XML reports? Which one gets uploaded?

There should be a way to run both of these in one pytest command, no?

NoahStapp: Good point, I'm not sure how the XML reports interact. I'll open a follow-up PR with a better solution.

ShaneHarvey: Have you opened a ticket for this yet? The second XML file overwrites the first and there are actually other problems too. Like any test that relies on `posargs` gets run twice, for example the test_auth_oidc.py file gets run twice: https://evergreen.mongodb.com/task/mongo_python_driver_oidc_auth_test__platform~macos_oidc_auth_test_patch_636603f89371e626c41e48504c986e53f5122aaf_6696cbe31505d30007b20aae_24_07_16_19_39_11

NoahStapp: I haven't, sorry--this slipped past me. I don't know if there is a way to run this all in one command while keeping the existing behavior: we want to run all tests in `test/` and `test/asynchronous` by default, but override both those directories with whatever tests are specified.

ShaneHarvey: No problem. Opened https://jira.mongodb.org/browse/PYTHON-4558.

PR #1702, opened 2024-06-25
PYTHON-4521 Example transition of an existing test to an async one

NoahStapp:
This PR shows the work required to convert https://github.com/mongodb/mongo-python-driver/blob/8fbf84d314d18c2ce6e897e5d1bdd21f2493e409/test/test_logger.py to be asynchronous as well as generate the synchronous version.


blink1073:
Can we remove the top level `test_logger.py` as part of this?

NoahStapp:
> Can we remove the top level `test_logger.py` as part of this?

Good point, I've added running of the `test/synchronous` directory to https://github.com/mongodb/mongo-python-driver/pull/1699 so we can start removing the top-level test files as they're converted.

blink1073:
@NoahStapp this should be ready for an update now.

Jibola (dismissed):
LGTM!

ShaneHarvey (approved):

PR #1718, opened 2024-06-27
PYTHON-4530 - Move synchronized test code into top-level test directory

NoahStapp:
Once all the tests are converted, we will move them in a single commit to `test/synchronous`.


test/shared.py
```diff
@@ -0,0 +1,367 @@
+# Copyright 2010-present MongoDB, Inc.
```
ShaneHarvey, line 1: 2024.

Any alternative names for this file? “Shared.py” is pretty generic.

ShaneHarvey (commented):
Nice!

test/shared.py
NoahStapp: What if we combine `test/utils.py` and this new file? Having multiple files that have effectively the same purpose seems unnecessary.

NoahStapp: Renamed to `test/helpers.py` for now, we can revisit down the road.

test/__init__.py
```diff
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Test suite for pymongo, bson, and gridfs."""
+"""Asynchronous test suite for pymongo, bson, and gridfs."""
```
blink1073, line 15: Isn't this synchronous?

blink1073 (approved):
LGTM!

PR #1724, opened 2024-07-01
PYTHON-4533 Convert essential tests to async for Async PyMongo beta release

NoahStapp:
Needs https://github.com/mongodb/mongo-python-driver/pull/1718 to be merged first.


pymongo/asynchronous/mongo_client.py
```diff
@@ -1517,6 +1521,9 @@ async def close(self) -> None:
             # TODO: PYTHON-1921 Encrypted MongoClients cannot be re-opened.
             await self._encrypter.close()
 
+    async def aclose(self) -> None:
+        await self.close()
```
blink1073, line 1525: Public API should have a docstring.

pymongo/synchronous/client_session.py
```diff
@@ -691,7 +696,10 @@ async def callback(session, custom_arg, custom_kwarg=None):
         while True:
             self.start_transaction(read_concern, write_concern, read_preference, max_commit_time_ms)
             try:
-                ret = callback(self)
+                if not _IS_SYNC and iscoroutinefunction(callback):
+                    ret = callback(self)  # type: ignore[assignment]
```
blink1073, line 700: Shouldn't this `await`?

test/pymongo_mocks.py
```diff
@@ -141,13 +142,32 @@ def __init__(
...
-        kwargs["_monitor_class"] = partial(MockMonitor, self)
+        kwargs["_monitor_class"] = partial(SyncMockMonitor, self)
 
         client_options = client_context.default_client_options.copy()
         client_options.update(kwargs)
 
         super().__init__(*args, **client_options)
 
+    @classmethod
+    def get_async_mock_client(
```
blink1073, line 153: Should this be `get_sync_mock_client`?

test/test_client.py
```diff
@@ -2332,7 +2398,7 @@ class TestClientPool(MockClientTest):
     @client_context.require_connection
     def test_rs_client_does_not_maintain_pool_to_arbiters(self):
         listener = CMAPListener()
-        c = MockClient(
+        c = MockClient.get_async_mock_client(
             standalones=[],
```
blink1073, line 2402: Should this be `get_sync_mock_client`?

test/test_session.py
```diff
@@ -780,385 +831,385 @@ def test_session_not_copyable(self):
...
-                        f"{f.__name__} received no $clusterTime with {succeeded.command_name}",
-                    )
-
-                    self.assertTrue(
-                        event.command["$clusterTime"]["clusterTime"]
-                        >= succeeded.reply["$clusterTime"]["clusterTime"],
-                        f"{f.__name__} sent wrong $clusterTime with {event.command_name}",
-                    )
-
+# class TestCausalConsistency(unittest.TestCase):
```
blink1073, line 834: TODO?

pymongo/synchronous/client_session.py
blink1073: oh, nvm, this is the sync file.  :smile:

ShaneHarvey (commented):
Does anyone else feel there's a bit too much going on in this one PR to review effectively? Can you split it up into smaller and more manageable pieces?

pymongo/asynchronous/mongo_client.py
ShaneHarvey: Why do we need aclose() when we already have close()?

pymongo/asynchronous/mongo_client.py
```diff
@@ -861,6 +861,10 @@ def __init__(
             # This will be used later if we fork.
             AsyncMongoClient._clients[self._topology._topology_id] = self
 
+    async def connect(self) -> None:
```
ShaneHarvey, line 864: "Connect" might not be the best name here since we're not actually connecting. I know we have the "connect=True" argument but I can't help but feel like "open()" would be more appropriate. If a user actually wanted to connect I would expect them to run something like "db.command('ping')". What do you think?

NoahStapp:
> Does anyone else feel there's a bit too much going on in this one PR to review effectively? Can you split it up into smaller and more manageable pieces?

Good point, let me try.

pymongo/asynchronous/mongo_client.py
NoahStapp: It feels a bit odd to me for the synchronous API to have a `connect=True` option but the equivalent async option is called `open` instead even though they do the same thing. Consistency of naming makes more sense to me here.

pymongo/asynchronous/mongo_client.py
NoahStapp: `contextlib` expects that asynchronous context managers use `aclose`. We could simplify by just setting `aclose = close` instead.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: In that case I think the docstring just needs to explain the concept more clearly.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: In that case should we remove "close()"?

pymongo/synchronous/mongo_client.py
```diff
@@ -860,6 +860,10 @@ def __init__(
             # This will be used later if we fork.
             MongoClient._clients[self._topology._topology_id] = self
 
+    def connect(self) -> None:
+        """Explicitly connect synchronously."""
+        self._get_topology()
```
ShaneHarvey, line 865: We should avoid adding connect() and aclose() to the sync client.

test/__init__.py
```diff
@@ -1051,3 +1111,95 @@ def print_running_clients():
...
+    cmd.update(**kwargs)
+    return authdb.command(cmd)
+
+
+def connected(client):
+    """Convenience to wait for a newly-constructed client to connect."""
+    with warnings.catch_warnings():
+        # Ignore warning that ping is always routed to primary even
+        # if client's read preference isn't PRIMARY.
+        warnings.simplefilter("ignore", UserWarning)
```
ShaneHarvey, line 1135: I've never seen this warning. What is it referring to?

test/asynchronous/__init__.py
```diff
@@ -1053,3 +1113,95 @@ def print_running_clients():
...
+    interval = sys.getswitchinterval()
+    sys.setswitchinterval(1e-6)
+
+    try:
+        yield
+    finally:
+        sys.setswitchinterval(interval)
+
+
+def lazy_client_trial(reset, target, test, get_client):
```
ShaneHarvey, line 1188: Does this even make sense to test in async? I would expect an async client to only be useable from a single thread since an I/O loop only runs on a single thread.

test/test_transactions.py
```diff
@@ -389,7 +392,7 @@ def test_transaction_direct_connection(self):
             with client.start_session() as s, s.start_transaction():
                 res = f(*args, session=s)  # type:ignore[operator]
                 if isinstance(res, (CommandCursor, Cursor)):
-                    list(res)
+                    res.to_list()
```
ShaneHarvey, line 395: Can you add a docstring and changelog entry for to_list?

ShaneHarvey (changes requested):

pymongo/asynchronous/client_session.py
```diff
@@ -598,7 +601,10 @@ def _inherit_option(self, name: str, val: _T) -> _T:
 
     async def with_transaction(
         self,
-        callback: Callable[[AsyncClientSession], _T],
+        callback: Union[
+            Callable[[AsyncClientSession], Coroutine[Any, Any, _T]],
```
ShaneHarvey, line 605: Should we use Coroutine here or Awaitable? Which is a better fit?

pymongo/asynchronous/client_session.py
```diff
@@ -598,7 +601,10 @@ def _inherit_option(self, name: str, val: _T) -> _T:
...
+            Callable[[AsyncClientSession], _T],
```
ShaneHarvey, line 606: Why are there two callables here?

pymongo/asynchronous/client_session.py
```diff
@@ -693,7 +699,10 @@ async def callback(session, custom_arg, custom_kwarg=None):
                 read_concern, write_concern, read_preference, max_commit_time_ms
             )
             try:
-                ret = callback(self)
+                if not _IS_SYNC and iscoroutinefunction(callback):
```
ShaneHarvey, line 702: Why check IS_SYNC or iscoroutinefunction here? Shouldn't it just be:
```python
ret = await callback(self)
```

ShaneHarvey (changes requested):

pymongo/asynchronous/mongo_client.py
NoahStapp: I think that's better, yeah.

test/pymongo_mocks.py
NoahStapp: Good catch, using `get_mock_client` seems better.

test/test_session.py
NoahStapp: Whoops yes good catch.

test/__init__.py
NoahStapp: This already exists in our test suite: https://github.com/mongodb/mongo-python-driver/blob/554ce7d9844ea3b49f71a7abb6c5ad6101dea131/test/utils.py#L866

pymongo/synchronous/mongo_client.py
NoahStapp: What if we rename them to be private? Like `aconnect` -> `_aconnect` and `aclose` -> `_aclose`? There isn't a clean way of having the synchro process just not convert specific methods.

test/asynchronous/__init__.py
NoahStapp: It isn't being tested at all in async, not sure why I moved it here. Putting back into `test/utils.py`.

pymongo/synchronous/client_session.py
```diff
@@ -598,7 +600,10 @@ def _inherit_option(self, name: str, val: _T) -> _T:
 
     def with_transaction(
         self,
-        callback: Callable[[ClientSession], _T],
+        callback: Union[
+            Callable[[ClientSession], _T],
+            Callable[[ClientSession], _T],
+        ],
```
Jibola, line 606: It looks like the type definition is the same twice here. Is that needed?

pymongo/synchronous/mongo_client.py
```diff
@@ -1512,6 +1516,9 @@ def close(self) -> None:
             # TODO: PYTHON-1921 Encrypted MongoClients cannot be re-opened.
             self._encrypter.close()
 
+    def aclose(self) -> None:
```
Jibola, line 1519: Same as the comment from @ShaneHarvey before. We can remove close

test/__init__.py
```diff
@@ -812,6 +811,12 @@ def require_no_api_version(self, func):
             func=func,
         )
 
+    def require_sync(self, func):
+        """Run a test only if using the synchronous API."""
+        return self._require(
+            lambda: _IS_SYNC, "This test only works with the synchronous API", func=func
```
Jibola, line 817: Since _IS_SYNC is defined as true, would the lambda ever fail?

test/__init__.py
```diff
@@ -919,6 +924,32 @@ def _target() -> None:
             self.assertEqual(proc.exitcode, 0)
 
 
+class UnitTest(PyMongoTestCase):
+    """Async base class for TestCases that don't require a connection to MongoDB."""
+
+    @classmethod
+    def setUpClass(cls):
+        if _IS_SYNC:
```
Jibola, line 932: Same comment. How does/may _IS_SYNC get redefined when tests are running?

test/test_transactions.py
NoahStapp: `list(cursor)` still works on synchronous cursors, but asynchronous cursors require the use of `Cursor.to_list()` instead. This change is purely so our tests can be synchronized smoothly.

test/__init__.py
```diff
@@ -1051,3 +1111,95 @@ def print_running_clients():
...
+        # if client's read preference isn't PRIMARY.
+        warnings.simplefilter("ignore", UserWarning)
+        client.admin.command("ping")  # Force connection.
+
+    return client
+
+
+def drop_collections(db: Database):
+    # Drop all non-system collections in this database.
+    for coll in db.list_collection_names(filter={"name": {"$regex": r"^(?!system\.)"}}):
```
Jibola, line 1143: Do we need to filter the non system collection names? I had assumed those couldn't be dropped.

test/__init__.py
ShaneHarvey: I did some digging and that warning was removed in https://jira.mongodb.org/browse/PYTHON-814 in PyMongo 3.0. Could you remove the catch_warnings block?

test/test_transactions.py
ShaneHarvey: Right, the only issue is that to_list() is undocumented.

test/__init__.py
```diff
@@ -1051,3 +1111,95 @@ def print_running_clients():
...
+    threads = []
+    for i in range(NTHREADS):
+        bound_target = partial(target, collection, i)
+        threads.append(threading.Thread(target=bound_target))
+
+    for t in threads:
+        t.start()
+
+    for t in threads:
+        t.join(60)
```
Jibola, line 1170: 60 seconds is quite a bit of time. Have we been doing 60 historically?

pymongo/synchronous/mongo_client.py
ShaneHarvey: How about something like this?:
```python
_IS_SYNC = False
class AsyncMongoClient:
    if not _IS_SYNC:
        async def aclose(self):
            pass
```

test/__init__.py
```diff
@@ -1051,3 +1111,95 @@ def print_running_clients():
...
+    `test` takes the lazily-connecting collection and asserts a
+    post-condition to prove `target` succeeded.
+    """
+    collection = client_context.client.pymongo_test.test
+
+    with frequent_thread_switches():
+        for _i in range(NTRIALS):
+            reset(collection)
+            lazy_client = get_client()
+            lazy_collection = lazy_client.pymongo_test.test
```
Jibola, line 1203: NIT: do we need to redefine the lazy_client each time?

Jibola (changes requested):
Halfway through! Left some comments for now.

test/__init__.py
NoahStapp: This file is generated from `test/asynchronous/__init__.py`, it always passes for a synchronous test, as intended.

test/__init__.py
NoahStapp: `_IS_SYNC` is always true for a synchronous file and always false for an asynchronous file.

pymongo/synchronous/mongo_client.py
ShaneHarvey: aclose -> close SGTM. For aconnect() I would prefer the approach I outlined above.

ShaneHarvey: Actually. renaming aconnect -> _aconnect is fine too.

NoahStapp: The synchro process will still convert the `aconnect` method within the `if not _IS_SYNC` block and add it to the sync client source. Is that acceptable?

ShaneHarvey: I'm good with any approach as long as aconnect() or connect() isn't a public method on MongoClient.

test/__init__.py
ShaneHarvey: Yes, the point of the repeating the test is to run with a new client each time.

test/__init__.py
ShaneHarvey: Yeah, 60 is just a fail safe in case something goes wrong.

test/__init__.py
ShaneHarvey: This is ancient code, probably it's to avoid attempting to drop a system collection and getting an OperationFailure because it's not allowed.

pymongo/asynchronous/client_session.py
NoahStapp: We've used `Coroutine` consistently so far, `Awaitable` is a less-specific type hint that allows several other objects besides asynchronous functions.

pymongo/asynchronous/client_session.py
NoahStapp: An artifact of the reasoning below, removed.

pymongo/asynchronous/client_session.py
NoahStapp: I was trying to be more lenient and allow both async and sync callbacks, but it's better to require all asynchronous callbacks for the Async API.

pymongo/synchronous/client_session.py
NoahStapp: No, this is an artifact of supporting both kinds of callbacks in the async API. Will be resolved in the separate PR for `test_transaction`

NoahStapp:
Split into #1730, #1731, #1732, and #1733.

PR #1726, opened 2024-07-06
PYTHON-4538 await _check_user_provided

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4538

Note that the SSL tests are still failing due to https://jira.mongodb.org/browse/PYTHON-4539.


NoahStapp:
Good catch, this is also fixed in https://github.com/mongodb/mongo-python-driver/pull/1724.

Jibola (approved):

PR #1727, opened 2024-07-09
Remove "/" from dbase

JoaoAssalim:
Remove the / to avoid communication errors and thus leave only the database name in the dbase variable

### DIFF
--- a/pymongo/uri_parser.py
+++ b/pymongo/uri_parser.py
@@ -502,6 +502,8 @@ def parse_uri(
 
     if dbase:
         dbase = unquote_plus(dbase)
+        if "/" == dbase[0]:
+            dbase = dbase[1:]
         if "." in dbase:
             dbase, collection = dbase.split(".", 1)
         if _BAD_DB_CHARS.search(dbase):


pymongo/uri_parser.py
```diff
@@ -502,6 +502,8 @@ def parse_uri(
 
     if dbase:
         dbase = unquote_plus(dbase)
+        if "/" == dbase[0]:
```
ShaneHarvey, line 505: When/why would the database start with a "/"?

ShaneHarvey (changes requested):

pymongo/uri_parser.py
ShaneHarvey: Can you give an example of your use case?

JoaoAssalim: This had never occurred to me before, but today implementing it on a new system, I had the problem with the dbase variable starting with "/" and by doing this the problem was solved

JoaoAssalim: example of my problem:

variables on .env:

```
CONNECTION_STRING=mongodb://user:password@localhost:27017
DATABASE_NAME=db_name
```

And the value of variable `host_plus_db_part` was `user:password@localhost:27017//db_name`.
So, when the code makes `host_part, _, dbase = host_plus_db_part.partition("/")`, the value in dbase is `/db_name`

ShaneHarvey: The only way for this to happen is if the database name portion incorrectly starts with a percent encoded `/`, like this:
```python
>>> uri_parser.parse_uri('mongodb://localhost/dbname')
{'nodelist': [('localhost', 27017)], 'username': None, 'password': None, 'database': 'dbname', 'collection': None, 'options': {}, 'fqdn': None}
>>> uri_parser.parse_uri('mongodb://localhost/%2Fdbname')
  File "/Users/shane/git/mongo-python-driver/pymongo/uri_parser.py", line 508, in parse_uri
    raise InvalidURI('Bad database name "%s"' % dbase)
pymongo.errors.InvalidURI: Bad database name "/dbname"
```

PyMongo is raising an InvalidURI error which is correct because MongoDB database names can't contain `/`.

JoaoAssalim: But this was a feature that worked in other versions. And I believe that this validator and remover of / in dbase can be something useful and will not bring harm to the code

ShaneHarvey: This was an intentional change in PyMongo 3.5 in https://jira.mongodb.org/browse/PYTHON-1282 circa 2017.

Before that change, pymongo would silently ignore the invalid database name and an error wouldn't be raised until get_default_database was called:
```python
>>> client = MongoClient('mongodb://localhost//dbname')
>>> client.get_default_database()
  File "/Users/shane/git/mongo-python-driver/pymongo/mongo_client.py", line 1225, in get_default_database
    return self[self.__default_database_name]
  File "/Users/shane/git/mongo-python-driver/pymongo/mongo_client.py", line 999, in __getitem__
    return database.Database(self, name)
  File "/Users/shane/git/mongo-python-driver/pymongo/database.py", line 106, in __init__
    _check_name(name)
  File "/Users/shane/git/mongo-python-driver/pymongo/database.py", line 46, in _check_name
    raise InvalidName("database names cannot contain the "
pymongo.errors.InvalidName: database names cannot contain the character '/'
```

Personally I don't like the idea of adding back this behavior, or the idea of selectively ignoring certain characters in the name.

I suggest updating your app to avoid generating the invalid URI. The [connection string spec is pretty clear that this isn't supported](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.md#auth-database-optional):

> Auth Database (optional)
> The database to authenticate against. If provided it is everything after the Host Information (ending with "/") and up to the first question mark ("?") or end of string. The auth database MUST be URL decoded by the parser.
> The following characters MUST NOT appear in the database name, once it has been decoded: slash ("/"), backslash (""), space (" "), double-quote ("""), or dollar sign ("$").

JoaoAssalim: Okay, no problem! Thanks for your time!

JoaoAssalim:
Close pull request

PR #1729, opened 2024-07-10
PYTHON-4458 Test against Python 3.13 beta

blink1073:


.github/workflows/test-python.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        python-version: ["3.8", "3.11", "pypy-3.9"]
+        python-version: ["3.8", "3.11", "pypy-3.9", "3.13"]
```
ShaneHarvey, line 54: Should we remove 3.11 now?

ShaneHarvey (changes requested):

.github/workflows/test-python.yml
blink1073: Sure, we're still testing it in EVG.

blink1073, line 54:
```suggestion
        python-version: ["3.8", "pypy-3.9", "3.13"]
```

ShaneHarvey (approved):

PR #1731, opened 2024-07-10
PYTHON-4533 - Convert test/test_cursor.py to async

NoahStapp:
Requires https://github.com/mongodb/mongo-python-driver/pull/1730 to be merged first.


test/test_cursor.py
```diff
@@ -23,13 +23,14 @@
 import sys
 import threading
 import time
+from test import IntegrationTest, client_context
```
ShaneHarvey, line 26: "test" imports should be grouped with the others lower in the file.

test/test_cursor.py
```diff
@@ -245,7 +253,7 @@ def test_max_await_time_ms(self):
         listener.reset()
 
         # Tailable_await with max_time_ms
-        list(coll.find(cursor_type=CursorType.TAILABLE_AWAIT).max_time_ms(99))
```
ShaneHarvey, line 248: We need to keep at least one test that uses `list(cursor)` to ensure that it's supported.

test/test_cursor.py
```diff
@@ -665,27 +692,27 @@ def test_skip(self):
             self.assertEqual(i["x"], 0)
             break
 
-        for i in db.test.find().skip(20):
+        for i in (db.test.find()).skip(20):
```
ShaneHarvey, line 695: It looks like a large percentage of these changes are the extra parenthesis because async collection defined find() as async. We should consider making it a sync method since it's not intended to perform any I/O until it's first iterated anyway.

ShaneHarvey (changes requested):

test/test_cursor.py
NoahStapp: `find` is only async because we need to call [_supports_exhaust](https://github.com/mongodb/mongo-python-driver/blob/d0193eb0456aebcd06615f4aa1647475bc79d57e/pymongo/asynchronous/cursor.py#L240) after creating a cursor. If we want to keep that existing behavior, we need to make `_server_property` and its use of `_get_topology()` and `select_server` synchronous.

ShaneHarvey: Instead of changing _server_property, AsyncCursor should delay checking exhaust support until the cursor is first iterated.

ShaneHarvey (changes requested):

NoahStapp:
Waiting on https://github.com/mongodb/mongo-python-driver/pull/1734 before updating the tests here.

test/asynchronous/test_cursor.py
```diff
@@ -260,8 +259,11 @@ async def test_max_await_time_ms(self):
         self.assertEqual(99, listener.started_events[1].command["maxTimeMS"])
         listener.reset()
 
-        # Tailable_await with max_time_ms
-        await (await coll.find(cursor_type=CursorType.TAILABLE_AWAIT)).max_time_ms(99).to_list()
+        # Tailable_await with max_time_ms and make sure list() works on synchronous cursors
+        if _IS_SYNC:
+            list(await coll.find(cursor_type=CursorType.TAILABLE_AWAIT).max_time_ms(99))
```
ShaneHarvey: This is good but I'd also like to see a dedicated test for list() on a regular cursor, with multiple batches. Like "test_iteration_with_list"

pymongo/asynchronous/cursor.py
```diff
@@ -242,6 +242,7 @@ def __init__(
             self._exhaust_checked = True
             self._supports_exhaust()  # type: ignore[unused-coroutine]
         else:
+            self._exhaust = cursor_type == CursorType.EXHAUST
```
ShaneHarvey, line 245: Can `self._exhaust` be removed? It looks unused.

ShaneHarvey (changes requested):

test/asynchronous/test_cursor.py
NoahStapp: Explicitly testing the behavior on a synchronous cursor? Makes sense, will add.

pymongo/asynchronous/cursor.py
ShaneHarvey: Oh so it is being used, then never mind.

ShaneHarvey (approved):

PR #1734, opened 2024-07-15
PYTHON-4552 - Make AsyncCollection.find synchronous

NoahStapp:


test/test_collection.py
```diff
@@ -2227,9 +2228,16 @@ def test_helpers_with_let(self):
         for let in [10, "str", [], False]:
             for helper, args in helpers:
                 with self.assertRaisesRegex(TypeError, "let must be an instance of dict"):
-                    helper(*args, let=let)  # type: ignore
+                    # c.find() is synchronous, can't await it
+                    if helper == c.find:
+                        helper(*args, let=let)  # type: ignore
+                    else:
+                        helper(*args, let=let)  # type: ignore
```
ShaneHarvey, line 2235: Should this have an await?

test/test_collection.py
```diff
@@ -2227,9 +2228,16 @@ def test_helpers_with_let(self):
...
         for helper, args in helpers:
-            helper(*args, let={})  # type: ignore
+            if helper == c.find:
+                helper(*args, let={})  # type: ignore
+            else:
+                helper(*args, let={})  # type: ignore
```
ShaneHarvey, line 2240: Same here

ShaneHarvey (changes requested):

test/asynchronous/test_collection.py
```diff
@@ -2258,9 +2251,16 @@ async def test_helpers_with_let(self):
         for let in [10, "str", [], False]:
             for helper, args in helpers:
                 with self.assertRaisesRegex(TypeError, "let must be an instance of dict"):
-                    await helper(*args, let=let)  # type: ignore
+                    # c.find() is synchronous, can't await it
```
ShaneHarvey, line 2254: It could be cleaner here to do this:
```python
async def afind(*args, **kawrgs):
    return c.find(*args, **kwargs)
    
helpers = [
...
            (afind, ({})),

]
```
That way we can always await here.

test/test_collection.py
ShaneHarvey: Forget this, I was looking at the wrong file.

ShaneHarvey (approved):
LGTM, the outstanding comment was just an idea and doesn't need to be implemented.

test/asynchronous/test_collection.py
NoahStapp: I like it, done.

ShaneHarvey (approved):

PR #1736, opened 2024-07-16
PYTHON-4537 - Use selector asyncio loop on windows

NoahStapp:


pymongo/__init__.py
```diff
@@ -108,6 +110,11 @@
 version = __version__
 """Current version of PyMongo."""
 
+# The default asyncio loop implementation on Windows has issues with sharing sockets across processes
+# We explicitly use a different loop implementation here to prevent that issue
+if sys.platform == "win32":
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
ShaneHarvey, line 116: This seems unreasonable to do as a python library. At best we could do this in the test suite, but not in pymongo itself. What is the actual underlying problem? Pymongo isn't supposed to be sharing sockets across processes (only across threads).

ShaneHarvey (changes requested):

pymongo/__init__.py
NoahStapp: Oh whoops this was meant to be in `test/__init__.py` not the top-level one, my bad.

NoahStapp: I'm not sure what the underlying problem is, the only information I could gather on the error we're seeing was that it is a known issue with asyncio and the default loop on Windows when dealing with sockets across processes. (https://github.com/encode/uvicorn/issues/529). Maybe the issue extends to sharing sockets across threads? Which tests fail and how many seems to change every run of the async tests on Windows.

NoahStapp: If that is concurrent sockets, I can't imagine we ever get close to 512 sockets being used at once.

NoahStapp: Running the async test suite on windows with this loop enabled passes without issue.

PR #1738, opened 2024-07-16
PYTHON-4558 Fix coverage and XML results for async tests

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4558

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -30,6 +30,7 @@ set -o xtrace
 
 AUTH=${AUTH:-noauth}
 SSL=${SSL:-nossl}
+TEST_SUITES="test/ test/asynchronous/"
 TEST_ARGS="${*:1}"
 
 export PIP_QUIET=1  # Quiet by default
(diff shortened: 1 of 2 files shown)


.evergreen/run-tests.sh
```diff
@@ -257,8 +254,7 @@ PIP_QUIET=0 python -m pip list
 if [ -z "$GREEN_FRAMEWORK" ]; then
     # Use --capture=tee-sys so pytest prints test output inline:
     # https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html
-    python -m pytest -v --capture=tee-sys --durations=5 --maxfail=10 $TEST_ARGS
-    python -m pytest -v --capture=tee-sys --durations=5 --maxfail=10 test/asynchronous/ $TEST_ARGS
+    python -m pytest -v --capture=tee-sys --durations=5 --maxfail=10 $TEST_SUITES $TEST_ARGS
```
NoahStapp, line 257: Nice, this is a clean solution. Do we still need `$TEST_ARGS`?

.github/workflows/test-python.yml
```diff
@@ -205,12 +205,11 @@ jobs:
...
+        - name: Run connect test from sdist
           shell: bash
           run: |
             cd sdist/test
             ls
             which python
             pip install -e ".[test]"
-            pytest -v
-            pytest -v test/asynchronous/
+            PYMONGO_MUST_CONNECT=1 pytest -v test/test_client_context.py
```
NoahStapp, line 215: Has this test been just running the wrong suite this whole time?

NoahStapp (changes requested):

.github/workflows/test-python.yml
ShaneHarvey: No but it just seems unnessecary to run the whole suite from the sdist, we already run the whole suite elsewhere on the same commit.

.evergreen/run-tests.sh
ShaneHarvey: TEST_ARGS is still used to add `--cov`.

NoahStapp (approved):

PR #1741, opened 2024-07-22
PYTHON-4565 - Use pytest markers for test suite configuration

NoahStapp:


NoahStapp:
I can't reproduce the typechecking failures locally, could it be a toolchain issue?

blink1073:
PyOpenSSL just had a release you might be using an older version locally.

NoahStapp:
> PyOpenSSL just had a release, you might be using an older version locally.

Just updated to the newest version, still no issue locally.

blink1073:
It isn't the toolchain, because it is happening on GitHub.

blink1073:
Note it also occurs in #1742

blink1073:
Ah, they just recently added type hints to the library itself, I assume we were using types-pyopenssl before.  Can you open a ticket?

https://github.com/pyca/pyopenssl/blob/main/src/OpenSSL/py.typed

NoahStapp:
> Ah, they just recently added type hints to the library itself, I assume we were using types-pyopenssl before. Can you open a ticket?
> 
> [pyca/pyopenssl@`main`/src/OpenSSL/py.typed](https://github.com/pyca/pyopenssl/blob/main/src/OpenSSL/py.typed?rgh-link-date=2024-07-23T19%3A04%3A44Z)

Did we explicitly include `types-pyopenssl` anywhere? I don't see any references.

blink1073:
> Did we explicitly include types-pyopenssl anywhere? I don't see any references.

It gets pulled in by `--install-types` if there is no `py.typed` in the package.

blink1073 (approved):
LGTM!

PR #1745, opened 2024-07-24
PYTHON-4550 Add MongoClient.bulk_write API

shruti-sridhar:
## Summary

This PR implements the `MongoClient.bulkWrite` API described in this [spec](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md). This API allows users to perform insert, update, and delete operations against mixed namespaces and receive detailed results for each operation performed. 

This new API is distinct from `MongoCollection.bulkWrite` in that it supports writes across many namespaces, and it uses the `bulkWrite` server command rather than individual `insert`/`update`/`delete` server commands.

### In `pymongo/asynchronous` and `pymongo/synchronous`:
* `mongo_client.py`: contains the user-facing `bulk_write()` function

* `client_bulk.py`: contains the `_AsyncClientBulk`/`_ClientBulk` class, which handles the internal details of parsing the input write models, sending batches of commands to the server, handling the response, and returning the result

### In `pymongo`:
* `client_bulk_shared.py`: contains the `ClientBulkWriteException` class and a helper to merge results across batches

* `message.py`: contains the `_ClientBulkWriteContext` class and associated functions to construct and send OP_MSG
    * the batch-splitting logic is in the function `_client_batched_op_msg_impl()`
    * the OP_MSG construction logic is in the function `_client_construct_op_msg()`
    
* `operations.py`: contains the write model classes

* `results.py`: contains the `ClientBulkWriteResult` class as well as individual write result classes

## Testing

All the added JSON files in this PR are spec tests that were synced from [this commit](https://github.com/mongodb/specifications/commit/10919c92cb784e9730551841e57541bee2e1af93). The unified test runner `test/unified_format.py` contains the changes required for these spec tests to run. 

The prose tests from that commit are implemented in `test/asynchronous/test_client_bulk_write.py` and `test/test_client_bulk_write.py`.


pymongo/asynchronous/mongo_client.py
```diff
@@ -2204,10 +2224,136 @@ async def drop_database(
...
+
+        Requests are passed as a list of write operation instances (
+        :class:`~pymongo.operations.ClientInsertOne`,
+        :class:`~pymongo.operations.ClientUpdateOne`,
+        :class:`~pymongo.operations.ClientUpdateMany`,
+        :class:`~pymongo.operations.ClientReplaceOne`,
+        :class:`~pymongo.operations.ClientDeleteOne`, or
+        :class:`~pymongo.operations.ClientDeleteMany`).
+
+          >>> for doc in db.test.find({}):
```
NoahStapp, line 2249: These inline examples should be asynchronous since this is in the asynchronous API.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2204,10 +2224,136 @@ async def drop_database(
...
+          ...
+          >>> for doc in db.coll.find({}):
+          ...     print(doc)
+          ...
+          {'x': 2, '_id': ObjectId('507f1f77bcf86cd799439011')}
+          {'y': 2, '_id': ObjectId('507f1f77bcf86cd799439012')}
+
+        :param models: A list of write operation instances.
+        :param session: (optional) An instance of
+            :class:`~pymongo.client_session.AsyncClientSession`.
```
NoahStapp, line 2291:
```suggestion
            :class:`~pymongo.asynchronous.client_session.AsyncClientSession`.
```

pymongo/message.py
```diff
@@ -878,6 +888,368 @@ def _do_batched_op_msg(
     return _batched_op_msg(operation, command, docs, ack, opts, ctx)
 
 
+class _ClientBulkWriteContext:
```
NoahStapp, line 891: This class has a lot of crossover with the existing `_BulkWriteContext`. Would a parent class with the common functionality to reduce code duplication make sense here?

pymongo/operations.py
```diff
@@ -501,6 +502,451 @@ def _add_to_bulk(self, bulkobj: _AgnosticBulk) -> None:
         )
 
 
+class ClientInsertOne(Generic[_DocumentType]):
```
NoahStapp, line 505: All of these Model classes look identical to the existing collection-level bulk write API ones except for having a namespace. In the interest of code consolidation, how about adding an optional `namespace` to the existing classes and a new `_add_to_client_bulk` method?

pymongo/results.py
```diff
@@ -81,6 +81,26 @@ def inserted_id(self) -> Any:
         return self.__inserted_id
 
 
+class ClientInsertOneResult(_WriteResult):
```
NoahStapp, line 84: These classes are also very similar to the existing ones, with the removal of `acknowledged`. Could consolidation also be done here by making `acknowledged` optional?

test/unified_format.py
```diff
@@ -288,11 +297,122 @@ def parse_bulk_write_result(result):
...
+    if op_type == "insert":
+        return {"insertedId": result.inserted_id}
+    if op_type == "update":
+        if result.upserted_id:
+            return {
+                "matchedCount": result.matched_count,
+                "modifiedCount": result.modified_count,
+                "upsertedId": result.upserted_id,
+            }
+        return {
```
NoahStapp, line 372: Add an `else` here for clarity.

NoahStapp (changes requested):
Great start, really impressive work! Some formatting and clarity comments and a few larger refactor questions.

pymongo/asynchronous/mongo_client.py
shruti-sridhar: Is it okay for the inline examples in the synchronous API to be async? (since they get overwritten by the lint-manual)

NoahStapp: The inline examples should be converted to synchronous as part of the synchro process, are you seeing some remain async?

shruti-sridhar: When I add `async for` or `await` as part of the inline examples, they remain there in the sync version as well

NoahStapp (changes requested):
A few minor comments, great refactor!

pymongo/asynchronous/mongo_client.py
```diff
@@ -2246,13 +2246,13 @@ async def bulk_write(
         :class:`~pymongo.operations.DeleteOne`, or
         :class:`~pymongo.operations.DeleteMany`).
 
-          >>> async for doc in await db.test.find({}):
+          >>> for doc in db.test.find({}):
```
NoahStapp, line 2249: Sorry, we still need the `async for` here and below since `find` returns an asynchronous cursor.

NoahStapp (changes requested):

pymongo/asynchronous/mongo_client.py
```diff
@@ -2204,10 +2224,136 @@ async def drop_database(
...
+            constant or closed expressions that do not reference document
+            fields. Parameters can then be accessed as variables in an
+            aggregate expression context (e.g. "$$var").
+        :param write_concern: (optional) The write concern to use for this bulk write.
+
+        :return: An instance of :class:`~pymongo.results.ClientBulkWriteResult`.
+
+        .. seealso:: :ref:`writes-and-ids`
+
+        .. note:: `bypass_document_validation` requires server version
```
blink1073, line 2314: We only support 3.6+, so we can remove this notes.  We should document that this features requires MongoDB 8.0+ instead.

pymongo/asynchronous/client_bulk.py
```diff
@@ -0,0 +1,765 @@
...
+        conn: AsyncConnection,
+    ) -> None:
+        """Execute all operations, returning no results (w=0)."""
+        if self.uses_collation:
+            raise ConfigurationError("Collation is unsupported for unacknowledged writes.")
+        if self.uses_array_filters:
+            raise ConfigurationError("arrayFilters is unsupported for unacknowledged writes.")
+        # Guard against unsupported unacknowledged writes.
+        unack = self.write_concern and not self.write_concern.acknowledged
+        if unack and self.uses_hint_delete and conn.max_wire_version < 9:
```
blink1073, line 724: I think we can remove these checks and instead add a top-level check for max_wire_version < 25 (MongoDB 8.0) and raise an error that it is not supported.

pymongo/message.py
```diff
@@ -878,6 +923,305 @@ def _do_batched_op_msg(
...
+        if doc_size > limit:
+            _raise_document_too_large(op_type, doc_size, limit)
+        return doc_size
+
+    max_bson_size = ctx.max_bson_size
+    max_write_batch_size = ctx.max_write_batch_size
+    max_message_size = ctx.max_message_size
+
+    command_doc = _dict_to_bson(command, False, opts)
+    # Don't include command-agnostic fields in document size calculations.
```
ShaneHarvey, line 1061: Is this behavior described in the spec? What is a command-agnostic field?

pymongo/message.py
```diff
@@ -878,6 +923,305 @@ def _do_batched_op_msg(
...
+        if ns_doc_to_send:
+            to_send_ns.append(ns_doc_to_send)
+            total_ns_length += ns_length
+        idx += 1
+        # We have enough documents, return this batch.
+        if idx == max_write_batch_size:
+            break
+
+    # Construct the entire OP_MSG.
+    length = _client_construct_op_msg(command_doc, to_send_ops, to_send_ns, ack, opts, buf)
```
ShaneHarvey, line 1135: command_doc is bytes here but _client_construct_op_msg accepts a `command: Mapping[str, Any]`. Which type is correct? Shouldn't the mypy test be failing on that?

pymongo/message.py
```diff
@@ -878,6 +923,305 @@ def _do_batched_op_msg(
...
+    total_ops_length = 0
+    total_ns_length = 0
+    idx = 0
+
+    for real_op_type, op_doc in operations:
+        op_type = real_op_type
+        # Check insert/replace document size if unacknowledged.
+        if real_op_type == "insert":
+            if not ack:
+                _check_doc_size_limits(real_op_type, op_doc["document"], max_bson_size)
```
ShaneHarvey, line 1092: We should only need to bson encode each operation once, but _check_doc_size_limits causes the op to be encoded a second time. Could you open a ticket to improve that? We don't need to do it in this PR since it's just a perf issue.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2204,10 +2224,135 @@ async def drop_database(
...
+          {'x': 1, '_id': ObjectId('54f62e60fba5226811f634f0')}
+          ...
+          >>> async for doc in db.coll.find({}):
+          ...     print(doc)
+          ...
+          {'x': 2, '_id': ObjectId('507f1f77bcf86cd799439011')}
+          ...
+          >>> # DeleteMany, UpdateOne, and UpdateMany are also available.
+          >>> from pymongo import InsertOne, DeleteOne, ReplaceOne
+          >>> models = [InsertOne("db.test", {'y': 1}),
```
ShaneHarvey, line 2262: The namespace has to be the last positional argument, or has to be passed by keyword.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2204,10 +2224,135 @@ async def drop_database(
...
+          ...           DeleteOne("db.test", {'x': 1}),
+          ...           InsertOne("db.coll", {'y': 2}),
+          ...           ReplaceOne("db.test", {'w': 1}, {'z': 1}, upsert=True)]
+          >>> client = AsyncMongoClient()
```
ShaneHarvey, line 2266: This calls above use `db.test` so we don't need to create a new client here. We can just do: `client.bulk_write` on the existing client.

pymongo/operations.py
```diff
@@ -80,48 +81,121 @@ class _Op(str, enum.Enum):
...
         self._doc = document
+        self._namespace = namespace
 
     def _add_to_bulk(self, bulkobj: _AgnosticBulk) -> None:
         """Add this operation to the _AsyncBulk/_Bulk instance `bulkobj`."""
         bulkobj.add_insert(self._doc)  # type: ignore[arg-type]
 
+    def _add_to_client_bulk(self, bulkobj: _AgnosticClientBulk) -> None:
+        """Add this operation to the _AsyncClientBulk/_ClientBulk instance `bulkobj`."""
+        bulkobj.add_insert(self._namespace, self._doc)  # type: ignore[arg-type]
```
ShaneHarvey, line 111: It should be an error to call client.bulk_write() with an operation that doesn't have a namespace. We should add a check and remove the type ignore here. Could you add a test case for this?
```python 
client.bulk_write([InsertOne({})]) # Should raise error due to missing namespace.
```

ShaneHarvey: Also related, does the spec say to perform any validation on the namespace?

pymongo/asynchronous/client_bulk.py
```diff
@@ -0,0 +1,763 @@
...
+            self.is_retryable,
+            retryable_bulk,
+            session,
+            operation,
+            bulk=self,  # type: ignore[arg-type]
+            operation_id=op_id,
+        )
+
+        if full_result["error"] or full_result["writeErrors"] or full_result["writeConcernErrors"]:
+            _throw_client_bulk_write_exception(full_result, self.verbose_results)
```
ShaneHarvey, line 636: What error gets raise when a batch fails with a top-level command error or a connection error (eg AutoReconnect)? The coll.bulk_write api suffers from a problem where any batch that fails with such an error (after retries) will abandon the bulk write exception and raise AutoReconnect directly, thus loosing information about previous batches that could have succeeded. 

If the new bulk_write suffers from the same problem we should open a follow up ticket to improve the behavior.

ShaneHarvey (changes requested):
Great work!

pymongo/message.py
shruti-sridhar: The spec says to consider the size of the `bulkWrite` command document, excluding command-agnostic / operation-agnostic fields (those that aren't specific to the `bulkWrite` command) like `lsid` and `txnNumber`. The behavior is outlined [here](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#total-message-size)

pymongo/message.py
shruti-sridhar: Yep will do!

pymongo/operations.py
shruti-sridhar: The spec says to ensure that the namespace is provided for each write request, but apart from that it doesn't mention any additional validation to perform

pymongo/asynchronous/client_bulk.py
blink1073: I opened https://jira.mongodb.org/browse/PYTHON-4598, so I have something to point to when skipping the relevant tests in [PYTHON-4242](https://jira.mongodb.org/browse/PYTHON-4242).

shruti-sridhar: A `ClientBulkWriteException` gets raised for any top-level command/connection errors, and the `error` field of the exception will contain info about the error. So if it's a connection error like AutoReconnect, the `error` field will contain the error itself. If it's a top-level command error, the `error` field will contain the raw server response. Any results from previous batches get stored in the `partial_result` field of the `ClientBulkWriteException`.

pymongo/message.py
shruti-sridhar: I think the types are correct since `_client_batched_op_msg_impl()` accepts a `command: Mapping[str, Any]` but `_client_construct_op_msg()` accepts a `command_doc: bytes`. I'll change them both to accept a `command: Mapping[str, Any]` so that it's less confusing!

pymongo/message.py
ShaneHarvey: Since we have the entire command document I don't think we should do this. Instead we should check the size of `command_doc`.

pymongo/message.py
ShaneHarvey: Oh I see, I misread then. No need to change the existing approach then.

pymongo/client_bulk_shared.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(self, results: _DocumentOut, verbose: bool) -> None:
+        super().__init__("batch op errors occurred", 65, results)
+        self.verbose = verbose
+
+    def __reduce__(self) -> tuple[Any, Any]:
+        return self.__class__, (self.details,)
+
+    @property
+    def error(self) -> Optional[Any]:
```
ShaneHarvey, line 43: We should add docstrings for all the public attributes here.

pymongo/client_bulk_shared.py
```diff
@@ -0,0 +1,114 @@
...
+
+from pymongo.errors import OperationFailure, WriteConcernError, WriteError
+from pymongo.helpers_shared import _get_wce_doc
+from pymongo.results import ClientBulkWriteResult
+
+if TYPE_CHECKING:
+    from pymongo.typings import _DocumentOut
+
+
+class ClientBulkWriteException(OperationFailure):
```
ShaneHarvey, line 29: What do you think about moving this errors.py where all the other exceptions live?

pymongo/asynchronous/mongo_client.py
```diff
@@ -2204,10 +2224,134 @@ async def drop_database(
...
+            except AttributeError:
+                raise TypeError(f"{model!r} is not a valid request") from None
+
+        return await blk.execute(session, _Op.BULK_WRITE)
+
 
 def _retryable_error_doc(exc: PyMongoError) -> Optional[Mapping[str, Any]]:
     """Return the server response from PyMongo exception or None."""
-    if isinstance(exc, BulkWriteError):
+    if isinstance(exc, (BulkWriteError, ClientBulkWriteException)):
```
ShaneHarvey, line 2354: Should this check this writeConcernErrors field or the ClientBulkWriteException.error field? If both fields are set, how do we know which to use?

pymongo/client_bulk_shared.py
shruti-sridhar: When I put it in errors.py I get a circular import error with results.py -- I think because I import `ClientBulkWriteResult` from results.py for the `partial_result` field of the exception

ShaneHarvey: We could make it a delayed import:
```
    @property
    def partial_result(self) -> Optional[ClientBulkWriteResult]:
        from pymongo.results import ClientBulkWriteResult
        ...
```

pymongo/message.py
shruti-sridhar: Are there benefits to checking the size of the entire `command_doc` vs. checking the abridged size and accounting for the excluded fields in the 1000 byte overhead allowance? I was wondering because the spec mentions [here](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#how-was-the-op_msg-overhead-allowance-determined) that even if it's possible to determine the exact size of the OP_MSG, we should follow the given overhead calculations in order to standardize things across drivers

pymongo/client_bulk_shared.py
shruti-sridhar: Got it I'll move it to errors.py!

pymongo/message.py
ShaneHarvey: We should still add the 1000 byte overhead, I'm only suggesting we remove the "abridged" concept since it's extra unneeded complexity.

shruti-sridhar: Ahh got it -- I think the only place the abridged concept is necessary is for a few of the prose tests about batch-splitting. For example, [this test](https://github.com/mongodb/specifications/tree/master/source/crud/tests#case-2-batch-splitting-required) relies on the command document's size being calculated based only on the bulkWrite-specific keys.

pymongo/asynchronous/mongo_client.py
shruti-sridhar: For checking for retryable write concern errors, I think we'd only need to check the ClientBulkWriteException.writeConcernErrors field. Since ClientBulkWriteException.error will only ever contain a top-level error I don't we'll need to check it for writeConcernErrors.

pymongo/client_bulk_shared.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2024-present MongoDB, Inc.
```
ShaneHarvey, line 1: Can you rename this file to `_client_bulk_shared.py` to make it more clearly private?

test/unified_format.py
```diff
@@ -288,11 +297,123 @@ def parse_bulk_write_result(result):
     }
 
 
+def parse_client_bulk_write_models(models):
```
ShaneHarvey, line 300: We already parse collection.bulk_write arguments into models here:
https://github.com/mongodb/mongo-python-driver/blob/f0e025a127ef5f7bd3f4511eeacb0c8298aed737/test/utils.py#L1264-L1268

Can we reuse that code and remove parse_client_bulk_write_models?

pymongo/message.py
ShaneHarvey: Does that test fail if we change:
```python
    # Don't include bulkWrite-command-agnostic fields in document size calculations.
    abridged_keys = ["bulkWrite", "errorsOnly", "ordered"]
    if command.get("bypassDocumentValidation"):
        abridged_keys.append("bypassDocumentValidation")
    if command.get("comment"):
        abridged_keys.append("comment")
    if command.get("let"):
        abridged_keys.append("let")
    command_abridged = {key: command[key] for key in abridged_keys}
    command_len_abridged = len(_dict_to_bson(command_abridged, False, opts))
```
to:
```python
    command_abridged = command
    command_len_abridged = len(_dict_to_bson(command, False, opts))
```

shruti-sridhar: Yep that causes the test to fail, since the command doc size is now larger than what's expected [here](https://github.com/mongodb/specifications/tree/master/source/crud/tests#details-on-size-calculations)

pymongo/synchronous/client_bulk.py
```diff
@@ -0,0 +1,786 @@
...
+            op_id,
+            listeners,  # type: ignore[arg-type]
+            session,
+            self.client.codec_options,
+        )
+
+        while self.idx_offset < self.total_ops:
+            # If this is the last possible batch, use the
+            # final write concern.
+            if self.total_ops - self.idx_offset <= bwc.max_write_batch_size:
```
ShaneHarvey, line 512: This condition does not represent the final batch, for example if there are 10 InsertOne ops of size 16MB then this condition is true but multiple batches are still required. Could you file a bug ticket for this?

pymongo/results.py
```diff
@@ -133,16 +139,18 @@ def raw_result(self) -> Optional[Mapping[str, Any]]:
     @property
     def matched_count(self) -> int:
         """The number of documents matched for this update."""
-        self._raise_if_unacknowledged("matched_count")
-        if self.upserted_id is not None:
-            return 0
+        if self.acknowledged is not None:
+            self._raise_if_unacknowledged("matched_count")
```
ShaneHarvey, line 143: It would be cleaner to put the `if self.acknowledged is not None:` inside _raise_if_unacknowledged to avoid duplicating it on every call.

ShaneHarvey (commented):
A few final comments. One more question, are there tests for retryable writes with client-bulk where a command fails mid batch? Like first batch succeeds, second batch fails and succeeds on retry?

pymongo/operations.py
```diff
@@ -432,11 +580,30 @@ def __init__(
         .. versionchanged:: 3.5
            Added the `collation` option.
         """
-        super().__init__(filter, update, upsert, collation, array_filters, hint)
+        super().__init__(filter, update, upsert, collation, array_filters, hint, namespace)
 
     def _add_to_bulk(self, bulkobj: _AgnosticBulk) -> None:
         """Add this operation to the _AsyncBulk/_Bulk instance `bulkobj`."""
+        if self._upsert is None:
+            self._upsert = False
```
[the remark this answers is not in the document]
shruti-sridhar: would it be okay to instead do this in _add_to_bulk?
```
bulkobj.add_update(
            self._filter,
            self._doc,
            True,
            self._upsert is True,
            collation=validate_collation_or_none(self._collation),
            array_filters=self._array_filters,
            hint=self._hint,
        )
```

pymongo/operations.py
```diff
@@ -399,14 +542,16 @@ def __init__(
         self,
         filter: Mapping[str, Any],
         update: Union[Mapping[str, Any], _Pipeline],
-        upsert: bool = False,
+        upsert: Optional[bool] = None,
```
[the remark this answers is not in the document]
shruti-sridhar: I made the change since the client.bulk_write tests expect `upsert` to be optional and only passed into the command doc if it's set by the user. whereas those for coll.bulk_write expect `upsert` to always be passed into the command doc (by default as false)

shruti-sridhar:
> A few final comments. One more question, are there tests for retryable writes with client-bulk where a command fails mid batch? Like first batch succeeds, second batch fails and succeeds on retry?

There aren't retryable writes tests that cover retries across multiple batches -- the tests only cover succeeding/failing on retry for a single batch

pymongo/operations.py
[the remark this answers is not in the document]
ShaneHarvey: Yep! But `bool(self._upsert)` or `self._upsert or False` would be more "pythonic".

pymongo/operations.py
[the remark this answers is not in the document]
ShaneHarvey: Oh I see, this is related to DRIVERS-604 which we never fully implemented.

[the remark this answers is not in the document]
ShaneHarvey: Opened https://jira.mongodb.org/browse/PYTHON-4608.

ShaneHarvey:
Can you open a DRIVERS ticket to cover testing retries across multiple batches? That's an important case that needs to be tested.

pymongo/results.py
```diff
@@ -28,15 +28,15 @@ class _WriteResult:
 
     __slots__ = ("__acknowledged",)
 
-    def __init__(self, acknowledged: bool) -> None:
+    def __init__(self, acknowledged: Optional[bool] = None) -> None:
```
ShaneHarvey, line 31: Why is acknowledged optional now? What does that case mean?

shruti-sridhar: `acknowledged` is None for an individual write result within the ClientBulkWriteResult, since the spec mentions that acknowledgement info is not relevant for individual ops within the bulk write. Would it make more sense to pass `acknowledged=True` in these cases?

ShaneHarvey: If the ClientBulkWriteResult is acknowledged=False, then there are no results to inspect at all. If the ClientBulkWriteResult is acknowledged, the all the children results should also be acknowledged=True. Does that sound correct?

shruti-sridhar: Yep that sounds correct!

NoahStapp (approved):
Fantastic work, Shruti! @ShaneHarvey any remaining comments?

ShaneHarvey (approved):
LGTM!

PR #1748, opened 2024-07-25
PYTHON-4537 - Use selector asyncio loop on windows tests

NoahStapp:


pymongo/asynchronous/periodic_executor.py
```diff
@@ -27,6 +27,12 @@
 
 _IS_SYNC = False
 
+# The default asyncio loop implementation on Windows
+# has issues with sharing sockets across loops (https://github.com/python/cpython/issues/122240)
+# We explicitly use a different loop implementation here to prevent that issue
+if not _IS_SYNC and sys.platform == "win32":
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())  # type: ignore[attr-defined]
```
ShaneHarvey, line 34: We can't call asyncio.set_event_loop_policy since that has global side effects. Instead we need to explicitly create our on loop, possibly only when the default would be the ProactorEventLoop.

ShaneHarvey (changes requested):

pymongo/asynchronous/periodic_executor.py
```diff
@@ -65,7 +65,21 @@ def __repr__(self) -> str:
...
-        asyncio.run(self._run())  # type: ignore[func-returns-value]
+        # The default asyncio loop implementation on Windows
+        # has issues with sharing sockets across loops (https://github.com/python/cpython/issues/122240)
+        # We explicitly use a different loop implementation here to prevent that issue
+        if (
+            not _IS_SYNC
+            and sys.platform == "win32"
+            and asyncio.get_event_loop_policy() == asyncio.WindowsProactorEventLoopPolicy
+        ):
+            loop = asyncio.WindowsSelectorEventLoopPolicy().new_event_loop()  # type: ignore[attr-defined]
```
ShaneHarvey, line 76: How about we just always use WindowsSelectorEventLoopPolicy on windows for now. I'm  worried the WindowsProactorEventLoopPolicy check won't work for custom policies that use  the proactor loop.

ShaneHarvey (changes requested):

pymongo/synchronous/periodic_executor.py
```diff
@@ -65,7 +65,21 @@ def __repr__(self) -> str:
...
-        asyncio.run(self._run())  # type: ignore[func-returns-value]
+        # The default asyncio loop implementation on Windows
+        # has issues with sharing sockets across loops (https://github.com/python/cpython/issues/122240)
+        # We explicitly use a different loop implementation here to prevent that issue
+        if (
+            not _IS_SYNC
+            and sys.platform == "win32"
+            and asyncio.get_event_loop_policy() == asyncio.WindowsProactorEventLoopPolicy
+        ):
+            loop = asyncio.WindowsSelectorEventLoopPolicy().new_event_loop()  # type: ignore[attr-defined]
```
ShaneHarvey, line 76: Let's call asyncio.SelectorEventLoop() directly.

pymongo/asynchronous/periodic_executor.py
ShaneHarvey: Let's also call asyncio.SelectorEventLoop() directly.

pymongo/asynchronous/periodic_executor.py
```diff
@@ -65,7 +65,17 @@ def __repr__(self) -> str:
         return f"<{self.__class__.__name__}(name={self._name}) object at 0x{id(self):x}>"
 
     def _run_async(self) -> None:
-        asyncio.run(self._run())  # type: ignore[func-returns-value]
+        # The default asyncio loop implementation on Windows
+        # has issues with sharing sockets across loops (https://github.com/python/cpython/issues/122240)
+        # We explicitly use a different loop implementation here to prevent that issue
+        if not _IS_SYNC and sys.platform == "win32":
```
ShaneHarvey, line 71: We can remove `not _IS_SYNC` here.

ShaneHarvey (approved):

PR #1749, opened 2024-07-26
PYTHON-4549 - Optimize Cursor.to_list

NoahStapp:
This optimization is showing improvement of between 70%-120% compared to the naive `to_list` implementation.

### DIFF
--- a/gridfs/asynchronous/grid_file.py
+++ b/gridfs/asynchronous/grid_file.py
@@ -1821,6 +1821,9 @@ class _AsyncGridOutChunkIterator:
 
     __anext__ = next
 
+    async def to_list(self) -> list[Mapping[str, Any]]:
+        return [x async for x in self]  # noqa: C416,RUF100
+
     async def close(self) -> None:
         if self._cursor:
             await self._cursor.close()
@@ -1841,6 +1844,9 @@ class AsyncGridOutIterator:
         return bytes(chunk["data"])
 
     __anext__ = next
+
+    async def to_list(self) -> list[bytes]:
+        return [x async for x in self]  # noqa: C416,RUF100
 
 
 class AsyncGridOutCursor(AsyncCursor):
@@ -1891,6 +1897,9 @@ class AsyncGridOutCursor(AsyncCursor):
         _disallow_transactions(self.session)
         next_file = await super().next()
         return AsyncGridOut(self._root_collection, file_document=next_file, session=self.session)
+
+    async def to_list(self) -> list[AsyncGridOut]:
+        return [x async for x in self]  # noqa: C416,RUF100
 
     __anext__ = next
 
--- a/gridfs/asynchronous/grid_file.py
+++ b/gridfs/asynchronous/grid_file.py
@@ -1821,9 +1821,6 @@ class _AsyncGridOutChunkIterator:
 
     __anext__ = next
 
-    async def to_list(self) -> list[Mapping[str, Any]]:
-        return [x async for x in self]  # noqa: C416,RUF100
-
     async def close(self) -> None:
         if self._cursor:
             await self._cursor.close()
@@ -1844,9 +1841,6 @@ class AsyncGridOutIterator:
         return bytes(chunk["data"])
 
     __anext__ = next
-
-    async def to_list(self) -> list[bytes]:
-        return [x async for x in self]  # noqa: C416,RUF100
 
 
 class AsyncGridOutCursor(AsyncCursor):
--- a/pymongo/asynchronous/cursor.py
+++ b/pymongo/asynchronous/cursor.py
@@ -1260,19 +1260,19 @@ class AsyncCursor(Generic[_DocumentType]):
         else:
             raise StopAsyncIteration
 
-    async def _next_batch(self, result: list) -> list[_DocumentType]:
-        """Get all documents from the cursor."""
+    async def _next_batch(self, result: list) -> bool:
+        """Get all available documents from the cursor."""
         if not self._exhaust_checked:
             self._exhaust_checked = True
             await self._supports_exhaust()
         if self._empty:
-            raise StopAsyncIteration
+            return False
         if len(self._data) or await self._refresh():
             result.extend(self._data)
             self._data.clear()
-            return result
+            return True
         else:
-            raise StopAsyncIteration
+            return False
 
     async def __anext__(self) -> _DocumentType:
         return await self.next()
@@ -1288,11 +1288,9 @@ class AsyncCursor(Generic[_DocumentType]):
 
     async def to_list(self) -> list[_DocumentType]:
         res: list[_DocumentType] = []
-        try:
-            while self.alive:
-                await self._next_batch(res)
-        except StopAsyncIteration:
-            pass
+        while self.alive:
+            if not await self._next_batch(res):
+                break
         return res
 
 
--- a/pymongo/asynchronous/cursor.py
+++ b/pymongo/asynchronous/cursor.py
@@ -1260,6 +1260,19 @@ class AsyncCursor(Generic[_DocumentType]):
         else:
             raise StopAsyncIteration
 
+    async def _get_all(self, result) -> list[_DocumentType]:
+        """Get all documents from the cursor."""
+        if not self._exhaust_checked:
+            self._exhaust_checked = True
+            await self._supports_exhaust()
+        if self._empty:
+            raise StopAsyncIteration
+        while self.alive:
+            if len(self._data) or await self._refresh():
+                result.extend(self._data)
+                self._data.clear()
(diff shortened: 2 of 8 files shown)


pymongo/asynchronous/cursor.py
```diff
@@ -1275,6 +1288,9 @@ async def __aexit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
     async def to_list(self) -> list[_DocumentType]:
         return [x async for x in self]  # noqa: C416,RUF100
 
+    async def to_list_batch(self) -> list[_DocumentType]:
```
ShaneHarvey, line 1291: What is to_list_batch, was it only added for testing purposes?

pymongo/asynchronous/cursor.py
```diff
@@ -1260,6 +1260,19 @@ async def next(self) -> _DocumentType:
...
             raise StopAsyncIteration
 
+    async def _get_all(self, result: list) -> list[_DocumentType]:
+        """Get all documents from the cursor."""
+        if not self._exhaust_checked:
+            self._exhaust_checked = True
+            await self._supports_exhaust()
+        if self._empty:
+            raise StopAsyncIteration
+        while self.alive:
```
ShaneHarvey, line 1270: This loop seems problematic for tailable cursors. I would expect `cursor.to_list()` to has the same behavior as `[x async for x in cursor]`. The former would run forever on a tailable cursor while the later stops when the first empty batch is returned. See https://pymongo.readthedocs.io/en/latest/examples/tailable.html#tailable-cursors for an example.

ShaneHarvey (changes requested):

ShaneHarvey (changes requested):
We need to make this change in CommandCursor as well.

pymongo/asynchronous/cursor.py
NoahStapp: Forgot to rename it, fixed.

pymongo/asynchronous/cursor.py
NoahStapp: Makes sense, fixed.

NoahStapp:
> We need to make this change in CommandCursor as well.

Done. I've chosen to explicitly give the various `GridFS` cursors the old, slower implementation due to the complexity of their `next` implementations. We can address those in a separate ticket if desired.

gridfs/asynchronous/grid_file.py
```diff
@@ -1842,6 +1845,9 @@ async def next(self) -> bytes:
 
     __anext__ = next
 
+    async def to_list(self) -> list[bytes]:
```
ShaneHarvey, line 1848: AsyncGridOutIterator/GridOutIterator are never used so I vote we remove to_list. I'm not sure why GridOutIterator even exists, maybe for backwards compat? We don't even have tests for it.

pymongo/asynchronous/cursor.py
ShaneHarvey: Could you add a test case for this?

gridfs/asynchronous/grid_file.py
NoahStapp: Three tests fail if this one is removed:

```
FAILED test/asynchronous/test_session.py::TestSession::test_gridfs - AttributeError: 'dict' object has no attribute 'read'
FAILED test/asynchronous/test_session.py::TestSession::test_gridfs_bucket - AttributeError: 'dict' object has no attribute 'read'
FAILED test/asynchronous/test_session.py::TestSession::test_gridfsbucket_cursor - AttributeError: 'dict' object has no attribute 'read'
```

NoahStapp: Whoops sorry wrong one, this change is fine.

ShaneHarvey: Are you sure? AsyncGridOutIterator is never used anywhere so I don't see how that's possible.

NoahStapp: Yeah I read the wrong line numbers for this change, removed the new `to_list` methods that aren't for `AsyncGridOutCursor`.

test/asynchronous/test_cursor.py
```diff
@@ -1380,6 +1380,17 @@ async def test_getMore_does_not_send_readPreference(self):
         self.assertEqual("getMore", started[1].command_name)
         self.assertNotIn("$readPreference", started[1].command)
 
+    @async_client_context.require_sync
```
ShaneHarvey, line 1383: This test should run on both async and sync.

test/asynchronous/test_cursor.py
```diff
@@ -1380,6 +1380,17 @@ async def test_getMore_does_not_send_readPreference(self):
...
+    def test_to_list_when_empty(self):
+        self.db.drop_collection("test")
+
+        c = self.db.test.find().limit(0)
+        _ = c[0]
```
ShaneHarvey, line 1388: What is the purpose of these two lines?

test/asynchronous/test_cursor.py
```diff
@@ -1380,6 +1380,17 @@ async def test_getMore_does_not_send_readPreference(self):
         self.assertEqual("getMore", started[1].command_name)
         self.assertNotIn("$readPreference", started[1].command)
 
+    @async_client_context.require_sync
+    def test_to_list_when_empty(self):
```
ShaneHarvey, line 1384: This test should cover calling to_list on tailable cursors. See https://pymongo.readthedocs.io/en/latest/examples/tailable.html#tailable-cursors for an example.

I still believe the current version of to_list will incorrectly loop forever on a tailable cursor.

ShaneHarvey (changes requested):

test/asynchronous/test_cursor.py
NoahStapp: Your comment suggested testing the case where the `self._empty == True`, which only happens when `__getitem__` is called in a case like this.

test/asynchronous/test_cursor.py
NoahStapp: `__getitem__` is not supported for async cursors and is the only way to trigger `self._empty == True`.

test/asynchronous/test_cursor.py
NoahStapp: Ah, got it. The following test passes:

```
    async def test_to_list_tailable(self):
        self.db.drop_collection("test")

        await self.db.test.insert_many([{"x": i} for i in range(100)])

        c = self.db.test.find(cursor_type=pymongo.CursorType.TAILABLE_AWAIT)

        docs = await c.to_list()

        self.assertEqual(100, len(docs))  # type: ignore[arg-type]
```

test/asynchronous/test_cursor.py
ShaneHarvey: I was referring to tailable cursors but sure the empty case should be tested too. In that case this test should be calling `c.to_list()` instead of calling to_list on a new cursor.

test/asynchronous/test_cursor.py
ShaneHarvey: Sorry you have to create a tailable, await data cursor that yields an infinite number of documents, like tailing the oplog or running $chanegStream.

NoahStapp: Do we support running `$changeStream` separately from using our internal `change_stream` cursor? I see that cursor also has its own behavior for `try_next` that would need to be implemented in its own `_next_batch` method.

NoahStapp: This test using `oplog` also passes: 

```
    async def test_to_list_tailable(self):
        client = await async_rs_or_single_client()
        self.addAsyncCleanup(client.aclose)

        c = client.local.oplog.rs.find(
                   cursor_type=pymongo.CursorType.TAILABLE_AWAIT,
                   oplog_replay=True)

        docs = await c.to_list()

        self.assertGreater(len(docs), 0)  # type: ignore[arg-type]
```

test/asynchronous/test_cursor.py
```diff
@@ -1380,6 +1380,27 @@ async def test_getMore_does_not_send_readPreference(self):
         self.assertEqual("getMore", started[1].command_name)
         self.assertNotIn("$readPreference", started[1].command)
 
+    @async_client_context.require_replica_set
+    async def test_to_list_tailable(self):
```
ShaneHarvey, line 1384: We should test these cases with CommandCursor as well.

test/asynchronous/test_cursor.py
ShaneHarvey: It's possible to run $changeStream directly via aggregate().

ShaneHarvey (approved):

PR #1751, opened 2024-07-27
PYTHON-4245 Convert CRUD v1 spec tests to unified test format

blink1073:


NoahStapp:
How should I approach reviewing this? Do we need to check every file and line?

blink1073:
> How should I approach reviewing this? Do we need to check every file and line?

No, ignore the .json file changes, they're direct copies.  The only substantial change is in `test/test_crud_v1.py`, which removes specific code to handle the legacy tests.

NoahStapp (approved):

PR #1755, opened 2024-07-30
PYTHON-4589 - Add async API docs

NoahStapp:
Many of the inline docstrings and examples are incorrect: there are async class links in the synchronous API and sync class links + examples in the asynchronous API. These will all be fixed in [PYTHON-4592](https://jira.mongodb.org/browse/PYTHON-4592).


blink1073 (approved):
LGTM!

PR #1756, opened 2024-07-30
PYTHON-4592 - Synchronize inline examples and docstrings

NoahStapp:

PR #1757, opened 2024-07-30
PYTHON-4594 - Add to_list documentation

NoahStapp:


pymongo/asynchronous/command_cursor.py
```diff
@@ -382,6 +382,12 @@ async def __aexit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
         await self.close()
 
     async def to_list(self) -> list[_DocumentType]:
+        """Converts the contents of this cursor to a list.
+
+        If the cursor is empty or has no more results, an empty list will be returned.
```
ShaneHarvey, line 387: Suggest adding an example and saying this is a more efficient verison of `[doc async for doc in cursor]`.

NoahStapp: Done. The synchronization to the sync versions will be handled by https://github.com/mongodb/mongo-python-driver/pull/1756.

pymongo/synchronous/command_cursor.py
```diff
@@ -382,6 +382,16 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
         self.close()
 
     def to_list(self) -> list[_DocumentType]:
+        """Converts the contents of this cursor to a list more efficiently than ``[doc async for doc in cursor]``.
```
Jibola, line 385:
```suggestion
        """Converts the contents of this cursor to a list more efficiently than ``[doc for doc in cursor]``.
```

I don't think we need to supply that language here. That or we just say above.

pymongo/synchronous/cursor.py
```diff
@@ -1285,6 +1285,16 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
         self.close()
 
     def to_list(self) -> list[_DocumentType]:
+        """Converts the contents of this cursor to a list more efficiently than ``[doc async for doc in cursor]``.
```
Jibola, line 1288:
```suggestion
        """Converts the contents of this cursor to a list more efficiently than ``[doc for doc in cursor]``.
```

Jibola (changes requested):

ShaneHarvey (approved):

pymongo/synchronous/command_cursor.py
NoahStapp: Sorry, which language?

pymongo/synchronous/cursor.py
NoahStapp: https://github.com/mongodb/mongo-python-driver/pull/1756 will fix the inline examples not being synchronized.

pymongo/synchronous/command_cursor.py
Jibola: The `async` part. I see you addressed it earlier.

Jibola (approved):

PR #1758, opened 2024-07-31
PYTHON-4600 Handle round trip time being negative because time.monotonic() is not monotonic

morotti:
Hello,

This fixes this long standing issue that's been opened since 2017
https://jira.mongodb.org/browse/PYTHON-1271
https://jira.mongodb.org/browse/PYTHON-3616
https://jira.mongodb.org/browse/PYTHON-4298

We've been getting regular failures from the mongo client in CI, since updating to a more recent python and mongo client.
image

Symptom is the client crashing with round_trip_time of None.
This is a very long standing issue as we've found bug tickets going back to 2017 regarding this problem.

Oddly enough, we did not see that frequently on older python version and older client, although the bug is present. I suspect there's a subtle side effect from python 3.11+ being noticeably faster, which makes it a lot more common to get sub millisecond measurements.

root cause:
After many days of debugging, it turns out the issue is pretty simple: time.monotonic() is in fact not monotonic!
Basic code to measure the round trip time `start = time.monotonic() ... end = time.monotonic()` doesn't work as expected, it can give a negative duration once in a while.

The build nodes run NTP. I think NTP is adjusting the time and causing the issue. As per linux documentation CLOCK_MONOTONIC is subject to time adjustments by NTP and adjtime(). https://man7.org/linux/man-pages/man2/clock_gettime.2.html

Related. The code to average round trip times had a condition to silently ignore negative durations `sample < 0`, which contributed to making the issue extremely difficult to debug. I've corrected the check to ensure that negative durations cannot be provided.

There may be other places that are using time.monotonic() incorrectly. I've corrected the ones that crashed my production systems and the ones I could find by searching the code base. There are more usage to measure timeout or deadline which I think are not affected it the clock goes backward by a millisecond.

Cheers.

### DIFF
--- a/pymongo/asynchronous/monitor.py
+++ b/pymongo/asynchronous/monitor.py
@@ -48,6 +48,15 @@ def _sanitize(error: Exception) -> None:
     error.__cause__ = None
 
 
+def _monotonic_duration(start: float) -> float:
+    """Return the duration since the given start time.
+
+    Accounts for buggy platforms where time.monotonic() is not monotonic.
+    See PYTHON-4600.
+    """
+    return max(0.0, time.monotonic() - start)
+
+
 class MonitorBase:
     def __init__(self, topology: Topology, name: str, interval: int, min_interval: float):
         """Base class to do periodic work on a background thread.
@@ -247,9 +256,7 @@ class Monitor(MonitorBase):
             _sanitize(error)
             sd = self._server_description
             address = sd.address
-            end = time.monotonic()
-            # time.monotonic() is not monotonic
-            duration = max(0.0, end - start)
+            duration = _monotonic_duration(start)
             if self._publish:
                 awaited = bool(self._stream and sd.is_server_type_known and sd.topology_version)
                 assert self._listeners is not None
@@ -319,9 +326,7 @@ class Monitor(MonitorBase):
         else:
             # New connection handshake or polling hello (MongoDB <4.4).
             response = await conn._hello(cluster_time, None, None)
-        end = time.monotonic()
-        # time.monotonic() is not monotonic
-        duration = max(0.0, end - start)
+        duration = _monotonic_duration(start)
         return response, duration
 
 
@@ -446,9 +451,7 @@ class _RttMonitor(MonitorBase):
                 raise Exception("_RttMonitor closed")
             start = time.monotonic()
             await conn.hello()
-            end = time.monotonic()
-            # time.monotonic() is not monotonic
-            duration = max(0.0, end - start)
+            duration = _monotonic_duration(start)
             return duration
 
 
--- a/pymongo/asynchronous/monitor.py
+++ b/pymongo/asynchronous/monitor.py
@@ -247,7 +247,9 @@ class Monitor(MonitorBase):
             _sanitize(error)
             sd = self._server_description
             address = sd.address
-            duration = time.monotonic() - start
+            end = time.monotonic()
+            # time.monotonic() is not monotonic
+            duration = max(0.0, end - start)
             if self._publish:
                 awaited = bool(self._stream and sd.is_server_type_known and sd.topology_version)
                 assert self._listeners is not None
@@ -317,7 +319,10 @@ class Monitor(MonitorBase):
         else:
             # New connection handshake or polling hello (MongoDB <4.4).
             response = await conn._hello(cluster_time, None, None)
-        return response, time.monotonic() - start
+        end = time.monotonic()
+        # time.monotonic() is not monotonic
+        duration = max(0.0, end - start)
+        return response, duration
 
 
 class SrvMonitor(MonitorBase):
@@ -441,7 +446,10 @@ class _RttMonitor(MonitorBase):
                 raise Exception("_RttMonitor closed")
             start = time.monotonic()
             await conn.hello()
-            return time.monotonic() - start
+            end = time.monotonic()
+            # time.monotonic() is not monotonic
+            duration = max(0.0, end - start)
+            return duration
 
 
 # Close monitors to cancel any in progress streaming checks before joining
--- a/pymongo/asynchronous/monitor.py
+++ b/pymongo/asynchronous/monitor.py
@@ -451,8 +451,7 @@ class _RttMonitor(MonitorBase):
                 raise Exception("_RttMonitor closed")
             start = time.monotonic()
             await conn.hello()
-            duration = _monotonic_duration(start)
-            return duration
+            return _monotonic_duration(start)
 
 
 # Close monitors to cancel any in progress streaming checks before joining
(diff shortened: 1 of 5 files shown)


ShaneHarvey:
> After many days of debugging, it turns out the issue is pretty simple: time.monotonic() is in fact not monotonic!
Basic code to measure the round trip time start = time.monotonic() ... end = time.monotonic() doesn't work as expected, it can give a negative duration once in a while.

If true, this is a great find, although I'm a bit suspicious. The api is very clear that time.monotonic never goes backwards:

> Return the value (in fractional seconds) of a monotonic clock, i.e. a clock that cannot go backwards. The clock is not affected by system clock updates. 
https://docs.python.org/3/library/time.html#time.monotonic

Linux is also clear in this regard:

> CLOCK_MONOTONIC
>   A nonsettable system-wide clock that represents monotonic
>   time since—as described by POSIX—"some unspecified point
>   in the past".  On Linux, that point corresponds to the
>   number of seconds that the system has been running since
>   it was booted.
>   
>   The CLOCK_MONOTONIC clock is not affected by discontinuous
>   jumps in the system time (e.g., if the system
>   administrator manually changes the clock), but is affected
>   by frequency adjustments.  This clock does not count time
>   that the system is suspended. **All CLOCK_MONOTONIC
>   variants guarantee that the time returned by consecutive
>   calls will not go backwards, but successive calls may—
>   depending on the architecture—return identical (not-
>   increased) time values.**


What you're describing is a bug in either the CPython and/or Linux version you are using since time.monotonic is never supposed to go backwards. What system are you running on? Please include the Python version and Linux version.

morotti:
> What you're describing is a bug in either the CPython and/or Linux version you are using since time.monotonic is never supposed to go backwards. What system are you running on? Please include the Python version and Linux version.

It's on CentOS 7 with Python 3.11. 

I appreciate it can be surprising 😄 
As surprising as it may be, monotonic is really going backward. I can reproduce consistency by triggering dozens of builds on my CI servers. I've added print statements to show the values of time.monotonic() before mongo crashes, it is going backward.

You can go through the old tickets were customers reported the issue, it all makes sense once you comprehend that the time difference can be negative.
The mongo code to measure rtt is really just two calls to time.monotonic() a bit apart (until v4.7). The resulting difference is negative. It goes to `if sample < 0: return None` in the code to average time and a None is propagated. Explosion.

I think the doc of the interpreter was overoptimistic for saying time.monotonic() is monotonic. The cpython implementation is using CLOCK_MONOTONIC which doesn't guarantee as much as the name implies. 
There are many threads on the internet saying it's not as monotonic as it says. It is subject to some time adjustments and NTP adjustments.

There seem to be a different clock CLOCK_MONOTONIC_RAW that may more monotonic.
However raw was only introduced in Linux 2.6.28, which was too new about 10 years ago when all these functions were made.
And raw is some microseconds slower to invoke, until kernel v4.x that introduced an optimized CPU call.

There may be more at play. I have my suspicions that my hardware or kernel or CPU are drifting more than they ought to be. 
CentOS 7 is aging, it's possible old versions of NTP or the kernel were doing larger adjustments or one-off adjustments that new versions are not doing. It's possible they have some bugs. Who knows.
Doing some research, NTP only does small adjustments, but small adjustments can mean a whole millisecond. It's more than the connection time to measure, on an internal network. 

I'm gonna do more research to see if I can dig up something and maybe raise the bug to the interpreter.

ShaneHarvey:
> I think the doc of the interpreter was overoptimistic for saying time.monotonic() is monotonic. The cpython implementation is using CLOCK_MONOTONIC which doesn't guarantee as much as the name implies.
There are many threads on the internet saying it's not as monotonic as it says. It is subject to some time adjustments and NTP adjustments.

CLOCK_MONOTONIC and time.monotonic have to be monotonic even in the face of NTP and clock adjustments by design. It's right in the name :). Any different means the clock implementation is broken.


That said, clearly these broken systems still exist so I'm fine with merging a fix for this in pymongo.

> It's on CentOS 7 with Python 3.11.

Can you elaborate the exact versions `python -c "import sys; print(sys.version)"` and `hostnamectl` to get the linux kernel version?

pymongo/asynchronous/monitor.py
```diff
@@ -247,7 +247,9 @@ async def _check_server(self) -> ServerDescription:
             _sanitize(error)
             sd = self._server_description
             address = sd.address
-            duration = time.monotonic() - start
+            end = time.monotonic()
+            # time.monotonic() is not monotonic
+            duration = max(0.0, end - start)
```
ShaneHarvey, line 252: I like it but would you mind refactoring the 3 calls in this file like this:
```python
def _monotonic_duration(start: float) -> float:
    """Return the duration since the given start time.
    
    Accounts for buggy platforms where time.monotonic() is actually not monotonic. See PYTHON-4600.
    """
    return max(0.0, time.monotonic() - start))
```

ShaneHarvey (changes requested):
Nice work! Just one small change suggestion to centralize the workaround.

pymongo/asynchronous/monitor.py
morotti: done

morotti:
I've resend the patch, adding common code to a function.

> Can you elaborate the exact versions python -c "import sys; print(sys.version)" and hostnamectl to get the linux kernel version?

I am seeing some information I am not sure I can share from these commands. I can give you this instead.

uname -a
Linux 185b71dc801a 3.10.0-1160.114.2.el7.x86_64 #1 SMP Wed Mar 20 15:54:52 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

python version 3.11.9

failing tests hitting the issue consistency are running in a centos 7 container, the host VM is running on centos 7 too.
the VM itself is running on a physical machine, I don't know what.

I've found a few related bugs on https://bugzilla.redhat.com/ with a range of root cause. 
There are both known issues due to NTP and known issues with hardware clocks going bonkers on particular machines.
I am running some diagnostic trying to measure time warp but I don't think I will get much further.

I've seeing cases where the clock goes back years between two calls. A bit flip with the CPU / hardware clock or an overflow with floats would make a lot of sense, as insane as that may be. ¯\_(ツ)_/¯

pymongo/asynchronous/monitor.py
```diff
@@ -317,7 +326,8 @@ async def _check_with_socket(self, conn: AsyncConnection) -> tuple[Hello, float]
         else:
             # New connection handshake or polling hello (MongoDB <4.4).
             response = await conn._hello(cluster_time, None, None)
-        return response, time.monotonic() - start
+        duration = _monotonic_duration(start)
+        return response, duration
```
ShaneHarvey, line 330: Our linter says these should be written as:
```python
return response, _monotonic_duration(start)
```

```
pymongo/asynchronous/monitor.py:455:20: RET504 Unnecessary assignment to `duration` before `return` statement
pymongo/synchronous/monitor.py:455:20: RET504 Unnecessary assignment to `duration` before `return` statement
```

morotti:
should be passing the ruff checks now

ShaneHarvey (approved):
Thanks for solving this long standing cold case @morotti!

morotti:
Thank you so much for replying so quickly and accepting the fix :) @ShaneHarvey

PR #1762, opened 2024-07-31
PYTHON-4021 - Publish TopologyDescriptionChangedEvent on topology close

NoahStapp:
The rename of `TopologyOpenedEvent` to `TopologyOpeningEvent` is for consistency with the [spec](https://github.com/mongodb/specifications/blob/89edd8129d751a982d058632b5d9203dd9826356/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.md?plain=1#L129) and the unified tests.

### DIFF
--- a/pymongo/asynchronous/topology.py
+++ b/pymongo/asynchronous/topology.py
@@ -77,7 +77,7 @@ _IS_SYNC = False
 _pymongo_dir = str(Path(__file__).parent)
 
 
-def process_events_queue(queue_ref: weakref.ReferenceType[queue.Queue]) -> bool:
+def process_events_queue(queue_ref: weakref.ReferenceType[queue.Queue], closed: bool) -> bool:
     q = queue_ref()
     if not q:
         return False  # Cancel PeriodicExecutor.
@@ -86,6 +86,9 @@ def process_events_queue(queue_ref: weakref.ReferenceType[queue.Queue]) -> bool:
         try:
             event = q.get_nowait()
         except queue.Empty:
+            # Publish all remaining events before closing the event publishing thread
+            if closed:
+                return False
             break
         else:
             fn, args = event
@@ -157,7 +160,7 @@ class Topology:
             weak: weakref.ReferenceType[queue.Queue]
 
             async def target() -> bool:
-                return process_events_queue(weak)
+                return process_events_queue(weak, self._closed)
 
             executor = periodic_executor.PeriodicExecutor(
                 interval=common.EVENTS_QUEUE_FREQUENCY,
@@ -685,15 +688,6 @@ class Topology:
             )
             self._events.put((self._listeners.publish_topology_closed, (self._topology_id,)))
         if self._publish_server or self._publish_tp:
-            # Publish all remaining events before closing the event publishing thread
-            while True:
-                try:
-                    event = self._events.get_nowait()  # type: ignore[attr-defined]
-                except queue.Empty:
-                    break
-                else:
-                    fn, args = event
-                    fn(*args)
             self.__events_executor.close()
 
     @property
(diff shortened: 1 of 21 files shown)


pymongo/monitoring.py
```diff
@@ -1281,7 +1281,7 @@ def __repr__(self) -> str:
         )
 
 
-class TopologyOpenedEvent(TopologyEvent):
+class TopologyOpeningEvent(TopologyEvent):
```
ShaneHarvey, line 1284: This is a breaking change so we have to live with the "TopologyOpenedEvent" name.

ShaneHarvey (changes requested):

pymongo/asynchronous/topology.py
```diff
@@ -664,8 +664,36 @@ async def close(self) -> None:
...
         if self._publish_server or self._publish_tp:
+            # Publish all remaining events before closing the event publishing thread
+            while True:
+                try:
+                    event = self._events.get_nowait()  # type: ignore[attr-defined]
+                except queue.Empty:
+                    break
+                else:
+                    fn, args = event
+                    fn(*args)
```
ShaneHarvey, line 696: These events should only be published by the `__events_executor` otherwise the user could see the events out of order. How about we either:
- wait until the events are published, or
- make `__events_executor` keep running until it's queue is empty.

NoahStapp: I like the second option here.

pymongo/monitoring.py
ShaneHarvey: We already do that here: https://github.com/mongodb/mongo-python-driver/blob/272538513a589d8bf795ba0b7591f37414e96919/test/unified_format.py#L917-L919C23

pymongo/asynchronous/topology.py
```diff
@@ -77,7 +77,7 @@
 _pymongo_dir = str(Path(__file__).parent)
 
 
-def process_events_queue(queue_ref: weakref.ReferenceType[queue.Queue]) -> bool:
+def process_events_queue(queue_ref: weakref.ReferenceType[queue.Queue], closed: bool) -> bool:
```
ShaneHarvey, line 80: If the executor is sleeping while it's closed, I don't think process_events_queue will every be called.

NoahStapp: Do we have an existing way to close an executor after it runs its `target` method one more time? Otherwise we'll have to have `Topology.close` just wait until its event queue is empty to close the executor and finish its execution.

pymongo/asynchronous/topology.py
```diff
@@ -664,9 +664,31 @@ async def close(self) -> None:
...
+                        self._description,
+                        self._topology_id,
+                    ),
+                )
+            )
             self._events.put((self._listeners.publish_topology_closed, (self._topology_id,)))
         if self._publish_server or self._publish_tp:
+            # Make sure the events executor thread is fully closed before publishing the remaining events
             self.__events_executor.close()
+            self.__events_executor.join()
```
ShaneHarvey, line 690: Need a join timeout here  as a fail safe, I'd suggest 1 second.

ShaneHarvey (changes requested):

pymongo/asynchronous/topology.py
ShaneHarvey: We discussed in person. We don't have that functionality.

NoahStapp:
> LGTM, can you mention the new closed event in the changelog?

New closed event? I don't think any events were added.

ShaneHarvey (approved):

ShaneHarvey:
Ah that's right, I misunderstood the newly added event. LGTM.

PR #1765, opened 2024-08-01
PYTHON-4379 Test that durations are included on relevant pool events

blink1073:


test/connection_monitoring/pool-checkin-make-available.json
```diff
@@ -22,7 +22,8 @@
     {
       "type": "ConnectionCheckedOut",
       "connectionId": 1,
-      "address": 42
+      "address": 42,
+      "duration": 42
```
ShaneHarvey, line 26: Do we need to update the runner to assert on the new duration field or does that code already exist?

blink1073: Yes, we handle it with `self.assertIsNotNone(actual_val)` (found by debugging).

ShaneHarvey (approved):

PR #1772, opened 2024-08-06
PYTHON-4618 - Fix TypeError: Socket cannot be of type SSLSocket

NoahStapp:


pymongo/network_layer.py
```diff
@@ -35,15 +36,31 @@
...
-    socket.sendall(buf)
+async def _async_sendall_ssl(sock: Union[socket.socket, _sslConn], buf: bytes) -> None:
+    view = memoryview(buf)
+    total_length = len(buf)
+    total_sent = 0
+    while total_sent < total_length:
+        try:
+            sent = sock.send(view[total_sent:])
+        except BLOCKING_IO_ERRORS:
+            await asyncio.sleep(0.5)
```
ShaneHarvey, line 60: Is there any way to poll here rather than sleep?

pymongo/network_layer.py
```diff
@@ -35,15 +39,65 @@
...
+
+    while sent < len(buf):
+        try:
+            sent += sock.send(buf)
+        except BLOCKING_IO_ERRORS as exc:
+            fd = sock.fileno()
+            # Check for closed socket.
+            if fd == -1:
+                raise SSLError("Underlying socket has been closed") from None
+            if isinstance(exc, SSLWantReadError):
```
ShaneHarvey, line 79: Do these checks work for the pyopenssl exceptions?

pymongo/network_layer.py
```diff
@@ -35,15 +39,65 @@
...
-def sendall(socket: Union[socket.socket, _sslConn], buf: bytes) -> None:
-    socket.sendall(buf)
+async def _async_sendall_ssl(
+    sock: Union[socket.socket, _sslConn], buf: bytes, loop: AbstractEventLoop
+) -> None:
+    fd = sock.fileno()
+    sent = 0
+
+    def _is_ready(fut: Future) -> None:
+        if fut.done():
```
ShaneHarvey, line 65: Should we remove reader/writer before checking done()?

pymongo/network_layer.py
```diff
@@ -35,15 +39,65 @@
...
+        except BLOCKING_IO_ERRORS as exc:
+            fd = sock.fileno()
+            # Check for closed socket.
+            if fd == -1:
+                raise SSLError("Underlying socket has been closed") from None
+            if isinstance(exc, SSLWantReadError):
+                fut = loop.create_future()
+                loop.add_reader(fd, _is_ready, fut)
+                await fut
+            if isinstance(exc, SSLWantWriteError):
```
ShaneHarvey, line 83: How should we handle pyopenssl's WantX509LookupError?

pymongo/network_layer.py
```diff
@@ -35,15 +39,65 @@
...
+                fut = loop.create_future()
+                loop.add_writer(fd, _is_ready, fut)
+                await fut
+
+
+async def _async_sendall_ssl_windows(sock: Union[socket.socket, _sslConn], buf: bytes) -> None:
```
ShaneHarvey, line 89: Could you add a comment that this is needed because the proactor event loop doesn't support add_writer/add_reader?

ShaneHarvey (changes requested):

pymongo/network_layer.py
NoahStapp: Good catch.

pymongo/network_layer.py
NoahStapp: According to the docs, it should be called again with the same arguments: https://www.pyopenssl.org/en/23.0.0/api/ssl.html#OpenSSL.SSL.WantX509LookupError. Maybe we register both a reader and writer?

pymongo/network_layer.py
NoahStapp: Fixed.

caseyclements (approved):
lgtm

pymongo/network_layer.py
```diff
@@ -18,32 +18,105 @@
 import asyncio
 import socket
 import struct
+import sys
+from asyncio import AbstractEventLoop, Future
+from ssl import SSLError, SSLSocket
```
ShaneHarvey, line 23: This has to be wrapped with try/except to account for when the ssl module is not available.

pymongo/ssl_support.py
```diff
@@ -53,6 +53,9 @@
     IPADDR_SAFE = True
     SSLError = _ssl.SSLError
     BLOCKING_IO_ERRORS = _ssl.BLOCKING_IO_ERRORS
+    BLOCKING_IO_READ_ERROR = _ssl.BLOCKING_IO_READ_ERROR
+    BLOCKING_IO_WRITE_ERROR = _ssl.BLOCKING_IO_READ_ERROR
```
ShaneHarvey, line 57: BLOCKING_IO_READ_ERROR -> BLOCKING_IO_WRITE_ERROR

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #1773, opened 2024-08-06
PYTHON-4610 More robust to_list tests

ShaneHarvey:
PYTHON-4610

### DIFF
--- a/test/asynchronous/test_cursor.py
+++ b/test/asynchronous/test_cursor.py
@@ -1380,41 +1380,36 @@ class TestCursor(AsyncIntegrationTest):
         self.assertEqual("getMore", started[1].command_name)
         self.assertNotIn("$readPreference", started[1].command)
 
+    @async_client_context.require_version_min(4, 0)
     @async_client_context.require_replica_set
     async def test_to_list_tailable(self):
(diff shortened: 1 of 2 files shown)


test/asynchronous/test_cursor.py
```diff
@@ -1380,41 +1380,39 @@ async def test_getMore_does_not_send_readPreference(self):
...
         self.assertNotIn("$readPreference", started[1].command)
 
+    @async_client_context.require_version_min(4, 0)
     @async_client_context.require_replica_set
     async def test_to_list_tailable(self):
         oplog = self.client.local.oplog.rs
         last = await oplog.find().sort("$natural", pymongo.DESCENDING).limit(-1).next()
         ts = last["ts"]
-
+        # Set maxAwaitTimeMS=1 to speed up the test and avoid blocking on the noop writer.
```
ShaneHarvey, line 1389: maxAwaitTimeMS defaults to 1000, which gives the noop writer a chance to write a document to the oplog. What happens is the to_list call repeatedly call getMore which blocks for 1 second and returns 1 noop document each time. That's why the test was sometimes taking 20+ seconds.

NoahStapp (approved):

PR #1564, opened 2024-03-27
PYTHON-4305 Fix bson size check

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4305

PR #1778, opened 2024-08-08
backport the security patch of CVE-2024-21506

Crispy-fried-chicken:
Here is a vulnerability which is mentioned in https://github.com/mongodb/mongo-python-driver/pull/1564 and fixed in the [v4.6](https://github.com/mongodb/mongo-python-driver/compare/v4.6)branch https://github.com/mongodb/mongo-python-driver/commit/56b6b6dbc267d365d97c037082369dabf37405d2, but is not fixed in the branch of v4.4, maybe it should be backported?

### DIFF
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -2289,6 +2289,7 @@ static PyObject* get_value(PyObject* self, PyObject* name, const char* buffer,
             uint32_t c_w_s_size;
             uint32_t code_size;
             uint32_t scope_size;
+            uint32_t len;
             PyObject* code;
             PyObject* scope;
             PyObject* code_type;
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Hi @Crispy-fried-chicken, we don't support the v4.4 version anymore. Is there any reason you cannot upgrade to 4.6.3+, 4.7+ or 4.8.0? We follow semantic version so your app should be able to upgrade without issue.

blink1073:
> could potentially cause various supply chain security challenges

We only have one direct dependency, `dnspython`, and it did not change between the two versions.  We don't have the infrastructure set up to make a release against this branch.

blink1073:
Unfortunately we are unable accept this change.

PR #1782, opened 2024-08-09
PYTHON-4644 Use a random name for hatchenv

blink1073:


.evergreen/hatch.sh
```diff
@@ -13,8 +13,10 @@ if $PYTHON_BINARY -m hatch --version; then
       $PYTHON_BINARY -m hatch run "$@"
     }
 else # No toolchain hatch present, set up virtualenv before installing hatch
-    createvirtualenv "$PYTHON_BINARY" hatchenv
-    trap "deactivate; rm -rf hatchenv" EXIT HUP
+    ENV_NAME=hatchenv-$RANDOM
+    createvirtualenv "$PYTHON_BINARY" $ENV_NAME
+    # shellcheck disable=SC2064
+    trap "deactivate; rm -rf $ENV_NAME" EXIT HUP
```
ShaneHarvey, line 19: Why do we need a random name? IIRC we never had to use random venv names before.

ShaneHarvey: I see, could you add a comment like: "# Use a random venv name because the encryption tasks run this script multiple times in the same run."

blink1073: Done

ShaneHarvey (approved):

Jibola (approved):

PR #1784, opened 2024-08-09
PYTHON-4649 Skip CSOT tests on slow Windows and macOS hosts

ShaneHarvey:
PYTHON-4649

### DIFF
--- a/test/test_csot.py
+++ b/test/test_csot.py
@@ -77,6 +77,7 @@ class TestCSOT(IntegrationTest):
     @client_context.require_change_streams
     def test_change_stream_can_resume_after_timeouts(self):
         coll = self.db.test
+        coll.insert_one({})
         with coll.watch() as stream:
             with pymongo.timeout(0.1):
                 with self.assertRaises(PyMongoError) as ctx:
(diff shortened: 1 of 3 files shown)


test/test_csot.py
```diff
@@ -77,6 +77,7 @@ def test_timeout_nested(self):
     @client_context.require_change_streams
     def test_change_stream_can_resume_after_timeouts(self):
         coll = self.db.test
+        coll.insert_one({})
```
ShaneHarvey, line 80: This was needed because when I ran this test alone on a fresh cluster it failed with " cannot start $changeStream on non-existent database pymongo_test".

ShaneHarvey:
I confirmed that the CSOT spec tests are skipped on windows/macos and are still run on other platforms.

test/test_csot.py
Jibola: NIT: Should we use `create_collection`, or does it not matter?

ShaneHarvey: We could but insert_one is easier since it would fail if the collection already exists.

PR #1792, opened 2024-08-13
PYTHON-4660 Fix AttributeError when MongoClient.bulk_write batch fails with InvalidBSON

shruti-sridhar:
[PYTHON-4660](https://jira.mongodb.org/browse/PYTHON-4660)

### DIFF
--- a/test/asynchronous/test_client_bulk_write.py
+++ b/test/asynchronous/test_client_bulk_write.py
@@ -19,7 +19,13 @@ import sys
 
 sys.path[0:0] = [""]
 
-from test.asynchronous import AsyncIntegrationTest, async_client_context, unittest
+from test.asynchronous import (
+    AsyncIntegrationTest,
+    AsyncMockClientTest,
+    async_client_context,
+    unittest,
+)
+from test.asynchronous.pymongo_mocks import AsyncMockClient
 from test.utils import (
     OvertCommandListener,
     async_rs_or_single_client,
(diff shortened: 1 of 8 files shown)


ShaneHarvey (approved):

ShaneHarvey (changes requested):
Forgot one thing, is it possible to add a test case for this?

shruti-sridhar:
> Forgot one thing, is it possible to add a test case for this? IIUC we need to make a batch fail with a non-PyMongoError

Do you know if there's a way to trigger a non-PyMongoError within a failpoint? I've tested that it works by manually causing a TypeError somewhere in the code for instance, but I'm not sure how to automatically trigger a non-PyMongoError.

ShaneHarvey:
No it's not possible with a failpoint. I believe it would have to be mocked.

test/asynchronous/test_client_bulk_write.py
```diff
@@ -577,3 +583,22 @@ async def test_timeout_in_multi_batch_bulk_write(self):
             if event.command_name == "bulkWrite":
                 bulk_write_events.append(event)
         self.assertEqual(len(bulk_write_events), 2)
+
+
+class TestClientBulkWriteMock(AsyncMockClientTest):
+    @async_client_context.require_version_min(8, 0, 0, -24)
+    async def test_handles_non_pymongo_error(self):
+        mock_client = await AsyncMockClient.get_async_mock_client(
```
ShaneHarvey, line 591: Instead of using a mock client, what do you think about using a real client with unittest.mock.patch on _AsyncClientBulk.write_command (or even Connection.write_command)?

ShaneHarvey (changes requested):

test/asynchronous/test_client_bulk_write.py
shruti-sridhar: Sure I can change that!

ShaneHarvey (approved):

ShaneHarvey:
Nice!

PR #1796, opened 2024-08-15
PYTHON-4578 Benchmark collection and client bulk write

shruti-sridhar:
[PYTHON-4578](https://jira.mongodb.org/browse/PYTHON-4578)

### DIFF
--- a/test/performance/perf_test.py
+++ b/test/performance/perf_test.py
@@ -63,13 +63,9 @@ from bson import decode, encode, json_util
 from gridfs import GridFSBucket
 from pymongo import (
     DeleteMany,
-    DeleteOne,
     InsertOne,
     MongoClient,
     ReplaceOne,
-    UpdateMany,
-    UpdateOne,
-    WriteConcern,
 )
 
 pytestmark = pytest.mark.perf
--- a/test/performance/perf_test.py
+++ b/test/performance/perf_test.py
@@ -474,8 +474,8 @@ class TestSmallDocBulkMixedOps(SmallDocMixedTest, unittest.TestCase):
         self.models = []
         for doc in self.documents:
             self.models.append(InsertOne(document=doc.copy()))
-            self.models.append(ReplaceOne(filter={}, replacement=doc.copy(), upsert=True))
-            self.models.append(DeleteMany(filter={}))
+            self.models.append(ReplaceOne(filter={}, replacement=doc.copy(), upsert=True))  # type: ignore[arg-type]
+            self.models.append(DeleteMany(filter={}))  # type: ignore[arg-type]
 
     def do_task(self):
         self.corpus.bulk_write(self.models, ordered=True)
@@ -491,8 +491,8 @@ class TestSmallDocClientBulkMixedOps(SmallDocMixedTest, unittest.TestCase):
                 ReplaceOne(
                     namespace="perftest.corpus", filter={}, replacement=doc.copy(), upsert=True
                 )
-            )
-            self.models.append(DeleteMany(namespace="perftest.corpus", filter={}))
+            )  # type: ignore[arg-type]
+            self.models.append(DeleteMany(namespace="perftest.corpus", filter={}))  # type: ignore[arg-type]
 
     def do_task(self):
         self.client.bulk_write(self.models, ordered=True)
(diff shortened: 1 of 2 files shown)


blink1073 (dismissed):
LGTM!

test/performance/perf_test.py
```diff
@@ -429,11 +457,141 @@ def do_task(self):
...
+        self.models = []
+        for doc in self.documents:
+            self.models.append(InsertOne(namespace="perftest.corpus", document=doc.copy()))
+
+    @client_context.require_version_min(8, 0, 0, -24)
+    def do_task(self):
+        self.client.bulk_write(self.models, ordered=True)
+
+
+class TestLargeDocBulkMixedOps(LargeDocMixedTest, unittest.TestCase):
```
ShaneHarvey, line 543: For the MixedOps and Namespace tests, we don't need to test both small and large documents. Let's only test small. This will lower the runtime but more importantly reduce the number of benchmarks results we need to look through.

test/performance/perf_test.py
```diff
@@ -429,11 +457,141 @@ def do_task(self):
...
+    @client_context.require_version_min(8, 0, 0, -24)
+    def setUp(self):
+        super().setUp()
+        self.models = []
+        for doc in self.documents:
+            self.models.append(InsertOne(namespace="perftest.corpus", document=doc.copy()))
+            self.models.append(
+                ReplaceOne(
+                    namespace="perftest.corpus", filter={}, replacement=doc.copy(), upsert=True
+                )  # type: ignore[arg-type]
```
ShaneHarvey, line 496: What's the type ignore for?

test/performance/perf_test.py
```diff
@@ -429,11 +457,141 @@ def do_task(self):
...
+
+
+class TestSmallDocBulkMixedOps(SmallDocMixedTest, unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.models = []
+        for doc in self.documents:
+            self.models.append(InsertOne(document=doc.copy()))
+            self.models.append(ReplaceOne(filter={}, replacement=doc.copy(), upsert=True))  # type: ignore[arg-type]
+            self.models.append(DeleteMany(filter={}))  # type: ignore[arg-type]
```
ShaneHarvey, line 480: For the mixed tests can we use DeleteOne instead of DeleteMany?

test/performance/perf_test.py
```diff
@@ -429,11 +457,141 @@ def do_task(self):
...
+                )  # type: ignore[arg-type]
+            )
+            self.models.append(DeleteMany(namespace="perftest.corpus", filter={}))  # type: ignore[arg-type]
+
+    @client_context.require_version_min(8, 0, 0, -24)
+    def do_task(self):
+        self.client.bulk_write(self.models, ordered=True)
+
+
+class TestSmallDocMultiNamespaceBulkInsert(SmallDocMixedTest, unittest.TestCase):
```
ShaneHarvey, line 505: Is it useful to benchmark bulk_write with 2 namespaces?

ShaneHarvey (changes requested):

test/performance/perf_test.py
shruti-sridhar: Appending the ReplaceOne and DeleteMany to self.models was causing a type error (since it was expecting self.models to be a list of InsertOne's after the first model). I can add a type hint to self.models and remove the type ignores.

test/performance/perf_test.py
shruti-sridhar: For the MultiNamespace cases I was curious to compare the two bulk_writes, since coll.bulk_write would have to make 2 server commands while client.bulk_write needs to make only 1. But yeah I'm not sure if it's that useful to benchmark!

ShaneHarvey: Let's remove the MultiNamespace ones. The main purpose of this benchmark should be to test the CPU perf of pymongo's ability to encode the bulk write commands. Mixing the namespace doesn't really change much there.

ShaneHarvey (approved):

ShaneHarvey:
Here are the results on the new benchmarks from [this run](https://spruce.mongodb.com/task/mongo_python_driver_perf_tests_perf_8.0_standalone_patch_adf8817df8f3fd47d82cf4c5e3bd4e93cfdfb602_66be6f408984260007ccd052_24_08_15_21_12_35/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC):
Name | MB/sec
-- | --
LargeDocClientBulkInsert | 21.086811
LargeDocBulkInsert | 104.5071745
SmallDocClientBulkMixedOps | 1.353895781
SmallDocBulkMixedOps | 0.6469863368
SmallDocClientBulkInsert | 8.437455728
SmallDocBulkInsert | 41.5301894659

So it looks like we have a lot of work to do to make client.bulk_write's perf on par with coll.bulk_write.

PR #1798, opened 2024-08-16
PYTHON-4666 Fix handling of large documents in client.bulk_write

shruti-sridhar:
[PYTHON-4666](https://jira.mongodb.org/browse/PYTHON-4666)

### DIFF
--- a/pymongo/message.py
+++ b/pymongo/message.py
@@ -1119,6 +1119,10 @@ def _client_batched_op_msg_impl(
         new_message_size = total_ops_length + total_ns_length + op_length + ns_length
         # We have enough data, return this batch.
         if new_message_size > max_doc_sequences_bytes:
+            if idx == 0:
+                raise InvalidOperation(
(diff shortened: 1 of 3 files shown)


pymongo/message.py
```diff
@@ -1119,6 +1119,10 @@ def _check_doc_size_limits(
         new_message_size = total_ops_length + total_ns_length + op_length + ns_length
         # We have enough data, return this batch.
         if new_message_size > max_doc_sequences_bytes:
+            if idx == 0:
+                raise InvalidOperation(
+                    "No operations can be executed -- given operation is too large"
+                )
```
ShaneHarvey, line 1125: We should consider raising the same error as collection.bulk_write in this case.

ShaneHarvey (changes requested):

ShaneHarvey (commented):
Nice could you comment on the PR with what the new error looks like?

ShaneHarvey (approved):

PR #1800, opened 2024-08-19
PYTHON-4668 Improve performance of client.bulk_write

shruti-sridhar:
[PYTHON-4668](https://jira.mongodb.org/browse/PYTHON-4668)

### DIFF
--- a/doc/examples/client_bulk.rst
+++ b/doc/examples/client_bulk.rst
@@ -145,7 +145,7 @@ For example, a duplicate key error on the third operation below aborts the remai
     'idx': 2,
     'code': 11000,
     'errmsg': 'E11000 duplicate key error ... dup key: { _id: 3 }', ...
-    'op': {'insert': 'db.test_three', 'document': {'_id': 3}}}]
+    'op': {'insert': 0, 'document': {'_id': 3}}}]
   >>> exception.partial_result.inserted_count
   2
   >>> exception.partial_result.deleted_count
@@ -181,7 +181,7 @@ For example, the fourth and fifth write operations below get executed successful
     'idx': 2,
     'code': 11000,
     'errmsg': 'E11000 duplicate key error ... dup key: { _id: 5 }', ...
-    'op': {'insert': 'db.test_five', 'document': {'_id': 5}}}]
+    'op': {'insert': 0, 'document': {'_id': 5}}}]
   >>> exception.partial_result.inserted_count
   3
   >>> exception.partial_result.deleted_count
--- a/test/asynchronous/test_client_bulk_write.py
+++ b/test/asynchronous/test_client_bulk_write.py
@@ -73,6 +73,24 @@ class TestClientBulkWrite(AsyncIntegrationTest):
             self.assertIsInstance(context.exception.error, TypeError)
             self.assertFalse(hasattr(context.exception.error, "details"))
 
+    @async_client_context.require_version_min(8, 0, 0, -24)
+    async def test_formats_write_error_correctly(self):
+        client = await async_rs_or_single_client()
+        self.addAsyncCleanup(client.close)
+
+        models = [
+            InsertOne(namespace="db.coll", document={"_id": 1}),
+            InsertOne(namespace="db.coll", document={"_id": 1}),
+        ]
+        self.addAsyncCleanup(client.db["coll"].drop)
+
+        with self.assertRaises(ClientBulkWriteException) as context:
+            await client.bulk_write(models=models)
+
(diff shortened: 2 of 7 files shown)


doc/examples/client_bulk.rst
```diff
@@ -181,7 +181,7 @@ For example, the fourth and fifth write operations below get executed successful
     'idx': 2,
     'code': 11000,
     'errmsg': 'E11000 duplicate key error ... dup key: { _id: 5 }', ...
-    'op': {'insert': 'db.test_five', 'document': {'_id': 5}}}]
+    'op': {'insert': 0, 'document': {'_id': 5}}}]
```
ShaneHarvey, line 184: Interesting, do we have any tests that asserts we send the correct bulkWrite format here?

ShaneHarvey (changes requested):
Could you also remove the doc.copy()'s from the TestSmallDocClientBulkInsert/TestSmallDocBulkMixedOps/TestSmallDocClientBulkMixedOps/TestLargeDocClientBulkInsert setup?

doc/examples/client_bulk.rst
shruti-sridhar: I can add a test for that!

test/asynchronous/test_client_bulk_write.py
```diff
@@ -73,6 +73,24 @@ async def test_handles_non_pymongo_error(self):
             self.assertIsInstance(context.exception.error, TypeError)
             self.assertFalse(hasattr(context.exception.error, "details"))
 
+    @async_client_context.require_version_min(8, 0, 0, -24)
+    async def test_formats_write_error_correctly(self):
+        client = await async_rs_or_single_client()
```
ShaneHarvey, line 78: Nice test! Just a few nits.

Let's use the self.client here instead of creating a new one for each test. The tests run more efficiently that way because creating a new client is expensive. We only create a new client if we need to change a client level setting like registering an event listener. 

Could you make this change in test_returns_error_if_no_namespace_provided and test_handles_non_pymongo_error too?

test/asynchronous/test_client_bulk_write.py
```diff
@@ -73,6 +73,24 @@ async def test_handles_non_pymongo_error(self):
...
+        self.addAsyncCleanup(client.close)
+
+        models = [
+            InsertOne(namespace="db.coll", document={"_id": 1}),
+            InsertOne(namespace="db.coll", document={"_id": 1}),
+        ]
+        self.addAsyncCleanup(client.db["coll"].drop)
```
ShaneHarvey, line 85: We can remove this drop call. Drop is expensive so we try to avoid it.

test/performance/perf_test.py
```diff
@@ -461,8 +461,8 @@ def setUp(self):
         super().setUp()
         self.models: list[Union[InsertOne, ReplaceOne, DeleteOne]] = []
         for doc in self.documents:
-            self.models.append(InsertOne(document=doc.copy()))
-            self.models.append(ReplaceOne(filter={}, replacement=doc.copy(), upsert=True))
+            self.models.append(InsertOne(document=doc))
+            self.models.append(ReplaceOne(filter={}, replacement=doc, upsert=True))
```
ShaneHarvey, line 465: The replacement=doc.copy() one might actually need to stay due to _id issues.

test/performance/perf_test.py
```diff
@@ -475,11 +475,9 @@ def setUp(self):
         super().setUp()
         self.models: list[Union[InsertOne, ReplaceOne, DeleteOne]] = []
         for doc in self.documents:
-            self.models.append(InsertOne(namespace="perftest.corpus", document=doc.copy()))
+            self.models.append(InsertOne(namespace="perftest.corpus", document=doc))
             self.models.append(
-                ReplaceOne(
-                    namespace="perftest.corpus", filter={}, replacement=doc.copy(), upsert=True
-                )
+                ReplaceOne(namespace="perftest.corpus", filter={}, replacement=doc, upsert=True)
```
ShaneHarvey, line 480: The replacement=doc.copy() one might actually need to stay due to _id issues.

ShaneHarvey (changes requested):

test/asynchronous/test_client_bulk_write.py
shruti-sridhar: Got it I'll change that!

ShaneHarvey (approved):
LGTM!

caseyclements (approved):
LGTM. Thx for the chat!

PR #1802, opened 2024-08-20
PYTHON-4672 Clarify Reauthentication and Speculative Authentication combination behavior

blink1073:


test/auth_oidc/test_auth_oidc.py
```diff
@@ -1019,6 +1015,54 @@ def fetch(self, _):
...
+            client.test.test.insert_one({})
+
+        # Assert that the callback was called once.
+        self.assertEqual(self.request_called, 1)
+
+        # Assert there were `SaslStart` commands executed.
+        found = False
+        for event in listener.started_events:
+            if event.command_name.lower() != "saslstart":
+                found = True
```
Jibola, line 1060: should this be `event.command_name.lower() == "saslstart"`?

test/auth_oidc/test_auth_oidc.py
```diff
@@ -1019,6 +1015,54 @@ def fetch(self, _):
...
+        assert found
```
Jibola, line 1061: NIT: (note I've still left it != for now)
```suggestion
        assert any([event.command_name.lower()  != "saslstart" for event in listener.started_events])
```

Jibola (changes requested):

test/auth_oidc/test_auth_oidc.py
blink1073: Yup, thanks!

Jibola (approved):
LGTM

PR #1811, opened 2024-08-23
PYTHON-4691 Fix non-UTC timezones with DATETIME_CLAMP/DATETIME_AUTO

ShaneHarvey:
PYTHON-4691 Fix non-UTC timezones with DATETIME_CLAMP/DATETIME_AUTO

### DIFF
--- a/bson/datetime_ms.py
+++ b/bson/datetime_ms.py
@@ -114,17 +114,29 @@ class DatetimeMS:
         return self._value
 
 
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
Note we still have to remove the functools.lru_cache to fix the bug in PYTHON-4663 but it turned out to make more sense to me to fix this bug first.

blink1073 (approved):
LGTM!

PR #1813, opened 2024-08-24
[DRIVERS-2926] [PYTHON-4577] BSON Binary Vector Subtype Support

caseyclements:
Adds support for new subtype of Binary BSON subtype to be used for densely-packed 1d arrays (vectors).
Corresponding changes are in specifications repo PR [#1658](https://github.com/mongodb/specifications/pull/1658)


bson/vector.py
```diff
@@ -0,0 +1,131 @@
...
+    One byte defines the data type, described by the bson.binary.vector.DTYPE Enum.
+    Another byte declares the number of bits at the end that should be ignored,
+    in the case that a vector's length and type do not require a whole number of bytes.
+    This number is referred to as padding.
+    """
+
+    dtype: str
+    padding: int = 0
+
+    def __new__(cls, data: Any, dtype: Union[DTYPES, bytes], padding: int = 0) -> BinaryVector:
```
blink1073, line 44: Can padding be inferred from the data?

bson/vector.py
```diff
@@ -0,0 +1,131 @@
...
+        if isinstance(dtype, bytes):
+            dtype = DTYPE_FROM_HEX[dtype]
+        assert dtype in DTYPES
+        self.dtype = dtype
+        self.padding = padding
+        return self
+
+    @classmethod
+    def from_list(
+        cls: Type[BinaryVector], num_list: List, dtype: DTYPES, padding: int = 0
```
blink1073, line 55: We should specify the explicit List types, e.g. `Union[List[int], List[float]]`.

bson/__init__.py
```diff
@@ -388,6 +392,12 @@ def _get_binary(
     # Decode subtype 0 to 'bytes'.
     if subtype == 0:
         value = data[position:end]
+    # Decode subtype 9 to vector.
+    elif subtype == 9:
+        dtype, padding = struct.unpack_from("<sB", data, position)
+        position += 2
+        end = position + length
+        value = BinaryVector(data[position:end], dtype=dtype, padding=padding)
```
ShaneHarvey, line 400: We shouldn't have any custom vector encoding or decoding logic here. vector subtypes  should be decoded to regular bson.Binary instances the same as any other subtype.

bson/vector.py
```diff
@@ -0,0 +1,131 @@
...
+    INT8 = b"\x03"
+    FLOAT32 = b"\x27"
+    BOOL = b"\x10"
+
+
+# Map from bytes to enum value, for decoding.
+DTYPE_FROM_HEX = {key.value: key for key in DTYPES}
+
+
+class BinaryVector(Binary):
```
ShaneHarvey, line 26: Let's not add a new class, instead we should add from_list/as_list to Binary.

ShaneHarvey (changes requested):

bson/__init__.py
ShaneHarvey: I disagree. The Binary class needs to contain the actual bson data otherwise we're breaking the Liskov Substitution Principle. We could always extend the api with a new Vector by adding a new decoding setting like codec_options.vector_conversion`.

blink1073: Ah, good call out.  Yeah, I like the idea of `codec_options.vector_conversion`.

ShaneHarvey: I'm not against adding a vector_conversion but I don't think we should do it in this project. Right now we don't have any real use cases so we should stick with the simple Binary.from_list/as_list apis. Once we have more users we can tackle extending the design via vector_conversion. What do you think?

blink1073: Yeah, in the interest of getting this into the 4.9 release I think de-scoping it make sense.

bson/vector.py
caseyclements: Only for data types whose bit size is a multiple of 8.

bson/binary.py
```diff
@@ -337,6 +384,92 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
...
+        """Create a BSON Binary Vector subtype from a list of python objects.
+
+        The data type and byte padding are prepended to the vector itself.
+
+        :param vector: List of values
+        :param dtype: Data type of the values
+        :param padding: For fractional bytes, number of bits to ignore at end of vector.
+        :return: Binary packed data identified by dtype and padding.
+        """
+        if dtype == BinaryVectorDtype.INT8:  # pack ints in [-128, 127] as signed int8
```
blink1073, line 403: I suggest moving these comments to docstrings in the enum, like we do for `QueryType`.

bson/binary.py
```diff
@@ -191,21 +194,64 @@ class UuidRepresentation:
...
+
+# Map from bytes to enum value, for decoding.
+DTYPE_FROM_HEX = {key.value: key for key in BinaryVectorDtype}
+
+
+@dataclass
+class BinaryVector:
+    """Vector of numbers along with metadata for binary interoperability.
+
+    dtype specifies the data type stored in binary.
```
blink1073, line 230: I suggest moving these descriptions to docstrings under each property.

caseyclements: Like the following?

```
@dataclass
class BinaryVector:
    """Vector of numbers along with metadata for binary interoperability.

    :param: data: Sequence of numbers representing the mathematical vector.
    :param dtype:  specifies the data type stored in binary.
    :param: padding: specifies the number of bits in the final byte that are to be ignored
      when a vector element's size is less than a byte and the length of the vector is not a multiple of 8.
      """

    data: (Sequence[float | int]
    dtype: BinaryVectorDtype
    padding: Optional[int] = 0
```

caseyclements: i.e. Try """ under each var

caseyclements:
Hi @Shane. Do you still have changes that haven't been addressed?

jyemin (changes requested):

test/bson_binary_vector/vector-test-cases.json
```diff
@@ -0,0 +1,87 @@
...
+  "description": "Basic Tests of Binary Vectors, subtype 9",
+  "test_key": "vector",
+  "tests": [
+    {
+      "description": "Simple Vector INT8",
+      "valid": true,
+      "vector": [127, 7],
+      "dtype_hex": "0x03",
+      "dtype_alias": "INT8",
+      "padding": 0,
```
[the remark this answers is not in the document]
caseyclements: It's not required. There are now tests that include non-zero padding (both invlaid and valid ones).

test/bson_binary_vector/vector-test-cases.json
```diff
@@ -0,0 +1,87 @@
+{
+  "description": "Basic Tests of Binary Vectors, subtype 9",
+  "test_key": "vector",
+  "tests": [
+    {
+      "description": "Simple Vector INT8",
+      "valid": true,
```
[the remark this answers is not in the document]
caseyclements: This one's out of date.

test/bson_binary_vector/vector-test-cases.json
```diff
@@ -0,0 +1,87 @@
+{
+  "description": "Basic Tests of Binary Vectors, subtype 9",
+  "test_key": "vector",
+  "tests": [
```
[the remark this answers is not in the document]
caseyclements: The format now follows one of the standards I saw.

test/bson_binary_vector/vector-test-cases.json
```diff
@@ -0,0 +1,87 @@
...
+    {
+      "description": "Simple Vector INT8",
+      "valid": true,
+      "vector": [127, 7],
+      "dtype_hex": "0x03",
```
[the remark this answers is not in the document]
caseyclements: We need to distinguish, for example, between int8, packed_bit, or even float32.

test/bson_binary_vector/vector-test-cases.json
```diff
@@ -0,0 +1,87 @@
...
+      "valid": true,
+      "vector": ["-inf", 0.0, "inf"],
+      "dtype_hex": "0x27",
+      "dtype_alias": "FLOAT32",
+      "padding": 0,
+      "canonical_bson": "2000000005766563746F72000E000000092700000080FF000000000000807F00",
+      "canonical_extjson": "{\"vector\": {\"$binary\": {\"base64\": \"JwAAAID/AAAAAAAAgH8=\", \"subType\": \"09\"}}}"
+    }
+  ],
+  "invalid": [
```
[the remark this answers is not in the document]
caseyclements: Those tests are still in bson corpus tests in binary. These focus on roundtrip from numbers and dtypes.

jyemin:
Removing myself from review of this PR.  I'm just focusing on the specifications repository PR, but that should not hold up merging this one.

bson/binary.py
```diff
@@ -191,21 +194,76 @@ class UuidRepresentation:
...
 
-    Raises TypeError if `data` is not an instance of :class:`bytes`
-    or `subtype` is not an instance of :class:`int`.
+    Subtype 9 provides a space-efficient representation of 1-dimensional vector data.
+    Its data is prepended with two bytes of metadata.
+    The first (dtype) describes its data type, such as float32 or int8.
+    The second (padding) prescribes the number of bits to ignore in the final byte.
+    This is relevant when the element size of the dtype is not a multiple of 8.
+
+    Raises TypeError if subtype` is not an instance of :class:`int`.
```
blink1073, line 266:
```suggestion
    Raises TypeError if `subtype` is not an instance of :class:`int`.
```

bson/binary.py
```diff
@@ -337,6 +398,90 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
...
+
+    def as_vector(self, uncompressed: Optional[bool] = False) -> BinaryVector:
+        """From the Binary, create a list of numbers, along with dtype and padding.
+
+
+        :param uncompressed: If true, return the true mathematical vector.
+            This is only necessary for datatypes where padding is applicable.
+            For example, setting this to True for a PACKED_BIT vector will result
+            in a List[int] of zeros and ones.
+        :return: List of numbers, along with dtype and padding.
```
blink1073, line 445: This is misleading, it should say that it returns a BinaryVector object.

doc/changelog.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Changes in Version 4.10.0
+-------------------------
+
+- Provisional **(BETA)** support for a new Binary BSON subtype used for efficient storage and retrieval of vectors: densely packed arrays of numbers, all of the same type.
```
blink1073, line 7: Suggest adding: "New methods :meth:`bson.binary.Binary.from_vector` and
  :meth:`bson.binary.Binary.as_vector`"

blink1073: We should also call out which subtype it is, e.g. "BSON binary subtype 4" elsewhere in the changelog.

bson/binary.py
```diff
@@ -337,6 +398,90 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
...
+        if dtype == BinaryVectorDtype.INT8:
+            dtype_format = "b"
+            format_string = f"{n_values}{dtype_format}"
+            vector = list(struct.unpack_from(format_string, self, position))
+            return BinaryVector(vector, dtype, padding)
+
+        elif dtype == BinaryVectorDtype.FLOAT32:
+            n_bytes = len(self) - position
+            n_values = n_bytes // 4
+            assert n_bytes % 4 == 0
```
ShaneHarvey, line 465: We should not use `assert` for validation data. If the argument type is incorrect we raise a TypeError, if the type is correct but the value is invalid we raise a ValueError.

bson/binary.py
```diff
@@ -337,6 +398,90 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
...
+        :param vector: List of values
+        :param dtype: Data type of the values
+        :param padding: For fractional bytes, number of bits to ignore at end of vector.
+        :return: Binary packed data identified by dtype and padding.
+
+        .. versionadded:: 4.9
+        """
+        if dtype == BinaryVectorDtype.INT8:  # pack ints in [-128, 127] as signed int8
+            format_str = "b"
+            assert not padding, f"padding does not apply to {dtype=}"
```
ShaneHarvey, line 424: Same comment about assert.

bson/binary.py
```diff
@@ -337,6 +398,90 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
...
+        elif dtype == BinaryVectorDtype.PACKED_BIT:  # pack ints in [0, 255] as unsigned uint8
+            format_str = "B"
+        elif dtype == BinaryVectorDtype.FLOAT32:  # pack floats as float32
+            format_str = "f"
+            assert not padding, f"padding does not apply to {dtype=}"
```
ShaneHarvey, line 429: Same comment about assert.

bson/binary.py
```diff
@@ -337,6 +398,90 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
...
+
+        :param uncompressed: If true, return the true mathematical vector.
+            This is only necessary for datatypes where padding is applicable.
+            For example, setting this to True for a PACKED_BIT vector will result
+            in a List[int] of zeros and ones.
+        :return: List of numbers, along with dtype and padding.
+
+        .. versionadded:: 4.9
+        """
+
```
ShaneHarvey, line 449: We need to validate `self.subtype == 9` here before attempting to decode.

ShaneHarvey (changes requested):

bson/binary.py
```diff
@@ -191,21 +194,76 @@ class UuidRepresentation:
...
+    The `PACKED_BIT` value represents a special case where vector values themselves
+    can only be of two values (0 or 1) but these are packed together into groups of 8,
+    a byte. In Python, these are displayed as ints in range [0, 255]
+
+    Each value is of type bytes with a length of one.
+
+    .. versionadded:: 4.9
+    """
+
+    INT8 = b"\x03"
```
ShaneHarvey, line 225: An enum of bytes is a bit unusual. Can we change it to use ints? eg `INT8 = 3`

bson/binary.py
```diff
@@ -191,21 +194,76 @@ class UuidRepresentation:
...
+    :param data: Sequence of numbers representing the mathematical vector.
+    :param dtype:  The data type stored in binary
+    :param padding: The number of bits in the final byte that are to be ignored
+      when a vector element's size is less than a byte
+      and the length of the vector is not a multiple of 8.
+
+    .. versionadded:: 4.9
+    """
+
+    data: Sequence[float | int]
```
ShaneHarvey, line 247: Add `__slots__ = ("data", "dtype", "padding")` to reduce the memory overhead of using this class.

caseyclements: Sure. How do I do type annotation for that?

bson/binary.py
caseyclements: The naming convention that Geert used has meaning in both the first and last 4 bits, so I'd prefer it to stay as-is.

bson/binary.py
caseyclements: Changed, though I put the description here as it was specifically to complement the typing.

bson/binary.py
blink1073: None needed, e.g. https://github.com/mongodb/mongo-python-driver/blob/821811e80d72d2ae822d11bac30e1c7a935208c2/bson/timestamp.py#L31

bson/binary.py
caseyclements: @ShaneHarvey How do I do that?

bson/binary.py
blink1073: We could pass `slots=True` instead: https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass

caseyclements: Strange. And you're cool with that?

blink1073: It is part of the reason to use a `dataclass`, to have it provide these idioms for us.

bson/binary.py
blink1073: Something like:

```python
if self.subtype != VECTOR_SUBTYPE:
   raise ValueError(...)
```

bson/binary.py
blink1073: Whoops, slots keyword was added in Python 3.10.

caseyclements: Hmmm...  __slots__ with a default value in a dataclass?

bson/binary.py
ShaneHarvey: I don't see how the interpretation of the values here matters. Are you saying this is a bit flag?

bson/binary.py
```diff
@@ -191,21 +194,76 @@ class UuidRepresentation:
...
+    :param padding: The number of bits in the final byte that are to be ignored
+      when a vector element's size is less than a byte
+      and the length of the vector is not a multiple of 8.
+
+    .. versionadded:: 4.10
+    """
+
+    data: Sequence[float | int]
+    dtype: BinaryVectorDtype
+    padding: Optional[int] = 0
```
ShaneHarvey, line 249: `Optional[int]` -> `int`, unless padding=None is considered valid.

bson/binary.py
```diff
@@ -337,6 +398,95 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
             f"cannot decode subtype {self.subtype} to {UUID_REPRESENTATION_NAMES[uuid_representation]}"
         )
 
+    @classmethod
+    def from_vector(
+        cls: Type[Binary],
+        vector: list[int, float],
+        dtype: BinaryVectorDtype,
+        padding: Optional[int] = 0,
```
ShaneHarvey, line 406: `Optional[int]` -> `int`

bson/binary.py
```diff
@@ -337,6 +398,95 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
...
+            if padding:
+                raise ValueError(f"padding does not apply to {dtype=}")
+        else:
+            raise NotImplementedError("%s not yet supported" % dtype)
+
+        metadata = struct.pack("<sB", dtype.value, padding)
+        data = struct.pack(f"{len(vector)}{format_str}", *vector)
+        return cls(metadata + data, subtype=VECTOR_SUBTYPE)
+
+    def as_vector(self, uncompressed: Optional[bool] = False) -> BinaryVector:
```
ShaneHarvey, line 439: `Optional[bool]` -> `bool` unless `uncompressed=None` is valid.

bson/binary.py
blink1073: We have to add them manually instead of using `slots=True`.

bson/binary.py
```diff
@@ -337,6 +398,95 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
...
+        elif dtype == BinaryVectorDtype.PACKED_BIT:
+            # data packed as uint8
+            dtype_format = "B"
+            unpacked_uint8s = list(struct.unpack_from(f"{n_values}{dtype_format}", self, position))
+            if not uncompressed:
+                return BinaryVector(unpacked_uint8s, dtype, padding)
+            else:
+                bits = []
+                for uint8 in unpacked_uint8s:
+                    bits.extend([int(bit) for bit in f"{uint8:08b}"])
```
ShaneHarvey, line 483: Why add an `uncompressed` option here at all? Is it useful? It looks like this option is irreversible because from_vector does not support the same option.

ShaneHarvey (changes requested):

bson/binary.py
caseyclements:
```
class BinaryVector:
    """**(BETA)** Vector of numbers along with metadata for binary interoperability.

    :param data (Sequence[float | int]): Sequence of numbers representing the mathematical vector.
    :param dtype (:class:`bson.Binary.BinaryVectorDtype`):  The data type stored in binary
    :param padding (Optional[int] = 0): The number of bits in the final byte that are to be ignored
      when a vector element's size is less than a byte
      and the length of the vector is not a multiple of 8. Default is 0.

    .. versionadded:: 4.10
    """

    __slots__ = ("data", "dtype", "padding")

    def __init__(self, data, dtype, padding=0):
        self.data = data
        self.dtype = dtype
        self.padding = padding
```

Is this right? @blink1073

bson/binary.py
blink1073: Yeah, I think we should leave out `uncompressed`, it was a quality of life addition, but is not likely to be standardized.

bson/binary.py
blink1073: padding=None is expected when it is not a packed type

bson/binary.py
blink1073: The user doesn't see the value locally either.  Given that the bson spec itself always uses integers I think we should use the integer form.

bson/binary.py
blink1073: yep

bson/binary.py
caseyclements: Yes. It's irreversible. We don't yet provide an API to go from a full vector of zeros and ones to a packed bit vector.

caseyclements: Going in, you'd be using something like `numpy.packbits`

ShaneHarvey: Let's omit it since it adds unneeded complexity at this point.

bson/binary.py
ShaneHarvey: Oh I didn't realize `__slots__` with dataclass was problematic.

bson/binary.py
ShaneHarvey: Look at `as_uuid` it should be the same as that validation:
```python
        if self.subtype not in ALL_UUID_SUBTYPES:
            raise ValueError(f"cannot decode subtype {self.subtype} as a uuid")
```

bson/binary.py
blink1073: It should be sorted now

bson/binary.py
caseyclements: Done

ShaneHarvey:
Fixing the test suite and the Enum issue are my last remaining comments.

blink1073:
We're pressing ahead with the hex values in the Enum because they simplify the implementation of encode and decode.

blink1073:
Again, the user will only ever see `BinaryVectorDtype.PACKED_BIT` in their code.

blink1073 (approved):
LGTM!

PR #1815, opened 2024-08-27
PYTHON-4669 - Update More APIs for Motor Compatibility

NoahStapp:


blink1073 (approved):
LGTM!

gridfs/asynchronous/grid_file.py
```diff
@@ -1484,6 +1486,17 @@ def __init__(
     _file: Any
     _chunk_iter: Any
 
+    async def __anext__(self) -> bytes:
+        return super().__next__()
```
ShaneHarvey, line 1490: This is incorrect. We can't call `super().__next__()` because that does blocking I/O.

gridfs/synchronous/grid_file.py
```diff
@@ -1472,6 +1474,15 @@ def __init__(
     _file: Any
     _chunk_iter: Any
 
+    def __next__(self) -> bytes:
+        return super().__next__()
+
+    def __next__(self) -> bytes:  # noqa: F811, RUF100
```
ShaneHarvey, line 1480: Any way to avoid the duplicate `def __next__(self)` definitions?

gridfs/asynchronous/grid_file.py
NoahStapp: Hmm, good catch. We don't have an async equivalent here unless we write one ourselves.

ShaneHarvey: IOBase implements next using readline:
> [IOBase](https://docs.python.org/3/library/io.html#io.IOBase) (and its subclasses) supports the iterator protocol, meaning that an [IOBase](https://docs.python.org/3/library/io.html#io.IOBase) object can be iterated over yielding the lines in a stream. Lines are defined slightly differently depending on whether the stream is a binary stream (yielding bytes), or a text stream (yielding character strings). See [readline()](https://docs.python.org/3/library/io.html#io.IOBase.readline) below.

https://docs.python.org/3/library/io.html#io.IOBase

gridfs/synchronous/grid_file.py
ShaneHarvey: Right, but can we workaround that? The duplicate code is strange the read. There's also a runtime perf cost to overriding a method just to call the super class.

NoahStapp: There isn't a simple way to workaround it, no. We could change the definition to be less confusing, like this:

```
    async def __anext__(self) -> bytes:
        return super().__next__()

    if not _IS_SYNC:

        def __next__(self) -> bytes:  # noqa: F811, RUF100
            raise TypeError(
                "AsyncGridOut does not support synchronous iteration. Use `async for` instead"
            )
```

Which would synchronize to 

```
    def __next__(self) -> bytes:
        return super().__next__()

    if not _IS_SYNC:

        def __next__(self) -> bytes:  # noqa: F811, RUF100
            raise TypeError("GridOut does not support synchronous iteration. Use `for` instead")
```

We can also add a comment explaining why the duplicate def exists.

gridfs/asynchronous/grid_file.py
NoahStapp: There isn't an `asyncio` version of `readline`, so we'd need to write our own. The canonical way to do so appears to be with threads (https://stackoverflow.com/questions/34699948/does-asyncio-supports-asynchronous-i-o-for-file-operations), at which point I question if the performance gained by not blocking the loop is more than the cost of thread overhead.

gridfs/synchronous/grid_file.py
ShaneHarvey: Good idea, but how about:
```python

    if not _IS_SYNC:
        async def __anext__(self) -> bytes:
            return await self.readline()

        def __next__(self) -> bytes:  # noqa: F811, RUF100
            raise TypeError(
                "AsyncGridOut does not support synchronous iteration. Use `async for` instead"
            )
```

NoahStapp: I forgot we had our own async `readline`, good catch. This looks like we'd read a single line or every byte if the file wasn't line-delimited. Is that the intended behavior for iteration here?

ShaneHarvey: Yeah that's what IOBase is supposed to do and AsyncGridOut iteration should match the sync version. We also need to remove IOBase from the async.

NoahStapp: Done.

PR #1819, opened 2024-08-28
PYTHON-4651: Migrate test_client_context.py to async

sleepyStick:


sleepyStick:
I think the failing tests are all related to [PYTHON-4701](https://jira.mongodb.org/browse/PYTHON-4701)

ShaneHarvey:
Strange the rhel8-pr-assign-reviewer task ran but didn't assign a reviewer.

sleepyStick:
Oh i assumed it was cuz the tests/checks weren't passing? Because it won't let me assign / add reviews either..?

test/asynchronous/test_client_context.py
```diff
@@ -0,0 +1,74 @@
...
+
+from test.asynchronous import AsyncUnitTest, SkipTest, async_client_context, unittest
+
+_IS_SYNC = False
+
+
+class TestAsyncClientContext(AsyncUnitTest):
+    @classmethod
+    async def _setup_class(cls):
+        pass
```
ShaneHarvey, line 29: Can we remove _setup_class since it's just a `pass`?

sleepyStick: `AsyncUnitTest` assumes that `_setup_class` would be implemented..without this the code would infinitely recurse and eventually error. 

To my knowledge this is the first time that a class inherits from `AsyncUnitTest` that doesn't "need" a `_setup_class`. I don't know if we want to change `AsyncUnitTest` to accommodate this class's use case or not...

ShaneHarvey: AsyncUnitTest should `pass` instead of trying to recursively call itself.

blink1073:
> Strange the rhel8-pr-assign-reviewer task ran but didn't assign a reviewer.

That's intentional in draft mode

ShaneHarvey:
But the PR was moved out of draft mode. Shouldn't the reviewer be auto-assigned in that case?

blink1073:
> But the PR was moved out of draft mode. Shouldn't the reviewer be auto-assigned in that case?

Not unless the EVG task is re-run.

ShaneHarvey:
Oh TIL. I thought it worked the other way.

blink1073:
> Oh TIL. I thought it worked the other way.

Not until we can have other GitHub events trigger EVG runs, or we make our own service listener with a webhook.

test/asynchronous/test_client_context.py
sleepyStick: That work's too, don't know why i didn't think of that earlier! Just changed :)

NoahStapp (changes requested):
Great work! The async `test_client_context.py` should also be run [here](https://github.com/mongodb/mongo-python-driver/blob/e6b95f65953e01e435004af069a6976473eaf841/.github/workflows/test-python.yml#L212) as part of the `test_sdist` PR hook.

ShaneHarvey (dismissed):

.github/workflows/test-python.yml
```diff
@@ -209,4 +209,4 @@ jobs:
             ls
             which python
             pip install -e ".[test]"
-            PYMONGO_MUST_CONNECT=1 pytest -v test/test_client_context.py
+            PYMONGO_MUST_CONNECT=1 pytest -k client_context
```
NoahStapp, line 212:
```suggestion
            PYMONGO_MUST_CONNECT=1 pytest -v -k client_context
```

NoahStapp: For consistency across our test runs.

NoahStapp (changes requested):

.github/workflows/test-python.yml
ShaneHarvey: Is that something we can set in pyproject?

ShaneHarvey (approved):

NoahStapp (approved):

PR #1820, opened 2024-08-29
PYTHON-4590 - Add type guards to async API methods

NoahStapp:


ShaneHarvey (commented):
Do we need to do this? Doesn't the user already get an error if they pass in the wrong type? Could you give some examples?

NoahStapp:
> Do we need to do this? Doesn't the user already get an error if they pass in the wrong type? Could you give some examples?

The idea is that the current behavior looks like this: 

```
>>> d = Database(AsyncMongoClient(), "test")
>>> d.list_collections()
<coroutine object AsyncMongoClient._retryable_read at 0x103023c40>
```

This change gives us this:

```
>>> d = Database(AsyncMongoClient(), "test")
  File "/Users/nstapp/Github/mongo-python-driver/pymongo/synchronous/database.py", line 128, in __init__
    raise TypeError(f"MongoClient required but {client} is an instance of {type(client)}")
TypeError: MongoClient required but AsyncMongoClient(host=['localhost:27017'], document_class=dict, tz_aware=False, connect=True) is an instance of <class 'pymongo.asynchronous.mongo_client.AsyncMongoClient'>
```

pymongo/asynchronous/encryption.py
```diff
@@ -194,9 +194,7 @@ async def kms_request(self, kms_context: MongoCryptKmsContext) -> None:
             # Wrap I/O errors in PyMongo exceptions.
             _raise_connection_failure((host, port), error)
 
-    async def collection_info(
-        self, database: AsyncDatabase[Mapping[str, Any]], filter: bytes
-    ) -> Optional[bytes]:
+    async def collection_info(self, database: str, filter: bytes) -> Optional[bytes]:
```
blink1073, line 197: Was the signature wrong before?

NoahStapp: We appear to pass a database name and not an actual database here.

blink1073: And we didn't catch this before because it wasn't in the mypy checks?

NoahStapp: We don't call `collection_info` anywhere in our own codebase, only in libmongocrypt. That's probably why it didn't get caught.

pymongo/asynchronous/client_session.py
```diff
@@ -176,6 +176,7 @@
 if TYPE_CHECKING:
     from types import TracebackType
 
+    from pymongo.asynchronous.mongo_client import AsyncMongoClient
```
blink1073, line 179: It looks like this is no longer used

pymongo/asynchronous/bulk.py
```diff
@@ -240,6 +240,7 @@ async def write_command(
         client: AsyncMongoClient,
     ) -> dict[str, Any]:
         """A proxy for SocketInfo.write_command that handles event publishing."""
+
```
ShaneHarvey, line 243: Unneeded.

pymongo/asynchronous/database.py
```diff
@@ -119,9 +119,16 @@ def __init__(
...
         )
 
+        from pymongo.asynchronous.mongo_client import AsyncMongoClient
+
         if not isinstance(name, str):
             raise TypeError("name must be an instance of str")
 
+        if not isinstance(client, AsyncMongoClient):
+            raise TypeError(
+                f"AsyncMongoClient required but {client} is an instance of {type(client)}"
```
ShaneHarvey, line 129: Do we need to include both the repr(client) and the type(client)? It could be safer to only add the type to avoid adding topology/hostname info into these errors.

pymongo/asynchronous/server.py
```diff
@@ -157,7 +157,9 @@ async def run_operation(
         :param unpack_res: A callable that decodes the wire protocol response.
         :param client: An AsyncMongoClient instance.
         """
+
```
ShaneHarvey, line 160: unneeded

pymongo/synchronous/collection.py
```diff
@@ -231,6 +231,12 @@ def __init__(
         )
         if not isinstance(name, str):
             raise TypeError("name must be an instance of str")
+        from pymongo.synchronous.database import Database
+
+        if not isinstance(database, Database):
```
ShaneHarvey, line 236: I'm paranoid this might break apps that Mock pymongo classes but it should be fine. Anything that's mocking should inherit from our base classes anyway (instead of trying to ducktype).

ShaneHarvey (changes requested):

pymongo/synchronous/collection.py
NoahStapp: Good point, we'll have to watch out for bug reports with this.

pymongo/asynchronous/bulk.py
blink1073: These we added by ruff it looks like

blink1073 (approved):

pymongo/asynchronous/collection.py
```diff
@@ -228,6 +228,12 @@ def __init__(
         )
         if not isinstance(name, str):
             raise TypeError("name must be an instance of str")
+        from pymongo.asynchronous.database import AsyncDatabase
+
+        if not isinstance(database, AsyncDatabase):
+            raise TypeError(
+                f"AsyncCollection requires an AsyncDatabase, {database} is an instance of {type(database)}"
```
ShaneHarvey, line 235: We should make the same `{client}` error message change to all the Database ones as well since Database's repr includes the client.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #1821, opened 2024-08-29
PYTHON-4669 - Update Async GridFS APIs for Motor Compatibility

NoahStapp:


gridfs/asynchronous/grid_file.py
```diff
@@ -1486,13 +1489,14 @@ def __init__(
...
+    if not _IS_SYNC:
 
-    def __next__(self) -> bytes:  # noqa: F811, RUF100
-        if _IS_SYNC:
-            return super().__next__()
-        else:
+        async def __anext__(self) -> bytes:
+            return await self.readline()
+
+        # This is a duplicate definition of __next__ for the synchronous API
```
ShaneHarvey, line 1497: synchronous -> asynchronous

gridfs/asynchronous/grid_file.py
```diff
@@ -1486,13 +1489,14 @@ def __init__(
...
-    async def __anext__(self) -> bytes:
-        return super().__next__()
+    if not _IS_SYNC:
 
-    def __next__(self) -> bytes:  # noqa: F811, RUF100
-        if _IS_SYNC:
-            return super().__next__()
-        else:
+        async def __anext__(self) -> bytes:
+            return await self.readline()
```
ShaneHarvey, line 1495: We need to handle the end-of-stream and raise StopAsyncIteration here. Could you add a test for this?

gridfs/asynchronous/grid_file.py
```diff
@@ -1486,13 +1489,14 @@ def __init__(
...
+
+        # This is a duplicate definition of __next__ for the synchronous API
+        # due to the limitations of our synchro process
+        def __next__(self) -> bytes:  # noqa: F811, RUF100
```
ShaneHarvey, line 1499: Let's just remove `__next__` altogether.

ShaneHarvey (changes requested):

pymongo/synchronous/topology.py
```diff
@@ -521,7 +521,7 @@ def _process_change(
             if server:
                 server.pool.reset(interrupt_connections=interrupt_connections)
 
-        # Wake waiters in select_servers().
+        # Wake witers in select_servers().
```
blink1073, line 524:
```suggestion
        # Wake waiters in select_servers().
```

test/test_grid_file.py
```diff
@@ -380,29 +411,30 @@ def test_multi_chunk_file(self):
...
-
-            g = GridOut(self.db.fs, f._id)
-            self.assertEqual(data, g.read())
-
-            g = GridOut(self.db.fs, f._id)
-            self.assertEqual(data, g.read(10) + g.read(10))
-            return True
-
-        qcheck.check_unittest(self, helper, qcheck.gen_string(qcheck.gen_range(0, 20)))
+    # # TODO: convert qcheck to async
```
blink1073, line 414: New ticket?

NoahStapp: https://jira.mongodb.org/browse/PYTHON-4708, put it in the comment.

blink1073 (dismissed):
LGTM!

gridfs/asynchronous/grid_file.py
```diff
@@ -1616,18 +1613,37 @@ async def read(self, size: int = -1) -> bytes:
...
 
-    async def readline(self, size: int = -1) -> bytes:  # type: ignore[override]
+    async def readline(self, size: int = -1) -> bytes:
         """Read one line or up to `size` bytes from the file.
 
         :param size: the maximum number of bytes to read
         """
         return await self._read_size_or_line(size=size, line=True)
 
+    async def readlines(self, size: int = -1) -> list[bytes]:
```
ShaneHarvey, line 1623: This should be async only too since we intentionally get it for free  from IOBase in the sync version.

gridfs/asynchronous/grid_file.py
```diff
@@ -1194,19 +1194,9 @@ def __setattr__(self, name: str, value: Any) -> None:
                 )
```
ShaneHarvey, line 1194: Rather than having `__setattr__` always raise an error in async, we should allow it as long as the file is not closed and only raise an error if the file is closed:
```python
                # All other attributes are part of the document in db.fs.files.
                # Store them to be sent to server on close() or if closed, send
                # them now.
                self._file[name] = value
                if self._closed:        
                    if _IS_SYNC:
                        self._coll.files.update_one({"_id": self._file["_id"]}, {"$set": {name: value}})
                    else:
                        raise AttributeError(
                            "AsyncGridIn does not support __setattr__ after being closed(). Set the attribute before closing the file or use AsyncGridIn.set() instead")
```

ShaneHarvey (changes requested):

test/test_grid_file.py
```diff
@@ -380,29 +411,30 @@ def test_multi_chunk_file(self):
...
+    #         self.assertEqual(self.chunks, self.db.fs.chunks.count_documents({}))
+    #
+    #         g = GridOut(self.db.fs, f._id)
+    #         self.assertEqual(data, g.read())
+    #
+    #         g = GridOut(self.db.fs, f._id)
+    #         self.assertEqual(data, g.read(10) + g.read(10))
+    #         return True
+    #
+    #     qcheck.check_unittest(self, helper, qcheck.gen_string(qcheck.gen_range(0, 20)))
```
ShaneHarvey, line 437: We shouldn't disable this test in sync. It should only be skipped in async.

test/test_grid_file.py
```diff
@@ -781,6 +848,7 @@ def test_zip(self):
         self.assertEqual(1, self.db.fs.chunks.count_documents({}))
 
         g = GridOut(self.db.fs, f._id)
+        g.open()
```
ShaneHarvey, line 851: Why add the open call here?

NoahStapp: Artifact from before I made this test sync-only, good catch.

test/test_grid_file.py
NoahStapp: Good catch, forgot to uncomment whoops.

test/test_grid_file.py
```diff
@@ -145,12 +150,19 @@ def test_grid_in_default_opts(self):
 
         self.assertEqual(None, a.filename)
         self.assertEqual(None, a.name)
-        a.filename = "my_file"
+        if _IS_SYNC:
+            a.filename = "my_file"
+        else:
+            a.set("filename", "my_file")
```
ShaneHarvey, line 156: Most (all?) of these can be change back to the `a.filename = "my_file"` version now, right?

NoahStapp: Trying to use `a.filename = "my_file"` gives the following error:

```
AttributeError: property 'filename' of 'AsyncGridIn' object has no setter
```

This is because all those properties are `_a_grid_in_property` instances, which don't have an async setter since setting does file IO.

ShaneHarvey: We should make the setters work now. They will only raise an error on async when the file is closed.

NoahStapp: Ah I misunderstood--makes sense, will do.

test/test_grid_file.py
```diff
@@ -213,12 +226,14 @@ def test_grid_out_default_opts(self):
 
         gout = GridOut(self.db.fs, 5)
         with self.assertRaises(NoFile):
+            gout.open()
```
ShaneHarvey, line 229: There's a lot of added open() calls in this file. Is that intended or should they be changed to `if not _IS_SYNC: open()`

NoahStapp: They should all be only for the async tests, fixed.

ShaneHarvey (approved):

ShaneHarvey:
LGTM but could you update the PR title and jira ticket to summarize the gridfs changes?

PR #1822, opened 2024-08-30
PYTHON-4667 Handle  $clusterTime from error responses in client Bulk Write

blink1073:


pymongo/asynchronous/client_bulk.py
```diff
@@ -312,6 +312,9 @@ async def write_command(
                 bwc._fail(request_id, failure, duration)
             # Top-level error will be embedded in ClientBulkWriteException.
             reply = {"error": exc}
+            # Propagate $clusterTime to the caller.
+            if "$clusterTime" in cmd:
+                reply["$clusterTime"] = cmd["$clusterTime"]
```
ShaneHarvey, line 317: $clusterTime needs to be parsed from the (potential) server error response, not the command. We also need to do the same for operationTime.

This is the method that handles it, it probably makes more sense to call this method at a lower level if we can (perhaps in write_command):
```
client._process_response(command_response, session)
```

Or another alternative.

blink1073: Makes sense, updated.

pymongo/asynchronous/bulk.py
```diff
@@ -308,6 +309,11 @@ async def write_command(
 
             if bwc.publish:
                 bwc._fail(request_id, failure, duration)
+            # Process the response from the server.
+            if isinstance(exc, OperationFailure):
+                await client._process_response(exc.details, bwc.session)  # type: ignore[arg-type]
+            else:
+                await client._process_response({}, bwc.session)  # type: ignore[arg-type]
```
ShaneHarvey, line 316: We don't need an `else` here.

pymongo/asynchronous/bulk.py
```diff
@@ -308,6 +309,11 @@ async def write_command(
 
             if bwc.publish:
                 bwc._fail(request_id, failure, duration)
+            # Process the response from the server.
+            if isinstance(exc, OperationFailure):
```
ShaneHarvey, line 313: `isinstance(exc, (NotPrimaryError, OperationFailure))`

pymongo/synchronous/bulk.py
```diff
@@ -308,6 +309,11 @@ def write_command(
 
             if bwc.publish:
                 bwc._fail(request_id, failure, duration)
+            # Process the response from the server.
+            if isinstance(exc, OperationFailure):
```
ShaneHarvey, line 313: Same comments here.

ShaneHarvey (changes requested):
Could you add a test to `test/mockupdb/test_cluster_time.py`? Something like:
```python
    def test_bulk_error(self):
        def callback(client: MongoClient[dict]) -> None:
            client.db.collection.bulk_write(
                [InsertOne({}), InsertOne({})]
            )

        self.cluster_time_conversation(
            callback,
            [{"ok": 0, "errmsg": "mock error"}],
        )
```

pymongo/asynchronous/bulk.py
blink1073: Removed

pymongo/asynchronous/bulk.py
blink1073: Done

pymongo/synchronous/bulk.py
blink1073: Automatically done.  :smile:

blink1073:
> Could you add a test

Done

test/mockupdb/test_cluster_time.py
```diff
@@ -166,6 +167,17 @@ def test_monitor(self):
...
+    def test_bulk_error(self):
+        def callback(client: MongoClient[dict]) -> None:
+            client.db.collection.bulk_write([InsertOne({}), InsertOne({})])
+
+        with self.assertRaises(OperationFailure) as context:
+            self.cluster_time_conversation(
+                callback,
+                [{"ok": 0, "errmsg": "mock error"}],
+            )
+        self.assertIn("$clusterTime", str(context.exception))
```
ShaneHarvey, line 179: This test is skipping the clusterTime asserts in cluster_time_conversation. We need to do:
```
        def callback(client: MongoClient[dict]) -> None:
                client.db.collection.bulk_write([InsertOne({}), InsertOne({})])

        self.cluster_time_conversation(
            callback,
            [{"ok": 0, "errmsg": "mock error"}],
        )
```

ShaneHarvey: Could you also add a similar test for client.bulk_write?

blink1073: Done and done

ShaneHarvey (approved):
LGTM once the tests pass!

PR #1826, opened 2024-09-03
PYTHON-4701 - Topology logging should use suppress_event

NoahStapp:


pymongo/asynchronous/topology.py
```diff
@@ -497,7 +497,7 @@ async def _process_change(
                     (td_old, self._description, self._topology_id),
                 )
             )
-        if _SDAM_LOGGER.isEnabledFor(logging.DEBUG):
+        if _SDAM_LOGGER.isEnabledFor(logging.DEBUG) and sd_old != server_description:
```
ShaneHarvey, line 500: Should this be ` and not suppress_event`?

ShaneHarvey:
Also could you add some more info to the failing assert message?

pymongo/asynchronous/topology.py
NoahStapp: `suppress_event` checks for event publishing flags, which we don't use for logging.

NoahStapp:
> Also could you add some more info to the failing assert message?

To which message?

pymongo/asynchronous/topology.py
ShaneHarvey: Could that be changed?
```
suppress_event = sd_old == server_description
```

ShaneHarvey (approved):

ShaneHarvey:
Could you update the PR title to describe the bug fix instead of the test failure name?

PR #1827, opened 2024-09-03
PYTHON-4704 Migrate test_bulk.py to async

sleepyStick:


test/test_bulk.py
```diff
@@ -929,13 +966,13 @@ def cause_wtimeout(self, requests, ordered):
 
         # Use the rsSyncApplyStop failpoint to pause replication on a
         # secondary which will cause a wtimeout error.
-        self.secondary.admin.command("configureFailPoint", "rsSyncApplyStop", mode="alwaysOn")
+        self.secondary.admin.command("configureFailPoint", "rsApplyStop", mode="alwaysOn")
```
NoahStapp, line 969: This needs to be `rsSyncApplyStop`.

NoahStapp (changes requested):

test/test_bulk.py
sleepyStick: thanks for catching that! fixed!

NoahStapp (approved):

PR #1828, opened 2024-09-04
PYTHON-4695 Fix test event loop policy and improve error traceback for ClientBulkWriteException

blink1073:
PYTHON-4695


ShaneHarvey (changes requested):
Great improvements but we still need to fix the bug where ClientBulkWriteException contains a raw TimeoutError. That should never be possible even if the event loop policy is incompatible. We should be catching TimeoutError and raising pymongo's NetworkTimeout.

ShaneHarvey (dismissed):

blink1073:
> Great improvements but we still need to fix the bug where ClientBulkWriteException contains a raw TimeoutError. That should never be possible even if the event loop policy is incompatible. We should be catching TimeoutError and raising pymongo's NetworkTimeout.

I opened [PYTHON-4740](https://jira.mongodb.org/browse/PYTHON-4740) to track that question separately.

ShaneHarvey (approved):

PR #1832, opened 2024-09-05
PYTHON-4708 - Convert test.qcheck to async

NoahStapp:


ShaneHarvey:
Did you see my question on the jira ticket: 
> IIRC test.qcheck is very slow. Do we need to migrate it to async?

I'm not sure we need this PR.

NoahStapp:
> Did you see my question on the jira ticket:
> 
> > IIRC test.qcheck is very slow. Do we need to migrate it to async?
> 
> I'm not sure we need this PR.

Sorry, missed your comment. If we don't care to have this one specific test, we can remove `qcheck` entirely from this test suite. The BSON tests are the only other place it's used.

ShaneHarvey:
Eh it's not as slow as I thought. This is fine.

PR #1833, opened 2024-09-05
PYTHON-3193 - Add ResourceWarning for unclosed MongoClients in __del__

NoahStapp:


pymongo/asynchronous/mongo_client.py
```diff
@@ -1173,6 +1175,22 @@ def __getitem__(self, name: str) -> database.AsyncDatabase[_DocumentType]:
         """
         return database.AsyncDatabase(self, name)
 
+    def __del__(self) -> None:
+        """Check that this AsyncMongoClient has been closed and issue a warning if not."""
+        # TODO: Remove in https://jira.mongodb.org/browse/PYTHON-4731
```
blink1073, line 1180: I didn't realize we removed `MongoClient.__del__`.  I think we should preserver the 4.8 behavior and change `connect=False` and remove the `__del__` behavior together in 5.0

pymongo/asynchronous/mongo_client.py
```diff
@@ -1173,6 +1175,22 @@ def __getitem__(self, name: str) -> database.AsyncDatabase[_DocumentType]:
...
+        try:
+            if not self._closed:
+                warnings.warn(
+                    f"Unclosed {self}",
+                    ResourceWarning,
+                    stacklevel=2,
+                    source=self,
+                )
+            if _IS_SYNC and self._opened:
+                self.close()  # type: ignore[unused-coroutine]
```
ShaneHarvey, line 1190: We can't call close() even in the sync api because it might deadlock.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1173,6 +1175,22 @@ def __getitem__(self, name: str) -> database.AsyncDatabase[_DocumentType]:
         """
         return database.AsyncDatabase(self, name)
 
+    def __del__(self) -> None:
+        """Check that this AsyncMongoClient has been closed and issue a warning if not."""
+        # TODO: Remove in https://jira.mongodb.org/browse/PYTHON-4731
+        try:
+            if not self._closed:
+                warnings.warn(
+                    f"Unclosed {self}",
```
ShaneHarvey, line 1184: We should add more information here so users know why this started.

ShaneHarvey (changes requested):

pymongo/asynchronous/mongo_client.py
ShaneHarvey: Wouldn't we want to keep the `__del__` warning forever?

blink1073: Oh wait, we never had a `__del__`.  Yes, I agree.

pymongo/asynchronous/mongo_client.py
NoahStapp: What kind of information should be here? Something like `Unclosed MongoClient. Call MongoClient.close() to safely shut down your client and free up resources.`?

NoahStapp: Do we want to pull that ticket up to 4.9, then? We should do them both in the same release.

ShaneHarvey:
@NoahStapp could you also do some history digging into `__del__`? I remember there was a reason we avoided adding it to client in the past. Something related to reference cycles.

NoahStapp:
> @NoahStapp could you also do some history digging into `__del__`? I remember there was a reason we avoided adding it to client in the past. Something related to reference cycles.

My understanding is that if you try to do any actual cleanup inside `__del__`, you risk potentially causing issues with the entire reference cycle your object is part of. All of the past tickets I found on this topic relate to attempting to do cleanup: since this version of `__del__` is only raising a warning if the client wasn't explicitly closed before garbage collection, we should avoid any issues.

ShaneHarvey:
Yes that's one issue but there's another unrelated issue. Adding `__del__` to a class changes the way CPython garbage collects the object. In particular if it's part of a reference cycle the GC can be significantly delayed because `__del__` is defined. I'm going based on memory so this is a bit fuzzy and CPython could work differently now.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1173,6 +1175,30 @@ def __getitem__(self, name: str) -> database.AsyncDatabase[_DocumentType]:
         """
         return database.AsyncDatabase(self, name)
 
+    def __del__(self) -> None:
+        """Check that this AsyncMongoClient has been closed and issue a warning if not."""
+        # TODO: Remove in https://jira.mongodb.org/browse/PYTHON-4731
+        try:
+            if not self._closed:
+                if _IS_SYNC:
```
ShaneHarvey, line 1183: I'd prefer to remove the `await` from the error message to reduce the duplication here.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1173,6 +1175,30 @@ def __getitem__(self, name: str) -> database.AsyncDatabase[_DocumentType]:
...
+                    msg = (
+                        f"Unclosed {type(self)}. "
+                        f"Call {type(self)}.close() to safely shut down your client and free up resources."
```
ShaneHarvey, line 1186: The error message can/should also include the traceback of where the client was opened via topology_settings._stack.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: We can remove this comment now.

ShaneHarvey (dismissed):

NoahStapp:
> Yes that's one issue but there's another unrelated issue. Adding `__del__` to a class changes the way CPython garbage collects the object. In particular if it's part of a reference cycle the GC can be significantly delayed because `__del__` is defined. I'm going based on memory so this is a bit fuzzy and CPython could work differently now.

I can't find an example of `__del__` delaying GC, but I did find this: the [garbage collector docs](https://docs.python.org/3/library/gc.html#gc.garbage) suggest that starting in 3.4, defining `__del__` shouldn't affect the GC behavior for that object, even if it is part of a reference cycle.

ShaneHarvey:
Thanks for looking into it. Could you?:
- add changelog entry
- merge master to get the tests green
- provide an example of the new warning behavior

NoahStapp:
Here's the current warning:

```
sys:1: ResourceWarning: Unclosed MongoClient opened at:
  File "/Users/nstapp/Github/mongo-python-driver/scratch.py", line 8, in <module>
    client = MongoClient()
  File "/Users/nstapp/Github/mongo-python-driver/pymongo/synchronous/mongo_client.py", line 847, in __init__
    self._topology_settings = TopologySettings(
  File "/Users/nstapp/Github/mongo-python-driver/pymongo/synchronous/settings.py", line 85, in __init__
    self._stack = "".join(traceback.format_stack())
```

ShaneHarvey:
What about the "Call {type(self).__name__}.close() to safely shut down your client and free up resources." part?

NoahStapp:
> What about the "Call {type(self).**name**}.close() to safely shut down your client and free up resources." part?

Fixed.

NoahStapp:
> Also I wonder if we can remove the last two frames from the trace to make it cleaner.

Return just the first frame? That should always be the line where the client was constructed, right?

ShaneHarvey:
I mean
```
self._stack = "".join(traceback.format_stack()[:-2])
```

doc/changelog.rst
```diff
@@ -42,6 +42,8 @@ PyMongo 4.9 brings a number of improvements including:
 - Fixed a bug where PyMongo would raise ``InvalidBSON: date value out of range``
   when using :attr:`~bson.codec_options.DatetimeConversion.DATETIME_CLAMP` or
   :attr:`~bson.codec_options.DatetimeConversion.DATETIME_AUTO` with a non-UTC timezone.
+- Added a warning to unclosed MongoClient instances
+  telling users to explicitly close clients when finished with them to avoid leaking resources .
```
ShaneHarvey, line 46: `resources .` -> `resources.`

Let's include an example of the warning here too.

doc/changelog.rst
```diff
@@ -42,6 +42,16 @@ PyMongo 4.9 brings a number of improvements including:
 - Fixed a bug where PyMongo would raise ``InvalidBSON: date value out of range``
   when using :attr:`~bson.codec_options.DatetimeConversion.DATETIME_CLAMP` or
   :attr:`~bson.codec_options.DatetimeConversion.DATETIME_AUTO` with a non-UTC timezone.
+- Added a warning to unclosed MongoClient instances
+  telling users to explicitly close clients when finished with them to avoid leaking resources.
+  Here's an example of what the warning looks like:
```
ShaneHarvey, line 47: Nit: `Here's an example of what the warning looks like:` -> `For example:`

blink1073 (approved):
LGTM!

PR #1837, opened 2024-09-05
PYTHON-4607 Use LogRecord.getMessage() not LogRecord.message

ShaneHarvey:
PYTHON-4607 Use [LogRecord.getMessage()](https://docs.python.org/3/library/logging.html#logging.LogRecord.getMessage) not LogRecord.message

LogRecord.message is documented like this:

> The logged message, computed as msg % args. This is set when Formatter.format() is invoked.

It's not set until Formatter.format() or getMessage() is called which explains the `AttributeError: 'LogRecord' object has no attribute 'message'` errors we've been seeing.


NoahStapp (approved):
Good find, thanks!

PR #1841, opened 2024-09-05
PYTHON-4617 Skip unified retryable writes tests on MMAPv1

ShaneHarvey:
PYTHON-4617 Skip unified retryable writes tests on MMAPv1

Also removes some unused legacy spec runner code.


blink1073 (approved):
LGTM!

PR #1853, opened 2024-09-11
PYTHON-4763 Migrate test_change_stream.py to async

sleepyStick:


test/test_change_stream.py
```diff
@@ -310,7 +312,7 @@ def _test_next_blocks(self, change_stream):
     @no_type_check
     def test_next_blocks(self):
         """Test that next blocks until a change is readable"""
-        # Use a short await time to speed up the test.
+        # Use a short time to speed up the test.
```
NoahStapp, line 315:
```suggestion
        # Use a short wait time to speed up the test.
```

test/test_change_stream.py
```diff
@@ -325,7 +327,7 @@ def test_aggregate_cursor_blocks(self):
     @no_type_check
     def test_concurrent_close(self):
         """Ensure a ChangeStream can be closed from another thread."""
-        # Use a short await time to speed up the test.
+        # Use a short time to speed up the test.
```
NoahStapp, line 330:
```suggestion
        # Use a short wait time to speed up the test.
```

NoahStapp (changes requested):
The async `test_change_stream.py` file is missing?

sleepyStick:
> The async `test_change_stream.py` file is missing?

omg one sec, i always forget that -a doesn't add the file... ;-;

test/test_change_stream.py
sleepyStick: changed :)

sleepyStick:
like fifty attempts later, its ready HAHA

test/test_change_stream.py
```diff
@@ -158,10 +162,14 @@ def test_try_next(self):
         with self.change_stream(max_await_time_ms=250) as stream:
             self.assertIsNone(stream.try_next())  # No changes initially.
             coll.insert_one({})  # Generate a change.
+
```
NoahStapp, line 165: Unneeded added space here.

test/test_change_stream.py
```diff
@@ -192,7 +200,11 @@ def test_try_next_runs_one_getmore(self):
 
             # Get at least one change before resuming.
             coll.insert_one({"_id": 2})
-            wait_until(lambda: stream.try_next() is not None, "get change from try_next")
+
+            def _wait_until():
+                return stream.try_next()
```
NoahStapp, line 205:
```suggestion
                return stream.try_next() is not None
```

test/test_change_stream.py
```diff
@@ -209,7 +221,11 @@ def test_try_next_runs_one_getmore(self):
 
             # Stream still works after a resume.
             coll.insert_one({"_id": 3})
-            wait_until(lambda: stream.try_next() is not None, "get change from try_next")
+
+            def _wait_until():
+                return stream.try_next()
```
NoahStapp, line 226:
```suggestion
                return stream.try_next() is not None
```

NoahStapp (changes requested):
A few minor comments, otherwise looks great!

test/test_change_stream.py
sleepyStick: Yeah...I'll try again for sure, but i tried to delete it and when i ran pre-commit it was added again??? not sure why though

blink1073: Because the async file overwrites the sync one

blink1073: ohh, I think it is because ruff is just doing that because the file has changed

sleepyStick: yeah so I found the space in the async file, deleted it, and then it got added back during pre-commit

NoahStapp: Ah in that case never mind, if the formatter adds an extra space that's fine.

sleepyStick: perfect, because I just gave up on trying to convince it otherwise

test/test_change_stream.py
sleepyStick: fixed :)

NoahStapp (approved):

PR #1854, opened 2024-09-12
PYTHON-1982 Update Invalid Document error message to include doc

navjots18:
Updated bson and cbson code to add the doc in the Invalid Document error message and added Unit test cases.
 
 Note: Make sure to run to `python _setup.py build_ext --inplace ` to install C file changes


blink1073:
Thanks @navjots18!  It looks like there is a segmentation fault in the C implementation.  Are you able to reproduce that failure locally?   I can take a closer look on Monday, but https://dnmtechs.com/handling-segmentation-fault-errors-in-python-3-programming/ is a good resource.

navjots18:
Hi @blink1073 , it was happening because i was assigning some variables to eType and eValue and then using Py_DECREF on the variables. Changed this which fixed the segmentation error on my local

bson/_cbsonmodule.c
```diff
@@ -1743,6 +1743,42 @@ int write_dict(PyObject* self, buffer_t buffer,
         while (PyDict_Next(dict, &pos, &key, &value)) {
             if (!decode_and_write_pair(self, buffer, key, value,
                                     check_keys, options, top_level)) {
+                if (PyErr_Occurred()) {
+                    PyObject *etype, *evalue, *etrace;
+                    PyErr_Fetch(&etype, &evalue, &etrace);
+                    PyObject *InvalidDocument = _error("InvalidDocument");
+
+                    if (PyErr_GivenExceptionMatches(etype, InvalidDocument)) {
```
blink1073, line 1751: Can you please reduce the nesting by using `if (InvalidDocument && PyErr_GivenExceptionMatches(etype, InvalidDocument))`?

bson/__init__.py
```diff
@@ -1006,7 +1006,10 @@ def _dict_to_bson(
             elements.append(_name_value_to_bson(b"_id\x00", doc["_id"], check_keys, opts))
         for key, value in doc.items():
             if not top_level or key != "_id":
-                elements.append(_element_to_bson(key, value, check_keys, opts))
+                try:
+                    elements.append(_element_to_bson(key, value, check_keys, opts))
+                except InvalidDocument as err:
+                    raise InvalidDocument(f"Invalid document {doc} | {err}") from err
```
ShaneHarvey, line 1012: I'm not sure this is the pattern we want, in the C version of this code we recursively call write_dict() for subdocuments which means if we end up erroring in a nested field the error will be something like:
`Invalid document {"a": {"b": {"c": ...}}} | Invalid document {"b": {"c": ...}} | Invalid document {"c": ...} | ...`

Is that the intended behavior?

bson/_cbsonmodule.c
```diff
@@ -1743,6 +1743,42 @@ int write_dict(PyObject* self, buffer_t buffer,
         while (PyDict_Next(dict, &pos, &key, &value)) {
             if (!decode_and_write_pair(self, buffer, key, value,
                                     check_keys, options, top_level)) {
+                if (PyErr_Occurred()) {
+                    PyObject *etype, *evalue, *etrace;
+                    PyErr_Fetch(&etype, &evalue, &etrace);
+                    PyObject *InvalidDocument = _error("InvalidDocument");
```
ShaneHarvey, line 1749: There's a lot of missing error/NULL checks in this code. I'm not comfortable merging this for 4.9 so it will need to wait for the next release.

navjots18: @blink1073 @ShaneHarvey  I took context from code written in the same file, would appreciate if can you point out some existing code from where i can see what all error/Nulls check to put here?

blink1073: Hi @navjots18, apologies for the delay.  I think all we're missing in this block is initializing the `etype`, `evalue`, and `etrace` to NULL as we do [here](https://github.com/mongodb/mongo-python-driver/blob/2f1227c504064d6859cb05d567999262f99b7937/bson/_cbsonmodule.c#L2515).  Otherwise I agree this matches the rest of the existing code block.

navjots18: Hey @blink1073 fixed these issues, can you check again?

blink1073 (approved):
LGTM, thank you!

PR #1855, opened 2024-09-12
PYTHON-4731 - Explicitly close all MongoClients opened during tests

NoahStapp:


blink1073:
`test_cleanup_executors_on_client_del` is failing because we have created a reference to the client through `self.addCleanup(client.close)` in this PR.   I think that test needs to be special cased and not use the builtin methods, but instead ignore the resource warning in the test.

blink1073:
The other half of the fix is to remove `source=self` from the warnings in `__del__`, which add another reference.

NoahStapp:
> `test_cleanup_executors_on_client_del` is failing because we have created a reference to the client through `self.addCleanup(client.close)` in this PR. I think that test needs to be special cased and not use the builtin methods, but instead ignore the resource warning in the test.

Yup that did it, fantastic catch!

pymongo/asynchronous/mongo_client.py
```diff
@@ -1185,7 +1185,7 @@ def __getitem__(self, name: str) -> database.AsyncDatabase[_DocumentType]:
     def __del__(self) -> None:
         """Check that this AsyncMongoClient has been closed and issue a warning if not."""
         try:
-            if not self._closed:
+            if self._opened and not self._closed:
```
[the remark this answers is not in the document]
NoahStapp: Good point, unless we think it's useful to provide the user with the instance of the unclosed client?

blink1073:
`test.asynchronous.test_bulk.AsyncTestBulk.test_numerous_inserts` needs updating.

blink1073:
Also `test.test_bulk.TestBulk.test_numerous_inserts`

NoahStapp:
> `test.asynchronous.test_bulk.AsyncTestBulk.test_numerous_inserts` needs updating.

Missed one of the branches in `async_client_context.init_client`, fixed.

blink1073:
I think you might need to pull from master again to pick up #1853

blink1073 (approved):
All green!

test/asynchronous/__init__.py
```diff
@@ -932,6 +937,188 @@ def _target() -> None:
...
+                not res["username"]
+                and not res["password"]
+                and "username" not in client_options
+                and "password" not in client_options
+            ):
+                client_options["username"] = db_user
+                client_options["password"] = db_pwd
+        client = AsyncMongoClient(uri, port, **client_options)
+        if client._options.connect:
+            await client.aconnect()
```
ShaneHarvey, line 1000: Can you explain what motivated calling connect() here? It makes it so that `_async_mongo_client()` behaves differently than `AsyncMongoClient()` which I worry could cause issues down the line.

NoahStapp: Several places that use a `_async_mongo_client()` wrapper expect the client it returns to already be connected. In the interest of minimizing the amount of changes, I wanted to keep that expectation where it existed.

Notably, `simple_client()` directly wraps `AsyncMongoClient()` instead without any of the extra wrapping we do in `_async_mongo_client()`.

ShaneHarvey: Makes sense, thanks!

PR #1858, opened 2024-09-16
 PYTHON-4764 Update to use current supported EVG hosts

blink1073:


.evergreen/config.yml
```diff
@@ -2795,7 +2795,7 @@ buildvariants:
 
 - matrix_name: "tests-windows-enterprise-auth"
   matrix_spec:
-    platform: windows-64-vsMulti-small
```
ShaneHarvey, line 2798: windows-64-vsMulti-small is/was a special distro that was built just for us to test with. I take it we don't need that anymore?

ShaneHarvey: https://jira.mongodb.org/browse/BUILD-9934

Ah it was for testing python <=3.4. So yea we don't need it anymore.

blink1073: This is just changing the name we give it, not the actual host

.evergreen/config.yml
```diff
@@ -2081,10 +2081,10 @@ axes:
...
       - id: rhel8
         display_name: "RHEL 8.x"
-        run_on: rhel87-small
+        run_on: rhel8.8-small
         batchtime: 10080  # 7 days
-      - id: rhel92-fips
-        display_name: "RHEL 9.2 FIPS"
+      - id: rhel9-fips
+        display_name: "RHEL 9 FIPS"
         run_on: rhel92-fips
```
Jibola, line 2088:
```suggestion
        run_on: rhel9-fips
```

Jibola: Does the run_on also need to be changed?

Jibola (changes requested):

.evergreen/config.yml
blink1073: No, the host is still rhel92-fips.

ShaneHarvey (approved):

PR #1859, opened 2024-09-16
PYTHON-3959 - NULL Initialize PyObjects

aclark4life:
NULL initialize PyObjects before calling PyErr_*

### DIFF
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -279,7 +279,7 @@ static PyObject* datetime_from_millis(long long millis) {
      * micros = diff * 1000                       111000
      * Resulting in datetime(1, 1, 1, 1, 1, 1, 111000) -- the expected result
      */
-    PyObject* datetime;
+    PyObject* datetime = NULL;
     int diff = (int)(((millis % 1000) + 1000) % 1000);
     int microseconds = diff * 1000;
     Time64_T seconds = (millis - diff) / 1000;
@@ -1790,7 +1790,7 @@ static PyObject* _cbson_dict_to_bson(PyObject* self, PyObject* args) {
     PyObject* result;
     unsigned char check_keys;
     unsigned char top_level = 1;
-    PyObject* options_obj;
+    PyObject* options_obj = NULL;
     codec_options_t options;
     buffer_t buffer;
     PyObject* raw_bson_document_bytes_obj;
@@ -2620,7 +2620,7 @@ static PyObject* _cbson_element_to_dict(PyObject* self, PyObject* args) {
     /* TODO: Support buffer protocol */
     char* string;
     PyObject* bson;
-    PyObject* options_obj;
+    PyObject* options_obj = NULL;
     codec_options_t options;
     unsigned position;
     unsigned max;
@@ -2732,7 +2732,7 @@ static PyObject* _cbson_bson_to_dict(PyObject* self, PyObject* args) {
     int32_t size;
     Py_ssize_t total_size;
     const char* string;
-    PyObject* bson;
+    PyObject* bson = NULL;
     codec_options_t options;
     PyObject* result = NULL;
     PyObject* options_obj;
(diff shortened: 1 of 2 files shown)


Jibola (approved):
LGTM

bson/_cbsonmodule.c
```diff
@@ -1790,7 +1790,7 @@ static PyObject* _cbson_dict_to_bson(PyObject* self, PyObject* args) {
     PyObject* result;
```
Jibola, line 1790: We should probably null initialize every `PyObject*`

Jibola (changes requested):
Actually, I see there are some non NULL-initialized PyObjects. I think we should target all the PyObject* instantiations as I'm assuming this has only targeted the ones that leverage `PyFetch`

ShaneHarvey:
The jira ticket, PYTHON-3959, mentions initializing vars for PyError_Fetch. Should this PR be updating those?

bson/_cbsonmodule.c
aclark4life: Right, I left them alone since the initial goal was to silence Coverity and I'd be curious to see a Coverity run before setting more NULLs to make sure I didn't miss any. Also, would be nice to have some metric to indicate that setting the rest of them to NULL makes some difference somewhere … unless we already know this to be a positive change and don't care to measure.

aclark4life:
> The jira ticket, [PYTHON-3959](https://jira.mongodb.org/browse/PYTHON-3959), mentions initializing vars for PyErr_Fetch. Should this PR be updating those?

[PYTHON-3959](https://jira.mongodb.org/browse/PYTHON-3959) mentions initializing `PyObject` variables so that `PyErr_*` calls are less noisy in Coverity reports. This PR only initializes `PyObject` variables that precede calls to `PyErr_*`.  As @Jibola [mentions above ](https://github.com/mongodb/mongo-python-driver/pull/1859#discussion_r1761631939) we could also initialize all `PyObject` variables too for "good measure", but if there are other proposed changes to be made in addition to that, I'm not aware of them yet.

ShaneHarvey:
So are we going to update the PyErr_Fetch calls?

bson/_cbsonmodule.c
aclark4life: Fixed!

aclark4life:
> So are we going to update the PyErr_Fetch calls?

Yes! Sorry, In my first pass I misread which PyObject variables were called by which PyErr_ functions. Fixed.

ShaneHarvey:
I know this is only changing NULLS but just to be safe we should wait until 4.9 is released to merge.

ShaneHarvey:
I'll defer to Jib for the rest.

Jibola (changes requested):
Missed two in `_cbsonmodule.c`
* [1](https://github.com/mongodb/mongo-python-driver/blob/d5d4a65889356e1064313eb8cfbb714d1f1488f7/bson/_cbsonmodule.c#L2515)
* [2](https://github.com/mongodb/mongo-python-driver/blob/d5d4a65889356e1064313eb8cfbb714d1f1488f7/bson/_cbsonmodule.c#L2588-L2589)


Thanks for the work thus far, though! I imagine navigating this c file is a bit pesky.

aclark4life:
> Thanks for the work thus far, though! I imagine navigating this c file is a bit pesky.

Got'em, thanks! C is a pleasure to work with 👼

Jibola (approved):
LGTM!!!

pymongo/_cmessagemodule.c
```diff
@@ -31,7 +31,7 @@ struct module_state {
     PyObject* _max_bson_size_str;
     PyObject* _max_message_size_str;
     PyObject* _max_write_batch_size_str;
-    PyObject* _max_split_size_str;
+    PyObject* _max_split_size_str;
```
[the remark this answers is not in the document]
aclark4life: Fixed in 20e1049d9026bb2e5da41fa02614fa94ce7db90e

blink1073:
Let's hold off on merging this until we address https://jira.mongodb.org/browse/PYTHON-4636

PR #1860, opened 2024-09-16
PYTHON-4585 Cursor.to_list does not apply client's timeoutMS setting

sleepyStick:


test/asynchronous/test_cursor.py
```diff
@@ -1444,6 +1455,21 @@ async def test_command_cursor_to_list_length(self):
         result = await db.test.aggregate([pipeline])
         self.assertEqual(len(await result.to_list(1)), 1)
 
+    @async_client_context.require_change_streams
+    @async_client_context.require_failCommand_fail_point
+    async def test_command_cursor_to_list_csot_applied(self):
```
ShaneHarvey, line 1460: 2 things:
- Let's use regular aggregate here since there's nothing special about $changeStream here and it complicates the test.
- The failCommands needs to be for "getMore", not "aggregate" since `await client.db.test.aggregate()` actually runs the aggregate command and to_list only runs getMore. We'll need to set batchSize to ensure at least 1 getMore even runs.

test/asynchronous/test_cursor.py
```diff
@@ -1415,6 +1417,15 @@ async def test_to_list_length(self):
...
         self.assertEqual(len(docs), 2)
 
+    async def test_to_list_csot_applied(self):
+        client = await async_single_client(timeoutMS=1000)
+        coll = client.pymongo.test
+        await coll.insert_many([{} for _ in range(5)])
+        cursor = coll.find({"$where": delay(2)})
+        with self.assertRaises(PyMongoError) as ctx:
+            await cursor.to_list()
+        self.assertTrue(ctx.exception.timeout)
```
ShaneHarvey, line 1427: nice.

ShaneHarvey (changes requested):

pymongo/asynchronous/cursor.py
```diff
@@ -294,6 +295,7 @@ def _clone(self, deepcopy: bool = True, base: Optional[AsyncCursor] = None) -> A
             "limit",
             "max_time_ms",
             "max_await_time_ms",
+            "timeout",
```
ShaneHarvey, line 298: We shouldn't need to add timeout here since it's already set implicitly in _clone_base.

blink1073:
@sleepyStick you'll need to resync from master and use the convenience class methods to create the clients.

pymongo/asynchronous/cursor.py
sleepyStick: hm i didn't at first but i was failing `test_clone` and adding this seemed to fix it >.< 
i'll undo this change in my next commit though in case i'm mis-remembering

ShaneHarvey (approved):
Nice!

PR #1861, opened 2024-09-17
PYTHON-4756 Add changelog note about dropping srv extra

blink1073:


doc/changelog.rst
```diff
@@ -93,6 +93,8 @@ Unavoidable breaking changes
 - Since we are now using ``hatch`` as our build backend, we no longer have a usable ``setup.py`` file
   and require installation using ``pip``.  Attempts to invoke the ``setup.py`` file will raise an exception.
   Additionally, ``pip`` >= 21.3 is now required for editable installs.
+- We no longer support the ``srv`` extra, since ``dnspython`` is included as a dependency in PyMongo 4.7+.
+  Instead of ``pip install pymongo[srv]``, use ``pip install pymongo``.
```
ShaneHarvey, line 97: Can you add the same for "tls"?

ShaneHarvey: The tls extra was only valid for Python 2 which we dropped in PYTHON-2133 pymongo 4.0.

blink1073: Updated

doc/changelog.rst
```diff
@@ -93,6 +93,10 @@ Unavoidable breaking changes
 - Since we are now using ``hatch`` as our build backend, we no longer have a usable ``setup.py`` file
   and require installation using ``pip``.  Attempts to invoke the ``setup.py`` file will raise an exception.
   Additionally, ``pip`` >= 21.3 is now required for editable installs.
+- We no longer support the ``srv`` or ``tls`` extras, since ``dnspython`` is included as a dependency in PyMongo 4.7+.
+  Instead of ``pip install pymongo[srv]`` or ``pip install pymongo[tls]``, use ``pip install pymongo``.
+- We no longer support the ``tls`` extra, which was only valid for Python 2.
+  Instead of ``pip install pymongo[tls]`` or ``pip install pymongo[tls]``, use ``pip install pymongo``.
```
ShaneHarvey, line 99: There's some duplication. How about:
```
- We no longer support the ``srv`` extra, since ``dnspython`` is included as a dependency in PyMongo 4.7+.
  Instead of ``pip install pymongo[srv]``, use ``pip install pymongo``.
- We no longer support the ``tls`` extra, which was only valid for Python 2.
  Instead of ``pip install pymongo[tls]``, use ``pip install pymongo``.
```

blink1073: Done

ShaneHarvey (approved):

PR #1862, opened 2024-09-17
PYTHON-4738 - Make test_encryption.TestClientSimple.test_fork sync-only

NoahStapp:
Asyncio and forking should not be mixed. I suspect the sync version was failing due to running after the async version caused a deadlock.


blink1073 (approved):
LGTM!

PR #1863, opened 2024-09-17
PYTHON-4737 Migrate test_binary.py to async

sleepyStick:
More like test_binary shouldn't have any client stuff in it....I've moved tests that require the client from `test_binary.py` to `test_client.py`


test/asynchronous/test_client.py
```diff
@@ -2020,6 +2023,110 @@ def test_dict_hints_sort(self):
     async def test_dict_hints_create_index(self):
         await self.db.t.create_index({"x": pymongo.ASCENDING})
 
+    @async_client_context.require_connection
+    async def test_legacy_java_uuid_roundtrip(self):
+        from_java = (
```
ShaneHarvey, line 2028: Any way we can share this blob with test_binary? Same for from_csharp.

ShaneHarvey (changes requested):

test/asynchronous/test_client.py
```diff
@@ -31,13 +32,15 @@
...
-from typing import Iterable, Type, no_type_check
+import uuid
+from typing import Any, Iterable, Type, no_type_check
 from unittest import mock
 from unittest.mock import patch
 
 import pytest
 import pytest_asyncio
 
+from bson.binary import *
```
ShaneHarvey, line 43: We try to avoid * imports for a few reasons: https://stackoverflow.com/questions/2386714/why-is-import-bad

test/asynchronous/test_client.py
```diff
@@ -57,6 +60,7 @@
     unittest,
 )
 from test.asynchronous.pymongo_mocks import AsyncMockClient
+from test.test_binary import TestBinary
```
ShaneHarvey, line 63: Importing TestBinary directly will actually cause pytest/unittest to pick up and run TestBinary in this file too. Instead we need to mangle to name, like: `from test.test_binary import TestBinary as BinaryBase` or move the test data out of TestBinary into to a global var that we can import.

test/asynchronous/test_client.py
```diff
@@ -2020,6 +2024,78 @@ def test_dict_hints_sort(self):
     async def test_dict_hints_create_index(self):
         await self.db.t.create_index({"x": pymongo.ASCENDING})
 
+    @async_client_context.require_connection
```
ShaneHarvey, line 2027: I beleive we can remove the `require_connection`s now that this is in test_client.

ShaneHarvey (changes requested):

sleepyStick:
My brain forgot to check up on this until now haha, but thanks for the feedback! I think I made the appropriate changes :)

ShaneHarvey (changes requested):
Screenshot 2024-09-24 at 12 13 11 PM
Hmm, my TestBinary as BinaryBase idea doesn't work. We'll have to do something else.

sleepyStick:
Ah good catch! I'll try something else!

ShaneHarvey (approved):
LGTM!

PR #1866, opened 2024-09-18
PYTHON-4770 Improve CPU overhead of async locks and latency on Windows TLS sendall

ShaneHarvey:
PYTHON-4770 Improve CPU overhead of async locks and latency on Windows TLS sendall

CPU usage baseline with the sync api:
```
$ time python bench-lock.py sync         
Python: 3.12.4, PyMongo: 4.9.0.dev0
python bench-lock.py sync  0.19s user 0.05s system 4% cpu 5.333 total
```

Before:
```
$ time python bench-lock.py async
Python: 3.12.4, PyMongo: 4.9.0.dev0
python bench-lock.py async  0.77s user 0.42s system 23% cpu 5.168 total
```

After this change (both user and system CPU are much lower):
```
$ time python bench-lock.py async
Python: 3.12.4, PyMongo: 4.9.0.dev0
python bench-lock.py async  0.34s user 0.09s system 8% cpu 5.196 total
```

### DIFF
--- a/pymongo/lock.py
+++ b/pymongo/lock.py
@@ -61,7 +61,7 @@ class _ALock:
                 return False
             if not blocking:
                 return False
-            await asyncio.sleep(0)
+            await asyncio.sleep(0.001)
 
     def release(self) -> None:
         self._lock.release()
@@ -96,7 +96,7 @@ class _ACondition:
                 return False
             if not blocking:
                 return False
-            await asyncio.sleep(0)
+            await asyncio.sleep(0.001)
 
     async def wait(self, timeout: Optional[float] = None) -> bool:
         if timeout is not None:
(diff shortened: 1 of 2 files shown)


blink1073:
There's another `asyncio.sleep(0)` in `asynchronous/network.py`.

ShaneHarvey:
I intentionally left the other sleep(0) because that code path is already blocking the I/O loop entirely: https://jira.mongodb.org/browse/PYTHON-4636

blink1073 (approved):
LGTM!

ShaneHarvey:
The loops that I changed here are different because each iteration performs a synchronous non-blocking call which turns the whole thing into a busy loop that eats CPU. wait_for_read is terrible in other ways but it doesn't have that CPU problem.

NoahStapp (approved):

ShaneHarvey:
It looks like this change makes the Windows test suite run a few minutes _slower_ so I did not merge it. Will need more investigation. Could be better to measure the Lock and sendall changes independently.

ShaneHarvey:
Closing because we already implemented the backoff in PYTHON-4636 and we're planning to replace these lock classes soon.

PR #1873, opened 2024-09-20
PYTHON-4781 Handle errors on Async PyMongo import

blink1073:


pymongo/__init__.py
```diff
@@ -105,6 +104,14 @@
 from pymongo.synchronous.mongo_client import MongoClient
 from pymongo.write_concern import WriteConcern
 
+try:
+    from pymongo.asynchronous.mongo_client import AsyncMongoClient
+except Exception as e:
+    # PYTHON-4781: Importing asyncio can fail on Windows.
+    import warnings
+
+    warnings.warn(f"Failed to import Async PyMongo: {e}", RuntimeWarning, stacklevel=2)
```
ShaneHarvey, line 113: 1) Should we use [ImportWarning](https://docs.python.org/3/library/exceptions.html#ImportWarning) instead? 
> ImportWarning: Base category for warnings triggered during the process of importing a module (ignored by default).

2) Suggest `{e!r}` to provide the repr which will include the exception class instead of just the message.

3) Suggest `import warnings as _warnings` to avoid polluting the pymongo namespace.

blink1073: Done

ShaneHarvey (approved):

ShaneHarvey:
Will this cause type errors now that AsyncMongoClient is not always defined?

blink1073:
> Will this cause type errors now that AsyncMongoClient is not always defined?

Perhaps on a system where importing asyncio is broken, but that system would have more concerning problems.

PR #1874, opened 2024-09-23
PYTHON-1714 Add c extension use to client metadata

aclark4life:
- Move `has_c` to common
- If `has_c` in pool_options update driver metadata
- If `has_c` in tests update driver metadata
- Add c replacement string to synchro

### DIFF
--- a/test/asynchronous/test_client.py
+++ b/test/asynchronous/test_client.py
@@ -95,7 +95,7 @@ from pymongo.asynchronous.pool import (
 from pymongo.asynchronous.settings import TOPOLOGY_TYPE
 from pymongo.asynchronous.topology import _ErrorContext
 from pymongo.client_options import ClientOptions
-from pymongo.common import _UUID_REPRESENTATIONS, CONNECT_TIMEOUT
+from pymongo.common import _UUID_REPRESENTATIONS, CONNECT_TIMEOUT, has_c
 from pymongo.compression_support import _have_snappy, _have_zstd
 from pymongo.driver_info import DriverInfo
 from pymongo.errors import (
@@ -343,7 +343,10 @@ class AsyncClientUnitTest(AsyncUnitTest):
 
     async def test_metadata(self):
         metadata = copy.deepcopy(_METADATA)
-        metadata["driver"]["name"] = "PyMongo|async"
+        if has_c():
+            metadata["driver"]["name"] = "PyMongo|c|async"
+        else:
+            metadata["driver"]["name"] = "PyMongo|async"
         metadata["application"] = {"name": "foobar"}
         client = self.simple_client("mongodb://foo:27017/?appname=foobar&connect=false")
         options = client.options
@@ -366,7 +369,10 @@ class AsyncClientUnitTest(AsyncUnitTest):
         with self.assertRaises(TypeError):
             self.simple_client(driver=("Foo", "1", "a"))
         # Test appending to driver info.
-        metadata["driver"]["name"] = "PyMongo|async|FooDriver"
+        if has_c():
+            metadata["driver"]["name"] = "PyMongo|c|async|FooDriver"
+        else:
+            metadata["driver"]["name"] = "PyMongo|async|FooDriver"
         metadata["driver"]["version"] = "{}|1.2.3".format(_METADATA["driver"]["version"])
         client = self.simple_client(
             "foo",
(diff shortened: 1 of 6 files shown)


test/asynchronous/test_client.py
```diff
@@ -341,71 +341,68 @@ async def test_read_preference(self):
...
+    async def test_metadata(self):
+        self._test_metadata({"driver": {"name": "async"},
+                             "application": {"name": "foobar"}},
+                            "mongodb://foo:27017/?appname=foobar&connect=false")
+        self._test_metadata({"driver": {"name": "async"}}, "foo", 27017,
+                            appname="foobar", connect=False)
+
         # No error
-        self.simple_client(appname="x" * 128)
+        self._test_metadata({}, appname="x" * 128)
```
ShaneHarvey, line 370: These simple tests seem better off using `simple_client` as before.

test/asynchronous/test_client.py
```diff
@@ -341,71 +341,68 @@ async def test_read_preference(self):
         )
         self.assertEqual(c.read_preference, ReadPreference.NEAREST)
 
-    async def test_metadata(self):
+
+    def _test_metadata(self, add_meta, *args, **kwargs):
```
ShaneHarvey, line 345: These tests don't actually verify the expected driver "name". They also no longer verify all the other expected metadata fields since we've removed this line:
`self.assertEqual(options.pool_options.metadata, metadata)`

I'd suggest we change this test back to how it was and deal with the `|c` issue like this:
```python
    async def test_metadata(self):
        if has_c():
            name = "PyMongo|c|async"
        else:
            name = "PyMongo|async"
        metadata = copy.deepcopy(_METADATA)
        metadata["driver"]["name"] = name
        metadata["application"] = {"name": "foobar"}
        client = self.simple_client("mongodb://foo:27017/?appname=foobar&connect=false")
        options = client.options
``` 

We'll need to add the `PyMongo|c|async` string to synchro.py too.

ShaneHarvey (changes requested):

test/asynchronous/test_client.py
aclark4life: > These tests don't actually verify the expected driver "name".

You mean you don't like the [`assertIn`](https://github.com/aclark4life/mongo-python-driver/blob/PYTHON-1714/test/test_client.py#L342) or … 

> They also no longer verify all the other expected metadata fields since we've removed this line:

I updated [`_test_handshake`](https://github.com/aclark4life/mongo-python-driver/blob/PYTHON-1714/test/test_client.py#L1890-L1904) to cover those.

ShaneHarvey: Right, the new test only checks for a substring while the old test checks the whole thing.

ShaneHarvey: Also the new _test_handshake is brittle since we need to update it every time we add a new field to the metadata. 

So the old way is better imo.

aclark4life: Done, thanks! Much simpler, I agree. That leaves:

```

FAILED test/asynchronous/test_client.py::AsyncClientUnitTest::test_metadata - AssertionError: {'dri[19 chars]ongo|c|async|FooDriver', 'version': '4.10.0.de[167 chars]ar'}} != {'dri[19 chars]ongo|async|FooDriver', 'version': '4.10.0....
FAILED test/test_client.py::ClientUnitTest::test_metadata - AssertionError: {'dri[19 chars]ongo|c|FooDriver', 'version': '4.10.0.dev0|1.2[161 chars]ar'}} != {'dri[19 chars]ongo|FooDriver', 'version': '4.10.0.dev0|1...
```

So I'll add the same conditional for the `FooDriver` test.

aclark4life: > We'll need to add the `PyMongo|c|async` string to synchro.py too.

Shouldn't we see a failure since it's not added there yet?

ShaneHarvey: Hmm I suppose something is already translating `PyMongo|c|async` to `PyMongo|c`? Perhaps unasync does that already?

ShaneHarvey: I was expecting to add a line here: https://github.com/mongodb/mongo-python-driver/blob/d0772f2/tools/synchro.py#L103

aclark4life: Yeah I don't understand why the tests pass with or without that string, maybe @blink1073 knows.

blink1073: Synchro hadn't been run, it failed the static check: https://github.com/mongodb/mongo-python-driver/actions/runs/11023004771/job/30613463481?pr=1874#step:5:67

ShaneHarvey (approved):

PR #1870, opened 2024-09-19
PYTHON-4636 - Avoid blocking I/O calls in async code paths

NoahStapp:

PR #1875, opened 2024-09-23
PYTHON-4782 Fix deadlock and blocking behavior in _ACondition.wait

ShaneHarvey:
PYTHON-4782 Fix deadlock and blocking behavior in _ACondition.wait

TODO:

- [x] Add unittests for _ACondition.
- [x] Add integration tests for blocking on client condition vars (server selection, maxConnecting, maxPoolSize). Edit: I'm deferring this to https://jira.mongodb.org/browse/PYTHON-4784

### DIFF
--- a/pymongo/lock.py
+++ b/pymongo/lock.py
@@ -46,6 +46,11 @@ def _release_locks() -> None:
             lock.release()
 
 
+# TODO: remove this.
+def _Lock(lock: threading.Lock) -> threading.Lock:
+    return lock
+
+
 class _ALock:
     __slots__ = ("_lock",)
 
--- a/pymongo/lock.py
+++ b/pymongo/lock.py
@@ -19,12 +19,14 @@ import os
 import threading
 import time
 import weakref
-from typing import Any, Optional
+from typing import Any, Callable, Optional, TypeVar
 
 _HAS_REGISTER_AT_FORK = hasattr(os, "register_at_fork")
 
 # References to instances of _create_lock
 _forkable_locks: weakref.WeakSet[threading.Lock] = weakref.WeakSet()
+
+_T = TypeVar("_T")
 
 
 def _create_lock() -> threading.Lock:
@@ -163,6 +165,20 @@ class _ACondition:
             self.notify(1)
             raise
 
+    async def wait_for(self, predicate: Callable[[], _T]) -> _T:
+        """Wait until a predicate becomes true.
+
+        The predicate should be a callable whose result will be
+        interpreted as a boolean value.  The method will repeatedly
+        wait() until it evaluates to true.  The final predicate value is
+        the return value.
+        """
+        result = predicate()
+        while not result:
+            await self.wait()
+            result = predicate()
+        return result
+
     def notify(self, n: int = 1) -> None:
         """By default, wake up one coroutine waiting on this condition, if any.
         If the calling coroutine has not acquired the lock when this method
@@ -203,6 +219,10 @@ class _ACondition:
         a RuntimeError is raised.
         """
         self.notify(len(self._waiters))
+
+    def locked(self) -> bool:
+        """Only needed for tests in test_locks."""
+        return self._condition._lock.locked()  # type: ignore[attr-defined]
 
     def release(self) -> None:
         self._condition.release()
--- a/pymongo/lock.py
+++ b/pymongo/lock.py
@@ -84,12 +84,18 @@ class _ALock:
         self.release()
 
 
+def _safe_set_result(fut: asyncio.Future) -> None:
+    # Ensure the future hasn't been cancelled before calling set_result.
+    if not fut.done():
+        fut.set_result(False)
+
+
 class _ACondition:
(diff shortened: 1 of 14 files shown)


ShaneHarvey:
Requesting review now, I'll keep investigating the test failures.

pymongo/lock.py
```diff
@@ -99,30 +113,116 @@ async def acquire(self, blocking: bool = True, timeout: float = -1) -> bool:
...
+        fut = loop.create_future()
+        self._waiters.append((loop, fut))
+        self.release()
+        try:
+            try:
+                try:
+                    await asyncio.wait_for(fut, timeout)
+                    return True
+                except asyncio.TimeoutError:
+                    return False  # Return false on timeout for sync pool compat.
```
NoahStapp, line 140: Do we still need to acquire the lock if we timeout here?

pymongo/lock.py
```diff
@@ -99,30 +113,116 @@ async def acquire(self, blocking: bool = True, timeout: float = -1) -> bool:
...
+        awakened, it re-acquires the lock and returns True.
+
+        This method may return spuriously,
+        which is why the caller should always
+        re-check the state and be prepared to wait() again.
+        """
+        loop = asyncio.get_running_loop()
+        fut = loop.create_future()
+        self._waiters.append((loop, fut))
+        self.release()
```
NoahStapp, line 133: Is this just to ensure we don't hold the lock while waiting for it?

pymongo/lock.py
```diff
@@ -99,30 +113,116 @@ async def acquire(self, blocking: bool = True, timeout: float = -1) -> bool:
...
+                except asyncio.TimeoutError:
+                    return False  # Return false on timeout for sync pool compat.
+            finally:
+                # Must re-acquire lock even if wait is cancelled.
+                # We only catch CancelledError here, since we don't want any
+                # other (fatal) errors with the future to cause us to spin.
+                err = None
+                while True:
+                    try:
+                        await self.acquire()
```
NoahStapp, line 148: Can this possibly loop forever?

NoahStapp (changes requested):

pymongo/lock.py
ShaneHarvey: Yes but only if something else holds the lock forever.

pymongo/lock.py
ShaneHarvey: Yes, the API contract for wait() says you MUST hold the lock before calling and you MUST still hold the lock when it returns, even on timeout.

pymongo/lock.py
ShaneHarvey: Yes without releasing the lock, this code would deadlock since nothing would be able to notify the waiter.

pymongo/lock.py
NoahStapp: Got it, so you always have to acquire the lock if you call wait and don't raise an error?

pymongo/lock.py
NoahStapp: And we're asserting that this new code can't do that, got it!

NoahStapp (dismissed):
Code looks good, approved contingent on tests!

pymongo/lock.py
NoahStapp: Can you add a comment linking to the API contract here? It would be good to make it more understandable for readers unfamiliar with it.

ShaneHarvey: There's already this comment:
```
# Must re-acquire lock even if wait is cancelled.
```

blink1073:
The failures seem relevant: `test.asynchronous.test_cursor.TestCursor.test_to_list_csot_applied`

blink1073:
It is the same failure across several tasks that all previously passed.

ShaneHarvey:
Yes I'm looking into test_to_list_csot_applied today.

ShaneHarvey:
I'm running a patch with both #1870 and this PR to see if the failures will work themselves out: https://spruce.mongodb.com/version/66f712658ea18700077002df/tasks

ShaneHarvey:
Bad timing for docker to start failing...
```
  ERROR: failed to solve: docker:stable: failed to resolve source metadata for docker.io/library/docker:stable: failed to authorize: failed to fetch oauth token: unexpected status from GET request to https://auth.docker.io/token?scope=repository%3Alibrary%2Fdocker%3Apull&service=registry.docker.io: 401 Unauthorized
```

ShaneHarvey:
New patch with both #1870 and this PR: https://spruce.mongodb.com/version/66f73f178ea187000770b298/

ShaneHarvey:
This is ready for another look.

NoahStapp (commented):
Increased timeouts make sense, still seeing test failures--do you think they're unrelated?

blink1073:
The Windows failures seem relevant, since they're in `test.asynchronous.test_locks.TestConditionStdlib`.

ShaneHarvey:
I fixed the remaining test failures.

blink1073 (approved):
LGTM!

PR #1878, opened 2024-09-27
PYTHON-4786 - Fix UpdateResult.did_upsert TypeError

NoahStapp:


blink1073 (dismissed):
LGTM!

pymongo/results.py
```diff
@@ -171,9 +171,10 @@ def upserted_id(self) -> Any:
 
     @property
     def did_upsert(self) -> bool:
-        """Whether or not an upsert took place."""
+        """Whether an upsert took place."""
         assert self.__raw_result is not None
-        return len(self.__raw_result.get("upserted", {})) > 0
+        result = self.__raw_result.get("upserted")
+        return result is not None and len(str(result)) > 0
```
ShaneHarvey, line 177: This doesn't handle the `None` case correctly.

ShaneHarvey: We need a few more tests. For both client.bulk_write and collection.update_one() that tests upserting a document with `"_id": None`.

NoahStapp: Do we expect something besides `False` if the result is `None`?

ShaneHarvey: yes None is a valid id

ShaneHarvey:
For example:
```python
>>> res = client.t.t.update_one({"_id": None}, {'$set': {'x': 1}}, upsert=True)
>>> res
UpdateResult({'n': 1, 'upserted': None, 'nModified': 0, 'ok': 1.0, 'updatedExisting': False}, acknowledged=True)
>>> res.upserted_id
>>> res.upserted_id is None
True
```

NoahStapp:
> For example:
> 
> ```python
> >>> res = client.t.t.update_one({"_id": None}, {'$set': {'x': 1}}, upsert=True)
> >>> res
> UpdateResult({'n': 1, 'upserted': None, 'nModified': 0, 'ok': 1.0, 'updatedExisting': False}, acknowledged=True)
> >>> res.upserted_id
> >>> res.upserted_id is None
> True
> >>> res.did_upsert
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/Users/shane/git/mongo-python-driver/pymongo/results.py", line 176, in did_upsert
>     return len(self.__raw_result.get("upserted", {})) > 0
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> TypeError: object of type 'NoneType' has no len()
> ```

I can't reproduce this locally, I get this result instead:

```python
>>> client = MongoClient()
>>> res = client.t.t.update_one({"_id": None}, {'$set': {'x': 1}}, upsert=True)
>>> res
UpdateResult({'n': 1, 'nModified': 0, 'ok': 1.0, 'updatedExisting': True}, acknowledged=True)
>>> res.upserted_id
>>> res.upserted_id is None
True
>>> res.did_upsert
False
```

ShaneHarvey:
The problem with your repro attempt is that it did not perform an upsert (see `'updatedExisting': True`) which means your collection already had a document with `_id:None`.

NoahStapp:
> The problem with your repro attempt is that it did not perform an upsert (see `'updatedExisting': True`) which means your collection already had a document with `_id:None`.

Good catch--I tried again with a fresh DB and successfully reproduced an incorrect result without an error:

```python
>>> client = MongoClient()
>>> res = client.t.t.update_one({"_id": None}, {'$set': {'x': 1}}, upsert=True)
>>> res
UpdateResult({'n': 1, 'upserted': None, 'nModified': 0, 'ok': 1.0, 'updatedExisting': False}, acknowledged=True)
>>> res.did_upsert
False
```

I'll add a test to catch this case with a fix.

test/test_results.py
```diff
@@ -122,6 +122,28 @@ def test_update_result(self):
...
+        self.assertEqual(result.did_upsert, True)
+
+        raw_result_3 = {
+            "n": 1,
+            "nModified": 1,
+        }
+        self.repr_test(UpdateResult, raw_result_3)
+
+        result = UpdateResult(raw_result_3, True)
+        self.assertEqual(result.did_upsert, False)
```
ShaneHarvey, line 146: This is good but we need integration tests for these cases too.

ShaneHarvey (changes requested):
Please add similar tests for client.bulk_write with verbose=True and coll.update_one.

blink1073:
drivers-pr-bot please backport to v4.9

blink1073 (dismissed):
LGTM!

test/test_client_bulk_write.py
```diff
@@ -550,6 +550,47 @@ def test_returns_error_if_auto_encryption_configured(self):
             "bulk_write does not currently support automatic encryption", context.exception._message
         )
 
+    @client_context.require_version_min(8, 0, 0, -24)
+    @client_context.require_no_serverless
+    @unittest.skipUnless(_HAVE_PYMONGOCRYPT, "pymongocrypt is not installed")
```
ShaneHarvey, line 555: Why require `_HAVE_PYMONGOCRYPT`?

NoahStapp: Whoops, artifact of copy-and-paste.

pymongo/results.py
```diff
@@ -171,9 +171,9 @@ def upserted_id(self) -> Any:
 
     @property
     def did_upsert(self) -> bool:
-        """Whether or not an upsert took place."""
+        """Whether an upsert took place."""
```
ShaneHarvey, line 174: Please add `versionadded: 4.9`

ShaneHarvey (approved):
LGTM once the tests pass.

blink1073:
drivers-pr-bot please backport to v4.9

PR #1880, opened 2024-09-30
PYTHON-4796 Update type checkers and handle with_options typing

blink1073:


caseyclements (approved):
LGTM.

I didn't catch the conversation in the meeting, but this pretty sweet. It comes with caveats?

blink1073:
>  It comes with caveats?

I split this off from [PYTHON-3992](https://jira.mongodb.org/browse/PYTHON-3992) and marked it as 5.0, because that change had some unexpected affects.

PR #1908, opened 2024-10-09
PYTHON-4574 - FaaS detection logic mistakenly identifies EKS as AWS Lambda

NoahStapp:


ShaneHarvey:
What about cases where a container is deployed on AWS Lambda? https://docs.aws.amazon.com/lambda/latest/dg/images-create.html

NoahStapp:
> What about cases where a container is deployed on AWS Lambda? [docs.aws.amazon.com/lambda/latest/dg/images-create.html](https://docs.aws.amazon.com/lambda/latest/dg/images-create.html)

If the correct environment variables are present for both AWS Lambda and a container,  we'd need to decide what the correct result is.

ShaneHarvey:
Let's move this convo to the drivers ticket/PR.

blink1073 (approved):
LGTM!

caseyclements (approved):
LGTM

PR #1909, opened 2024-10-09
PYTHON-4834 Add __repr__ to IndexModel, SearchIndexModel

aclark4life:
For use with [django-mongodb](https://github.com/mongodb-labs/django-mongodb) e.g.

```

(.venv) alex.clark@Mac tmp.0F48BsWTyQ % python manage.py sqlmigrate auth 0001
System check identified some issues:

WARNINGS:
?: (staticfiles.W004) The directory '/private/var/folders/5t/d3llmfm54v3dg712_4pks3lm0000gn/T/tmp.0F48BsWTyQ/frontend/build' in the STATICFILES_DIRS setting does not exist.
--
-- Create model Permission
--
db.create_collection('auth_permission')
db.auth_permission.create_indexes([IndexModel({'name': 'auth_permission_content_type_id_2f476e4b', 'key': {'content_type_id': 1}})])
--
-- Create model Group
--
db.create_collection('auth_group')
db.create_collection('auth_group_permissions')
db.auth_group_permissions.create_indexes([IndexModel({'name': 'auth_group_permissions_group_id_b120cbf9', 'key': {'group_id': 1}})])
db.auth_group_permissions.create_indexes([IndexModel({'name': 'auth_group_permissions_permission_id_84c5c92e', 'key': {'permission_id': 1}})])
--
-- Create model User
--
db.create_collection('auth_user')
db.create_collection('auth_user_groups')
db.auth_user_groups.create_indexes([IndexModel({'name': 'auth_user_groups_user_id_6a12ed8b', 'key': {'user_id': 1}})])
db.auth_user_groups.create_indexes([IndexModel({'name': 'auth_user_groups_group_id_97559544', 'key': {'group_id': 1}})])
db.create_collection('auth_user_user_permissions')
db.auth_user_user_permissions.create_indexes([IndexModel({'name': 'auth_user_user_permissions_user_id_a95ead1b', 'key': {'user_id': 1}})])
db.auth_user_user_permissions.create_indexes([IndexModel({'name': 'auth_user_user_permissions_permission_id_1fbb5f2c', 'key': {'permission_id': 1}})])
```

### DIFF
--- a/pymongo/operations.py
+++ b/pymongo/operations.py
@@ -773,6 +773,13 @@ class IndexModel:
         """
         return self.__document
 
+    def __repr__(self) -> str:
+        return "{}({}{})".format(
+            self.__class__.__name__,
+            self.document["key"],
+            "".join([f", {key}={value!r}" for key, value in self.document.items() if key != "key"]),
+        )
+
 
 class SearchIndexModel:
     """Represents a search index to create."""
@@ -812,3 +819,9 @@ class SearchIndexModel:
     def document(self) -> Mapping[str, Any]:
         """The document for this index."""
         return self.__document
+
+    def __repr__(self) -> str:
+        return "{}({})".format(
+            self.__class__.__name__,
+            ", ".join([f"{key}={value!r}" for key, value in self.document.items()]),
+        )
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1,0 +1,76 @@
+# Copyright 2009-present MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Test the operations module."""
+from __future__ import annotations
+
+from test import UnitTest, unittest
+
+from pymongo import ASCENDING, DESCENDING
+from pymongo.collation import Collation
+from pymongo.errors import OperationFailure
+from pymongo.operations import IndexModel, SearchIndexModel
+
+
+class TestOperationsBase(UnitTest):
+    """Base class for testing operations module."""
+
+    def assertRepr(self, obj):
+        new_obj = eval(repr(obj))
+        self.assertEqual(type(new_obj), type(obj))
+        self.assertEqual(repr(new_obj), repr(obj))
+
+
+class TestIndexModel(TestOperationsBase):
+    """Test IndexModel features."""
+
+    def test_repr(self):
+        # Based on examples in test_collection.py
+        self.assertRepr(IndexModel("hello"))
+        self.assertRepr(IndexModel([("hello", DESCENDING), ("world", ASCENDING)]))
+        self.assertRepr(
+            IndexModel([("hello", DESCENDING), ("world", ASCENDING)], name="hello_world")
+        )
+        # Test all the kwargs
+        self.assertRepr(IndexModel("name", name="name"))
+        self.assertRepr(IndexModel("unique", unique=False))
+        self.assertRepr(IndexModel("background", background=True))
+        self.assertRepr(IndexModel("sparse", sparse=True))
+        self.assertRepr(IndexModel("bucketSize", bucketSize=1))
+        self.assertRepr(IndexModel("min", min=1))
+        self.assertRepr(IndexModel("max", max=1))
+        self.assertRepr(IndexModel("expireAfterSeconds", expireAfterSeconds=1))
+        self.assertRepr(
+            IndexModel("partialFilterExpression", partialFilterExpression={"hello": "world"})
+        )
+        self.assertRepr(IndexModel("collation", collation=Collation(locale="en_US")))
+        self.assertRepr(IndexModel("wildcardProjection", wildcardProjection={"$**": 1}))
+        self.assertRepr(IndexModel("hidden", hidden=False))
+        # Test string literal
+        self.assertEqual(repr(IndexModel("hello")), "IndexModel({'hello': 1}, name='hello_1')")
+        self.assertEqual(
+            repr(IndexModel({"hello": 1, "world": -1})),
+            "IndexModel({'hello': 1, 'world': -1}, name='hello_1_world_-1')",
+        )
+
+
+class TestSearchIndexModel(TestOperationsBase):
+    """Test SearchIndexModel features."""
+
+    def test_repr(self):
+        self.assertRepr(SearchIndexModel("hello", key=1))
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1,4 +1,4 @@
-# Copyright 2009-present MongoDB, Inc.
+# Copyright 2009-2024 MongoDB, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -70,6 +70,10 @@ class TestSearchIndexModel(TestOperationsBase):
 
     def test_repr(self):
         self.assertRepr(SearchIndexModel({"hello": "hello"}, key=1))
+        self.assertEqual(
+            repr(SearchIndexModel({"hello": "hello"}, key=1)),
+            "SearchIndexModel(definition={'hello': 'hello'}, key=1)",
+        )
 
 
 if __name__ == "__main__":
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -69,7 +69,7 @@ class TestSearchIndexModel(TestOperationsBase):
     """Test SearchIndexModel features."""
 
     def test_repr(self):
-        self.assertRepr(SearchIndexModel("hello", key=1))
+        self.assertRepr(SearchIndexModel({"hello": "hello"}, key=1))
 
 
 if __name__ == "__main__":
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1,4 +1,4 @@
-# Copyright 2009-2024 MongoDB, Inc.
+# Copyright 2024-present MongoDB, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -17,6 +17,8 @@ PyMongo 4.11 brings a number of changes including:
 - :attr:`~pymongo.asynchronous.mongo_client.AsyncMongoClient.address` and
   :attr:`~pymongo.mongo_client.MongoClient.address` now correctly block when called on unconnected clients
   until either connection succeeds or a server selection timeout error is raised.
+- Added :func:`repr` support to :class:`pymongo.operations.IndexModel`.
+- Added :func:`repr` support to :class:`pymongo.operations.SearchIndexModel`.
 
 Issues Resolved
 ...............


test/test_collection.py
```diff
@@ -2253,6 +2253,10 @@ def afind(*args, **kwargs):
         for helper, args in helpers:
             helper(*args, let={})  # type: ignore
 
+    def test_repr(self):
+        im = IndexModel("hello")
+        self.assertEqual(im.__repr__(), f"{im.__class__.__name__}({im.document})")
```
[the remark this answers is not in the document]
aclark4life: > 1. Use `repr(im)` and use the actual string that's generated rather than an f-string.


Right, so this:

```
    def test_repr(self):    
        im = IndexModel("hello")
        self.assertEqual(repr(im), "IndexModel({'name': 'hello_1', 'key': {'hello': 1}})")
```

> 2. I didn't examine the organization of tests but the docstring for this file says, "Test the collection module." which may not be the right spot for a class that lives in `pymongo.operations`.


`test_index_management` was closer but only tested `SearchIndexModel` so I'm not sure either.


> 3. From [the docs for `repr()`](https://docs.python.org/3/library/functions.html#repr): "For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to [eval()](https://docs.python.org/3/library/functions.html#eval); otherwise, the representation is a string enclosed in angle brackets that contains the name of the type of the object together with additional information often including the name and address of the object."
> 
> Could you try to make it work as the docs suggest? Current (not working) behavior:
> 
> ```
> >>> a = IndexModel("name", name="foo", unique=True)
> >>> a.document
> {'name': 'foo', 'unique': True, 'key': {'name': 1}}
> >>> b = IndexModel(str(a.document))
> >>> b.document
> {'name': "{'name': 'foo', 'unique': True, 'key': {'name': 1}}_1", 'key': {"{'name': 'foo', 'unique': True, 'key': {'name': 1}}": 1}}
> ```


Sure, but why are you calling `.document` ? Shouldn't we test behavior with something (still broken) like this? 


```

>>> from pymongo.operations import IndexModel
>>> a = IndexModel("name", name="foo", unique=True)
IndexModel({'name': 'foo', 'unique': True, 'key': {'name': 1}})
>>> b = eval(repr(a))
IndexModel({'name': "name_foo_unique_True_key_{'name': 1}", 'key': {'name': 'foo', 'unique': True, 'key': {'name': 1}}})
```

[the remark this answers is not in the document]
ShaneHarvey: In other places we use this method to confirm that repr works correctly:
```python
    def assertRepr(self, obj):
        new_obj = eval(repr(obj))
        self.assertEqual(type(new_obj), type(obj))
        self.assertEqual(repr(new_obj), repr(obj))
```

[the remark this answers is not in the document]
timgraham: re: 3. My code was meant to replicate the current implementation and show that it doesn't work.

timgraham:
There are quite a lot of [possible IndexModel kwargs](https://pymongo.readthedocs.io/en/stable/api/pymongo/operations.html#pymongo.operations.IndexModel) besides the ones hardcoded here. Looking at [`IndexModel.__init__()`](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/operations.py#L760-L767)), it appears that besides the key/keys rename, all other document keys can be be used unaltered in `__repr__()`.

aclark4life:
> There are quite a lot of [possible IndexModel kwargs](https://pymongo.readthedocs.io/en/stable/api/pymongo/operations.html#pymongo.operations.IndexModel) besides the ones hardcoded here. Looking at [`IndexModel.__init__()`](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/operations.py#L760-L767)), it appears that besides the key/keys rename, all other document keys can be be used unaltered in `__repr__()`.

Thanks, fixed. I'm going to add some tests for the other known kwargs.

pymongo/operations.py
```diff
@@ -773,6 +773,19 @@ def document(self) -> dict[str, Any]:
         """
         return self.__document
 
+    def __repr__(self) -> str:
+        return "%s(%s%s)" % (
+            self.__class__.__name__,                                               
+            "" if not self.document.get("key") else "keys=%s" % self.document["key"],
```
ShaneHarvey, line 779: "key" always exists so we can remove that check. eg:
```
        return "%s(%s%s)" % (
            self.__class__.__name__,                                               
            self.document["key"],
            "".join(...)
```

test/test_collection.py
```diff
@@ -2253,6 +2253,29 @@ def afind(*args, **kwargs):
         for helper, args in helpers:
             helper(*args, let={})  # type: ignore
 
+    def assertRepr(self, obj):
+        new_obj = eval(repr(obj))
+        self.assertEqual(type(new_obj), type(obj))
+        self.assertEqual(repr(new_obj), repr(obj))
+
+    def test_index_model_repr(self):
+        self.assertRepr(IndexModel("hello"))
```
ShaneHarvey, line 2262: We should also have at least one assert with a literal string, eg:
```"IndexModel({'hello': 1, 'world': -1}, name='hello_world')"```

That way we'll know if the format changes by accident.

File: pymongo/operations.py (hunk 2 above)
Comment (aclark4life) [reply pymongo/operations.py:779]: How about 

```
            self.__class__.__name__,                                      
            "keys=%s" % self.document["key"],                                                  
            "".join(...)

```

So that `IndexError` gets `keys={'hello': 'world'}` instead of `{'hello': 'world'}`

?

File: test/test_collection.py (hunk 3 above)
Comment (aclark4life) [reply test/test_collection.py:2262]: https://github.com/mongodb/mongo-python-driver/pull/1909/commits/f4b255b42684e2436a4d9db0a7482521228f45e8#diff-97ad6fe624c2aee90c37ee01ad11452de5c87847a24567b0cab10c41824bf22fR2282

File: pymongo/operations.py (hunk 2 above)
Comment (ShaneHarvey) [reply pymongo/operations.py]: I prefer passing the first argument without `keys=` because we never use pass `keys=` in any doc examples.

File: test/test_collection.py (hunk 3 above)
Comment (ShaneHarvey) [reply test/test_collection.py:2262]: We need to compare the string literal value:

`self.assertEqual(repr(IndexModel(...)), "IndexModel({'hello': 1, 'world': -1}, name='hello_world')")`

File: pymongo/operations.py (hunk 2 above)
Comment (aclark4life) [reply pymongo/operations.py]: Copy that, thanks https://github.com/mongodb/mongo-python-driver/pull/1909/commits/569b706a26590f073ef50af9cd074a5576e20e43

**Comment by ShaneHarvey** [comment]
Ruff wants you to use f-string of .format() instead:
```
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1
- files were modified by this hook

pymongo/operations.py:777:16: UP031 Use format specifiers instead of percent format
pymongo/operations.py:782:21: UP031 Use format specifiers instead of percent format

Fixed 1 error:
- pymongo/operations.py:
    1 × SIM201 (negate-equal-op)

Found 3 errors (1 fixed, 2 remaining).
```

https://github.com/mongodb/mongo-python-driver/actions/runs/11335774791/job/31524560806?pr=1909#step:5:71

File: test/test_collection.py (hunk 3 above)
Comment (aclark4life) [reply test/test_collection.py:2262]: https://github.com/mongodb/mongo-python-driver/pull/1909/commits/09ef323e6c98d862e13a5f4a8331051880b2b380#diff-97ad6fe624c2aee90c37ee01ad11452de5c87847a24567b0cab10c41824bf22fR2282-R2283

**Comment by aclark4life** [comment]
> Ruff wants you to use f-string or .format() instead:
> 
> ```
> ruff.....................................................................Failed
> - hook id: ruff
> - exit code: 1
> - files were modified by this hook
> 
> pymongo/operations.py:777:16: UP031 Use format specifiers instead of percent format
> pymongo/operations.py:782:21: UP031 Use format specifiers instead of percent format
> 
> Fixed 1 error:
> - pymongo/operations.py:
>     1 × SIM201 (negate-equal-op)
> 
> Found 3 errors (1 fixed, 2 remaining).
> ```
> 
> https://github.com/mongodb/mongo-python-driver/actions/runs/11335774791/job/31524560806?pr=1909#step:5:71

Fair enough, ruff 👍 https://github.com/mongodb/mongo-python-driver/pull/1909/commits/ca3432219824bb125f2056a73fbc171b932c20c2

**Comment by aclark4life** [comment]
Thanks @timgraham and @ShaneHarvey, I think this is ready to go

File: pymongo/operations.py [hunk 4]
```diff
@@ -773,6 +773,19 @@ def document(self) -> dict[str, Any]:
...
 
+    def __repr__(self) -> str:
+        return "{}({}{})".format(
+            self.__class__.__name__,
+            self.document["key"],
+            "".join(
+                [
+                    ", {}={!r}".format(key, value)
+                    for key, value in self.document.items()
+                    if not key == "key"
```
timgraham, line 784: I'd use `!=` rather than `not ... ==`.

test/test_collection.py
```diff
@@ -2253,6 +2253,35 @@ def afind(*args, **kwargs):
...
+        self.assertRepr(IndexModel("unique", unique=False))
+        self.assertRepr(IndexModel("background", background=True))
+        self.assertRepr(IndexModel("sparse", sparse=True))
+        self.assertRepr(IndexModel("bucketSize", bucketSize=1))
+        self.assertRepr(IndexModel("min", min=1))
+        self.assertRepr(IndexModel("max", max=1))
+        self.assertRepr(IndexModel("expireAfterSeconds", expireAfterSeconds=1))
+        self.assertRepr(IndexModel("partialFilterExpression", partialFilterExpression={"hello": "world"}))
+
+        from pymongo.collation import Collation
```
timgraham, line 2276: Put import at top of file?

test/test_collection.py
```diff
@@ -2253,6 +2253,35 @@ def afind(*args, **kwargs):
...
+    def assertRepr(self, obj):
+        new_obj = eval(repr(obj))
+        self.assertEqual(type(new_obj), type(obj))
+        self.assertEqual(repr(new_obj), repr(obj))
+
+    def test_index_model_repr(self):
+        self.assertRepr(IndexModel("hello"))
+        self.assertRepr(IndexModel([("hello", DESCENDING), ("world", ASCENDING)]))
+        self.assertRepr(IndexModel([("hello", DESCENDING), ("world", ASCENDING)], name="hello_world"))
+
```
timgraham, line 2265: If there is some grouping to the tests such that a blank line is warranted, I'd use a comment to explain rather than a blank line (or even separate tests).

test/test_collection.py
```diff
@@ -2253,6 +2253,35 @@ def afind(*args, **kwargs):
         for helper, args in helpers:
             helper(*args, let={})  # type: ignore
 
+    def assertRepr(self, obj):
+        new_obj = eval(repr(obj))
+        self.assertEqual(type(new_obj), type(obj))
+        self.assertEqual(repr(new_obj), repr(obj))
+
+    def test_index_model_repr(self):
```
timgraham, line 2261: Probably this should at least be organized in a separate test class. Currently it lives in a subclass of `IntegrationTest` which seems unnecessary since it doesn't require database connectivity. As I said before, it seems like a new file `test_operations.py` might be appropriate. Just looking at `operations.py`, I see `SearchIndexModel` which could likely also benefit from `__repr__()` in the future (not sure if Django will use it), but that would be more tests that could live in such a file..

timgraham (commented):
LGTM besides some bikeshedding I won't speak definitively about.

aclark4life:
> LGTM besides some bikeshedding I won't speak definitively about.

Thanks! May as well bikeshed while we are in here. I have implemented your suggestions as well as moved `assertRepr` to the base test class.

test/__init__.py
```diff
@@ -69,6 +69,7 @@
 import pymongo
 import pymongo.errors
 from bson.son import SON
+from pymongo import IndexModel  # noqa
```
timgraham, line 72: Needed why?

test/test_operations.py
```diff
@@ -0,0 +1,66 @@
...
+"""Test the operations module."""
+
+from test import UnitTest
+from pymongo import ASCENDING, DESCENDING
+from pymongo.operations import IndexModel
+from pymongo.collation import Collation
+
+
+class TestOperations(UnitTest):
+    """Test Operations module including methods:
```
timgraham, line 24: "operations" ... no caps... (i.e. pymong.operations, right?)
"methods of"?

test/test_operations.py
```diff
@@ -0,0 +1,66 @@
...
+    SearchIndexModel"""
+
+    def test_index_model_repr(self):
+        # Based on examples in test_collection.py
+        self.assertRepr(IndexModel("hello"))
+        self.assertRepr(IndexModel([("hello", DESCENDING), ("world", ASCENDING)]))
+        self.assertRepr(
+            IndexModel([("hello", DESCENDING), ("world", ASCENDING)], name="hello_world")
+        )
+
```
timgraham, line 40: I meant the comments could replace the blank lines, just to keep things more compact.

test/test_operations.py
```diff
@@ -0,0 +1,66 @@
...
+from test import UnitTest
+from pymongo import ASCENDING, DESCENDING
+from pymongo.operations import IndexModel
+from pymongo.collation import Collation
+
+
+class TestOperations(UnitTest):
+    """Test Operations module including methods:
+    _Op, _DeleteOp, _UpdateOp,
```
timgraham, line 25: Not sure all this needs need be stated, especially when there are yet to be tests.. Frankly, I would just have a separate test class for each class. i.e. `TestIndexModel` with test method `test_repr`.

test/__init__.py
aclark4life: `self.assertRepr` test fails without it.

aclark4life: I haven't check yet but probably the calls to `self.assertRepr` from `test_connection…` will fail to without importing whatever type `repr` tries to use. So, we don't duplicate that code but we do have to import types being tested.

aclark4life: And… too many imports… duplicating `assertRepr` code for now.

timgraham: Right... `eval()` needs it. If `assertRepr` lived in the base class, I think you could put the "noqa" imports inside the method. I don't have a strong opinion about whether that's cleaner than duplicating the method.

test/test_operations.py
```diff
@@ -0,0 +1,61 @@
...
+"""Test the operations module."""
+
+from test import UnitTest
+from pymongo import ASCENDING, DESCENDING
+from pymongo.operations import IndexModel
+from pymongo.collation import Collation
+
+
+class TestOperations(UnitTest):
+    """Test the operations module."""
```
timgraham, line 24: Do you think the idea for a separate test class for each class is no good? It looks funny to have `"""Test the operations module."""` at the module-level and then to repeat it here. I would rather use the names `TestIndexModel` and `test_repr`, otherwise I think this is fine.

timgraham (approved):

pymongo/operations.py
```diff
@@ -773,6 +773,19 @@ def document(self) -> dict[str, Any]:
...
+                [
+                    ", {}={!r}".format(key, value)
+                    for key, value in self.document.items()
+                    if key != "key"
+                ]
+            ),
+        )
+
 
 class SearchIndexModel:
```
ShaneHarvey, line 790: Could we add the same repr+tests for SearchIndexModel?

ShaneHarvey (commented):
Could you add a note in the doc/changelog.rst? Here's an example:
```
- Added :func:`repr` support to :class:`bson.tz_util.FixedOffset`.
```

pymongo/operations.py
```diff
@@ -773,6 +773,19 @@ def document(self) -> dict[str, Any]:
         """
         return self.__document
 
+    def __repr__(self) -> str:
```
ShaneHarvey, line 776: ruff is still failing because it wants you to use f-strings. Do you have pre-commit set up?

aclark4life: Sorry, forgot my `pre-commit install`! I think I just need to add a more meaningful test to the `SearchIndexModel` test class to finish this.

test/test_operations.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2009-present MongoDB, Inc.
```
ShaneHarvey, line 1: 2009-> 2024

test/test_operations.py
```diff
@@ -0,0 +1,76 @@
...
+            repr(IndexModel({"hello": 1, "world": -1})),
+            "IndexModel({'hello': 1, 'world': -1}, name='hello_1_world_-1')",
+        )
+
+
+class TestSearchIndexModel(TestOperationsBase):
+    """Test SearchIndexModel features."""
+
+    def test_repr(self):
+        self.assertRepr(SearchIndexModel({"hello": "hello"}, key=1))
```
ShaneHarvey, line 72: Need an assert with a string literal for SearchIndexModel too.

ShaneHarvey (changes requested):
Please avoid force pushing because it makes it harder to review what's changed on the new commits.

test/test_operations.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2009-2024 MongoDB, Inc.
```
ShaneHarvey, line 1: 2009-2024 -> 2024-present

aclark4life: Hah 😂 fixing …

ShaneHarvey (approved):

PR #1913, opened 2024-10-10
PYTHON-4528 - Convert unified test runner to async

NoahStapp:


blink1073 (approved):
Well that was easy.  😅

NoahStapp:
> Well that was easy. 😅

Yeah, for some reason I expected more complexity but it was very simple.

test/asynchronous/unified_format.py
```diff
@@ -0,0 +1,1573 @@
...
+                opts["key_vault_namespace"],
+                opts["key_vault_client"],
+                DEFAULT_CODEC_OPTIONS,
+                opts.get("kms_tls_options", kms_tls_options),
+            )
+            return
+        elif entity_type == "thread":
+            name = spec["id"]
+            thread = SpecRunnerThread(name)
+            thread.start()
```
ShaneHarvey, line 379: We probably need to add an async version of SpecRunnerThread which creates a task instead of a thread.

PR #1916, opened 2024-10-11
PYTHON-4847 - Convert test.test_collection_management.py to async

NoahStapp:
Relies on https://github.com/mongodb/mongo-python-driver/pull/1913.


blink1073 (approved):
👍🏼

PR #1929, opened 2024-10-15
PYTHON-4870 - MongoClient.address should block until a connection suc…

NoahStapp:
…ceeds or fails

### DIFF
--- a/pymongo/asynchronous/mongo_client.py
+++ b/pymongo/asynchronous/mongo_client.py
@@ -1453,13 +1453,6 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
                 'Cannot use "address" property when load balancing among'
                 ' mongoses, use "nodes" instead.'
             )
-        if topology_type not in (
-            TOPOLOGY_TYPE.ReplicaSetWithPrimary,
-            TOPOLOGY_TYPE.Single,
-            TOPOLOGY_TYPE.LoadBalanced,
-            TOPOLOGY_TYPE.Sharded,
-        ):
-            return None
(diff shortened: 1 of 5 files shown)


blink1073 (approved):
LGTM!

pymongo/asynchronous/mongo_client.py
```diff
@@ -1453,13 +1453,6 @@ async def address(self) -> Optional[tuple[str, int]]:
                 'Cannot use "address" property when load balancing among'
                 ' mongoses, use "nodes" instead.'
             )
-        if topology_type not in (
-            TOPOLOGY_TYPE.ReplicaSetWithPrimary,
-            TOPOLOGY_TYPE.Single,
-            TOPOLOGY_TYPE.LoadBalanced,
-            TOPOLOGY_TYPE.Sharded,
-        ):
-            return None
```
ShaneHarvey, line 1462: I believe returning None when the TOPOLOGY_TYPE is ReplicaSetNoPrimary was intentional and should be reinstated. This also warrants a not in the changelog and a `.. versionchanged::...`

test/test_replica_set_reconfig.py
```diff
@@ -59,7 +59,8 @@ def test_client(self):
 
         with self.assertRaises(ServerSelectionTimeoutError):
             c.db.command("ping")
-        self.assertEqual(c.address, None)
+        with self.assertRaises(ServerSelectionTimeoutError):
+            _ = c.address
```
ShaneHarvey, line 63: Returning None here was intentional behavior.

PR #1930, opened 2024-10-15
PYTHON-4868 Generate server tests using shrub.py

blink1073:
- Break up the test matrix to cover all server versions, python versions, platforms, topologies, and auth/ssl with as little repetition as possible
- Break up sync and async tests on MacOS and Windows

Before: 1026 tasks across 39 build variants
After: 495 tasks across 37 build variants

image

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -2403,9 +2403,9 @@ buildvariants:
 # Server Tests for RHEL8.
 - name: test-rhel8-py3.9-auth-ssl
   tasks:
-    - name: .sharded_cluster
     - name: .standalone
     - name: .replica_set
+    - name: .sharded_cluster
   display_name: Test RHEL8 py3.9 Auth SSL
   run_on:
     - rhel87-small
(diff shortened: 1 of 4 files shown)


blink1073:
I'll need to update the GitHub PR aliases once this is merged.

blink1073:
I need to add coverage...

.evergreen/config.yml
```diff
@@ -2399,6 +2400,470 @@ axes:
...
+    SSL: ssl
+    TEST_SUITES: default_async
+    PYTHON_BINARY: C:/python/32/Python313/python.exe
+    SKIP_CSOT_TESTS: "true"
+- name: test-win32-py3.9-noauth-nossl-sync
+  tasks:
+    - name: .sharded_cluster
+  display_name: Test Win32 py3.9 NoAuth NoSSL Sync
+  run_on:
+    - windows-64-vsMulti-small
```
NoahStapp, line 2847: We run 32-bit python on a 64-bit machine?

NoahStapp (changes requested):
One clarification, otherwise looks great! This will be so nice for testing going forward

.evergreen/config.yml
blink1073: 32-bit versions of Python.  No one runs 32 bit machines anymore lol.

blink1073: But some apps still require 32 bit

NoahStapp: Just wanted to make sure, makes sense lol

blink1073:
evergreen retry

blink1073:
evergreen retry

PR #1932, opened 2024-10-15
Resync with master

NoahStapp:


blink1073 (approved):

NoahStapp:
The test failure is due to async SDAM conflicting with the async unified runner's assumptions about SDAM threads. Will fix in a follow-up ticket: https://jira.mongodb.org/browse/PYTHON-4871.

PR #1933, opened 2024-10-15
PYTHON-4871 - Fix async waitForEvent

NoahStapp:


ShaneHarvey (approved):

ShaneHarvey:
Should this be merged into master?

NoahStapp:
> Should this be merged into master?

This is only an issue when combined with async SDAM, so we don't need to merge this into master early.

PR #1928, opened 2024-10-15
PYTHON-4843 - All tests should isolate state within a single async IO loop

NoahStapp:

PR #1934, opened 2024-10-16
PYTHON-4860 - Async client should use asyncio.Lock and asyncio.Condition

NoahStapp:


NoahStapp:
Relies on https://github.com/mongodb/mongo-python-driver/pull/1928.

test/asynchronous/test_locks.py
```diff
@@ -1,513 +1,513 @@
...
-
-        for t in threads:
-            t.join()
-
-        self.assertEqual(results, [True] * nthreads)
-
-
-if __name__ == "__main__":
-    unittest.main()
+# # Copyright 2024-present MongoDB, Inc.
```
blink1073, line 1: TODO?

NoahStapp: However, the two issues fixed in 3.13 are still present on older versions of Python: (https://github.com/python/cpython/issues/111693 and https://github.com/python/cpython/issues/112202)

blink1073: I think we'll need to polyfill those fixes then

NoahStapp: Yeah, vendoring `asyncio.locks` seems like the best option here until we drop 3.12 support 😅

NoahStapp:
The test failures appear to be linux-specific, investigating.

pymongo/lock.py
```diff
@@ -11,15 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Lock and Condition classes vendored from https://github.com/python/cpython/blob/main/Lib/asyncio/locks.py
+to port 3.13 fixes to older versions of Python."""
```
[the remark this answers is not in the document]
blink1073: And we should make it a separate file with a note that it can one day be removed when we drop 3.12 support.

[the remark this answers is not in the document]
NoahStapp: Which parts of the CPython license do we need? Just the [PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2](https://github.com/python/cpython/tree/main?tab=License-1-ov-file#readme) section?

ShaneHarvey (changes requested):

pymongo/asynchronous/pool.py
```diff
@@ -1019,15 +1018,13 @@ def __init__(
         # The first portion of the wait queue.
         # Enforces: maxPoolSize
         # Also used for: clearing the wait queue
-        self.size_cond = _ACondition(threading.Condition(_lock))
```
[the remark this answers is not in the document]
NoahStapp: Good catch!

test/asynchronous/test_locks.py
```diff
@@ -1,513 +0,0 @@
-# Copyright 2024-present MongoDB, Inc.
```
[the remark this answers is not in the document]
NoahStapp: Couldn't we only run these tests on older versions of Python?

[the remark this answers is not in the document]
ShaneHarvey: Sure. By the way did you investigate the bugs to confirm that we actually need to vendor the locks? I'm wondering if we can just use the built in classes and accept any potential built in bugs.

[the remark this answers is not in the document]
ShaneHarvey: After looking, https://github.com/python/cpython/issues/111693 seems fine. 

https://github.com/python/cpython/issues/112202 would could result in deadlocks or timeouts so it seems serious enough to warrant vendoring.

pymongo/lock.py
```diff
@@ -11,15 +11,20 @@
...
 
 import asyncio
-import collections
 import os
+import sys
 import threading
-import time
 import weakref
-from typing import Any, Callable, Optional, TypeVar
+from asyncio import TimeoutError, wait_for
```
ShaneHarvey, line 24: Suggest using `asyncio.TimeoutError` to avoid shadowing the builtin TimeoutError.

pymongo/lock.py
```diff
@@ -28,6 +33,14 @@
 
 _T = TypeVar("_T")
 
+# Needed to support 3.13 asyncio fixes in older versions of Python
```
ShaneHarvey, line 36: Can you add `https://github.com/python/cpython/issues/112202` here? That's the most important bugfix we need.

pymongo/lock.py
```diff
@@ -39,209 +52,40 @@ def _create_lock() -> threading.Lock:
...
+def _async_create_lock() -> Lock:
+    """Represents an asyncio.Lock."""
+    return Lock()
+
+
+def _create_condition(
+    lock: threading.Lock, condition_class: Optional[Any] = None
+) -> threading.Condition:
+    """Represents a threading.Condition."""
+    if condition_class:
```
ShaneHarvey, line 64: Sigh, I can't believe we still have to write code for condition_class even though it's never used. I reopened https://jira.mongodb.org/browse/PYTHON-1149.

test/asynchronous/test_locks.py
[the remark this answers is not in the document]
NoahStapp: https://github.com/python/cpython/issues/111693 doesn't seem like an issue agreed.

pymongo/lock.py
NoahStapp: Yeah I wasn't sure if we actually still used it, removing it entirely would be great.

ShaneHarvey (approved):

blink1073 (approved):
LGTM!

PR #1935, opened 2024-10-16
PYTHON-4765 Resync server-selection spec

aclark4life:


blink1073 (approved):
LGTM!

test/server_selection_logging/operation-id.json
```diff
@@ -197,7 +197,7 @@
               }
             },
             {
-              "level": "debug",
+              "level": "info",
```
ShaneHarvey, line 200: How are we passing this test? We actually (intentionally) log this at info level. See PYTHON-4261.

ShaneHarvey: We're logging that line at "info" level while the spec test claims it should be at "debug" level. I would expect the test to fail. This indicates theres a bug in the test runner that we need to fix.

PR #1936, opened 2024-10-16
PYTHON-4865 Skip test_write_concern_failure tests temporarily

blink1073:


test/asynchronous/test_bulk.py
```diff
@@ -971,6 +971,9 @@ async def cause_wtimeout(self, requests, ordered):
     @async_client_context.require_replica_set
     @async_client_context.require_secondaries_count(1)
     async def test_write_concern_failure_ordered(self):
+        self.skipTest("Skipping until PYTHON-4865 is resolved.")
+        details = None
```
ShaneHarvey, line 975: What is details=None?

blink1073: To satisfy the type checker, since it is only defined in an exception block otherwise.

ShaneHarvey (approved):

test/asynchronous/test_bulk.py
ShaneHarvey: ack.

PR #1937, opened 2024-10-16
PYTHON-4873 Remove bson-stdint-win32.h from THIRD-PARTY-NOTICES

ShaneHarvey:
PYTHON-4873 Remove bson-stdint-win32.h from THIRD-PARTY-NOTICES


blink1073 (approved):
👍🏼

PR #1942, opened 2024-10-17
PYTHON-4874 Fix async Windows KMS support

ShaneHarvey:
PYTHON-4874 Fix async Windows KMS support

### DIFF
--- a/pymongo/network_layer.py
+++ b/pymongo/network_layer.py
@@ -215,11 +215,11 @@ else:
         while total_read < length:
             try:
                 read = conn.recv_into(mv[total_read:])
+                if read == 0:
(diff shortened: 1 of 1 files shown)


pymongo/network_layer.py
```diff
@@ -215,11 +215,11 @@ async def _async_receive_ssl(
         while total_read < length:
             try:
                 read = conn.recv_into(mv[total_read:])
+                if read == 0:
+                    raise OSError("connection closed")
```
ShaneHarvey, line 219: We have to check for a 0-read first.

NoahStapp (approved):

PR #1943, opened 2024-10-17
PYTHON-4881 - Use OvertCommandListener wherever sensitive events are not needed

NoahStapp:


blink1073 (approved):
LGTM!

blink1073 (approved):

blink1073 (approved):

NoahStapp:
I'm unable to reproduce any of the failures locally on a linux workstation. They're all limited to SSL tasks run with pypy3.10, but without reproduction I can't dig any deeper into the cause.

blink1073 (approved):

PR #1944, opened 2024-10-17
PYTHON-4879 Use shrub.py for compressor tests

blink1073:
Before: 288 tasks across 36 build variants
After: 78 tasks across 9 build variants

snappy compression RHEL8 py3 9

### DIFF
--- a/.evergreen/scripts/generate_config.py
+++ b/.evergreen/scripts/generate_config.py
@@ -351,10 +351,11 @@ def create_compression_variants():
     host = "rhel8"
     task_names = dict(snappy=[".standalone"], zlib=[".standalone"], zstd=[".standalone !.4.0"])
     variants = []
-    for compressor, python, c_ext in product(["snappy", "zlib", "zstd"], MIN_MAX_PYTHON, C_EXTS):
+    for ind, (compressor, c_ext) in enumerate(product(["snappy", "zlib", "zstd"], C_EXTS)):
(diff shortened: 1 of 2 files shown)


NoahStapp (approved):

.evergreen/scripts/generate_config.py
```diff
@@ -337,8 +345,49 @@ def create_load_balancer_variants():
...
+    return variants
+
+
 ##################
 # Generate Config
 ##################
 
-generate_yaml(variants=create_load_balancer_variants())
+variants = create_compression_variants()
+# print(len(variants))
```
NoahStapp, line 392:
```suggestion
```

NoahStapp (dismissed):

.evergreen/scripts/generate_config.py
NoahStapp: Should we keep it in this final merge into master?

blink1073: I'll just have to keep adding it back in every time while I'm building out these variants, doesn't seem worth doing

NoahStapp: Can you let me know when the last shrub conversion PR is then? That way I know when any debugging code should be gone.

blink1073: When I close the main ticket lol, a long way off

NoahStapp (approved):

PR #1945, opened 2024-10-17
PYTHON-4882 Use shrub.py for enterprise auth tests

blink1073:
Also added a marker for auth so we're not running the full test suite.

Before: 12 tasks across 12 build variants
After: 7 tasks across 7 build variants

image

### DIFF
--- a/test/asynchronous/test_auth.py
+++ b/test/asynchronous/test_auth.py
@@ -32,6 +32,8 @@ from test.asynchronous import (
 )
 from test.utils import AllowListEventListener, delay, ignore_deprecations
 
+import pytest
+
 from pymongo import AsyncMongoClient, monitoring
 from pymongo.asynchronous.auth import HAVE_KERBEROS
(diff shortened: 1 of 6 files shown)


test/asynchronous/test_auth.py
```diff
@@ -80,6 +82,7 @@ def run(self):
 
 
 class TestGSSAPI(AsyncPyMongoTestCase):
+    pytestmark = pytest.mark.auth
```
NoahStapp, line 85: We can put this once at the top of the file to mark the whole module rather than each class individually.

NoahStapp (changes requested):

test/asynchronous/test_auth.py
blink1073: Cool, I didn't think that would work with unittest classes, updated.

NoahStapp (approved):

PR #1946, opened 2024-10-18
PYTHON-4886 Use shrub.py for PyOpenSSL tests

blink1073:
Before: 154 tasks across 26 build variants
After: 77 tasks across 7 build variants

Pasted Graphic 7

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -3249,6 +3249,99 @@ buildvariants:
     AUTH: auth
     PYTHON_BINARY: /opt/python/pypy3.10/bin/python3
 
+# PyOpenSSL tests.
+- name: pyopenssl-macos-py3.9
+  tasks:
+    - name: .replica_set
+    - name: .7.0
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -3259,6 +3249,99 @@ buildvariants:
...
+- name: pyopenssl-macos-py3.9
+  tasks:
+    - name: .replica_set
+    - name: .7.0
+  display_name: PyOpenSSL macOS py3.9
+  run_on:
+    - macos-14
+  batchtime: 10080
+  expansions:
+    AUTH: noauth
```
NoahStapp, line 3262: Should this have auth as well? Every other task here does.

blink1073: No, there's a note in the generator that is carried over from the config.yml file

NoahStapp (approved):

NoahStapp (approved):

PR #1963, opened 2024-10-22
PYTHON-2926 Updated signature of Binary.from_vector to take a BinaryVector

caseyclements:
Backward-compatible change of `Binary.from_vector` to take the BinaryVector dataclass as an alternative to separate list, dtype, and padding.

### DIFF
--- a/bson/binary.py
+++ b/bson/binary.py
@@ -426,7 +426,7 @@ class Binary(bytes):
                 )
             dtype = vector.dtype
             padding = vector.padding
-            vector = vector.data
+            vector = vector.data  # type: ignore
 
         padding = 0 if padding is None else padding
         if dtype == BinaryVectorDtype.INT8:  # pack ints in [-128, 127] as signed int8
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
LGTM!

bson/binary.py
```diff
@@ -400,24 +400,35 @@ def as_uuid(self, uuid_representation: int = UuidRepresentation.STANDARD) -> UUI
     @classmethod
     def from_vector(
         cls: Type[Binary],
-        vector: list[int, float],
-        dtype: BinaryVectorDtype,
-        padding: int = 0,
+        vector: Union[BinaryVector, list[int, float]],
+        dtype: Optional[BinaryVectorDtype] = None,
+        padding: Optional[int] = None,
```
ShaneHarvey, line 405: Using a typing overload could be a cleaner solution here. Then you could properly convey to the type checker that dtype is required when vector is a list.

PR #1966, opened 2024-10-24
PYTHON-4902 Use shrub.py to generate tasks

blink1073:
Before:  1538 tasks across 194 build variants
After: 2285 tasks across 131 build variants

Ensures a consistent set of tests available for all python and server versions with the following combo:

```
SUB_TASKS = [
    ".sharded_cluster .auth .ssl",
    ".replica_set .noauth .ssl",
    ".standalone .noauth .nossl",
]
```

The full set of auth/ssl and topology is only tested on RHEL for the three main python versions.  
We split up the tasks on RHEL8 into async/sync as well, which accounts for a lot of the extra tasks, but the overall runtime will be similar.

image


blink1073:
I'll adjust the GitHub PR tasks to only run that subset on PRs.

NoahStapp (approved):

PR #1967, opened 2024-10-24
PYTHON-4903 Adds typing overloading to bson.binary.Binary.from_vector

caseyclements:
Follows suggestion from @ShaneHarvey that adding typing oveloading would make clear that dtype  is a required argument when the first argument, vector, is a list, not at BinaryVector instance.

### DIFF
--- a/test/test_bson.py
+++ b/test/test_bson.py
@@ -802,6 +802,13 @@ class TestBSON(unittest.TestCase):
         assert float_binary == Binary.from_vector(
             BinaryVector(list_vector, BinaryVectorDtype.FLOAT32)
         )
+        # Confirm kwargs cannot be passed when BinaryVector is provided
+        self.assertRaises(
+            ValueError,
+            Binary.from_vector,
+            BinaryVector(list_vector, BinaryVectorDtype.PACKED_BIT, padding),
+            {"dtype": BinaryVectorDtype.PACKED_BIT},
+        )
 
     def test_unicode_regex(self):
         """Tests we do not get a segfault for C extension on unicode RegExs.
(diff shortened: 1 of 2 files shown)


blink1073 (changes requested):
We should also remove the **BETA** and add a changelog entry

ShaneHarvey (changes requested):
Do we have a test that `from_vector(BinaryVector(), dtype=...)` raises an error and requires a type-ignore?

test/test_bson.py
```diff
@@ -802,6 +802,13 @@ def test_vector(self):
         assert float_binary == Binary.from_vector(
             BinaryVector(list_vector, BinaryVectorDtype.FLOAT32)
         )
+        # Confirm kwargs cannot be passed when BinaryVector is provided
+        self.assertRaises(
```
ShaneHarvey, line 806: Can you write this as a `with self.assertRaises():` so we can confirm a "type-ignore" is detected?

ShaneHarvey (changes requested):

test/test_bson.py
caseyclements: Sure can. How do we confirm that a type-ignore is detected?

```
        with self.assertRaises(ValueError):
            Binary.from_vector(
                BinaryVector(list_vector, BinaryVectorDtype.PACKED_BIT, padding),
                dtype=BinaryVectorDtype.PACKED_BIT
            )
```

caseyclements: Oh.... I see. This will fail linting without it.

ShaneHarvey: Yep that's what I was thinking!

ShaneHarvey (approved):

test/test_bson.py
caseyclements: TIL. Thanks @ShaneHarvey !  This was a very cool lesson.

blink1073 (approved):

PR #1976, opened 2024-10-28
PYTHON-4915 - Add guidance on adding _id fields to documents to CRUD spec, reorder client.bulk_write generated _id fields

NoahStapp:


ShaneHarvey (changes requested):
We don't need this change. For one it adds a likely non-trivial perf cost for documents with many top-level fields. We also already reorder the _id to be the first field when encoding a document to BSON, see: https://github.com/mongodb/mongo-python-driver/blob/3ef565fa43734dfef6bdbb7458b41a4d71451cb1/bson/__init__.py#L1005-L1006


Instead we should add a MockupDB test to assert that we actually send the `_id` field first, if such a test doesn't already exist.

NoahStapp:
> We don't need this change. For one it adds a likely non-trivial perf cost for documents with many top-level fields. We also already reorder the _id to be the first field when encoding a document to BSON, see:
> 
> https://github.com/mongodb/mongo-python-driver/blob/3ef565fa43734dfef6bdbb7458b41a4d71451cb1/bson/__init__.py#L1005-L1006
> 
> Instead we should add a MockupDB test to assert that we actually send the `_id` field first, if such a test doesn't already exist.

`insert` operations and collection-level `bulkWrite` correctly order their BSON documents, but client-level `bulkWrite` does not:

```
client.bulk_write([InsertOne(namespace="db.coll", document={'x2': 1})]) 

# Sends this to the server
{"bulkWrite": 1, "errorsOnly": true, "ordered": true, "$db": "admin", "ops": [{"insert": 0, "document": {"x2": 1, "_id": {"$oid": "671ff1d0fe637fbcdaa64254"}}}], "nsInfo": [{"ns": "db.coll"}]}
```

This is because the `document` is embedded within the operation mapping, so it does not put the `_id` field first.

ShaneHarvey:
Ah good find. Then we need to add the change back for client.bulkWrite. An alternative to copying the data is to use ChainMap:
```python
>>> from collections import ChainMap
>>> bson.encode({'subdoc': ChainMap({'_id': 1}, {'d': 'doc'})})
b'&\x00\x00\x00\x03subdoc\x00\x19\x00\x00\x00\x02d\x00\x04\x00\x00\x00doc\x00\x10_id\x00\x01\x00\x00\x00\x00\x00'
```

Although we may need to document this interaction with command monitoring. We'll also still need to add `_id` to the input for backwards compat.

test/mockupdb/test_id_ordering.py
```diff
@@ -0,0 +1,52 @@
+from __future__ import annotations
```
ShaneHarvey, line 1: Let's add the boilerplate License comment.

test/mockupdb/test_id_ordering.py
```diff
@@ -0,0 +1,52 @@
...
+except ImportError:
+    _HAVE_MOCKUPDB = False
+
+
+from bson.objectid import ObjectId
+
+pytestmark = pytest.mark.mockupdb
+
+
+class TestIdOrdering(PyMongoTestCase):
```
ShaneHarvey, line 22: Can you add a link to the crud spec that describes this test?

pymongo/asynchronous/client_bulk.py
```diff
@@ -133,7 +133,10 @@ def add_insert(self, namespace: str, document: _DocumentOut) -> None:
         validate_is_document_type("document", document)
         # Generate ObjectId client side.
         if not (isinstance(document, RawBSONDocument) or "_id" in document):
-            document["_id"] = ObjectId()
+            new_document = {"_id": ObjectId()}
+            new_document.update(document)
+            document.clear()
+            document.update(new_document)
```
ShaneHarvey, line 139: Thoughts on the perf implications of this vs ChainMap?

pymongo/asynchronous/client_bulk.py
```diff
@@ -133,7 +133,10 @@ def add_insert(self, namespace: str, document: _DocumentOut) -> None:
         validate_is_document_type("document", document)
         # Generate ObjectId client side.
         if not (isinstance(document, RawBSONDocument) or "_id" in document):
-            document["_id"] = ObjectId()
+            new_document = {"_id": ObjectId()}
+            new_document.update(document)
+            document.clear()
```
ShaneHarvey, line 138: The more I think about it the more I think it's problematic to call clear() and update() here. Those methods could have unintentional side effects aside from the perf problems. For example consider a user passing a custom mapping class which overrides clear()/update().

pymongo/asynchronous/client_bulk.py
```diff
@@ -133,7 +133,10 @@ def add_insert(self, namespace: str, document: _DocumentOut) -> None:
         validate_is_document_type("document", document)
         # Generate ObjectId client side.
         if not (isinstance(document, RawBSONDocument) or "_id" in document):
-            document["_id"] = ObjectId()
+            new_document = {"_id": ObjectId()}
```
ShaneHarvey, line 136: This implementation is also incomplete because it does not put the `_id` field first if the user supplies it. For example when inserting `{"a": 1, "_id": 2}`. We should add tests for this case for insert/bulk/clientBulk as well.

test/mockupdb/test_id_ordering.py
NoahStapp: Once the spec is merged, yes.

pymongo/asynchronous/client_bulk.py
NoahStapp: Using ChainMap is cleaner, encoding to RawBSONDocuments might have additional performance costs.

pymongo/asynchronous/client_bulk.py
NoahStapp: Using ChainMap makes sense, agreed. Explicitly modifying a user-supplied mapping will always carry some risks unfortunately, using the least amount of APIs as possible seems like a safer bet here.

pymongo/asynchronous/client_bulk.py
NoahStapp: We've decided to make re-ordering user-supplied `_id` fields optional due to the complexity of doing so across different driver implementations. We can do it in PyMongo if we want, but it won't be standard across all drivers.

pymongo/asynchronous/client_bulk.py
```diff
@@ -133,7 +134,9 @@ def add_insert(self, namespace: str, document: _DocumentOut) -> None:
         validate_is_document_type("document", document)
         # Generate ObjectId client side.
         if not (isinstance(document, RawBSONDocument) or "_id" in document):
-            document["_id"] = ObjectId()
+            document = ChainMap(document, {"_id": ObjectId()})
```
ShaneHarvey, line 137: We still need to add id to the input document here. Can you also add a test for that?

pymongo/asynchronous/client_bulk.py
```diff
@@ -133,7 +134,9 @@ def add_insert(self, namespace: str, document: _DocumentOut) -> None:
...
+        elif not isinstance(document, RawBSONDocument) and "_id" in document:
```
ShaneHarvey, line 138: Could you refactor these two if statements to avoid repeating the checks?
```python
if not isinstance(document, RawBSONDocument):
    if "_id" in document:
        ...
    else:
```

ShaneHarvey: It would also be worthwhile to add a comment explaining the ChainMap usage and why we have to use here but not in other insert code paths.

ShaneHarvey (changes requested):

pymongo/asynchronous/client_bulk.py
NoahStapp: Did some more thinking: is the added complexity and changing of the type to `ChainMap` here worth it over this much simpler approach:

```
  if "_id" in document:
      document = {"_id": document["_id"]} | document
  else:
      id = ObjectId()
      document["_id"] = id
      document = {"_id": id} | document
```

If the original document already had an `_id` field, this doesn't modify it, which is consistent with our other insert code paths. If we generated an `_id` field, we still add it to the original document, but we don't worry about the order of the original document.

ShaneHarvey: Simpler yes, but it's not performant:
```
$ python -m timeit -s 'd={str(k):k for k in range(10000)};from collections import ChainMap' 'ChainMap(d,{"_id":1})'
2000000 loops, best of 5: 163 nsec per loop
$ python -m timeit -s 'd={str(k):k for k in range(10000)}' '{"_id":1}|d'
2000 loops, best of 5: 143 usec per loop
```

NoahStapp: Excellent point!

NoahStapp: Yeah unwrapping it back into the original map makes sense.

pymongo/asynchronous/client_bulk.py
```diff
@@ -132,8 +133,16 @@ def add_insert(self, namespace: str, document: _DocumentOut) -> None:
...
+        if not isinstance(document, RawBSONDocument):
+            # Since the data document itself is nested within the insert document
+            # it won't be automatically re-ordered by the BSON conversion.
+            # We use ChainMap here to make the _id field the first field instead.
+            if "_id" in document:
+                document = ChainMap(document, {"_id": document["_id"]})
+            else:
+                id = ObjectId()
+                document["_id"] = id
+                document = ChainMap(document, {"_id": id})
```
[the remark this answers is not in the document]
NoahStapp: We'd still need to unwrap it, even with this change:

```python
# Started events
[{'bulkWrite': 1, 'errorsOnly': True, 'ordered': False, 'lsid': {'id': Binary(b'\xa3\xc7\x80\xdd\x07\x98L\x13\x81\xb8\xbcY\xe8\xa0\x04\xf3', 4)}, '$db': 'admin', 'ops': [{'insert': 0, 'document': ChainMap({'foo': 'bar', '_id': 5}, {'_id': 5})}, {'insert': 1, 'document': ChainMap({'foo': 'bar', '_id': 6}, {'_id': 6})}, {'insert': 0, 'document': ChainMap({'foo': 'bar', '_id': 5}, {'_id': 5})}, {'insert': 1, 'document': ChainMap({'foo': 'bar', '_id': 7}, {'_id': 7})}, {'delete': 0, 'filter': {'foo': 'bar', '_id': 5}, 'multi': False}], 'nsInfo': [{'ns': 'db.test_five'}, {'ns': 'db.test_six'}]}]

# Bulk write error
batch op errors occurred, full error: {'anySuccessful': True, 'error': None, 'writeErrors': [{'ok': 0.0, 'idx': 1, 'code': 11000, 'errmsg': 'E11000 duplicate key error collection: db.test_six index: _id_ dup key: { _id: 6 }', 'keyPattern': {'_id': 1}, 'keyValue': {'_id': 6}, 'n': 0, 'op': {'insert': 1, 'document': ChainMap({'foo': 'bar', '_id': 6}, {'_id': 6})}}, {'ok': 0.0, 'idx': 2, 'code': 11000, 'errmsg': 'E11000 duplicate key error collection: db.test_five index: _id_ dup key: { _id: 5 }', 'keyPattern': {'_id': 1}, 'keyValue': {'_id': 5}, 'n': 0, 'op': {'insert': 0, 'document': ChainMap({'foo': 'bar', '_id': 5}, {'_id': 5})}}, {'ok': 0.0, 'idx': 3, 'code': 11000, 'errmsg': 'E11000 duplicate key error collection: db.test_six index: _id_ dup key: { _id: 7 }', 'keyPattern': {'_id': 1}, 'keyValue': {'_id': 7}, 'n': 0, 'op': {'insert': 1, 'document': ChainMap({'foo': 'bar', '_id': 7}, {'_id': 7})}}], 'writeConcernErrors': [], 'nInserted': 1, 'nUpserted': 0, 'nMatched': 0, 'nModified': 0, 'nDeleted': 1, 'insertResults': {}, 'updateResults': {}, 'deleteResults': {}}
```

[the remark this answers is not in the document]
NoahStapp: I still like moving the ChainMap logic into `_client_batched_op_msg_impl` to make how we add `_id` fields consistent across insert methods.

pymongo/message.py
```diff
@@ -1111,6 +1112,12 @@ def _check_doc_size_limits(
         # key and the index of its namespace within ns_info as its value.
         op_doc[op_type] = ns_info[namespace]  # type: ignore[index]
 
+        # Since the data document itself is nested within the insert document
+        # it won't be automatically re-ordered by the BSON conversion.
+        # We use ChainMap here to make the _id field the first field instead.
+        if real_op_type == "insert":
+            op_doc["document"] = ChainMap(op_doc["document"], {"_id": op_doc["document"]["_id"]})  # type: ignore[index]
```
ShaneHarvey, line 1119: The only reason we need to unwrap ChainMap later is because we're mutating op_doc. Instead we can do this:
```python
        doc_to_encode = op_doc
        if real_op_type == "insert":
            doc = op_doc["document"]
            if not isinstance(doc, RawBSONDocument):
                doc_to_encode = op_doc.copy()  # Shallow copy
                doc_to_encode["document"] = ChainMap(doc, {"_id": doc["_id"]})  # type: ignore[index]
        op_doc_encoded = _dict_to_bson(doc_to_encode, False, opts)
```

We also still need the RawBSONDocument here. We should add a test to ensure RawBSONDocument is not inflated after a call to bulk_write.

ShaneHarvey (changes requested):

pymongo/message.py
NoahStapp: Ah, I see! That's a slick solution.

test/asynchronous/test_client_bulk_write.py
```diff
@@ -18,6 +18,8 @@
 import os
 import sys
 
+from bson import RawBSONDocument, encode
```
ShaneHarvey, line 21: RawBSONDocument needs to be imported from `bson.raw_bson`

test/asynchronous/test_client_bulk_write.py
```diff
@@ -84,6 +86,17 @@ async def test_formats_write_error_correctly(self):
...
+    @async_client_context.require_no_serverless
+    async def test_raw_bson_not_inflated(self):
+        doc = RawBSONDocument(encode({"a": "b" * 100}))
+        models = [
+            InsertOne(namespace="db.coll", document=doc),
+        ]
+        await self.client.bulk_write(models=models)
+
+        self.assertIsNone(doc._RawBSONDocument__inflated_doc)
+
```
ShaneHarvey, line 99: Nice!

test/mockupdb/test_id_ordering.py
ShaneHarvey: Was the spec merged?

NoahStapp: Yes. added!

ShaneHarvey:
When merging can you update the commit message to mention the client.bulk_write change?

PR #1978, opened 2024-10-29
PYTHON-4922 Remove Support for MONGODB-CR Authentication

blink1073:


test/test_uri_parser.py
```diff
@@ -294,32 +293,6 @@ def test_parse_uri(self):
         self.assertEqual(res, parse_uri("mongodb://localhost/?readPreference=secondary"))
 
         # Various authentication tests
-        res = copy.deepcopy(orig)
-        res["options"] = {"authmechanism": "MONGODB-CR"}
```
ShaneHarvey, line 298: Rather than removing these tests, can you update then to use SCRAM-SHA-256?

blink1073: Done

ShaneHarvey (approved):

PR #1980, opened 2024-10-29
PYTHON-4766 - Fix logic for determining whether to populate BulkWriteException.partialResult

NoahStapp:
Our implementation already used the correct logic, so this is purely adding the new spec tests (thanks Shruti!)


blink1073 (approved):
LGTM!

PR #1981, opened 2024-10-29
PYTHON-4209 Ensure that no error is raised for unknown auth mechanism

blink1073:


NoahStapp (approved):

test/mockupdb/test_handshake.py
```diff
@@ -229,6 +229,39 @@ def test_client_handshake_saslSupportedMechs(self):
...
+
+        primary_response = OpReply(
+            "ismaster",
+            True,
+            minWireVersion=2,
+            maxWireVersion=MIN_SUPPORTED_WIRE_VERSION,
+            saslSupportedMechs=["SCRAM-SHA-256", "does_not_exist"],
+        )
+        client = MongoClient(
+            server.uri, authmechanism="PLAIN", username="username", password="password"
```
ShaneHarvey, line 245: I think this test may be flawed. A client only negotiates the auth mechanism if authmechanism is not supplied but this test sets `authmechanism="PLAIN"`. Instead it should create the client without authmechanism or set `authmechanism="DEFAULT"`.

PR #1983, opened 2024-10-30
PYTHON-3906 & PYTHON-2867 Implement GSSAPI ServiceHost support and expand canonicalization options

blink1073:


NoahStapp (approved):

pymongo/auth_shared.py
```diff
@@ -78,7 +78,7 @@ def __hash__(self) -> int:
 
 
 GSSAPIProperties = namedtuple(
-    "GSSAPIProperties", ["service_name", "canonicalize_host_name", "service_realm"]
+    "GSSAPIProperties", ["service_name", "canonicalize_host_name", "service_realm", "service_host"]
```
ShaneHarvey, line 81: It doesn't look like we actually do anything with service_host. Why add it if it's not used?

pymongo/auth_shared.py
```diff
@@ -87,6 +87,16 @@ def __hash__(self) -> int:
 """Mechanism properties for MONGODB-AWS authentication."""
 
 
+def _validate_canonicalize_host_name(value: str | bool) -> str | bool:
+    valid_names = [False, True, "none", "forward", "forwardAndReverse"]
```
ShaneHarvey, line 91: So "none", "forward", and "forwardAndReverse" are all supposed to be treated the same as True? Is that accurate?

blink1073: No, you'll see in the follow up PR how they're used in practice.

PR #1985, opened 2024-10-31
PYTHON-3096 Finish implementation and tests for GSSAPI options

blink1073:


pymongo/asynchronous/auth.py
```diff
@@ -177,13 +177,20 @@ def _auth_key(nonce: str, username: str, password: str) -> str:
...
     # https://github.com/krb5/krb5/blob/d406afa363554097ac48646a29249c04f498c88e/src/util/k5test.py#L505-L520
+    if option in [False, None]:
+        return hostname
+
     af, socktype, proto, canonname, sockaddr = socket.getaddrinfo(
         hostname, None, 0, 0, socket.IPPROTO_TCP, socket.AI_CANONNAME
     )[0]
 
+    # For forward just to resolve the cname as dns.lookup() will not return it.
+    if option == "forward":
```
ShaneHarvey, line 191: It looks like the string `"none"` is also valid:
```
def _validate_canonicalize_host_name(value: str | bool) -> str | bool:
    valid_names = [False, True, "none", "forward", "forwardAndReverse"]
```

Should that have the same behavior as `False`? Or should validate convert "none" to False?

ShaneHarvey (changes requested):

pymongo/asynchronous/auth.py
blink1073: I changed it to convert "none" to false, so the three validated options are True, False, and "forward".

test/asynchronous/test_auth.py
```diff
@@ -268,6 +272,61 @@ async def test_gssapi_threaded(self):
...
+            GSSAPI_PORT,
+            username=GSSAPI_PRINCIPAL,
+            password=GSSAPI_PASS,
+            authMechanism="GSSAPI",
+            authMechanismProperties=props,
+        )
+        await client.server_info()
+
+        if props["CANONICALIZE_HOST_NAME"] == "none":
+            return
```
ShaneHarvey, line 301: This if-statement should be removed since it's redundant, or is there something missing?

pymongo/asynchronous/auth.py
```diff
@@ -177,13 +177,20 @@ def _auth_key(nonce: str, username: str, password: str) -> str:
...
+    if option in [False, "none"]:
+        return hostname
+
     af, socktype, proto, canonname, sockaddr = socket.getaddrinfo(
         hostname, None, 0, 0, socket.IPPROTO_TCP, socket.AI_CANONNAME
     )[0]
 
+    # For forward just to resolve the cname as dns.lookup() will not return it.
+    if option == "forward":
+        return canonname.lower()
```
ShaneHarvey, line 192: I'm still confused by this method. DRIVERS-1803 says:
> "forward" is equivalent to the current behaviour of true (a forward DNS lookup)

But here we are introducing new behavior for "forward". Is that because our behavior for "true" is actually "forwardAndReverse"? If so could you add a comment mentioning that?

ShaneHarvey (changes requested):

pymongo/asynchronous/auth.py
blink1073: Ah, I misinterpreted the Node implementation and got confused.   I'll redo the logic.

blink1073: No, it was the description in the ticket that was incorrect.  I verified by reading the kerberos spec, the drivers spec, and the node implementation that this is the correct implementation.

blink1073: I also updated the ticket description

test/asynchronous/test_auth.py
blink1073: removed

pymongo/asynchronous/auth.py
```diff
@@ -177,13 +177,20 @@ def _auth_key(nonce: str, username: str, password: str) -> str:
...
 
 
-def _canonicalize_hostname(hostname: str) -> str:
+def _canonicalize_hostname(hostname: str, option: str | bool) -> str:
     """Canonicalize hostname following MIT-krb5 behavior."""
     # https://github.com/krb5/krb5/blob/d406afa363554097ac48646a29249c04f498c88e/src/util/k5test.py#L505-L520
+    if option in [False, "none"]:
+        return hostname
+
     af, socktype, proto, canonname, sockaddr = socket.getaddrinfo(
```
ShaneHarvey, line 186: This should be using asyncio's loop.getaddrinfo. Can you open a new ticket to fix that? We might have the same problem in other places.

ShaneHarvey (approved):

pymongo/asynchronous/auth.py
blink1073: https://jira.mongodb.org/browse/PYTHON-5021

PR #2003, opened 2024-11-15
PYTHON-4977 Fix import time on Windows again

ShaneHarvey:
PYTHON-4977 Fix import time on Windows again

In https://github.com/mongodb/mongo-python-driver/pull/1678 we accidentally reverted the fix for PYTHON-4455.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1798,14 +1798,8 @@ buildvariants:
     - name: test_aws_lambda_task_group
 
 - name: rhel8-import-time
-  display_name: Import Time Linux
+  display_name: Import Time
   run_on: rhel87-small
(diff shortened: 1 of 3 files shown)


ShaneHarvey:
I added an import time check on Windows. Also I relized the import time check is pretty easy to miss because it will only fail once on the PR and once on the slow commit, and the it will pass on all commits after that. The only way I can think to avoid that problem is to use a known good version as the reference point and keep it fixed.

ShaneHarvey:
Okay rereading the original ticket, it turns out we can't reproduce the slow import on our Windows hosts anyway so I've removed the windows import check.

blink1073 (approved):
LGTM!

PR #2008, opened 2024-11-18
PYTHON-4356 Unskip spec tests for agg $out

ShaneHarvey:
PYTHON-4356 Unskip spec tests for agg $out

The spec tests themselves were already updated in PYTHON-4576.


blink1073 (approved):
LGTM!

PR #2012, opened 2024-11-19
PYTHON-4982 Remove redundant configureFailPoint

ShaneHarvey:
PYTHON-4982 Remove redundant configureFailPoint

This shouldn't resolve all the issues but it's worth removing the redundant code.

### DIFF
--- a/test/asynchronous/test_retryable_reads.py
+++ b/test/asynchronous/test_retryable_reads.py
@@ -174,9 +174,8 @@ class TestRetryableReads(AsyncIntegrationTest):
             retryReads=True,
         )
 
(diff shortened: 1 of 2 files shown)


test/asynchronous/test_retryable_reads.py
```diff
@@ -174,9 +174,8 @@ async def test_retryable_reads_in_sharded_cluster_multiple_available(self):
             retryReads=True,
         )
 
-        async with self.fail_point(fail_command):
```
ShaneHarvey, line 177: We already set the failpoint on both mongoses above.

blink1073 (approved):
LGTM!

PR #2015, opened 2024-11-20
PYTHON-4982 Fix spurious network errors when using asyncio

ShaneHarvey:
PYTHON-4982 Fix spurious network errors when using asyncio


ShaneHarvey:
This does not work.

PR #2017, opened 2024-11-25
PYTHON-4965 - Consolidate startup and teardown tasks

NoahStapp:


.evergreen/config.yml
```diff
@@ -1640,6 +1594,7 @@ tasks:
           params:
             binary: bash
             working_dir: src
+            include_expansions_in_env: ["PYTHON_BINARY"]
```
Jibola, line 1597: Is this a bug you caught during the fix?

Jibola (approved):
LGTM.

blink1073:
Docker teardown is handled in [drivers-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/teardown.sh#L29).  I'd suggest a single teardown script that runs the appropriate drivers-evergreen-tools teardowns: https://github.com/mongodb/mongo-go-driver/blob/master/.evergreen/config.yml#L143

.evergreen/config.yml
NoahStapp: We need to pass this explicitly where needed now to use the correct Python binary. Part of our later work will be to unify the use of `PYTHON_BINARY` to remove the need to pass it explicitly.

blink1073:
[.evergreen/scripts/teardown-docker.sh](https://github.com/mongodb/mongo-python-driver/pull/2017/files#diff-3f85f7d3bea43139f98a5c8a59562ce6438a0e2f076f76e0b0e2e37c708d0ecc) can be removed now.

NoahStapp:
> [.evergreen/scripts/teardown-docker.sh](https://github.com/mongodb/mongo-python-driver/pull/2017/files#diff-3f85f7d3bea43139f98a5c8a59562ce6438a0e2f076f76e0b0e2e37c708d0ecc) can be removed now.

As can https://github.com/mongodb/mongo-python-driver/pull/2017/files#diff-fa520e855fc44f5c073c5ddfca5ac589f408d6174e2b77946e91c3a1323c35ff!

blink1073 (approved):
LGTM!

PR #2037, opened 2025-01-01
PYTHON-5020 Fix behavior of network timeouts on pyopenssl connections

ShaneHarvey:
PYTHON-5020 Fix behavior of network timeouts on pyopenssl connections

The explanation is that previous to PYTHON-4292 we never put the connection in non-blocking mode so this bug in our pyopenssl wrapper code was never noticed.


blink1073 (approved):
LGTM!

PR #2051, opened 2025-01-08
PYTHON-5017 Fix post-publish step

blink1073:


blink1073:
> Is there any way to test these changes before merging?

Yes, but it is a huge PITA involving changing the default branch on a fork, so I don't do it often.

ShaneHarvey (approved):

PR #2052, opened 2025-01-08
PYTHON-5011 Fix behavior of TLS connection errors on PyPy

ShaneHarvey:
PYTHON-5011 Fix behavior of TLS connection errors on PyPy

This change reverts the wait_for_read change in PYTHON-4292 but only for PyPy to deal with the issues described in that ticket.


ShaneHarvey:
This is ready for review. The async tests are still facing the connection errors but that should be fixed by Noah's work in PYTHON-4493. The other pypy failures are tracked by other test failure tickets.

NoahStapp (approved):

PR #2054, opened 2025-01-10
PYTHON-5014 Fix handling of async socket errors in kms request

blink1073:
This addresses the MacOS encryption errors but not yet the Encryption PyOpenSSL errors.
It addresses the Encryption PyOpenSSL error by expanding the regex to include "SSL handshake failed".

### DIFF
--- a/pymongo/asynchronous/encryption.py
+++ b/pymongo/asynchronous/encryption.py
@@ -213,6 +213,9 @@ class _EncryptionIO(AsyncMongoCryptCallback):  # type: ignore[misc]
                     if not data:
                         raise OSError("KMS connection closed")
                     kms_context.feed(data)
+            # Async raises an OSError instead of returning empty bytes
+            except OSError as err:
+                raise OSError("KMS connection closed") from err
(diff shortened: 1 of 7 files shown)


pymongo/asynchronous/encryption.py
```diff
@@ -213,6 +213,9 @@ async def kms_request(self, kms_context: MongoCryptKmsContext) -> None:
                     if not data:
                         raise OSError("KMS connection closed")
                     kms_context.feed(data)
+            # Async raises an OSError instead of returning empty bytes
+            except OSError as err:
+                raise OSError("KMS connection closed") from err
```
ShaneHarvey, line 218: Doesn't OSError already get handled by the `except Exception as exc` block? It should hit the _raise_connection_failure line.

blink1073: Fixed

ShaneHarvey:
I wonder if some of this logic can be removed after the async streams/protocol work is merged?

PR #2056, opened 2025-01-10
PYTHON-4991 Fix perf client.bulkWrite perf regression

ShaneHarvey:
PYTHON-4991 Fix perf client.bulkWrite perf regression

Explanation: I noticed that the perf regression applied equally to both small and large documents which wouldn't make sense if the overhead was only attributed to the allocation cost of ChainMaps. For this regression to apply to larger documents, that must mean that there's also a performance cost when iterating over the ChainMap in the encoding step (something I had not expected). 

Instead we can encode the document directly so that it gets recognized as a "top-level" document, then the existing `_id` reordering logic will be applied. Running the perf tests on this PR will confirm.

Perf task running here: https://spruce.mongodb.com/task/mongo_python_driver_perf_tests_perf_8.0_standalone_patch_493fc2ab3e237c2155fde4400002ed2aafe9b2be_67817f2e6497de0007ec294b_25_01_10_20_12_36/logs?execution=0

### DIFF
--- a/pymongo/message.py
+++ b/pymongo/message.py
@@ -24,7 +24,6 @@ from __future__ import annotations
 import datetime
 import random
 import struct
-from collections import ChainMap
 from io import BytesIO as _BytesIO
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Results of this change: 
The perf regression for large documents is fixed but for small documents there's still a ~20% regression compare to pre-PYTHON-4915. Therefor I think we should go ahead and revert PYTHON-4915.

PR #2057, opened 2025-01-10
PYTHON-4975 Use justfile as the task runner

blink1073:
I ran a cross-section of tasks here: https://spruce.mongodb.com/version/67830641207aee000703f428/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC.

The justfile itself will get a lot cleaner once we switch to uv.  We'll be able to install `just` and `uv` in the system setup, and then use `uv run` instead of all these hard-coded paths.


blink1073:
zizmor failure is unrelated: https://github.com/woodruffw/zizmor/issues/431

blink1073:
@NoahStapp this is ready for review

.github/workflows/test-python.yml
```diff
@@ -73,18 +75,16 @@ jobs:
...
+        uses: extractions/setup-just@v2
       - name: Install dependencies
         run: |
-          pip install -U pip
-          if [[ "${{ matrix.python-version }}" == "3.13" ]]; then
-            pip install --pre cffi setuptools
-            pip install --no-build-isolation hatch
-          elif [[ "${{ matrix.python-version }}" == "3.13t" ]]; then
+         if [[ "${{ matrix.python-version }}" == "3.13t" ]]; then
             # Hatch can't be installed on 3.13t, use pytest directly.
```
NoahStapp, line 83:
```suggestion
            # Just can't be installed on 3.13t, use pytest directly.
```

pymongo/pyopenssl_context.py
```diff
@@ -277,7 +277,7 @@ def __get_options(self) -> None:
         # Calling set_options adds the option to the existing bitmask and
         # returns the new bitmask.
         # https://www.pyopenssl.org/en/stable/api/ssl.html#OpenSSL.SSL.Context.set_options
-        return self._ctx.set_options(0)
```
NoahStapp, line 280: Accidental regression? This incorrect type hint is fixed in master.

.evergreen/install-dependencies.sh
```diff
@@ -13,6 +13,32 @@ else
   SUDO="sudo"
 fi
 
+# Install just.
+if [ "${CI:-}" == "true" ]; then
```
NoahStapp, line 17: Is `CI` set somewhere by drivers-evergreen-tools on some platforms?

NoahStapp (changes requested):

.evergreen/install-dependencies.sh
blink1073: It is set on Evergreen jobs only.  I'll add a comment.

pymongo/pyopenssl_context.py
blink1073: Fixed

NoahStapp (approved):
Awesome work Steve!

PR #2059, opened 2025-01-14
PYTHON-5021 - Fix usages of getaddrinfo to be non-blocking

NoahStapp:


NoahStapp:
Typing failures fixed by https://github.com/mongodb/mongo-python-driver/pull/2060.

pymongo/asynchronous/auth.py
```diff
@@ -177,15 +178,28 @@ def _auth_key(nonce: str, username: str, password: str) -> str:
...
     if option in [False, "none"]:
         return hostname
 
-    af, socktype, proto, canonname, sockaddr = socket.getaddrinfo(
-        hostname, None, 0, 0, socket.IPPROTO_TCP, socket.AI_CANONNAME
-    )[0]
+    if not _IS_SYNC:
+        loop = asyncio.get_event_loop()
+        af, socktype, proto, canonname, sockaddr = (
+            await loop.getaddrinfo(
```
ShaneHarvey, line 190: FYI:
> Note Both getaddrinfo and getnameinfo internally utilize their synchronous versions through the loop’s default thread pool executor. When this executor is saturated, these methods may experience delays, which higher-level networking libraries may report as increased timeouts. To mitigate this, consider using a custom executor for other user tasks, or setting a default executor with a larger number of workers.

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.getaddrinfo

Which means our users will eventually run into this issue: https://github.com/python/cpython/issues/112169

This is still better than blocking the loop of course but I wonder if we need to warn of this potential problem or if we should test it explicitly.

pymongo/asynchronous/auth.py
```diff
@@ -177,15 +178,28 @@ def _auth_key(nonce: str, username: str, password: str) -> str:
...
     """Canonicalize hostname following MIT-krb5 behavior."""
     # https://github.com/krb5/krb5/blob/d406afa363554097ac48646a29249c04f498c88e/src/util/k5test.py#L505-L520
     if option in [False, "none"]:
         return hostname
 
-    af, socktype, proto, canonname, sockaddr = socket.getaddrinfo(
-        hostname, None, 0, 0, socket.IPPROTO_TCP, socket.AI_CANONNAME
-    )[0]
+    if not _IS_SYNC:
+        loop = asyncio.get_event_loop()
```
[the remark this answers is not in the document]
blink1073: Yes, "Because this function has rather complex behavior (especially when custom event loop policies are in use), using the get_running_loop() function is preferred to get_event_loop() in coroutines and callbacks.".

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

pymongo/asynchronous/auth.py
NoahStapp: Would explicitly testing it provide an actionable solution? We could increase the default number of workers to help mitigate this, but warning users might only confuse them since this is an internal API.

pymongo/asynchronous/auth.py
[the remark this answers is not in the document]
NoahStapp: Addressed for the rest of the codebase in https://github.com/mongodb/mongo-python-driver/pull/2063.

pymongo/asynchronous/auth.py
blink1073: Should we instead use `run_in_executor` and have our own executor?  We use `run_in_executor` in `_configured_socket` as well.

NoahStapp: Oh, you're saying we use our own `ThreadPoolExecutor` instance to avoid competing with the default executor? Does that make a difference when the underlying OS threads are still shared between the executors?

blink1073: It does if the OS thread limit is much higher than the default thread executor's thread limit.

NoahStapp: Good point, I like that idea then.

blink1073: I would also suggest making a utility function for `getaddrinfo` to avoid repeating this ugly block everywhere. ;)

pymongo/asynchronous/helpers.py
```diff
@@ -68,6 +70,16 @@ async def inner(*args: Any, **kwargs: Any) -> Any:
     return cast(F, inner)
 
 
+async def getaddrinfo(host, port, **kwargs):
+    if not _IS_SYNC:
+        loop = asyncio.get_running_loop()
+        return await loop.getaddrinfo(  # type: ignore[assignment]
```
blink1073, line 76: Shouldn't we be using `run_in_executor` here instead as well?

NoahStapp: Good catch sorry, juggling too many changes at once 😅

pymongo/_asyncio_executor.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+"""A separate ThreadPoolExecutor instance used internally to avoid competing for resources with the default asyncio ThreadPoolExecutor
+that user code will use."""
+
+from __future__ import annotations
+
+from concurrent.futures import ThreadPoolExecutor
+
+_PYMONGO_EXECUTOR = ThreadPoolExecutor(thread_name_prefix="PYMONGO_EXECUTOR-")
```
ShaneHarvey, line 22: I'm not sure I like this approach because we now have a thread pool that hangs around forever even after all clients have been closed.

My other comment was more around adding guidance for potential errors, not for changing our implementation. Like something that says "if your app runs into "XXX" error consider this may mean your app's default loop executor is under provisioned. Consider increasing the size of this thread pool or ..."

ShaneHarvey (changes requested):

pymongo/_asyncio_executor.py
NoahStapp: It could be difficult to distinguish when this issue occurs, having a separate thread pool for our internal use will help mitigate how common it is. An extra thread pool instance shouldn't be expensive to have a reference to for the lifetime of the application.

ShaneHarvey: Personally I prefer we go with the loop.getaddrinfo approach because it avoids adding the complexity of managing our own thread pool. It's not really kosher to leave a threadpool open even if the threads are "idle". The limitation in loop.getaddrinfo is also implementation detail that could be fixed at any point.

I expect it will be clear when this issue occurs because a timeout error caused by threadpool starvation looks different than a real DNS timeout error. It should be simple to add an example to our docs by:
1. saturating the executor with long running tasks
2. then attempting run a client command
3. record the error

NoahStapp: I don't see much complexity in managing our own thread pool, but I totally understand the desire to not have an extra pool lying around. I'll revert back to using `loop.getaddrinfo()` once I have a good example for our docs.

NoahStapp: After investigating, I believe the docs are slightly misleading: what actually happens when the executor pool is fully saturated is any `loop.getaddrinfo()` calls block until a thread is freed up for use. There's no timeout mechanism inherent to the executor pool. We could add our own timeout to every `loop.run_in_executor()` call to prevent users from accidentally blocking the driver forever if they saturate the default executor permanently, but then we would cause timeouts to occur whenever the response is too slow.

ShaneHarvey: Thanks for investigating, I agree with that. The cpython ticket referencing anyio so that could explain the difference.

pymongo/synchronous/helpers.py
```diff
@@ -68,6 +70,24 @@ def inner(*args: Any, **kwargs: Any) -> Any:
     return cast(F, inner)
 
 
+def getaddrinfo(
```
ShaneHarvey, line 73: nit: let's rename this `_getaddrinfo`.

ShaneHarvey (changes requested):

blink1073 (approved):
LGTM!

ShaneHarvey (approved):

PR #2070, opened 2025-01-21
PYTHON-5049 Drop support for PyPy 3.9

blink1073:


.evergreen/generated_configs/variants.yml
```diff
@@ -460,15 +460,6 @@ buildvariants:
       test_encryption: "true"
       test_encryption_pyopenssl: "true"
       PYTHON_BINARY: /opt/python/3.12/bin/python3
-  - name: encryption-rhel8-pypy3.9
```
blink1073, line 463: Note: there is still a pypy 3.10 variant above.

.evergreen/generated_configs/variants.yml
```diff
@@ -1164,17 +1135,6 @@ buildvariants:
     expansions:
       COVERAGE: coverage
       PYTHON_BINARY: /opt/python/3.12/bin/python3
-  - name: test-rhel8-pypy3.9
```
blink1073, line 1167: Same here

test/asynchronous/test_collection.py
```diff
@@ -133,13 +133,10 @@ def test_getattr(self):
 
     def test_iteration(self):
         coll = self.db.coll
-        if "PyPy" in sys.version and sys.version_info < (3, 8, 15):
-            msg = "'NoneType' object is not callable"
+        if _IS_SYNC:
```
NoahStapp, line 136: We can remove this `if-else` block now.

NoahStapp (changes requested):

test/asynchronous/test_collection.py
blink1073: Fixed

blink1073:
Link failure is unrelated, `https://www.pythonanywhere.com/` appears to be down (I'll hold off on creating a ticket in case it is just temporary).

PR #2074, opened 2025-01-23
PYTHON-5055 - Convert test_client.py to unittest

NoahStapp:
Main points of note:
1. All stateful `unittest` methods have been converted to `pytest` fixtures.
2. Every async test and fixture runs on the same loop that exists for the entire test session. This is separate from the pytest caching scope for fixtures, which is controlled by `scope` instead of `loop_scope`.
3. `ClientContext` has been kept as a metadata store for our tests served by a session-scoped fixture.
4. `unit` and `integration` pytest marks have been added. Going forward, all test classes should be marked with the appropriate one.


blink1073:
evergreen retry

sleepyStick (commented):
I think the Python changes look good to me. (I found things that I needed to change on my PR / test conversions as a result of reviewing this so that was helpful haha)

test/test_client.py
```diff
@@ -609,42 +611,47 @@ def test_detected_environment_logging(self, mock_get_hosts):
...
+                    for host in normal_hosts:
+                        simple_client(host)
+                    for host in srv_hosts:
+                        mock_get_hosts.return_value = [(host, 1)]
+                        simple_client(host)
+                    simple_client(multi_host)
+
+
+@pytest.mark.usefixtures("require_integration")
+@pytest.mark.integration
```
ShaneHarvey, line 651: Is there a simple way to avoid needing to mark the test as "integration" twice?

.evergreen/run-tests.sh
```diff
@@ -265,15 +266,39 @@ PIP_QUIET=0 uv run ${UV_ARGS[*]} --with pip pip list
 if [ -z "$GREEN_FRAMEWORK" ]; then
     # Use --capture=tee-sys so pytest prints test output inline:
     # https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html
-    PYTEST_ARGS="-v --capture=tee-sys --durations=5 $TEST_ARGS"
+    PYTEST_ARGS=("-v" "--capture=tee-sys" "--durations=5" "${TEST_ARGS[@]}")
     if [ -n "$TEST_SUITES" ]; then
-      PYTEST_ARGS="-m $TEST_SUITES $PYTEST_ARGS"
+      # Workaround until unittest -> pytest conversion is complete
+      if [[ "$TEST_SUITES" == *"default_async"* ]]; then
+        ASYNC_PYTEST_ARGS=("-m asyncio" "--junitxml=xunit-results/TEST-asyncresults.xml" "${PYTEST_ARGS[@]}")
```
ShaneHarvey, line 273: What about coverage? Does that pick up both test runs automatically?

test/test_client.py
NoahStapp: `@pytest.mark.usefixtures("require_integration")` isn't actually applying a mark, it's calling the `require_integration` as a requirement for the decorated test. 

`pytest -m require_integration` won't select any tests, but `pytest -m integration` will.

.evergreen/run-tests.sh
NoahStapp: The coverage runs pick up both test cases: https://spruce.mongodb.com/task/mongo_python_driver_test_rhel8_python3.9_cov_test_8.0_replica_set_noauth_ssl_sync_async_patch_dc182310dabff470dbbfc7da3c09eb6a4e08dfed_6797d0671d810000071bc171_25_01_27_18_28_57?execution=0&sortBy=STATUS&sortDir=ASC.

.evergreen/run-tests.sh
```diff
@@ -265,15 +266,39 @@ PIP_QUIET=0 uv run ${UV_ARGS[*]} --with pip pip list
...
+    # shellcheck disable=SC2048
+    uv run ${UV_ARGS[*]} pytest "${PYTEST_ARGS[@]}"
+    exit_code=$?
+
     # shellcheck disable=SC2048
-    uv run ${UV_ARGS[*]} pytest $PYTEST_ARGS
+    uv run ${UV_ARGS[*]} pytest "${ASYNC_PYTEST_ARGS[@]}" "--collect-only"
+    collected=$?
+    set -o errexit
+    # If we collected at least one async test, run all collected tests
```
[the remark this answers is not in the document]
NoahStapp: Good point. My original intention was to avoid the error-handling that would require, but this approach doesn't avoid that either.

sleepyStick (approved):

test/test_client.py
ShaneHarvey: I see this comment from the PR description is related:
> unit and integration pytest marks have been added. Going forward, all test classes should be marked with the appropriate one.

If integration tests need to be tagged with `require_integration`, what's the benefit of using the unit vs integration marks?

.evergreen/run-tests.sh
ShaneHarvey: This shows it's not working as expected. The second coverage run will overwrite the `.coverage` output file by default: https://coverage.readthedocs.io/en/7.6.10/cmd.html#data-file

test/test_client.py
NoahStapp: Using the `integration` and `unit` marks allows us to run only that set of tests. If we want to add additional subsets of tests that cross suites in the future, having those marks will be a good first step.

.evergreen/run-tests.sh
NoahStapp: Good catch, does it make sense to use `--append` to keep the single file?

NoahStapp: The alternative would be to use a separate file with `COVERAGE_FILE` and then `coverage combine` the two.

PR #2086, opened 2025-01-28
PYTHON-5071 Use one event loop for all asyncio tests

ShaneHarvey:
PYTHON-5071 Use one event loop for all asyncio tests

### DIFF
--- a/test/asynchronous/__init__.py
+++ b/test/asynchronous/__init__.py
@@ -869,6 +869,28 @@ class AsyncClientContext:
 # Reusable client context
 async_client_context = AsyncClientContext()
 
+# Global event loop for async tests.
+LOOP = None
+
+
+def get_loop() -> asyncio.AbstractEventLoop:
+    """Get the test suite's global event loop."""
+    global LOOP
+    if LOOP is None:
+        try:
+            LOOP = asyncio.get_running_loop()
+        except RuntimeError:
+            # no running event loop, fallback to get_event_loop.
+            try:
+                # Ignore DeprecationWarning: There is no current event loop
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore", DeprecationWarning)
+                    LOOP = asyncio.get_event_loop()
+            except RuntimeError:
+                LOOP = asyncio.new_event_loop()
+                asyncio.set_event_loop(LOOP)
+    return LOOP
+
 
 class AsyncPyMongoTestCase(unittest.TestCase):
     if not _IS_SYNC:
@@ -880,7 +902,6 @@ class AsyncPyMongoTestCase(unittest.TestCase):
         async def asyncTearDown(self):
             pass
 
-        # See IsolatedAsyncioTestCase.addAsyncCleanup.
         def addAsyncCleanup(self, func, /, *args, **kwargs):
             self.addCleanup(*(func, *args), **kwargs)
 
@@ -1232,31 +1253,10 @@ class AsyncMockClientTest(AsyncUnitTest):
         await super().asyncTearDown()
 
 
-LOOP = None
-
-
-def get_loop() -> asyncio.AbstractEventLoop:
-    global LOOP
-    if LOOP is None:
-        try:
-            LOOP = asyncio.get_running_loop()
-        except RuntimeError:
-            # no running event loop, fallback to get_event_loop.
-            try:
-                # Ignore DeprecationWarning: There is no current event loop
-                with warnings.catch_warnings():
-                    warnings.simplefilter("ignore", DeprecationWarning)
-                    LOOP = asyncio.get_event_loop()
-            except RuntimeError:
-                LOOP = asyncio.new_event_loop()
-                asyncio.set_event_loop(LOOP)
-    return LOOP
-
-
 async def async_setup():
     if not _IS_SYNC:
-        global LOOP
-        LOOP = asyncio.get_running_loop()
+        # Set up the event loop.
+        get_loop()
     await async_client_context.init()
     warnings.resetwarnings()
     warnings.simplefilter("always")
--- a/test/asynchronous/__init__.py
+++ b/test/asynchronous/__init__.py
@@ -874,25 +874,6 @@ class AsyncPyMongoTestCase(unittest.TestCase):
     if not _IS_SYNC:
         # An async TestCase that uses a single event loop for all tests.
         # Inspired by IsolatedAsyncioTestCase.
-        def __init__(self, methodName="runTest"):
-            super().__init__(methodName)
-            self._loop = None
-
-        @property
-        def loop(self):
-            if self._loop:
-                return self._loop
-            try:
-                with warnings.catch_warnings():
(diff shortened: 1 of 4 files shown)


NoahStapp:
The PyPy variant failure looks like the same asyncio SSL bug we've been tracking in https://github.com/pypy/pypy/issues/5131. The other failures are all `test.asynchronous.test_client.TestClient.test_continuous_network_errors`, but it's not clear why this change would cause that failure.

ShaneHarvey:
I suspect the test_continuous_network_errors errors are a side effect of having too many unopened clients or perhaps too many unawaited SDAM tasks. I removed the logic to reinit the client_context on each test. Let's see if that fixes it.

ShaneHarvey:
I also fixed an issue in loop management which was causing some tests to encounter `RuntimeError: <asyncio.locks.Condition object at 0x1095c2250 [locked]> is bound to a different event loop` in asyncTearDown. The fix was make the loop initialization lazy sot hat it gets created on the first async test, instead of in `__init__` which gets called at when collecting the tests.

ShaneHarvey:
I went back to package scope because I believe that's actually what we want. With "session" scope, the async teardown doesn't happen until after all the _sync_ tests run which causes problems because the client has been paused that whole time. This actually highlights a problem with async MongoClient, if the loop is paused for a while and then resumed the client's SDAM state will be out of date. This is similar to the sigstop/sigcont behavior.

test/asynchronous/__init__.py
```diff
@@ -1203,7 +1232,31 @@ async def asyncTearDown(self) -> None:
...
+            except RuntimeError:
+                LOOP = asyncio.new_event_loop()
+                asyncio.set_event_loop(LOOP)
+    return LOOP
+
+
 async def async_setup():
+    if not _IS_SYNC:
+        global LOOP
+        LOOP = asyncio.get_running_loop()
```
NoahStapp, line 1259: Is this supposed to be `get_loop` instead?

blink1073: No, this is the correct method to call within an async function.

NoahStapp: Good catch. I think we still have a loop mismatch here, although I might be misunderstanding: this will use the loop pytest creates for our package-scoped `test_setup_and_teardown` fixture. Then, our first async testcase will create its own separate loop that will be used for each test. So our async_client_context will be on one pytest-managed loop, and all of our test cases will be on one self-managed loop created by `get_loop`.

ShaneHarvey: Either one is fine because get_loop calls get_running_loop first but I updated to use get_loop to be more consistent.

ShaneHarvey: Attempting to use "package" in pyproject.toml results in this error:
```
____________________________________________________ ERROR at setup of TestCommon.test_mongo_client ____________________________________________________
file /Users/shane/git/mongo-python-driver/test/asynchronous/test_common.py, line 135
      async def test_mongo_client(self):
          pair = await async_client_context.pair
          m = await self.async_rs_or_single_client(w=0)
          coll = m.pymongo_test.write_concern_test
          await coll.drop()
          doc = {"_id": ObjectId()}
          await coll.insert_one(doc)
          self.assertTrue(await coll.insert_one(doc))
          coll = coll.with_options(write_concern=WriteConcern(w=1))
              await coll.insert_one(doc)

          m = await self.async_rs_or_single_client()
          coll = m.pymongo_test.write_concern_test
          new_coll = coll.with_options(write_concern=WriteConcern(w=0))
          self.assertTrue(await new_coll.insert_one(doc))
              await coll.insert_one(doc)

          m = await self.async_rs_or_single_client(
              f"mongodb://{pair}/", replicaSet=async_client_context.replica_set_name
          )

          coll = m.pymongo_test.write_concern_test
              await coll.insert_one(doc)
          m = await self.async_rs_or_single_client(
              f"mongodb://{pair}/?w=0", replicaSet=async_client_context.replica_set_name
          )

          coll = m.pymongo_test.write_concern_test
          await coll.insert_one(doc)

          # Equality tests
          direct = await connected(await self.async_single_client(w=0))
          direct2 = await connected(
              await self.async_single_client(f"mongodb://{pair}/?w=0", **self.credentials)
          )
          self.assertEqual(direct, direct2)
          self.assertFalse(direct != direct2)
file /Users/shane/git/mongo-python-driver/test/asynchronous/conftest.py, line 25
  @pytest_asyncio.fixture(scope="package", autouse=True)
  async def test_setup_and_teardown():
      await async_setup()
      yield
      await async_teardown()
E       fixture 'test/asynchronous::<event_loop>' not found
>       available fixtures: _session_event_loop, _unittest_setUpClass_fixture_TestCommon, anyio_backend, anyio_backend_name, anyio_backend_options, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, doctest_namespace, event_loop, event_loop_policy, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, test/asynchronous/test_common.py::<event_loop>, test/asynchronous/test_common.py::TestCommon::<event_loop>, test_setup_and_teardown, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, unused_tcp_port, unused_tcp_port_factory, unused_udp_port, unused_udp_port_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

"session" seems to be the right thing for what we want there.

ShaneHarvey:
Ready for another look.

test/asynchronous/__init__.py
NoahStapp: I opened https://github.com/pytest-dev/pytest-asyncio/issues/1052 about that exact error, it persists even if you explicitly mark a fixture as `package`. 

With any pytest-asyncio scope, we'll still have two separate loops. That doesn't seem to be causing test failures, but it's something to be aware of for future debugging 😅

ShaneHarvey: Thanks I subscribed to that issue. However I did verify that the current code only ever uses 1 loop for the whole async test suite. Printing the loop that pytest-asyncio creates shows this: `event_loop_fixture_id='_session_event_loop'` which means the `asyncio_default_fixture_loop_scope="session"` in pyproject is behaving as expected.

NoahStapp: You mean that `get_loop` is picking up and using pytest's `_session_event_loop`? Or that that loop is separate from the loop the tests use that `get_loop` creates?

NoahStapp (approved):

test/asynchronous/__init__.py
ShaneHarvey: The first one.

seifertm: > I opened [pytest-dev/pytest-asyncio#1052](https://github.com/pytest-dev/pytest-asyncio/issues/1052) about that exact error, it persists even if you explicitly mark a fixture as `package`.

pytest-asyncio v1.0.0 addresses this issue :)

PR #2088, opened 2025-01-28
PYTHON-5075 Convert test.test_csot to async

sleepyStick:
Convert test.test_csot to async


test/asynchronous/test_csot.py
```diff
@@ -0,0 +1,118 @@
...
+from __future__ import annotations
+
+import os
+import sys
+from pathlib import Path
+
+sys.path[0:0] = [""]
+
+from test.asynchronous import AsyncIntegrationTest, async_client_context, unittest
+from test.unified_format import generate_test_classes
```
NoahStapp, line 25:
```suggestion
from test.asynchronous.unified_format import generate_test_classes
```

NoahStapp: The synchronous one doesn't work with async tests.

NoahStapp (changes requested):

test/asynchronous/test_csot.py
sleepyStick: didn't realize that, thanks for pointing it out!

NoahStapp (approved):

NoahStapp:
The async CSOT tests are continuing to be extremely flakey, even on consecutive runs of the same variant and platform. Some failures are due to the test runner never picking up `commandStartedEvents`, some are due to the expected `TimeoutError` never occurring.

test/asynchronous/unified_format.py
```diff
@@ -1387,15 +1387,16 @@ async def run_scenario(self, spec, uri=None):
...
-        self.addAsyncCleanup(self.kill_all_sessions)
 
         if "csot" in self.id().lower():
             # Retry CSOT tests up to 2 times to deal with flakey tests.
             attempts = 3
             for i in range(attempts):
                 try:
                     return await self._run_scenario(spec, uri)
-                except AssertionError:
+                except (AssertionError, OperationFailure) as exc:
```
NoahStapp, line 1397: Do we want to do this for all CSOT tests, or only the asynchronous ones?

NoahStapp (changes requested):

test/asynchronous/unified_format.py
sleepyStick: OHHH good point, probably just the async ones right? I don't think I've seen flakey sync tests failing with OperationFailure

NoahStapp: Agreed, no need to change the synchronous behavior if it's not seeing the same issue.

sleepyStick: not sure if it was the best way to only retry in async, but i just changed the logic of the if statement inside the except

test/asynchronous/unified_format.py
```diff
@@ -1387,15 +1387,20 @@ async def run_scenario(self, spec, uri=None):
...
+                    if not (
+                        isinstance(exc, OperationFailure)
+                        and not _IS_SYNC
+                        and "failpoint" in exc._message
```
NoahStapp, line 1401: Try something like this instead:

```
if isinstance(exc, OperationFailure) and (_IS_SYNC or "failpoint" not in exc._message):
	raise
```

That way it's clear that we raise any OperationFailures that are either thrown by a sync test or don't have the failpoint message.

NoahStapp (changes requested):

test/asynchronous/unified_format.py
sleepyStick: ah yeah, that's cleaner

NoahStapp (approved):
Great work!

PR #2094, opened 2025-01-29
PYTHON-4864 - Create async version of SpecRunnerThread

NoahStapp:


test/asynchronous/utils_spec_runner.py
sleepyStick, line 1: not loving the duplicated code between sync and async but its because the async version needs some more methods? If so, then i understand and can live with it >.<

NoahStapp: The async version doesn't implement `threading.Thread`, but it still needs to match the same API as the synchronous version. Let me see if I can reduce some of the duplication though.

NoahStapp: Did some refactoring, much less duplication now.

sleepyStick (approved):
awesome work!

test/asynchronous/utils_spec_runner.py
```diff
@@ -47,46 +47,68 @@
...
+    async def run(self):
         while not self.stopped or self.ops:
             if not self.ops:
-                with self.cond:
-                    self.cond.wait(10)
+                async with self.cond:
+                    if _IS_SYNC:
+                        await self.cond.wait(10)  # type: ignore[call-arg]
+                    else:
+                        await asyncio.wait_for(self.cond.wait(), timeout=10)  # type: ignore[arg-type]
```
ShaneHarvey, line 104: Let's use our `_async_cond_wait` compat function here to avoid the branching.

test/asynchronous/unified_format.py
```diff
@@ -1177,16 +1177,16 @@ def primary_changed() -> bool:
 
         wait_until(primary_changed, "change primary", timeout=timeout)
 
-    def _testOperation_runOnThread(self, spec):
+    async def _testOperation_runOnThread(self, spec):
         """Run the 'runOnThread' operation."""
         thread = self.entity_map[spec["thread"]]
-        thread.schedule(lambda: self.run_entity_operation(spec["operation"]))
+        await thread.schedule(functools.partial(self.run_entity_operation, spec["operation"]))
```
ShaneHarvey, line 1183: Are we running any async unified tests that use runOnThread?

ShaneHarvey (changes requested):

test/asynchronous/unified_format.py
NoahStapp: The SDAM unified tests are the only ones that use `runOnThread`. Those are currently slated to be converted to async, yes.

test/asynchronous/utils_spec_runner.py
```diff
@@ -47,46 +47,77 @@
...
+if _IS_SYNC:
+    PARENT = threading.Thread
+else:
+    PARENT = object
 
-class SpecRunnerThread(threading.Thread):
-    def __init__(self, name):
-        super().__init__()
+
+class ConcurrentRunner(PARENT):
```
ShaneHarvey, line 64: Let's move this to utils.py since it will be used by non-spec tests.

test/asynchronous/utils_spec_runner.py
```diff
@@ -47,46 +47,77 @@
...
+    def __init__(self, name, *args, **kwargs):
+        if _IS_SYNC:
+            super().__init__(*args, **kwargs)
         self.name = name
+        self.stopped = False
```
ShaneHarvey, line 69: Stopped is never set to True by this class.

ShaneHarvey (changes requested):

test/asynchronous/utils_spec_runner.py
NoahStapp: Let's delay that move to [PYTHON-5113](https://jira.mongodb.org/browse/PYTHON-5113), it'll involve a lot of git changes.

ShaneHarvey: Okay then what about helpers.py?

ShaneHarvey (approved):

PR #2096, opened 2025-01-29
PYTHON-5079 Convert test.test_dns to async

sleepyStick:


test/asynchronous/test_dns.py
```diff
@@ -0,0 +1,222 @@
...
+            if hostname == "localhost":
+                copts = async_client_context.default_client_options.copy()
+                # Remove tls since SRV parsing should add it automatically.
+                copts.pop("tls", None)
+                if async_client_context.tls:
+                    # Our test certs don't support the SRV hosts used in these
+                    # tests.
+                    copts["tlsAllowInvalidHostnames"] = True
+
+                client = AsyncPyMongoTestCase.unmanaged_simple_client(uri, **copts)
```
NoahStapp, line 145:
```suggestion
                client = self.simple_client(uri, **copts)
```

test/asynchronous/test_dns.py
```diff
@@ -0,0 +1,222 @@
...
+                    )
+                if num_hosts is not None:
+                    await async_wait_until(
+                        lambda: num_hosts == len(client.nodes), "wait to connect to num_hosts"
+                    )
+                if test_case.get("ping", True):
+                    await client.admin.command("ping")
+                # XXX: we should block until SRV poller runs at least once
+                # and re-run these assertions.
+                await client.close()
```
NoahStapp, line 162: This can be removed since our clients are closed automatically when using the helper methods.

NoahStapp (changes requested):

test/asynchronous/test_dns.py
sleepyStick: yeah, make sense :)

test/asynchronous/test_dns.py
sleepyStick: how do we decide when to use helper methods vs the unmanaged clients?

NoahStapp: Generally you should always use helper methods unless we need a client somewhere entirely out of a unittest class context. Those situations should be very rare in our test suites.

NoahStapp (approved):

PR #2097, opened 2025-01-29
PYTHON-5080 - Convert test.test_examples to async

NoahStapp:


NoahStapp:
The remaining failure is 
[test.asynchronous.test_concurrency.TestAsyncConcurrency.test_concurrency](https://spruce.mongodb.com/task/mongo_python_driver_test_rhel8_python3.13_cov_test_4.4_sharded_cluster_auth_ssl_sync_async_patch_1784e2c4b9c7e5efbed1796e81e37fa49f8845f0_679b9a02056e0a00073a4a08_25_01_30_15_25_56?execution=0&sortBy=STATUS&sortDir=ASC), which is a known flakey test that depends on network and platform conditions.

test/test_examples.py
```diff
@@ -736,58 +741,108 @@ def test_delete(self):
...
+                cursor.close()
+            finally:
+                done = True
+                t
 
     def test_aggregate_examples(self):
         db = self.db
 
         # Start Aggregation Example 1
+
```
sleepyStick, line 845: doesn't matter that much, but new line?

sleepyStick (approved):

test/test_examples.py
NoahStapp: Good catch!

sleepyStick (approved):

test/test_examples.py
```diff
@@ -1347,20 +1401,37 @@ def test_snapshot_query(self):
...
-                [{"$match": {"adoptable": True}}, {"$count": "adoptableDogsCount"}], session=s
-            ).next()["adoptableDogsCount"]
+            adoptablePetsCount = (
+                (
+                    db.cats.aggregate(
+                        [{"$match": {"adoptable": True}}, {"$count": "adoptableCatsCount"}],
+                        session=s,
+                    )
+                ).next()
+            )["adoptableCatsCount"]
```
ShaneHarvey, line 1425: Is this really ruff's formatting? Ew.

test/test_examples.py
```diff
@@ -736,53 +741,102 @@ def test_delete(self):
...
+            db = self.db
+            done = False
+
+            def insert_docs():
+                nonlocal done
+                while not done:
+                    db.inventory.insert_one({"username": "alice"})
+                    db.inventory.delete_one({"username": "alice"})
+
+            t = asyncio.create_task(insert_docs())
```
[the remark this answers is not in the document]
NoahStapp: Similar to how `SpecRunnerThread/SpecRunnerTask` work? We could move that abstraction up to be a generic class we use everywhere we need a similar pattern.

test/test_examples.py
NoahStapp: Yeah it's not great.

test/test_examples.py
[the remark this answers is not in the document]
NoahStapp: Updated https://github.com/mongodb/mongo-python-driver/pull/2094 for this use case.

sleepyStick (approved):

test/asynchronous/helpers.py
```diff
@@ -404,5 +404,8 @@ def is_alive(self):
 
     async def run(self):
         if self.target:
-            await self.target()
+            if self.args:
+                await self.target(*self.args)
+            else:
+                await self.target()
```
ShaneHarvey, line 410: `await self.target(*self.args)` will work even if self.args is an empty list so this code can be simplified.

test/asynchronous/helpers.py
```diff
@@ -404,5 +404,8 @@ def is_alive(self):
 
     async def run(self):
         if self.target:
```
ShaneHarvey, line 406: When would "target" ever be None?

ShaneHarvey (changes requested):

test/asynchronous/helpers.py
NoahStapp: Good point, we'd want it to throw an error here instead of being a silent no-op.

test/test_examples.py
```diff
@@ -747,7 +753,7 @@ def insert_docs():
                 db.inventory.insert_one({"username": "alice"})
                 db.inventory.delete_one({"username": "alice"})
```
ShaneHarvey, line 754: We should probably put a short sleep in here (5ms maybe) to avoid a tight loop here.

ShaneHarvey (changes requested):

test/asynchronous/helpers.py
```diff
@@ -403,6 +403,5 @@ def is_alive(self):
             return not self.stopped
 
     async def run(self):
-        if self.target:
-            await self.target()
+        await self.target(*self.args)
         self.stopped = True
```
ShaneHarvey, line 407: Do we need to set stopped to try even when target raises? If so I suggest using try/finally.

ShaneHarvey (commented):
Sorry one last question.

test/asynchronous/helpers.py
NoahStapp: Good point, I'd say so.

ShaneHarvey (approved):

PR #2100, opened 2025-01-30
PYTHON-5084 - Convert test.test_heartbeat_monitoring to async

NoahStapp:


test/test_heartbeat_monitoring.py
```diff
@@ -50,20 +56,16 @@ def _check_with_socket(self, *args, **kwargs):
...
-                self.assertEqual(actual.connection_id, responses[0])
-                if expected != "ServerHeartbeatStartedEvent":
-                    if isinstance(actual.reply, Hello):
-                        self.assertEqual(actual.duration, 99)
-                        self.assertEqual(actual.reply._doc, responses[1])
-                    else:
-                        self.assertEqual(actual.reply, responses[1])
-
-        finally:
-            m.close()
```
sleepyStick, line 66: huh a try finally block is...wacky HAHA i wonder how that came to be

sleepyStick (approved):
assuming evergreen tests pass, lgtm!

test/test_heartbeat_monitoring.py
NoahStapp: `try...finally` without an except is something we've used to ensure a resource is cleaned up even if an exception occurs unexpectedly. It looks pretty goofy though yeah.

sleepyStick: i see, that makes sense

sleepyStick (approved):

PR #2101, opened 2025-01-30
PYTHON-5085 - Convert test.test_index_management to async

NoahStapp:


test/test_index_management.py
```diff
@@ -82,23 +90,25 @@ class SearchIndexIntegrationBase(PyMongoTestCase):
 
     @classmethod
     def setUpClass(cls) -> None:
-        super().setUpClass()
```
sleepyStick, line 85: just confirming, but i'm guessing super doesn't have a setUpClass?

sleepyStick (approved):
assuming evergreen tests pass, lgtm!

PR #2103, opened 2025-01-30
PYTHON-5087 - Convert test.test_load_balancer to async

NoahStapp:


test/test_load_balancer.py
```diff
@@ -26,15 +28,27 @@
...
 
 # Location of JSON test specifications.
-TEST_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)), "load_balancer")
+if _IS_SYNC:
+    _TEST_PATH = os.path.join(pathlib.Path(__file__).resolve().parent, "load_balancer")
+else:
+    _TEST_PATH = os.path.join(pathlib.Path(__file__).resolve().parent.parent, "load_balancer")
 
 # Generate unified tests.
-globals().update(generate_test_classes(TEST_PATH, module=__name__))
```
sleepyStick, line 37: i don't think it really matters but `_TEST_PATH` vs `TEST_PATH`?

//should i be changing them to `TEST_PATH` (my brain would appreciate it if it was consistent across the entire test suite haha)

sleepyStick (approved):
great work

test/test_load_balancer.py
NoahStapp: Good catch. Let's stick with TEST_PATH for now. I'll open a ticket to rename all the ones named `_TEST_PATH` in one go.

NoahStapp:
Note: I've opened [PYTHON-5113](https://jira.mongodb.org/browse/PYTHON-5113) to refactor our test utils once all conversions are complete.

sleepyStick (dismissed):

test/utils.py
```diff
@@ -927,6 +928,22 @@ def run(self):
             raise
 
 
+class ExceptionCatchingTask:
+    """A Task that stores any exception encountered from its task."""
+
+    def __init__(self, target):
+        self.exc = None
+        self.target = target
+        self.task = create_task(self.run())
```
ShaneHarvey, line 937: create_task should be called in run() to mirror the behavior of Thread.

test/utils.py
```diff
@@ -927,6 +928,22 @@ def run(self):
...
+
+    async def run(self):
+        try:
+            await self.target()
```
ShaneHarvey, line 941: Awaiting the task should happen in join() to mirror the behavior of Thread.

test/asynchronous/test_load_balancer.py
```diff
@@ -0,0 +1,244 @@
...
+        await client.test_session_gc.test.find_one({}, session=session)
+        # Cleanup the transaction left open on the server unless we're
+        # testing serverless which does not support killSessions.
+        if not async_client_context.serverless:
+            self.addAsyncCleanup(self.client.admin.command, "killSessions", [session.session_id])
+        if async_client_context.load_balancer:
+            self.assertEqual(pool.active_sockets, 1)  # Pinned.
+
+        if _IS_SYNC:
+            thread = PoolLocker(pool)
```
ShaneHarvey, line 168: If we make the ExceptionCatchingTask changes I suggested, will that let use remove this branch?

ShaneHarvey (changes requested):

test/utils.py
NoahStapp: It should be called in `start()`  to mirror Thread, right?

ShaneHarvey: Yeah that's right. A Thread doesn't start running until start() is called.

test/asynchronous/test_load_balancer.py
NoahStapp: I believe so, yes.

ShaneHarvey (approved):
LGTM!

PR #2106, opened 2025-01-30
PYTHON-5090 Convert test.test_monitor to async

sleepyStick:


sleepyStick:
Okay i don't know why, but when sometimes i run pre-commit, syncho wants to change like..35 files most of which are one's i haven't touched obviously (its all like spacing related stuff) and i cant quite piece together when it works as i expect and when it doesn't work as expected??

test/asynchronous/test_monitor.py
```diff
@@ -0,0 +1,108 @@
...
+        executors.append(server._monitor._rtt_monitor._executor)
+    executors.append(client._kill_cursors_executor)
+    executors.append(client._topology._Topology__events_executor)
+    return [e for e in executors if e is not None]
+
+
+class TestMonitor(AsyncIntegrationTest):
+    async def create_client(self):
+        listener = ServerAndTopologyEventListener()
+        client = await self.unmanaged_async_single_client(event_listeners=[listener])
```
NoahStapp, line 56:
```suggestion
        client = await self.async_single_client(event_listeners=[listener])
```

Avoid using the `unmanaged` helper methods wherever possible.

NoahStapp (changes requested):

test/asynchronous/test_monitor.py
sleepyStick: looking at the tests, i thought it was intentional to use unmanaged because the tests want to explicitly close the client and test behavior after client close (or client del)?

NoahStapp: Does it cause failures to switch the call from unmanaged?

ShaneHarvey: Yes we have to used unmanaged here, otherwise the client will never get garbage collected.

ShaneHarvey: Adding client.close to the test cleanup will keep a reference to the client and prevent it from being GCd.

NoahStapp: Ah, the reference inside the cleanups list will cause the `del client` call to not GC the client?

NoahStapp: Oh wait: a GC'd AsyncMongoClient will trigger the unclosed client warning.

ShaneHarvey: We still want to make sure that a GC'd async client's tasks eventually get cleaned up.

sleepyStick:
okay in summary, we need the unmanaged async client for the test, because we want to test that the executors get GCd on client del and client.close. if we used managed, the client will not get GCd. on pypy GC is non-deterministic hence the failure. with that said, i believe the only failure is because of that and we should either skip this test in pypy (if so how?) or do something else to mitigate the failing test?

test/asynchronous/test_monitor.py
```diff
@@ -0,0 +1,108 @@
...
+            # Each executor stores a weakref to itself in _EXECUTORS.
+            executor_refs = [(r, r()._name) for r in _EXECUTORS.copy() if r() in executors]
+
+            del executors
+            del client
+
+            for ref, name in executor_refs:
+                await async_wait_until(
+                    partial(unregistered, ref), f"unregister executor: {name}", timeout=5
+                )
```
ShaneHarvey, line 76: Can we add an extra async_wait_until here that waits for the `__del__` warning to be raised? That would avoid the warning being accidentally raised later on in a subsequent test like we see in the most recent run: https://github.com/mongodb/mongo-python-driver/actions/runs/13466188001/job/37632412774?pr=2106

ShaneHarvey (changes requested):

test/asynchronous/test_monitor.py
ShaneHarvey: Like:
```
with warnings.catch_warnings(record=True) as w:
   ...
   wait until the "Call AsyncMongoClient.close() to safely shut down your client and free up resources" warning appears in w.
```

NoahStapp (approved):
Great work! LGTM, assuming passing tests.

sleepyStick:
The windows test failures are from `test_connection_monitoring` and `test_async_cancellation`, neither of which relate to this PR to my knowledge :)

ShaneHarvey (approved):
LGTM nice work!

PR #2107, opened 2025-01-31
PYTHON-5089 Convert test.test_mongos_load_balancing to async

sleepyStick:


sleepyStick:
Note: I believe the failing evergreen tests are unrelated to the changes I've made

test/asynchronous/test_mongos_load_balancing.py
```diff
@@ -0,0 +1,223 @@
...
+from pymongo.errors import AutoReconnect, InvalidOperation
+from pymongo.server_selectors import writable_server_selector
+from pymongo.topology_description import TOPOLOGY_TYPE
+
+_IS_SYNC = False
+
+
+@async_client_context.require_connection
+@async_client_context.require_no_load_balancer
+def asyncSetUpModule():
```
NoahStapp, line 39: Sadly there isn't an async version of `unittest.setUpModule`. I'd just add these decorators to an `asyncSetUp` method for `TestMongosLoadBalancing` instead since it's the only test class in this file.

test/asynchronous/test_mongos_load_balancing.py
```diff
@@ -0,0 +1,223 @@
...
+    pass
+
+
+if not _IS_SYNC:
```
NoahStapp, line 43: Let's standardize these checks to `if _IS_SYNC` for clarity.

test/asynchronous/test_mongos_load_balancing.py
```diff
@@ -0,0 +1,223 @@
...
+
+    class SimpleOp:
+        def __init__(self, client):
+            self.task: asyncio.Task
```
NoahStapp, line 47:
```suggestion
            self.task: asyncio.Task = None
```

test/asynchronous/test_mongos_load_balancing.py
```diff
@@ -0,0 +1,223 @@
...
+            self.client = client
+            self.passed = False
+
+        def run(self):
+            self.client.db.command("ping")
+            self.passed = True  # No exception raised.
+
+
+async def do_simple_op(client, nthreads):
+    threads = [SimpleOp(client) for _ in range(nthreads)]
```
NoahStapp, line 74: What do you think about using `tasks` here instead of `threads`?

test/asynchronous/test_mongos_load_balancing.py
sleepyStick: Okay yes, that's what I **wanted** (and did initially) but for some reason if its `if _IS_SYNC` first, type-checker assumes both definitions of `SimpleOp` inherit from threading.Thread and then insist that both implementations adhere to the thread api (in this case `join` must return a value) 
is that preferred? it felt weird to return a value for the sake of it in the async version of SimpleOp simply because its not used at all

test/asynchronous/test_mongos_load_balancing.py
sleepyStick: would I need to modify synchro to change `tasks` to `threads` to get this to work?

test/asynchronous/test_mongos_load_balancing.py
sleepyStick: Ah good to know, makes sense

test/asynchronous/test_mongos_load_balancing.py
sleepyStick: hmmm I don't love that because it'd mess with the type of `self.task` (specifically it'd have to be `Optional[asyncio.Task]`) in the interest of not making it optional, I moved the assignment from `start()` to the init. Does that work? Is that better?

test/asynchronous/test_mongos_load_balancing.py
NoahStapp: Our typing checks do, or the IDE's own highlighting does? Either way, we should have all checks be for `_IS_SYNC` whenever possible for clarity.

test/asynchronous/test_mongos_load_balancing.py
NoahStapp: We'd need to have it only do so within test files and with a specific token set to not also change things that happen to be called threads but aren't what we want to synchro. 

I'd rather have concurrent executors always be called `tasks` for consistency rather than have them be called `threads` in the async code.

test/asynchronous/test_mongos_load_balancing.py
NoahStapp: Can you show what you mean exactly?

sleepyStick: (I thought I committed and pushed HAHA sorry about that actually pushed now)

test/asynchronous/test_mongos_load_balancing.py
sleepyStick: makes sense, its tasks now :)

test/asynchronous/test_mongos_load_balancing.py
sleepyStick: both, our typing checks is what caught my attention first actually

test/asynchronous/test_mongos_load_balancing.py
```diff
@@ -34,17 +34,11 @@
...
-def asyncSetUpModule():
-    pass
-
-
 if not _IS_SYNC:
 
     class SimpleOp:
         def __init__(self, client):
-            self.task: asyncio.Task
+            self.task = asyncio.create_task(self.run())
```
NoahStapp, line 41: I prefer the earlier version, sorry for the churn.

NoahStapp (changes requested):

test/test_mongos_load_balancing.py
```diff
@@ -30,33 +31,48 @@
...
+
+    class SimpleOp(threading.Thread):
+        def __init__(self, client):
+            super().__init__()
+            self.client = client
+            self.passed = False
+
+        def run(self):
+            self.client.db.command("ping")
+            self.passed = True  # No exception raised.
```
ShaneHarvey, line 64: Can we settle on a common pattern for the threading.Thread -> Task conversions? I made some suggestions in https://github.com/mongodb/mongo-python-driver/pull/2103/.

NoahStapp: Thoughts on the approach in https://github.com/mongodb/mongo-python-driver/pull/2094?

test/asynchronous/test_mongos_load_balancing.py
sleepyStick: nope all good, i offered it as an alternative :)

test/test_mongos_load_balancing.py
sleepyStick: (i think my tab wasn't fully refreshed earlier so i'm just seeing Noah's comment right now) but I actually really like that approach!

NoahStapp (changes requested):
Can you refactor this to use `ConcurrentRunner` added by https://github.com/mongodb/mongo-python-driver/pull/2094?

test/asynchronous/helpers.py
```diff
@@ -395,7 +395,7 @@ def __init__(self, **kwargs):
         async def start(self):
             self.task = create_task(self.run(), name=self.name)
 
-        async def join(self, timeout: float | None = 0):  # type: ignore[override]
```
NoahStapp, line 398: Great find on this bug. It's not at all obvious and causes a lot of confusing behavior that's hard to spot.

test/test_mongos_load_balancing.py
```diff
@@ -30,14 +32,10 @@
 from pymongo.server_selectors import writable_server_selector
 from pymongo.topology_description import TOPOLOGY_TYPE
 
+_IS_SYNC = True
 
-@client_context.require_connection
-@client_context.require_no_load_balancer
-def setUpModule():
```
NoahStapp, line 36: We can keep these to improve performance when skipping this test suite.

test/asynchronous/test_mongos_load_balancing.py
```diff
@@ -0,0 +1,204 @@
...
+            try:
+                await client.db.command("ping")
+            except AutoReconnect:
+                # Second attempt succeeds.
+                await client.db.command("ping")
+
+            passed.append(True)
+
+        if _IS_SYNC:
+            threads = [threading.Thread(target=f) for _ in range(nthreads)]
```
NoahStapp, line 125: Can we use `ConcurrentRunner` for these as well?

NoahStapp (changes requested):

test/asynchronous/helpers.py
sleepyStick: yup 😭 it took about 20 print debug statements and a good nights rest and a whole lot of "this makes NO sense" until i figured it out HAHA but now that I've noticed it, i'm like why didn't i see it sooner!

test/test_mongos_load_balancing.py
sleepyStick: I moved it to the class's `setUp` because on the async side it'd require setUpModule to be awaited and i couldn't find an easy way to achieve that. I figured it was the only class in this module so it didn't make a difference to just move it into the class. But if you know how i could await this in async, I have no hesitations to bring it back!

NoahStapp: Oh whoops my bad, forgot how the wrappers interact with async. You're right!

sleepyStick: All good, wasn't sure if I was missing / forgetting what the trick was!

NoahStapp (approved):
Great work!

PR #2116, opened 2025-01-31
PYTHON-5071 Workaround PyPy SSL bug again

ShaneHarvey:
The previous changes in PYTHON-5071 somehow made the issue in PYTHON-5011 crop up more frequently. This change attempts to reduce the test failures caused by that pypy ssl bug.


ShaneHarvey:
This didn't help to resolve the PyPy issues. Closing.

PR #2117, opened 2025-02-01
PYTHON-5099 Convert test.test_sdam_monitoring_spec to async

sleepyStick:


test/test_sdam_monitoring_spec.py
```diff
@@ -267,32 +274,27 @@ class TestSdamMonitoring(IntegrationTest):
     test_client: MongoClient
     coll: Collection
 
-    @classmethod
     @client_context.require_failCommand_fail_point
-    def setUpClass(cls):
-        super().setUp(cls)
+    def setUp(self):
```
NoahStapp, line 278: We can keep `setUpClass` and have it handle the knobs and listener configuration. Those don't do network I/O and are constant across every testcase. The client configuration should be moved to `setUp` as you've done here.

NoahStapp (changes requested):

test/test_sdam_monitoring_spec.py
sleepyStick: got it, makes sense!

NoahStapp (approved):

PR #2123, opened 2025-02-01
PYTHON 5104 - Convert test.test_sessions_unified to async

sleepyStick:


test/asynchronous/test_sessions_unified.py
```diff
@@ -0,0 +1,40 @@
...
+from __future__ import annotations
+
+import os
+import sys
+from pathlib import Path
+
+sys.path[0:0] = [""]
+
+from test import unittest
+from test.unified_format import generate_test_classes
```
NoahStapp, line 25:
```suggestion
from test.asynchronous.unified_format import generate_test_classes
```

NoahStapp (changes requested):

NoahStapp (approved):

tools/synchro.py
```diff
@@ -226,6 +226,7 @@ def async_only_test(f: str) -> bool:
     "test_retryable_writes.py",
     "test_retryable_writes_unified.py",
     "test_session.py",
+    "test_session_unified.py",
```
NoahStapp, line 229:
```suggestion
    "test_sessions_unified.py",
```

NoahStapp (changes requested):

tools/synchro.py
sleepyStick: good catch! sorry about that!

NoahStapp (approved):

PR #2126, opened 2025-02-03
PYTHON-5107 - Convert test.test_streaming_protocol to async

NoahStapp:


test/test_streaming_protocol.py
```diff
@@ -78,7 +79,7 @@ def marked_unknown():
             def rediscovered():
                 return len(listener.matching(_discovered_node)) >= 1
 
-            # Topology events are published asynchronously
+            # Topology events are published synchronously
```
sleepyStick, line 82: comment changed?

NoahStapp: Good catch, changed to not get caught by the synchro script.

sleepyStick (approved):
lgtm!

PR #2127, opened 2025-02-03
PYTHON-5053 - AsyncMongoClient.close() should await all background tasks

NoahStapp:


ShaneHarvey:
I scheduled the windows standalone jobs to see if this fixes the "OSError: [WinError 10038] An operation was attempted on something that is not a socket" errors.

pymongo/asynchronous/monitor.py
```diff
@@ -111,6 +111,8 @@ async def close(self) -> None:
         open() restarts the monitor after closing.
         """
         self.gc_safe_close()
+        if not _IS_SYNC:
+            await self._executor.join()
```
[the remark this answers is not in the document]
ShaneHarvey: So we either need to remove self calls to close() or move the join() logic to another method.

ShaneHarvey (changes requested):

pymongo/asynchronous/monitor.py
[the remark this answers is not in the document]
NoahStapp: Good catch. It makes more sense to me to have a separate `Monitor.join()` method that we call in the async API whenever we call `Monitor.close()` from a non-monitor task.

pymongo/asynchronous/monitor.py
```diff
@@ -189,6 +189,10 @@ def gc_safe_close(self) -> None:
         self._rtt_monitor.gc_safe_close()
         self.cancel_check()
 
+    async def join(self, timeout: Optional[int] = None) -> None:
```
ShaneHarvey, line 192: Do we ever pass a timeout here?

pymongo/asynchronous/topology.py
```diff
@@ -520,6 +520,8 @@ async def _process_change(
             and self._description.topology_type not in SRV_POLLING_TOPOLOGIES
         ):
             await self._srv_monitor.close()
+            if not _IS_SYNC:
+                await self._srv_monitor.join()
```
ShaneHarvey, line 524: I think we should have a separate code path for join() all the way down. That way we first signal everything to shutdown, then we wait for everything to exit. Joining each task individually will slowdown the close().

ShaneHarvey (changes requested):

pymongo/asynchronous/topology.py
ShaneHarvey: I meant to put this comment on Topology.close().

ShaneHarvey: I think this can deadlock. It's not safe to call join() while holding the topology lock because the task we're attempting to join() may be blocking on acquiring the same lock.

pymongo/asynchronous/monitor.py
NoahStapp: Not currently, but we likely should once we're productionizing the async API for release. Currently there's too much variability between platforms and test suites to pick a good timeout number.

ShaneHarvey: Let's remove the timeout parameter since it's not used.

pymongo/asynchronous/topology.py
NoahStapp: You're suggesting that we make `executor.close()` not call `task.cancel()` and we do it inside the `join()` instead?

ShaneHarvey: No I'm suggesting we never call join() while holding a lock.

NoahStapp: Ah, something like this inside `Topology`?

```
async def join(self):
    # Join all monitors
    ...
```
And then we call `topology.join()` inside `AsyncMongoClient.close()`?

ShaneHarvey:
Also we should make sure that when a server is removed from the Topology for other reasons (besides close) that we also join() it properly. For example when a server is no longer found in the replica set config.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1565,6 +1565,10 @@ async def close(self) -> None:
             # TODO: PYTHON-1921 Encrypted MongoClients cannot be re-opened.
             await self._encrypter.close()
         self._closed = True
+        if not _IS_SYNC:
+            self._topology._monitor_tasks.append(self._kill_cursors_executor)  # type: ignore[arg-type]
+            join_tasks = [t.join() for t in self._topology._monitor_tasks]  # type: ignore[func-returns-value]
```
ShaneHarvey, line 1570: Reading `_monitor_tasks` like this is not thread safe. Anytime we iterate it we need to guard against the list being mutated from another thread, something like:
```
tasks = []
try:
    while self._topology._monitor_tasks:
        tasks.append(self._topology._monitor_tasks.pop())
except IndexError:
    pass
    
```

pymongo/synchronous/topology.py
```diff
@@ -241,6 +245,7 @@ def open(self) -> None:
                 # Close servers and clear the pools.
                 for server in self._servers.values():
                     server.close()
+                    self._monitor_tasks.append(server._monitor)
```
ShaneHarvey, line 248: I believe we only want to record these tasks on async. Otherwise we'll have an unbounded list of threads in the sync version.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1565,6 +1565,10 @@ async def close(self) -> None:
             # TODO: PYTHON-1921 Encrypted MongoClients cannot be re-opened.
             await self._encrypter.close()
         self._closed = True
+        if not _IS_SYNC:
+            self._topology._monitor_tasks.append(self._kill_cursors_executor)  # type: ignore[arg-type]
```
ShaneHarvey, line 1569: Let's avoid appending to the topology's private state here.

pymongo/asynchronous/monitor.py
```diff
@@ -189,6 +189,10 @@ def gc_safe_close(self) -> None:
         self._rtt_monitor.gc_safe_close()
         self.cancel_check()
 
+    async def join(self) -> None:
+        await self._executor.join()
+        await self._rtt_monitor.join()
```
ShaneHarvey, line 194: This should use gather too right?

ShaneHarvey (changes requested):

pymongo/asynchronous/mongo_client.py
NoahStapp: Are we supporting multithreaded async workloads? My understanding was that we are explicitly not supporting such use cases.

pymongo/synchronous/topology.py
NoahStapp: Good catch, missed this one.

pymongo/asynchronous/mongo_client.py
NoahStapp: Or is this a futureproofing suggestion for when we do the same joining process for synchronous tasks?

NoahStapp: Made this change in the interest of covering our bases and reducing future changes for the sync API.

ShaneHarvey: Oh sorry, I was let my sync brain bleed into the async code. Yeah async is single threaded so it's safe to iterate the list as along as there no yield points.

pymongo/asynchronous/topology.py
```diff
@@ -288,10 +294,22 @@ async def select_servers(
                 selector, server_timeout, operation, operation_id, address
             )
 
-            return [
+            servers = [
                 cast(Server, self.get_server_by_address(sd.address)) for sd in server_descriptions
             ]
 
+        if not _IS_SYNC and self._monitor_tasks:
```
ShaneHarvey, line 301: Worth putting a comment here to explain why this code is here. Also this should happen _before_ selecting the server. Doing it after will increase the risk of returning stale information.

pymongo/asynchronous/mongo_client.py
NoahStapp: Any objection to doing your suggested change anyway for the reasons I stated above?

pymongo/asynchronous/topology.py
```diff
@@ -288,10 +294,22 @@ async def select_servers(
...
+            join_tasks = []
+            try:
+                while self._monitor_tasks:
+                    join_tasks.append(self._monitor_tasks.pop())
+            except IndexError:
+                pass
+            join_tasks = [t.join() for t in join_tasks]  # type: ignore[func-returns-value]
+            await asyncio.gather(*join_tasks)
```
ShaneHarvey, line 309: Could you refactor this into a method that MongoClient.Close can call too?

ShaneHarvey (changes requested):

pymongo/asynchronous/topology.py
NoahStapp: The added delay added by the cleanup between selecting the server and actually returning it being the risk? Makes sense.

ShaneHarvey: Yep that's it.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1565,6 +1565,10 @@ async def close(self) -> None:
             # TODO: PYTHON-1921 Encrypted MongoClients cannot be re-opened.
             await self._encrypter.close()
         self._closed = True
+        if not _IS_SYNC:
+            await asyncio.gather(
```
ShaneHarvey, line 1569: I believe we should be using return_exceptions=True on all these gather() calls:
> If return_exceptions is True, exceptions are treated the same as successful results, and aggregated in the result list.

https://docs.python.org/3/library/asyncio-task.html#asyncio.gather

Otherwise we may accidentally propagate an exception we don't care about.

ShaneHarvey (changes requested):

pymongo/asynchronous/mongo_client.py
ShaneHarvey: At this point should we be using asyncio.wait() instead of gather()?

NoahStapp: Good point. I think we can safely use `asyncio.wait()` instead here, since we don't need the aggregated result returned by `gather`.

NoahStapp: Oh wait, `asyncio.wait()` explicitly does not support waiting for coroutines, only `Task` or `Future` objects. Sticking with gather seems a little simpler, even with `return_exceptions=True`.

pymongo/asynchronous/monitor.py
```diff
@@ -189,6 +189,11 @@ def gc_safe_close(self) -> None:
         self._rtt_monitor.gc_safe_close()
         self.cancel_check()
 
+    async def join(self) -> None:
+        await asyncio.gather(
+            *[self._executor.join(), self._rtt_monitor.join()], return_exceptions=True
```
ShaneHarvey, line 194: One last nit: could you remove the `*[]`?

ShaneHarvey (approved):
LGTM!

PR #2132, opened 2025-02-05
PYTHON-4993 - Reevaluate handling of asyncio.CancelledError

NoahStapp:


NoahStapp:
Some of these could be added back in with https://jira.mongodb.org/browse/PYTHON-4745 depending on what we decide on.

pymongo/_client_bulk_shared.py
```diff
@@ -75,5 +76,7 @@ def _throw_client_bulk_write_exception(
             )
             raise OperationFailure(errmsg, code, full_result)
     if isinstance(full_result["error"], BaseException):
+        if isinstance(full_result["error"], asyncio.CancelledError):
```
ShaneHarvey, line 79: Is this actually needed? Like do we catch BaseException where _throw_client_bulk_write_exception is called? Or is this added just to be safe?

NoahStapp: Good catch. We don't catch `BaseException` when `_throw_client_bulk_write_exception` is called, so this can be removed.

ShaneHarvey (dismissed):

ShaneHarvey (commented):
Are the test failures expected?

NoahStapp:
> Are the test failures expected?

No, looking into now.

ShaneHarvey (approved):

PR #2135, opened 2025-02-06
PYTHON-5111 Update datetime_conversion in docstrings of MongoClients

caseyclements:
Reflects the fact that it is now in kwargs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -178,7 +178,7 @@ documentation including narrative docs, and the [Sphinx docstring format](https:
 You can build the documentation locally by running:
 
 ```bash
-just docs-build
+just docs
 ```
 
 When updating docs, it can be helpful to run the live docs server as:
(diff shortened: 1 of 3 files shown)


ShaneHarvey:
Screenshot 2025-02-06 at 12 00 31 PM
The format looks a little off. Perhaps we need a colon at the end? And could we bold the line?

caseyclements:
The format was off before. The text in the last :param gets indented. It's weird, but  I played with it to make it look pretty decent, adding kwargs as a param, and then having the list below.

image

ShaneHarvey (changes requested):
There have been no code changes since my last review.

caseyclements:
> There have been no code changes since my last review.

That's right. What I am saying is that Sphinx adds a carriage return to the last :param. This was how it was before, too. My picture shows the docs as they have existing for however long. When I moved datetime_conversion down, this weird behavior appears for :param type_registry.  This is why I added :param kwargs, specifically because it included a big indented list before. There previously was a bold line, but it was redundant if any text was added to kwargs. 

Suggestions are welcome, given the constraint. I'm pretty happy with how it looks now.

ShaneHarvey:
Oh I get it now. Could we at least add a colon and bold "Additional optional parameters available as keyword arguments:". That would better match with the other sections like "Write Concern options:"

caseyclements:
@ShaneHarvey Done. Bold with colon.

ShaneHarvey (approved):

PR #2136, opened 2025-02-06
PYTHON-4745 - Test behavior of async task cancellation

NoahStapp:
…ation

### DIFF
--- a/pymongo/asynchronous/change_stream.py
+++ b/pymongo/asynchronous/change_stream.py
@@ -391,7 +391,7 @@ class AsyncChangeStream(Generic[_DocumentType]):
             if not _resumable(exc) and not exc.timeout:
                 await self.close()
             raise
-        except Exception:
+        except BaseException:
             await self.close()
             raise
 
--- a/pymongo/asynchronous/change_stream.py
+++ b/pymongo/asynchronous/change_stream.py
@@ -391,6 +391,7 @@ class AsyncChangeStream(Generic[_DocumentType]):
             if not _resumable(exc) and not exc.timeout:
                 await self.close()
             raise
+        # Catch KeyboardInterrupt, CancelledError, etc. and cleanup.
         except BaseException:
             await self.close()
             raise
--- a/test/asynchronous/test_async_cancellation.py
+++ b/test/asynchronous/test_async_cancellation.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Test that async cancellation performed by users raises the expected error."""
+"""Test that async cancellation performed by users clean up resources correctly."""
 from __future__ import annotations
 
 import asyncio
@@ -71,7 +71,7 @@ class TestAsyncCancellation(AsyncIntegrationTest):
 
         self.assertFalse(session.in_transaction)
 
-    async def test_async_cancellation_kills_cursor(self):
+    async def test_async_cancellation_closes_cursor(self):
         client = await self.async_rs_or_single_client()
         await connected(client)
         for _ in range(2):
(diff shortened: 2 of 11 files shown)


test/asynchronous/test_async_cancellation.py
```diff
@@ -0,0 +1,136 @@
...
+from test.utils import async_get_pool, delay, one
+
+sys.path[0:0] = [""]
+
+from test.asynchronous import AsyncIntegrationTest, async_client_context, connected
+
+
+class TestAsyncCancellation(AsyncIntegrationTest):
+    async def test_async_cancellation_closes_connection(self):
+        client = await self.async_rs_or_single_client(maxPoolSize=1)
```
ShaneHarvey, line 29: I think all of these tests can reuse the global test client instead of creating new ones.

ShaneHarvey (changes requested):

pymongo/asynchronous/change_stream.py
```diff
@@ -391,7 +391,7 @@ async def try_next(self) -> Optional[_DocumentType]:
             if not _resumable(exc) and not exc.timeout:
                 await self.close()
             raise
-        except Exception:
+        except BaseException:
```
ShaneHarvey, line 394: Everywhere we catch BaseException, can you add a one line comment to explain its intentional? Something like:
```
# Catch KeyboardInterrupt, CancelledError, etc. and cleanup.
```

test/asynchronous/test_async_cancellation.py
```diff
@@ -0,0 +1,132 @@
...
+from test.asynchronous import AsyncIntegrationTest, async_client_context, connected
+
+
+class TestAsyncCancellation(AsyncIntegrationTest):
+    async def test_async_cancellation_closes_connection(self):
+        pool = await async_get_pool(self.client)
+        await connected(self.client)
+        conn = one(pool.conns)
+        await self.client.db.test.insert_one({"x": 1})
+        self.addAsyncCleanup(self.client.db.test.drop)
```
ShaneHarvey, line 33: Let's use delete_many() instead of drop() for better perf.

test/asynchronous/test_async_cancellation.py
```diff
@@ -0,0 +1,132 @@
...
+        with self.assertRaises(asyncio.CancelledError):
+            await task
+
+        self.assertFalse(session.in_transaction)
+
+    @async_client_context.require_failCommand_blockConnection
+    async def test_async_cancellation_closes_cursor(self):
+        await connected(self.client)
+        for _ in range(2):
+            await self.client.db.test.insert_one({"x": 1})
```
ShaneHarvey, line 76: Here and below, use insert_many instead of insert_one in a for-loop for better perf.

test/asynchronous/test_async_cancellation.py
```diff
@@ -0,0 +1,132 @@
...
+
+sys.path[0:0] = [""]
+
+from test.asynchronous import AsyncIntegrationTest, async_client_context, connected
+
+
+class TestAsyncCancellation(AsyncIntegrationTest):
+    async def test_async_cancellation_closes_connection(self):
+        pool = await async_get_pool(self.client)
+        await connected(self.client)
```
ShaneHarvey, line 30: We can remove `connected` from these tests for better perf.

ShaneHarvey (changes requested):

test/asynchronous/test_async_cancellation.py
NoahStapp: There is no asynchronous `setUpClass/tearDownClass` equivalent in unittest sadly.

ShaneHarvey (approved):

PR #2145, opened 2025-02-12
PYTHON-5092 - Convert test.test_pooling to async

NoahStapp:


test/asynchronous/test_pooling.py
```diff
@@ -0,0 +1,599 @@
...
+        await super().asyncSetUp()
+        self.c = await self.async_rs_or_single_client()
+        db = self.c[DB]
+        await db.unique.drop()
+        await db.test.drop()
+        await db.unique.insert_one({"_id": "jesse"})
+        await db.test.insert_many([{} for _ in range(10)])
+
+    async def asyncTearDown(self):
+        await self.c.close()
```
sleepyStick, line 162: is this close needed? I thought `async_rs_or_single_client()` would close it?

test/test_pooling.py
```diff
@@ -15,18 +15,20 @@
 """Test built in connection-pooling with threads."""
 from __future__ import annotations
 
+import asyncio
```
ShaneHarvey, line 18: I thought synchro.py was supposed to remove asyncio imports?

test/test_pooling.py
```diff
@@ -15,18 +15,20 @@
...
 import gc
 import random
 import socket
 import sys
-import threading
 import time
+from test.helpers import ConcurrentRunner
```
ShaneHarvey, line 24: please move this down next to the other test imports.

test/asynchronous/test_pooling.py
NoahStapp: Good catch! Removed.

test/test_pooling.py
NoahStapp: It doesn't remove asyncio imports. We still have some files that have `_IS_SYNC` branches that make use of `asyncio`.

ShaneHarvey (approved):

sleepyStick (approved):

PR #2161, opened 2025-02-27
PYTHON-5126 Resync bson vector spec tests following additions

caseyclements:
Qingyang added more tests of invalidity. In particular, malformed BSON that won't decode.

https://github.com/mongodb/specifications/pull/1750

PR #2162, opened 2025-02-27
Adds __repr__ to BinaryVector dataclass

caseyclements:
No ticket. I just noticed this while working on #2161

### DIFF
--- a/bson/binary.py
+++ b/bson/binary.py
@@ -247,6 +247,9 @@ class BinaryVector:
         self.dtype = dtype
         self.padding = padding
 
+    def __repr__(self):
+        return f"BinaryVector - {self.dtype=}, {self.dtype},{self.padding=}, {self.data=}"
+
 
 class Binary(bytes):
     """Representation of BSON binary data.


caseyclements:
@NoahStapp  One question in my mind. Is using an f-string an issue? If the vector is really large, will `__repr__` create the string even when it's not called?

blink1073:
Can you add the original PYTHON ticket for BSON Binary Vector if we're not adding a new ticket?

bson/binary.py
```diff
@@ -247,6 +247,9 @@ def __init__(self, data: Sequence[float | int], dtype: BinaryVectorDtype, paddin
         self.dtype = dtype
         self.padding = padding
 
+    def __repr__(self) -> str:
+        return f"BinaryVector - {self.dtype=}, {self.dtype},{self.padding=}, {self.data=}"
```
ShaneHarvey, line 251: This needs to follow Python's convention for repr: https://docs.python.org/3/library/functions.html#repr

And we should add tests for this like we do for our other custom type reprs.

ShaneHarvey (commented):
Let's add a new python ticket for this because it's a new feature.

NoahStapp:
> @NoahStapp One question in my mind. Is using an f-string an issue? If the vector is really large, will `__repr__` create the string even when it's not called?

f-strings are always evaluated, even if never used. In this case, I don't know if the performance benefit of f-strings is offset by a very large vector being parsed into a string at runtime. It's probably worth running some quick benchmarks to see if there is a performance difference here.

NoahStapp:
> > > > @NoahStapp One question in my mind. Is using an f-string an issue? If the vector is really large, will `__repr__` create the string even when it's not called?
> > > 
> > > 
> > > f-strings are always evaluated, even if never used. In this case, I don't know if the performance benefit of f-strings is offset by a very large vector being parsed into a string at runtime. It's probably worth running some quick benchmarks to see if there is a performance difference here.
> > 
> > 
> > Yeah. I'm going to switch to format. These vectors could get massive. I don't want users to think mongodb is slow when they've simply got something in a logger.warning statement..
> 
> Our pre-commit automatically converts str.format into f-string! What's the workaround?

You can ignore specific ruff rules with ` # noqa: <rule>` on the offending line.

caseyclements:
> > > > > @NoahStapp One question in my mind. Is using an f-string an issue? If the vector is really large, will `__repr__` create the string even when it's not called?
> > > > 
> > > > 
> > > > f-strings are always evaluated, even if never used. In this case, I don't know if the performance benefit of f-strings is offset by a very large vector being parsed into a string at runtime. It's probably worth running some quick benchmarks to see if there is a performance difference here.
> > > 
> > > 
> > > Yeah. I'm going to switch to format. These vectors could get massive. I don't want users to think mongodb is slow when they've simply got something in a logger.warning statement..
> > 
> > 
> > Our pre-commit automatically converts str.format into f-string! What's the workaround?
> 
> You can ignore specific ruff rules with ` # noqa: <rule>` on the offending line.

Should we consider adding the following to pyproject.toml?
```
[tool.ruff]
ignore = ["RUF100"]
```

NoahStapp:
> > > > > > @NoahStapp One question in my mind. Is using an f-string an issue? If the vector is really large, will `__repr__` create the string even when it's not called?
> > > > > 
> > > > > 
> > > > > f-strings are always evaluated, even if never used. In this case, I don't know if the performance benefit of f-strings is offset by a very large vector being parsed into a string at runtime. It's probably worth running some quick benchmarks to see if there is a performance difference here.
> > > > 
> > > > 
> > > > Yeah. I'm going to switch to format. These vectors could get massive. I don't want users to think mongodb is slow when they've simply got something in a logger.warning statement..
> > > 
> > > 
> > > Our pre-commit automatically converts str.format into f-string! What's the workaround?
> > 
> > 
> > You can ignore specific ruff rules with ` # noqa: <rule>` on the offending line.
> 
> Should we consider adding the following to pyproject.toml?
> 
> ```
> [tool.ruff]
> ignore = ["RUF100"]
> ```

Avoiding cluttering the codebase with unused directives is a pattern we follow consistently, are you seeing linting failures without this change?

caseyclements:
> > > > > > > @NoahStapp One question in my mind. Is using an f-string an issue? If the vector is really large, will `__repr__` create the string even when it's not called?
> > > > > > 
> > > > > > 
> > > > > > f-strings are always evaluated, even if never used. In this case, I don't know if the performance benefit of f-strings is offset by a very large vector being parsed into a string at runtime. It's probably worth running some quick benchmarks to see if there is a performance difference here.
> > > > > 
> > > > > 
> > > > > Yeah. I'm going to switch to format. These vectors could get massive. I don't want users to think mongodb is slow when they've simply got something in a logger.warning statement..
> > > > 
> > > > 
> > > > Our pre-commit automatically converts str.format into f-string! What's the workaround?
> > > 
> > > 
> > > You can ignore specific ruff rules with ` # noqa: <rule>` on the offending line.
> > 
> > 
> > Should we consider adding the following to pyproject.toml?
> > ```
> > [tool.ruff]
> > ignore = ["RUF100"]
> > ```
> 
> Avoiding cluttering the codebase with unused directives is a pattern we follow consistently, are you seeing linting failures without this change?

Wrong rule. That was from ChatGPT. `UP032` is the f-string one. I don't know. It feels like when someone is using `format` they probably have chosen to do so for a reason.

bson/binary.py
```diff
@@ -247,6 +247,11 @@ def __init__(self, data: Sequence[float | int], dtype: BinaryVectorDtype, paddin
         self.dtype = dtype
         self.padding = padding
 
+    def __repr__(self) -> str:
+        return "BinaryVector(dtype={}, padding={}, data={})".format(  # noqa: UP032
```
ShaneHarvey, line 251: > will __repr__ create the string even when it's not called?

f-string vs format makes no difference. If anything f-string will be faster. `__repr__` only runs when repr() is called.

ShaneHarvey (changes requested):

bson/binary.py
caseyclements: > `__` only runs when repr() is called.

So if one has logger.debug(vector), is it not true that str.format won't be called, but the f-string will be created?

ShaneHarvey: Logging behavior is unrelated to this PR since we're just adding a new method.

caseyclements:
Test failures are async ones, unrelated to these.

test/test_bson.py
```diff
@@ -809,6 +809,30 @@ def test_vector(self):
...
+            repr(two), f"BinaryVector(dtype=BinaryVectorDtype.FLOAT32, padding=0, data={data})"
+        )
+        three = BinaryVector(data, BinaryVectorDtype.FLOAT32, padding=0)
+        self.assertEqual(
+            repr(three), f"BinaryVector(dtype=BinaryVectorDtype.FLOAT32, padding=0, data={data})"
+        )
+        four = BinaryVector(data, BinaryVectorDtype.PACKED_BIT, padding=3)
+        self.assertEqual(
+            repr(four), f"BinaryVector(dtype=BinaryVectorDtype.PACKED_BIT, padding=3, data={data})"
+        )
```
ShaneHarvey, line 834: Could you add asserts using assertRepr like we do in test_connection_monitoring.py?:
```python
    def assertRepr(self, obj):
        new_obj = eval(repr(obj))
        self.assertEqual(type(new_obj), type(obj))
        self.assertEqual(repr(new_obj), repr(obj))
```

caseyclements: Done. That's very cool, that one can create an object using eval(repr(obj))

bson/binary.py
```diff
@@ -247,6 +247,9 @@ def __init__(self, data: Sequence[float | int], dtype: BinaryVectorDtype, paddin
         self.dtype = dtype
         self.padding = padding
 
+    def __repr__(self) -> str:
+        return f"BinaryVector(dtype={self.dtype}, padding={self.padding}, data={self.data})"
+
```
ShaneHarvey, line 252: Hold up, the real bug here is that we're using `@dataclass` but also defining `__init__`. Why is that? The whole point of dataclass is that it provides `__init__` and `__repr__` automatically.

ShaneHarvey (changes requested):

bson/binary.py
caseyclements: I believe that we had issues with the combo of {dataclass, slots,  defaults}. So I've just removed the dataclass decorator.

ShaneHarvey: Are there any other features that will break by removing @dataclass? We may be stuck with it until 5.0 if it's a breaking change.

caseyclements: I don't think so. As mentioned in the closed PR, we've done __init__, and __repr__, and we don't want to rely on equality except via the binary representation.

caseyclements: The code seems happy to be @dataclass or not. I have no strong preference.

caseyclements: @ShaneHarvey - Steve's in agreement. Dump the decorator.

ShaneHarvey: So what happens with `==` and `>` before and after this change?

caseyclements: https://jira.mongodb.org/browse/DRIVERS-3123

caseyclements: Sure. It's easy to try. Put a breakpoint on line 83 of test_bson_binary. 

```
# With the decorator:
vector_obs == vector_exp.as_vector()
True
vector_obs > vector_exp.as_vector()
# TypeError: '>' not supported between instances of 'BinaryVector' and 'BinaryVector'

# Without the decorator:
vector_obs == vector_exp.as_vector()
False
vector_obs > vector_exp.as_vector()
# TypeError: '>' not supported between instances of 'BinaryVector' and 'BinaryVector'

```

ShaneHarvey: `==` is already broken with our current `@dataclass` approach:
```python
>>> from bson.binary import *
>>> BinaryVector([1], BinaryVectorDtype.INT8) == BinaryVector([2], BinaryVectorDtype.INT8)
True
```

So we need to remove dataclass and implement `==` as well as `__repr__`.

ShaneHarvey (changes requested):
Looking good! Could you add a note for this in the changelog?

caseyclements:
> Looking good! Could you add a note for this in the changelog?

Sure!  Is this targeting 4.11 or 4.12? If the former, should I bump the micro version and add one line for description and another for jira? If the latter, should I additionallyx start up a whole new section?

ShaneHarvey (approved):
On second thought, let's add this to changelog later since I'm adding the new 4.12 section in another PR right now and we may end up being backporting this as a bug fix. Could you update the PR title and Jira ticket to say this is a bug fix for `repr` and `==`?

ShaneHarvey (changes requested):

caseyclements:
> On second thought, let's add this to changelog later since I'm adding the new 4.12 section in another PR right now and we may end up being backporting this as a bug fix. Could you update the PR title and Jira ticket to say this is a bug fix for `repr` and `==`?

Updated names. In the JIRA, I marked it as bug instead of improvement. Please adjust if that's not what you meant.

ShaneHarvey (approved):

PR #2164, opened 2025-02-27
PYTHON-5166 Allow Database.command to run bulkWrite commands

ShaneHarvey:
PYTHON-5166 Fix db.command bulkWrite bug

### DIFF
--- a/pymongo/message.py
+++ b/pymongo/message.py
@@ -105,7 +105,7 @@ _FIELD_MAP = {
     "insert": "documents",
     "update": "updates",
     "delete": "deletes",
-    "bulkWrite": "bulkWrite",
+    "bulkWrite": "ops",
 }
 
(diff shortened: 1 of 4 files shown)


pymongo/message.py
```diff
@@ -105,7 +105,7 @@
     "insert": "documents",
     "update": "updates",
     "delete": "deletes",
-    "bulkWrite": "bulkWrite",
+    "bulkWrite": "ops",
```
ShaneHarvey, line 108: There's a larger question here on if we should even be doing this behavior anymore at all but I'll open a new ticket to investigate that.

blink1073 (approved):
LGTM!

ShaneHarvey:
drivers-pr-bot please backport to v4.11

PR #2167, opened 2025-03-03
PYTHON-5136 Add check-json to pre-commit checks

blink1073:


sleepyStick (approved):
you were up *early* today!

PR #2170, opened 2025-03-03
PYTHON-5177 - Add instructions for enabling debug logs to CONTRIBUTING.md

NoahStapp:


blink1073 (commented):
Is there an environment variable we can set in bashrc?

NoahStapp:
> Is there an environment variable we can set in bashrc?

Pytest only provides the cli or `pyproject.toml` config options to configure logging, unfortunately.

PR #2191, opened 2025-03-11
PYTHON-3636 MongoClient should perform SRV resolution lazily

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-3636


pymongo/asynchronous/mongo_client.py
```diff
@@ -1117,14 +1216,20 @@ def options(self) -> ClientOptions:
 
     def __eq__(self, other: Any) -> bool:
         if isinstance(other, self.__class__):
-            return self._topology == other._topology
+            if hasattr(self, "_topology"):
+                return self._topology == other._topology
+            else:
+                raise InvalidOperation("Cannot perform operation until client is connected")
```
NoahStapp, line 1222:
```suggestion
                raise InvalidOperation("Cannot compare client equality until both clients are connected")
```

pymongo/asynchronous/mongo_client.py
```diff
@@ -1117,14 +1216,20 @@ def options(self) -> ClientOptions:
 
     def __eq__(self, other: Any) -> bool:
         if isinstance(other, self.__class__):
-            return self._topology == other._topology
+            if hasattr(self, "_topology"):
+                return self._topology == other._topology
```
NoahStapp, line 1220: We also need to check if the other client has a topology here.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1117,14 +1216,20 @@ def options(self) -> ClientOptions:
...
 
     def __ne__(self, other: Any) -> bool:
         return not self == other
 
     def __hash__(self) -> int:
-        return hash(self._topology)
+        if hasattr(self, "_topology"):
+            return hash(self._topology)
+        else:
+            raise InvalidOperation("Cannot perform operation until client is connected")
```
NoahStapp, line 1232:
```suggestion
            raise InvalidOperation("Cannot hash client until it is connected")
```

pymongo/asynchronous/mongo_client.py
```diff
@@ -1160,7 +1265,9 @@ def option_repr(option: str, value: Any) -> str:
         return ", ".join(options)
 
     def __repr__(self) -> str:
-        return f"{type(self).__name__}({self._repr_helper()})"
+        if hasattr(self, "_topology"):
```
NoahStapp, line 1268: A client with no topology should still have a valid `__repr__` that can be passed to `eval` to re-create it.

pymongo/uri_parser.py
```diff
@@ -550,6 +572,69 @@ def parse_uri(
...
+        "username": user,
+        "password": passwd,
+        "database": dbase,
+        "collection": collection,
+        "options": options,
+        "fqdn": fqdn,
+    }
+
+
+def _parse_srv(
```
NoahStapp, line 599: The `dns_resolver.get_hosts()` call within this function needs to be asynchronous to avoid blocking the loop. https://dnspython.readthedocs.io/en/latest/async-resolver-class.html is the asynchronous version of the `dns.resolver` class we currently use.

NoahStapp (changes requested):
Some small clarity tweaks and async-specific changes, but great start!

pymongo/asynchronous/mongo_client.py
sleepyStick: not sure what you mean by re-create _it_? 
But I changed it to always give a repr now, lmk what you think

NoahStapp: Our `__repr__` policy is that `my_object == eval(my_object.__repr__())`.

pymongo/asynchronous/uri_parser.py
```diff
@@ -0,0 +1,286 @@
...
+            normalize,
+            connect_timeout,
+            srv_service_name,
+            srv_max_hosts,
+        )
+    )
+    return result
+
+
+def _validate_uri(
```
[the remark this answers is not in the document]
sleepyStick: wouldn't that introduce a circular import? Because this file imports from `uri_parser_shared.py` ?

test/asynchronous/helpers.py
```diff
@@ -47,7 +47,7 @@
 from pymongo import common, message
 from pymongo.read_preferences import ReadPreference
 from pymongo.ssl_support import HAVE_SSL, _ssl  # type:ignore[attr-defined]
-from pymongo.uri_parser import parse_uri
+from pymongo.synchronous.uri_parser import parse_uri
```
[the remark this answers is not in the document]
sleepyStick: Wait this one was intentional,, because parse_uri is called for some test setup

pymongo/asynchronous/mongo_client.py
```diff
@@ -1117,14 +1218,22 @@ def options(self) -> ClientOptions:
...
 
     def __ne__(self, other: Any) -> bool:
         return not self == other
 
     def __hash__(self) -> int:
-        return hash(self._topology)
+        if hasattr(self, "_topology"):
+            return hash(self._topology)
+        else:
+            raise InvalidOperation("Cannot hash client until it is connected")
```
ShaneHarvey, line 1236: It seems odd to disallow comparison and hashing without "connecting" first. I wonder if there's a better way to define these without relying on SRV or TXT resolution. Like could be just compare some subset of MongoClient arguments?

pymongo/asynchronous/mongo_client.py
```diff
@@ -820,87 +822,186 @@ def __init__(
...
+        return opts
 
     async def aconnect(self) -> None:
         """Explicitly connect to MongoDB asynchronously instead of on the first operation."""
         await self._get_topology()
 
-    def _init_background(self, old_pid: Optional[int] = None) -> None:
+    def _init_background(self, old_pid: Optional[int] = None, first: bool = False) -> None:
         self._topology = Topology(self._topology_settings)
+        if first and _HAS_REGISTER_AT_FORK:
```
ShaneHarvey, line 1001: What does "first" mean here?

pymongo/asynchronous/mongo_client.py
```diff
@@ -820,87 +822,186 @@ def __init__(
...
+                username, password, self._resolve_srv_info["dbase"], opts, _IS_SYNC
+            )
+
+            self._init_based_on_options(seeds, srv_max_hosts, srv_service_name)
+
+    def _init_based_on_options(
+        self, seeds: Collection[tuple[str, int]], srv_max_hosts: Any, srv_service_name: Any
+    ) -> None:
+        self._event_listeners = self._options.pool_options._event_listeners
+        super().__init__(
```
ShaneHarvey, line 940: I don't think we can call `super().__init__` outside of `__init__`. That would mean we have undefined behavior in between. Like attempting to access `client.codec_options` will fail until the client becomes "connected".

pymongo/asynchronous/mongo_client.py
```diff
@@ -1140,13 +1249,16 @@ def option_repr(option: str, value: Any) -> str:
...
+            options = [
+                "host=%r"
+                % [
+                    "%s:%d" % (host, port) if port is not None else host
+                    for host, port in self._topology_settings.seeds
+                ]
             ]
-        ]
+        else:
+            options = ["host={self._host}", "port={self._port}"]
```
ShaneHarvey, line 1261: This is not safe. `self._host` is the entire URI so doing this would leak the user credentials.

ShaneHarvey (changes requested):

pymongo/srv_resolver.py
```diff
@@ -1,148 +1,22 @@
-# Copyright 2019-present MongoDB, Inc.
+# Copyright 2025-present MongoDB, Inc.
```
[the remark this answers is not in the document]
sleepyStick: //is this commenting on the file / line as above?

pymongo/asynchronous/mongo_client.py
sleepyStick: oh! good call out, i'm not 100% sure we need it anymore, but i added it as a "oh this is the first time we're calling init_background" because when i initially was doing this, i think i called it twice if we did SRV resolution...I'll try getting rid of it!

pymongo/asynchronous/mongo_client.py
sleepyStick: oh, i see, what should I use in the repr here instead then?

pymongo/uri_parser.py
```diff
@@ -1,639 +1,31 @@
...
-if __name__ == "__main__":
-    import pprint
-
-    try:
-        pprint.pprint(parse_uri(sys.argv[1]))  # noqa: T203
-    except InvalidURI as exc:
-        print(exc)  # noqa: T201
-    sys.exit(0)
+from pymongo.synchronous.uri_parser import *  # noqa: F403
+from pymongo.synchronous.uri_parser import __doc__ as original_doc
```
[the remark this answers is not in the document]
sleepyStick: Err, I'm down to add one (maybe in a diff PR tho since this PR is already fairly big?) if we think it's needed?

pymongo/srv_resolver.py
```diff
@@ -1,148 +1,22 @@
...
-        _, nodes = self._get_srv_response_and_hosts(True)
-        return nodes
-
-    def get_hosts_and_min_ttl(self) -> tuple[list[tuple[str, Any]], int]:
-        results, nodes = self._get_srv_response_and_hosts(False)
-        rrset = results.rrset
-        ttl = rrset.ttl if rrset else 0
-        return nodes, ttl
+__doc__ = original_doc
+__all__ = ["maybe_decode", "_SrvResolver"]
```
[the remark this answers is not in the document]
sleepyStick: Oh, i wasn't aware that it was entirely a private api, which is why i did this, but looking at motor, it seems like motor imports `_resolve` and` _SrvResolver` in synchro? 
https://github.com/mongodb/motor/blob/a879ce528ad60601853758e9d710a79f67277025/synchro/__init__.py#L107

I'm not too familar with how motor works, but based on that i think i should modify the above export to be `_resolve` and` _SrvResolver`, right?

pymongo/asynchronous/mongo_client.py
sleepyStick: update: not needed, and now removed :)

pymongo/asynchronous/mongo_client.py
ShaneHarvey: What about this?:
```python
AsyncMongoClient(host='mongodb+srv://fqdn', ...
```

test/asynchronous/test_client.py
```diff
@@ -1911,28 +1911,37 @@ async def test_service_name_from_kwargs(self):
             srvServiceName="customname",
             connect=False,
         )
+        await client.aconnect()
```
ShaneHarvey, line 1914: Are these open/close calls still needed?

pymongo/srv_resolver.py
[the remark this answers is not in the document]
ShaneHarvey: Thanks for checking, let's remove it.

pymongo/uri_parser.py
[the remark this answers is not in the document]
ShaneHarvey: By `__doc__` I mean a module level docstring like:
```python
"""Tools to parse and validate a MongoDB URI."""
```

pymongo/asynchronous/mongo_client.py
```diff
@@ -1115,16 +1224,30 @@ def options(self) -> ClientOptions:
...
+            else:
+                return self._topology == other._topology
         return NotImplemented
 
     def __ne__(self, other: Any) -> bool:
         return not self == other
 
     def __hash__(self) -> int:
-        return hash(self._topology)
+        if self._topology is None:
```
ShaneHarvey, line 1247: We still need to address hashing and equality. It's broken to change the hash of an object since it means that a single client could be added twice to a dictionary or set. The hash (and eq) methods need to use a single deterministic approach:
```python
set = {}
c = AsyncMongoClient('mongodb+srv://...')
set.add(c)
await c.aconnect()
set.add(c)
assert len(set) == 1
```

This means we need to use the eq_props() approach in all cases.

ShaneHarvey (changes requested):

test/asynchronous/test_client.py
sleepyStick: yes because these are srv uris

pymongo/asynchronous/mongo_client.py
```diff
@@ -750,6 +755,9 @@ def __init__(
             port = self.PORT
         if not isinstance(port, int):
             raise TypeError(f"port must be an instance of int, not {type(port)}")
+        self._host = host
+        self._port = port
+        self._topology: Optional[Topology] = None
```
ShaneHarvey, line 760: Adding `assert self._topology is not None` so many times feels a bit off since it adds runtime overhead for a static typing issue. What if we do this instead?:
```python
        self._topology: Topology = None  # type: ignore[assignment]
```

pymongo/asynchronous/mongo_client.py
```diff
@@ -1552,24 +1677,25 @@ async def close(self) -> None:
...
-            # TODO: PYTHON-1921 Encrypted MongoClients cannot be re-opened.
-            await self._encrypter.close()
-        self._closed = True
-        if not _IS_SYNC:
-            await asyncio.gather(
-                self._topology.cleanup_monitors(),  # type: ignore[func-returns-value]
-                self._kill_cursors_executor.join(),  # type: ignore[func-returns-value]
-                return_exceptions=True,
-            )
+        if self._topology is not None:
```
ShaneHarvey, line 1680: Rather than indenting this code (which causes code churn) how about we do this?:
```python
if self._topology is not None:
    return
session_ids = self._topology.pop_all_sessions()
...
```

ShaneHarvey (changes requested):

test/asynchronous/test_dns.py
```diff
@@ -185,35 +186,24 @@ def create_tests(cls):
...
-        )
-        self.assertRaisesRegex(
-            ConfigurationError,
-            "Invalid URI host: an IP address is not",
-            self.simple_client,
-            "mongodb+srv://[::1]",
-        )
+        with self.assertRaisesRegex(ConfigurationError, "Invalid URI host: mongodb is not"):
+            client = self.simple_client("mongodb+srv://mongodb")
+            await client.aconnect()
```
ShaneHarvey, line 191: This behavior change is important to call out in the changelog and jira ticket.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1140,13 +1257,16 @@ def option_repr(option: str, value: Any) -> str:
             return f"{option}={value!r}"
 
         # Host first...
-        options = [
-            "host=%r"
-            % [
-                "%s:%d" % (host, port) if port is not None else host
-                for host, port in self._topology_settings.seeds
+        if self._topology is None:
+            options = self._resolve_srv_info["seeds"]
```
ShaneHarvey, line 1261: Do we have tests for this? If not could we add some. I'd like to see the behavior difference in repr().

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM! Nice work!

pymongo/asynchronous/srv_resolver.py
```diff
@@ -0,0 +1,158 @@
...
+            return resolver.resolve(*args, **kwargs)
+        # dnspython 1.X
+        return resolver.query(*args, **kwargs)
+    else:
+        from dns import asyncresolver
+
+        if hasattr(asyncresolver, "resolve"):
+            # dnspython >= 2
+            return await asyncresolver.resolve(*args, **kwargs)  # type:ignore[return-value]
+        raise ConfigurationError("Upgrade to dnspython version >= 2.0")
```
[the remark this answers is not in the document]
ShaneHarvey: Could you add a suggestion here?

test/test_default_exports.py
```diff
@@ -69,6 +69,7 @@ def test_bson(self):
 
     def test_pymongo_imports(self):
         import pymongo
+        from pymongo.asynchronous.uri_parser import parse_uri
```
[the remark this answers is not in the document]
ShaneHarvey: Good catch. We need to revert all the changes to this test.

pymongo/asynchronous/srv_resolver.py
NoahStapp, line 64:
```suggestion
        raise ConfigurationError("Upgrade to dnspython version >= 2.0 to use AsyncMongoClient with mongodb+srv:// connections.")
```

NoahStapp (changes requested):

ShaneHarvey (approved):
LGTM once the tests pass!

ShaneHarvey (approved):

NoahStapp (approved):

PR #2192, opened 2025-03-11
PYTHON-5202 WaitQueueTimeoutError should not clear the pool

ShaneHarvey:
PYTHON-5202 WaitQueueTimeoutError should not clear the pool

### DIFF
--- a/pymongo/asynchronous/topology.py
+++ b/pymongo/asynchronous/topology.py
@@ -41,6 +41,7 @@ from pymongo.errors import (
     OperationFailure,
     PyMongoError,
     ServerSelectionTimeoutError,
+    WaitQueueTimeoutError,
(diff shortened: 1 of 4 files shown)


NoahStapp (approved):
How did you come across this bug? Great find!

blink1073:
drivers-pr-bot please backport to v4.11

ShaneHarvey:
I discovered this while working on a customer repro involving CSOT. Using `pymongo.timeout()` will cause WaitQueueTimeoutError when the timeout expires while waiting to checkout a connection from the pool. After this happens, subsequent operations failed with `AutoReconnect("connection pool paused")` errors which was unexpected. It would have been easier to diagnose if we had DRIVERS-1998.

PR #2194, opened 2025-03-12
PYTHON-5196 Convert OIDC tests to use new test scripts

blink1073:
Passing build: https://spruce.mongodb.com/version/67d1a6e179aa780007b9fbe6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


.evergreen/scripts/oidc_tester.py
```diff
@@ -0,0 +1,101 @@
...
+        run_command(f"bash {target_dir}/run-self-test.sh")
+        return None
+
+    source_file = None
+    if sub_test_name == "test":
+        source_file = f"{target_dir}/secrets-export.sh"
+    elif sub_test_name == "azure-remote":
+        source_file = "./env.sh"
+    elif sub_test_name == "gcp-remote":
+        source_file = "./secrets-export.sh"
```
NoahStapp, line 50: Why do the `azure-remote` and `gcp-remote` envs live in different files?

.evergreen/generated_configs/tasks.yml
```diff
@@ -1042,6 +1042,50 @@ tasks:
           TEST_NAME: ocsp
     tags: [ocsp, ocsp-rsa]
 
+  # Oidc tests
+  - name: test-auth-oidc-test
+    commands:
+      - func: run tests
+        vars:
+          TEST_NAME: auth_oidc
+          SUB_TEST_NAME: test
```
NoahStapp, line 1051: Can we change the `SUB_TEST_NAME` from `test` to something like `default` for clarity?

NoahStapp (changes requested):

.evergreen/generated_configs/tasks.yml
blink1073: That name corresponds to the OIDC_ENV variable, which is already established.

.evergreen/scripts/oidc_tester.py
blink1073: They're put there by the scripts in drivers-tools.

NoahStapp: I meant more why do they end up in two different places? I would expect them to be put into the same file by the scripts.

blink1073: I blame the author (past me).

.evergreen/generated_configs/tasks.yml
blink1073: Using `default` works: https://spruce.mongodb.com/version/67d1d5d9b4e9050007f28472/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

blink1073: I'll wait to push until we merge the drivers-tools PR and I remove the checkout of my fork.

blink1073:
Ready for another look

NoahStapp (approved):

PR #2199, opened 2025-03-13
PYTHON-5208 Add spec test for wait queue timeout errors do not clear the pool

ShaneHarvey:
PYTHON-5208 Add spec test for wait queue timeout errors do not clear the pool

Implementation of https://github.com/mongodb/specifications/pull/1768


test/asynchronous/unified_format.py
```diff
@@ -1443,11 +1440,12 @@ async def _run_scenario(self, spec, uri=None):
...
             self.TEST_SPEC.get("createEntities", []), uri=uri
         )
+        self._cluster_time = None
         # process initialData
         if "initialData" in self.TEST_SPEC:
             await self.insert_initial_data(self.TEST_SPEC["initialData"])
-            self._cluster_time = (await self.client.admin.command("ping")).get("$clusterTime")
-            await self.entity_map.advance_cluster_times()
+            self._cluster_time = self.client._topology.max_cluster_time()
+            await self.entity_map.advance_cluster_times(self._cluster_time)
```
ShaneHarvey, line 1448: I refactored this to remove the "ping" command because it was causing problems with the failpoint for this new test. We can get the clusterTime via `self.client._topology.max_cluster_time()` instead.

test/asynchronous/unified_format.py
```diff
@@ -1443,11 +1440,12 @@ async def _run_scenario(self, spec, uri=None):
         await self.entity_map.create_entities_from_spec(
             self.TEST_SPEC.get("createEntities", []), uri=uri
         )
+        self._cluster_time = None
```
NoahStapp, line 1443: Do we need to store this at the class level at all, or can this be a fully local variable?

NoahStapp (changes requested):

test/asynchronous/unified_format.py
ShaneHarvey: Yes we need it because the spec says it's the clusterTime from directly after initialData is inserted.

NoahStapp (approved):

test/asynchronous/unified_format.py
ShaneHarvey: > the test runner MUST advance the cluster time of any session entities created during the test using the cluster time collected from processing [initialData](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#initialData). See [MigrationConflict Errors on Sharded Clusters](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#migrationconflict-errors-on-sharded-clusters) for more information.

PR #2206, opened 2025-03-15
PYTHON-5216 & PYTHON-5185 Convert perf tests to use new scripts and use specific python version

blink1073:
Patch build: https://spruce.mongodb.com/version/67d4e62414a2060007c2bc62/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


.evergreen/scripts/generate_config.py
```diff
@@ -942,6 +950,26 @@ def create_enterprise_auth_tasks():
     return [EvgTask(name=task_name, tags=tags, commands=[server_func, assume_func, test_func])]
 
 
+def create_perf_tasks():
+    tasks = []
+    for version, ssl, sync in product(["6.0", "8.0"], ["ssl", "nossl"], ["sync", "async"]):
```
ShaneHarvey, line 955: We can remove the 6.0 tests now that we are pinning to v8.0-perf.

ShaneHarvey (changes requested):

.evergreen/scripts/generate_config.py
blink1073: Done

PR #2210, opened 2025-03-18
PYTHON-5046 Support $lookup in CSFLE and QE

sleepyStick:


sleepyStick:
pyproject.toml and uv.lock will need to change once https://github.com/mongodb/libmongocrypt/pull/989 is merged :)

test/asynchronous/test_encryption.py
```diff
@@ -2419,6 +2420,318 @@ async def test_05_roundtrip_encrypted_unindexed(self):
...
+        self.assertTrue(isinstance(doc["qe"], Binary))
+        await self.encrypted_client.db.qe2.insert_one({"qe2": "qe2"})
+        doc = await self.unencrypted_client.db.qe2.find_one()
+        self.assertTrue(isinstance(doc["qe2"], Binary))
+        await self.encrypted_client.db.no_schema.insert_one({"no_schema": "no_schema"})
+        await self.encrypted_client.db.no_schema2.insert_one({"no_schema2": "no_schema2"})
+
+    @async_client_context.require_version_min(8, 1, -1)
+    async def test_1(self):
+        self.encrypted_client = await self.async_rs_or_single_client(
```
ShaneHarvey, line 2488: Why do we recreate the client in each test? Can't we reuse the existing `self.encrypted_client`?

test/asynchronous/test_encryption.py
```diff
@@ -2419,6 +2420,318 @@ async def test_05_roundtrip_encrypted_unindexed(self):
...
+        doc = await self.unencrypted_client.db.qe.find_one()
+        self.assertTrue(isinstance(doc["qe"], Binary))
+        await self.encrypted_client.db.qe2.insert_one({"qe2": "qe2"})
+        doc = await self.unencrypted_client.db.qe2.find_one()
+        self.assertTrue(isinstance(doc["qe2"], Binary))
+        await self.encrypted_client.db.no_schema.insert_one({"no_schema": "no_schema"})
+        await self.encrypted_client.db.no_schema2.insert_one({"no_schema2": "no_schema2"})
+
+    @async_client_context.require_version_min(8, 1, -1)
+    async def test_1(self):
```
ShaneHarvey, line 2487: For each test could you add a descriptive name, like `test_1_csfle_joins_no_schema`

ShaneHarvey (changes requested):
Nice work! Could you add a note to the changelog since this is a new feature?

test/asynchronous/test_encryption.py
ShaneHarvey: Oh I see the spec says:
> Recreate encryptedClient with the same AutoEncryptionOpts as the setup. (Recreating prevents schema caching from impacting the test).

Can we close the asyncSetup client (self.encrypted_client)? Could we even make it a local var to avoid confusion?

sleepyStick: I made them all local vars (even the one in setup since its never reused outside of the setup) do you still want me to explicitly close them?

uv.lock
```diff
@@ -997,7 +997,6 @@ sdist = { url = "https://files.pythonhosted.org/packages/07/e9/ae44ea7d7605df9e5
 
 [[package]]
 name = "pymongo"
-version = "4.12.0.dev0"
```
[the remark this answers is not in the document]
sleepyStick: oops nope! should be fixed now though!

test/asynchronous/test_encryption.py
```diff
@@ -2419,6 +2420,307 @@ async def test_05_roundtrip_encrypted_unindexed(self):
...
+        doc = await unencrypted_client.db.csfle2.find_one()
+        self.assertTrue(isinstance(doc["csfle2"], Binary))
+        await encrypted_client.db.qe.insert_one({"qe": "qe"})
+        doc = await unencrypted_client.db.qe.find_one()
+        self.assertTrue(isinstance(doc["qe"], Binary))
+        await encrypted_client.db.qe2.insert_one({"qe2": "qe2"})
+        doc = await unencrypted_client.db.qe2.find_one()
+        self.assertTrue(isinstance(doc["qe2"], Binary))
+        await encrypted_client.db.no_schema.insert_one({"no_schema": "no_schema"})
+        await encrypted_client.db.no_schema2.insert_one({"no_schema2": "no_schema2"})
```
ShaneHarvey, line 2473: Could you add:
```python
await encrypted_client.close()
await unencrypted_client.close()
```

doc/changelog.rst
```diff
@@ -8,6 +8,7 @@ PyMongo 4.12 brings a number of changes including:
 
 - Support for configuring DEK cache lifetime via the ``key_expiration_ms`` argument to
   :class:`~pymongo.encryption_options.AutoEncryptionOpts`.
+- Support for $lookup in CSFLE and QE.pr
```
ShaneHarvey, line 11: `QE.pr` -> `QE supported on MongoDB 8.1+.`

ShaneHarvey (changes requested):

test/asynchronous/test_encryption.py
sleepyStick: ah, fair enough, they're all explicitly closed now! 
I initially thought that since we use self.async_rs_or_single_client to create the client, we wouldn't need to explicitly call close.

test/asynchronous/test_encryption.py
```diff
@@ -2499,6 +2502,7 @@ async def test_1_csfle_joins_no_schema(self):
             )
         )
         self.assertEqual(doc, {"csfle": "csfle", "matched": [{"no_schema": "no_schema"}]})
+        await encrypted_client.close()
```
ShaneHarvey, line 2505: Closing the clients in the tests isn't needed since they're closed at tear down anyway. The reason I asked to close the clients in asyncSetUp is because otherwise they remain open for the entire test and compete with other tasks that need to be scheduled.

ShaneHarvey (changes requested):

test/asynchronous/test_encryption.py
sleepyStick: ahh okay, i see. fixed

test/asynchronous/test_encryption.py
```diff
@@ -2419,6 +2420,310 @@ async def test_05_roundtrip_encrypted_unindexed(self):
...
+            auto_encryption_opts=AutoEncryptionOpts(
+                key_vault_namespace="db.keyvault",
+                kms_providers={"local": {"key": LOCAL_MASTER_KEY}},
+            )
+        )
+        await encrypted_client.drop_database("db")
+
+        key_doc = json_data("etc", "data", "lookup", "key-doc.json")
+        key_vault = await create_key_vault(encrypted_client.db.keyvault, key_doc)
+        self.addCleanup(key_vault.drop)
```
ShaneHarvey, line 2439: addCleanup -> addAsyncCleanup

Also can we use the client_context client here to drop the entire database instead of just the keyvault? This will fix the `InvalidOperation: Cannot use AsyncMongoClient after close"` errors (which were introduced now that we close these clients).

ShaneHarvey (changes requested):

test/asynchronous/test_encryption.py
ShaneHarvey: Alternatively we can just leave the data there and skip the cleanup altogether.

sleepyStick: Ah good catch, and yes! done!

pymongo/asynchronous/encryption.py
```diff
@@ -251,14 +251,12 @@ async def collection_info(self, database: str, filter: bytes) -> Optional[bytes]
         :param database: The database on which to run listCollections.
         :param filter: The filter to pass to listCollections.
 
-        :return: The first document from the listCollections command response as BSON.
+        :return: The all documents from the listCollections command response as BSON.
```
ShaneHarvey, line 254: "The all document" -> "All documents"

ShaneHarvey (changes requested):
One last thing.

sleepyStick:
I opened https://jira.mongodb.org/browse/PYTHON-5225 because I don't think those test failures are related to the work I did? But if I'm wrong, lmk!

pymongo/asynchronous/encryption.py
sleepyStick: Oop good catch, fixed!

ShaneHarvey (approved):
LGTM!

PR #2213, opened 2025-03-24
PYTHON-5169 - Deprecate Hedged Reads option

NoahStapp:


pymongo/read_preferences.py
```diff
@@ -103,6 +104,11 @@ def _validate_hedge(hedge: Optional[_Hedge]) -> Optional[_Hedge]:
     if not isinstance(hedge, dict):
         raise TypeError(f"hedge must be a dictionary, not {hedge!r}")
 
+    warnings.warn(
+        "Hedged reads are deprecated starting in server version 8.0.",
```
ShaneHarvey, line 108: Mentioning the server side deprecation is good but we should first mention the driver side deprecation. Something like:
`the read preference "hedge" option is deprecated in PyMongo 4.12+ because hedged reads are deprecated in MongoDB version 8.0+. Support for "hedge" will be removed in PyMongo 5.0.`

pyproject.toml
```diff
@@ -119,6 +119,7 @@ filterwarnings = [
     "module:please use dns.resolver.Resolver.resolve:DeprecationWarning",
     # https://github.com/dateutil/dateutil/issues/1314
     "module:datetime.datetime.utc:DeprecationWarning",
+    "module:Hedged reads are deprecated:DeprecationWarning",
```
ShaneHarvey, line 122: Rather than using this approach we should manually catch/silence hedge errors only in tests that are intended to use it. We should also add a new test that we do raise the warning when hedge is used.

ShaneHarvey (changes requested):

ShaneHarvey (commented):
Could you update the changelog as well?

doc/changelog.rst
```diff
@@ -9,6 +9,12 @@ PyMongo 4.12 brings a number of changes including:
 - Support for configuring DEK cache lifetime via the ``key_expiration_ms`` argument to
   :class:`~pymongo.encryption_options.AutoEncryptionOpts`.
 - Support for $lookup in CSFLE and QE supported on MongoDB 8.1+.
+- Deprecated the ``hedge`` parameter for
+  :class:`~pymongo.read_preferences.PrimaryPreferred`,
+  :class:`~pymongo.read_preferences.Secondary`,
+  :class:`~pymongo.read_preferences.SecondaryPreferred`,
+  :class:`~pymongo.read_preferences.Nearest`. Support for ``hedge`` will be removed in PyMongo 5.0.
```
ShaneHarvey, line 16: Can hedge also appear in the URI/MongoClient kwarg?

NoahStapp: It's not listed in our docs page, so I don't believe so. You can create a read preference instance and pass that as a kwarg, but that will still pass through the read preference constructor.

ShaneHarvey: Thanks for checking

pymongo/read_preferences.py
```diff
@@ -143,6 +149,11 @@ def document(self) -> dict[str, Any]:
         if self.__max_staleness != -1:
             doc["maxStalenessSeconds"] = self.__max_staleness
         if self.__hedge not in (None, {}):
+            warnings.warn(
```
ShaneHarvey, line 152: The `document` helper should not raise a warning. We already raised a warning in the constructor. We do need to keep the warning in `hedge()` since we will remove that api in 5.0.

ShaneHarvey (changes requested):

pymongo/read_preferences.py
```diff
@@ -103,6 +104,11 @@ def _validate_hedge(hedge: Optional[_Hedge]) -> Optional[_Hedge]:
     if not isinstance(hedge, dict):
         raise TypeError(f"hedge must be a dictionary, not {hedge!r}")
 
+    warnings.warn(
+        "The read preference 'hedge' option is deprecated in PyMongo 4.12+ because hedged reads are deprecated in MongoDB version 8.0+. Support for 'hedge' will be removed in PyMongo 5.0.",
+        DeprecationWarning,
+        stacklevel=2,
```
ShaneHarvey, line 110: One last question. I'm wondering about the stacklevel here. Can you provide an example of the warning in the ReadPreference constructor?

ShaneHarvey (changes requested):

pymongo/read_preferences.py
NoahStapp: `stacklevel=2`:
```
  /Users/nstapp/Github/mongo-python-driver/pymongo/read_preferences.py:131: DeprecationWarning: The read preference 'hedge' option is deprecated in PyMongo 4.12+ because hedged reads are deprecated in MongoDB version 8.0+. Support for 'hedge' will be removed in PyMongo 5.0.
    self.__hedge = _validate_hedge(hedge)
```

`stacklevel=4`:
```
  /Users/nstapp/Github/mongo-python-driver/test/asynchronous/test_read_preferences.py:583: DeprecationWarning: The read preference 'hedge' option is deprecated in PyMongo 4.12+ because hedged reads are deprecated in MongoDB version 8.0+. Support for 'hedge' will be removed in PyMongo 5.0.
    cls(hedge={"enabled": True})
```

NoahStapp: I think we want `stacklevel=4` to provide a useful traceback.

ShaneHarvey: Let's make it so.

ShaneHarvey (approved):

Jibola (approved):

PR #2214, opened 2025-03-24
PYTHON-5135 - Rename WriteConcernFailed code name to WriteConcernTimeout

NoahStapp:
We already did a chunk of this work in February, but the spec tests hadn't been fully updated yet.

### DIFF
--- a/pymongo/asynchronous/client_session.py
+++ b/pymongo/asynchronous/client_session.py
@@ -458,10 +458,10 @@ def _max_time_expired_error(exc: PyMongoError) -> bool:
 
 
 # From the transactions spec, all the retryable writes errors plus
-# WriteConcernTimedOut.
+# WriteConcernTimeout.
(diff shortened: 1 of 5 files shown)


pymongo/asynchronous/client_session.py
```diff
@@ -458,10 +458,10 @@ def _max_time_expired_error(exc: PyMongoError) -> bool:
 
 
 # From the transactions spec, all the retryable writes errors plus
-# WriteConcernFailed.
+# WriteConcernTimedOut.
 _UNKNOWN_COMMIT_ERROR_CODES: frozenset = _RETRYABLE_ERROR_CODES | frozenset(
     [
-        64,  # WriteConcernFailed
+        64,  # WriteConcernTimedOut
```
ShaneHarvey, line 464: These should say WriteConcernTimeout.

ShaneHarvey (approved):

PR #2215, opened 2025-03-24
PYTHON-5121 - Use canonical Extended JSON for BSON binary vector spec…

NoahStapp:
… tests

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -507,29 +507,6 @@ def loads(s: Union[str, bytes, bytearray], *args: Any, **kwargs: Any) -> Any:
     return json.loads(s, *args, **kwargs)
 
 
-def load(fp: Any, *args: Any, **kwargs: Any) -> Any:
-    """Helper function that wraps :func:`json.load`.
-
-    Automatically passes the object_hook for BSON type conversion.
-
-    Raises ``TypeError``, ``ValueError``, ``KeyError``, or
-    :exc:`~bson.errors.InvalidId` on invalid MongoDB Extended JSON.
-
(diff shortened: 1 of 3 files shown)


test/test_bson_binary_vector.py
```diff
@@ -27,6 +27,15 @@
 _TEST_PATH = Path(__file__).parent / "bson_binary_vector"
 
 
+def convert_extended_json(vector) -> float:
```
ShaneHarvey, line 30: Can we load the entire JSON file using json_util.loads() instead?

NoahStapp: I added a new `json_util.load()` method for this exact purpose.

bson/json_util.py
```diff
@@ -507,6 +507,29 @@ def loads(s: Union[str, bytes, bytearray], *args: Any, **kwargs: Any) -> Any:
     return json.loads(s, *args, **kwargs)
 
 
+def load(fp: Any, *args: Any, **kwargs: Any) -> Any:
```
ShaneHarvey, line 510: Let's open a new ticket for this. Adding this increases the scope of this ticket too much and opens other questions like "do we also need json_util.dump?"

I also question if this is valuable enough to add since it's pretty simple to use json_util.loads:
```python
json_util.loads(fp.read())
```

ShaneHarvey (changes requested):

bson/json_util.py
NoahStapp: Good point--I don't think we need `json_util.load()` when we can use `json_util.loads(fp.read())`.

PR #2216, opened 2025-03-24
PYTHON-4990 - Remove deprecated field from GridFS unified tests

NoahStapp:


aclark4life (approved):
RIP `md5`, `contentType`, `aliases`, LGTM

PR #2219, opened 2025-03-24
PYTHON-4946 - Add GridFSBucket.rename_by_name

NoahStapp:
…lename' feature - rename_by_name


gridfs/asynchronous/grid_file.py
```diff
@@ -1021,6 +1021,36 @@ async def rename(
...
+
+        .. versionadded:: 4.12
+        """
+        _disallow_transactions(session)
+        result = await self._files.update_many(
+            {"filename": filename}, {"$set": {"filename": new_filename}}, session=session
+        )
+        if not result.matched_count:
+            raise NoFile(
+                f"no files could be renamed {new_filename} because none matched filename {filename}"
```
ShaneHarvey, line 1051: Suggest: `f"no files could be renamed {new_filename!r} because none matched filename {filename!r}"` to make the strings more obvious.

gridfs/asynchronous/grid_file.py
```diff
@@ -1021,6 +1021,36 @@ async def rename(
                 "matched file_id %i" % (new_filename, file_id)
             )
 
+    async def rename_by_name(
```
ShaneHarvey, line 1024: Instead of adding a new method for this, should we consider overloading the existing rename method? For example:
```python
>>> fs.rename(file_id, "new") # rename by id
>>> fs.rename(filename="old", new_filename="new") # rename by name
```

The pro is we have less apis to maintain. I guess a con is that the user could accidentally mix up the file_id and filename params.

gridfs/asynchronous/grid_file.py
```diff
@@ -1021,6 +1021,36 @@ async def rename(
...
+          fs = AsyncGridFSBucket(my_db)
+          await fs.upload_from_stream("test_file", "data I want to store!")
+          await fs.rename_by_name("test_file", "new_test_name")
+
+        Raises :exc:`~gridfs.errors.NoFile` if no file with file_id exists.
+
+        :param filename: The filename of the file to be renamed.
+        :param new_filename: The new name of the file.
+        :param session: a
+            :class:`~pymongo.client_session.AsyncClientSession`
```
ShaneHarvey, line 1041: our line length allows this to be one line now right?
```
        :param session: a :class:`~pymongo.client_session.AsyncClientSession`
```

gridfs/asynchronous/grid_file.py
```diff
@@ -1021,6 +1021,36 @@ async def rename(
...
+        """Renames the stored file with the specified filename.
+
+        For example::
+
+          my_db = AsyncMongoClient().test
+          fs = AsyncGridFSBucket(my_db)
+          await fs.upload_from_stream("test_file", "data I want to store!")
+          await fs.rename_by_name("test_file", "new_test_name")
+
+        Raises :exc:`~gridfs.errors.NoFile` if no file with file_id exists.
```
ShaneHarvey, line 1036: `with file_id` -> `with the given filename`

test/asynchronous/unified_format.py
```diff
@@ -630,6 +630,9 @@ def process_error(self, exception, spec):
                 self.assertNotIsInstance(error, NotPrimaryError)
             elif isinstance(error, (InvalidOperation, ConfigurationError, EncryptionError)):
                 pass
+            # gridfs NoFile errors are considered client errors.
+            elif isinstance(error, NoFile):
```
ShaneHarvey, line 634: Suggest adding this to the previous elif: `InvalidOperation, ConfigurationError, EncryptionError, NoFile)):`

gridfs/synchronous/grid_file.py
```diff
@@ -1015,6 +1015,36 @@ def rename(
                 "matched file_id %i" % (new_filename, file_id)
             )
 
+    def rename_by_name(
```
ShaneHarvey, line 1018: Let's add some hand written tests for this too like test_rename in test_gridfs_bucket. That way we get some signal on typing and the UX.

ShaneHarvey (changes requested):

gridfs/asynchronous/grid_file.py
NoahStapp: I like this suggestion--both methods have nearly identical bodies, so combining them makes sense. For https://github.com/mongodb/mongo-python-driver/pull/2218, the two methods have slightly different bodies, but still similar enough that we could do the same thing there. Thoughts?

ShaneHarvey: What would the method signatures looks like?

NoahStapp: After trying this out a bit, I think it's a little too confusing/unclear to combine them.

ShaneHarvey: Could you share the signatures for posterity?

doc/changelog.rst
```diff
@@ -9,6 +9,8 @@ PyMongo 4.12 brings a number of changes including:
 - Support for configuring DEK cache lifetime via the ``key_expiration_ms`` argument to
   :class:`~pymongo.encryption_options.AutoEncryptionOpts`.
 - Support for $lookup in CSFLE and QE supported on MongoDB 8.1+.
+- Added :meth:`gridfs.asynchronous.grid_file.AsyncGridFSBucket.rename_by_name` and :meth:`gridfs.grid_file.GridFSBucket.rename_by_name`
+  for more performant renaming of file revisions.
```
ShaneHarvey, line 13: `of file revisions.` -> `of a file with multiple revisions.`

ShaneHarvey (changes requested):
One minor comment. Otherwise LGTM.

gridfs/asynchronous/grid_file.py
NoahStapp: `async def rename(file_id: Optional[Any], new_filename: str, filename: Optional[str] = None,...)`

ShaneHarvey (approved):

ShaneHarvey (approved):
LGTM, test failures are unrelated.

PR #2221, opened 2025-03-25
PYTHON-4939 - Sync CSOT runCursorCommand test

NoahStapp:


ShaneHarvey (approved):

ShaneHarvey:
Removed "bug" from the title to avoid confusing this with an actual bug fix.

PR #2227, opened 2025-03-25
PYTHON-4933 - Allow drivers to set bypassDocumentValidation: false on…

NoahStapp:
… write commands

### DIFF
--- a/pymongo/asynchronous/bulk.py
+++ b/pymongo/asynchronous/bulk.py
@@ -87,7 +87,7 @@ class _AsyncBulk:
         self,
         collection: AsyncCollection[_DocumentType],
         ordered: bool,
-        bypass_document_validation: Optional[bool] = None,
+        bypass_document_validation: Optional[bool],
         comment: Optional[str] = None,
         let: Optional[Any] = None,
     ) -> None:
--- a/pymongo/asynchronous/bulk.py
+++ b/pymongo/asynchronous/bulk.py
@@ -87,7 +87,7 @@ class _AsyncBulk:
         self,
         collection: AsyncCollection[_DocumentType],
         ordered: bool,
-        bypass_document_validation: bool,
+        bypass_document_validation: Optional[bool] = None,
         comment: Optional[str] = None,
         let: Optional[Any] = None,
     ) -> None:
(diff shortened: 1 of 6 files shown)


pymongo/asynchronous/bulk.py
```diff
@@ -87,7 +87,7 @@ def __init__(
         self,
         collection: AsyncCollection[_DocumentType],
         ordered: bool,
-        bypass_document_validation: bool,
+        bypass_document_validation: Optional[bool] = None,
```
ShaneHarvey, line 90: This should be `bypass_document_validation: Optional[bool],` the arg should still be required.

pymongo/asynchronous/collection.py
```diff
@@ -800,8 +800,8 @@ async def _insert_one(
         ordered: bool,
         write_concern: WriteConcern,
         op_id: Optional[int],
-        bypass_doc_val: bool,
         session: Optional[AsyncClientSession],
+        bypass_doc_val: Optional[bool] = None,
```
ShaneHarvey, line 804: This should be `bypass_doc_val: Optional[bool],`. It should still be required positional arg in all our internal APIs since the default is never used anyway.

pymongo/asynchronous/collection.py
```diff
@@ -3356,7 +3357,13 @@ async def find_one_and_delete(
...
             kwargs["comment"] = comment
         return await self._find_and_modify(
-            filter, projection, sort, let=let, hint=hint, session=session, **kwargs
+            filter,
+            projection,
+            sort,
+            let=let,
+            hint=hint,
+            session=session,
+            **kwargs,
```
ShaneHarvey, line 3366: Was this change intentional? It seems unneeded.

ShaneHarvey (changes requested):

pymongo/asynchronous/collection.py
NoahStapp: Not intended, good catch.

pymongo/asynchronous/collection.py
NoahStapp: Our current internal APIs expect it to be a keyword argument and not positional. Is changing it to a positional argument intentional here?

ShaneHarvey: Previously it was positional:
```python
        write_concern: WriteConcern,
        op_id: Optional[int],
        bypass_doc_val: bool,
        session: Optional[AsyncClientSession],
```
There's no need to change it to have a default of None because we always pass a value anyway so the default will never be used.

ShaneHarvey: Let's just keep this one as positional and leave the rest.

ShaneHarvey (approved):

PR #2230, opened 2025-03-26
PYTHON-3674 Simplify transaction options in convenient API doc example code

blink1073:
From the ticket: "The [Convenient API for Transactions](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst) [examples in the MongoDB manual](https://www.mongodb.com/docs/manual/core/transactions/) use the following options:

"local" read concern
"primary" read preference
"majority" write concern with wTimeoutMS=1000
These no longer seem relevant and I propose removing them."


aclark4life (approved):
LGTM

PR #2234, opened 2025-03-26
PYTHON-4575 Allow valid SRV hostnames with less than 3 parts

aclark4life:


test/test_uri_parser.py
```diff
@@ -553,6 +554,11 @@ def test_port_with_whitespace(self):
         with self.assertRaisesRegex(ValueError, r"Port contains whitespace character: '\\n'"):
             parse_uri("mongodb://localhost:27\n017")
 
+    def test_allow_srv_hosts_with_fewer_than_three_dot_separated_parts(self):
```
ShaneHarvey, line 557: Could you implement the tests described here?: https://github.com/mongodb/specifications/commit/8e9d768dc2840eafdb9e2cb4788c608247c56a96#diff-dfcf948e6a987b880ea85e1907d5b7ce8fb98c9dab752f42ccdaf1e656b8848f

ShaneHarvey (changes requested):

test/test_uri_parser.py
aclark4life: Working on the last one now

pymongo/asynchronous/srv_resolver.py
```diff
@@ -96,8 +96,10 @@ def __init__(
         except Exception:
             raise ConfigurationError(_INVALID_HOST_MSG % (fqdn,)) from None
         self.__slen = len(self.__plist)
-        if self.__slen < 2:
-            raise ConfigurationError(_INVALID_HOST_MSG % (fqdn,))
+        if fqdn == self._resolve_uri(True)[0].target.to_text():
```
ShaneHarvey, line 99: We can't call resolve() here because we can't do I/O in the constructor.

ShaneHarvey (changes requested):

pymongo/asynchronous/srv_resolver.py
aclark4life: Thanks, fixed in 19d31c79443a3f4350d92d040a53fb782f4775d0

test/test_uri_parser.py
aclark4life: Yes, although the middle test here fails (by not raising a configuration error), not sure why yet.

```

            {
                "query": "_mongodb._tcp.localhost",                  
                "mock_target": "test_1.cluster_1localhost",        
                "expected_error": "Invalid SRV host",                     
            },                                                       
            {                   
                "query": "_mongodb._tcp.mongo.local",
                "mock_target": "test_1.my_hostmongo.local",
                "expected_error": "Invalid SRV host",
            },
            {
                "query": "_mongodb._tcp.blogs.mongodb.com",
                "mock_target": "cluster.testmongodb.com",
                "expected_error": "Invalid SRV host",
            },                     

```

test/test_uri_parser.py
```diff
@@ -553,6 +554,72 @@ def test_port_with_whitespace(self):
         with self.assertRaisesRegex(ValueError, r"Port contains whitespace character: '\\n'"):
             parse_uri("mongodb://localhost:27\n017")
```
ShaneHarvey, line 556: For the new "prose tests" could you add a comment with a link to the spec:
```
# https://github.com/mongodb/specifications/blob/0a7a8b5/source/initial-dns-seedlist-discovery/tests/README.md#prose-tests
```

And also prefix the test names with the prose test number like `test_1_allow_...`?

test/test_uri_parser.py
```diff
@@ -553,6 +554,72 @@ def test_port_with_whitespace(self):
...
 
+    def test_allow_srv_hosts_with_fewer_than_three_dot_separated_parts(self):
+        with patch("dns.resolver.resolve"):
+            parse_uri("mongodb+srv://localhost/")
+            parse_uri("mongodb+srv://mongo.local/")
+
+    def test_error_when_return_address_does_not_end_with_srv_domain(self):
```
ShaneHarvey, line 562: Could you split this into 3 separate tests so we can correspond them with the spec?

test/test_uri_parser.py
```diff
@@ -553,6 +554,72 @@ def test_port_with_whitespace(self):
...
+            with patch("dns.resolver.resolve") as mock_resolver:
+
+                def mock_resolve(query, record_type, *args, **kwargs):
+                    mock_srv = MagicMock()
+                    mock_srv.target.to_text.return_value = case["mock_target"]
+                    return [mock_srv]
+
+                mock_resolver.side_effect = mock_resolve
+                domain = case["query"].split("._tcp.")[1]
+                connection_string = f"mongodb+srv://{domain}"
```
ShaneHarvey, line 615: What do you think about making the test param include the full SRV address:
```
            {
                "uri": "mongodb://localhost",
                "mock_response": "localhost.mongodb",
                "expected_error": "Invalid SRV host",
            },
```
This way we don't need to parse the `_mongodb._tcp` part or dynamically generate a SRV address.

ShaneHarvey (changes requested):

test/test_uri_parser.py
aclark4life: Fixed in 577cd7f5e28dadf595c767debe59d75570ef7ac2

test/test_uri_parser.py
aclark4life: Not sure but I think I just fixed the broken test case in 577cd7f5e28dadf595c767debe59d75570ef7ac2. This could go in now AFAICT, so additional changes would be refactors.

aclark4life: Oh wait, forgot to check evergreen 😂

aclark4life: LGTM

pymongo/asynchronous/srv_resolver.py
```diff
@@ -92,12 +92,14 @@ def __init__(
             pass
 
         try:
-            self.__plist = self.__fqdn.split(".")[1:]
+            self.__plist = (
+                self.__fqdn.split(".")[1:]
+                if len(self.__fqdn.split(".")) > 2
+                else self.__fqdn.split(".")
+            )
```
ShaneHarvey, line 99: Can you refactor this so we only call split() once?

pymongo/asynchronous/srv_resolver.py
```diff
@@ -131,6 +133,11 @@ async def _get_srv_response_and_hosts(
     ) -> tuple[resolver.Answer, list[tuple[str, Any]]]:
         results = await self._resolve_uri(encapsulate_errors)
 
+        if self.__fqdn == results[0].target.to_text():
+            raise ConfigurationError(
+                "Invalid SRV host: return address is identical to SRV hostname"
+            )
```
ShaneHarvey, line 139: Don't we need to validate this for all results? Can you refactor this to happen in the "Validate hosts" loop below? That way we don't need to call target.to_text() multiple times.

test/test_uri_parser.py
```diff
@@ -24,6 +24,7 @@
 sys.path[0:0] = [""]
 
 from test import unittest
+from unittest.mock import MagicMock, patch
```
ShaneHarvey, line 27: I would have expected this test to be async. Can you check if we already have a ticket for converting this file and open one if we don't?

test/test_uri_parser.py
```diff
@@ -553,6 +554,119 @@ def test_port_with_whitespace(self):
...
+
+                mock_resolver.side_effect = mock_resolve
+                domain = case["query"].split("._tcp.")[1]
+                connection_string = f"mongodb+srv://{domain}"
+                try:
+                    parse_uri(connection_string)
+                except ConfigurationError as e:
+                    self.assertIn(case["expected_error"], str(e))
+                else:
+                    self.fail(f"ConfigurationError was not raised for query: {case['query']}")
```
ShaneHarvey, line 599: Could you refactor these tests to use a helper method to reduce the code duplication?

test/test_uri_parser.py
```diff
@@ -553,6 +554,119 @@ def test_port_with_whitespace(self):
...
+                "mock_target": "blogs.evil.com",
+                "expected_error": "Invalid SRV host",
+            },
+            {
+                "query": "_mongodb._tcp.blogs.mongo.local",
+                "mock_target": "test_1.evil.com",
+                "expected_error": "Invalid SRV host",
+            },
+        ]
+        for case in test_cases:
```
ShaneHarvey, line 583: This for loop looks to be identical between these 3 tests. Can you refactor them to reduce the code duplication?

aclark4life: Done in 5c823fcd4037e68a8b46b286f9fc0e7f030c483e

pymongo/asynchronous/srv_resolver.py
aclark4life: Done in b6920f1110b3602d9525232169050fb39ed7baf1

test/test_uri_parser.py
aclark4life: IIRC, @NoahStapp said we don't need an async test …

ShaneHarvey: How's that? Don't we need to test the async code path for SRV parsing?

pymongo/asynchronous/srv_resolver.py
aclark4life: Done in f72cf8725f3064fd5293a8920c782044a86e9538, although now I wonder why the test cases passed prior to this change … also did someone miss committing a synchro change?

```
On branch PYTHON-4575
Your branch is up to date with 'origin/PYTHON-4575'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	gridfs/synchronous/__init__.py

```

test/test_uri_parser.py
aclark4life: Done in 5c823fcd4037e68a8b46b286f9fc0e7f030c483e

test/test_uri_parser.py
aclark4life: 🤷  I'll follow up on Slack and open a ticket as needed

test/asynchronous/test_dns.py
```diff
@@ -39,6 +40,26 @@
 _IS_SYNC = False
 
 
+def run_initial_dns_seedlist_discovery_prose_tests(self, test_cases):
+    for case in test_cases:
+        with patch("dns.resolver.resolve") as mock_resolver:
```
ShaneHarvey, line 45: The async version of this test needs to patch the async resolve method.

ShaneHarvey (changes requested):

test/asynchronous/test_dns.py
```diff
@@ -39,6 +40,26 @@
 _IS_SYNC = False
 
 
+def run_initial_dns_seedlist_discovery_prose_tests(self, test_cases):
```
ShaneHarvey, line 43: Nit: this should be a method on the test class.

pymongo/asynchronous/srv_resolver.py
```diff
@@ -138,7 +136,12 @@ async def _get_srv_response_and_hosts(
         ]
 
         # Validate hosts
+        srv_host = results[0].target.to_text()
         for node in nodes:
+            if self.__fqdn == srv_host:
```
ShaneHarvey, line 141: Shouldn't this be validating each returned host in "nodes"? like this:
```python
        for node in nodes:
            if self.__fqdn == node[0].lower():
```

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #2242, opened 2025-03-27
Update common.py

AARN0LD:
when the common.py tries to import SOM from bson it fails due to the fact the import needs to be from bson.som import SOM.

This is actually referenced correctly later on in some of the exception handling where it mentions bson.som.SOM

### DIFF
--- a/pymongo/common.py
+++ b/pymongo/common.py
@@ -36,7 +36,7 @@ from typing import (
 )
 from urllib.parse import unquote_plus
 
-from bson import SON
+from bson.som import SON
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
> it fails due to the fact the import needs to be from bson.som import SOM.

What do you mean it fails? Both imports should work:
```python
>>> from bson import SON
>>> SON
<class 'bson.son.SON'>
>>> from bson.son import SON
>>> SON
<class 'bson.son.SON'>
```

PR #2244, opened 2025-03-28
PYTHON-4924 - PoolClearedError should have TransientTransactionError …

NoahStapp:
…label appended to it

### DIFF
--- a/pymongo/asynchronous/pool.py
+++ b/pymongo/asynchronous/pool.py
@@ -190,6 +190,10 @@ def _raise_connection_failure(
 ) -> NoReturn:
     """Convert a socket.error to ConnectionFailure and raise it."""
     host, port = address
+    if isinstance(error, PyMongoError) and error._error_labels:
+        labels = error._error_labels
+    else:
+        labels = None
     # If connecting to a Unix socket, port will be None.
     if port is not None:
         msg = "%s:%d: %s" % (host, port, error)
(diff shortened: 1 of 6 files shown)


pymongo/asynchronous/pool.py
```diff
@@ -190,6 +190,10 @@ def _raise_connection_failure(
 ) -> NoReturn:
     """Convert a socket.error to ConnectionFailure and raise it."""
     host, port = address
+    if isinstance(error, PyMongoError) and error._error_labels:
```
ShaneHarvey, line 193: I don't think we need any pymongo changes here. We already add the error label to PoolClearedErrors here: https://github.com/mongodb/mongo-python-driver/blob/eea8a3725799c6e4d7286336ceb229c5025ebb25/pymongo/synchronous/mongo_client.py#L2607

ShaneHarvey (changes requested):

pymongo/asynchronous/pool.py
NoahStapp: Would we expect the added test to still pass without the code changes then? It fails without these changes.

NoahStapp: Got it, got it. We can go ahead and close this PR and ticket it sounds like then.

ShaneHarvey: Can you confirm that we have a test for "pool clear error" having the label? If not we should add one.

NoahStapp: Our `Start Transactions Retry Example 1`  section of `TestTransactionExamples` in `test_examples.py` refers to the label, but I don't think it is intended to fail every time and check for the label: https://github.com/mongodb/mongo-python-driver/blob/58a41ae7f5811e4a76ee66aecc2fe875321e294e/test/asynchronous/test_examples.py#L1007.

ShaneHarvey (approved):

PR #2245, opened 2025-03-28
PYTHON-4833 - Add Collection bulk_write benchmarks

NoahStapp:
Shruti had already written the client bulk_write benchmarks, so only the non-mixed collection ones needed to be added.


blink1073 (approved):
LGTM!

blink1073 (approved):

PR #2248, opened 2025-03-31
PYTHON-4557 - Add log message for retried commands

NoahStapp:


pymongo/asynchronous/mongo_client.py
```diff
@@ -2825,6 +2834,11 @@ async def _write(self) -> T:
                     # not support sessions raise the last error.
                     self._check_last_error()
                     self._retryable = False
+                if self._retrying:
+                    _debug_log(
+                        _COMMAND_LOGGER,
+                        message=f"Retrying write attempt number {self._attempt_number}",
```
ShaneHarvey, line 2840: Is there any fields we should add here that can be used to correlate the two retry attempts? Maybe _operation and/or _operation_id?

ShaneHarvey (changes requested):

pymongo/asynchronous/mongo_client.py
NoahStapp: With those fields added, this is what the message looks like:

`{"message": "Retrying read attempt number <x> for operation 'op_name' with id <id>}"}`

pymongo/asynchronous/mongo_client.py
```diff
@@ -2837,7 +2837,7 @@ async def _write(self) -> T:
                 if self._retrying:
                     _debug_log(
                         _COMMAND_LOGGER,
-                        message=f"Retrying write attempt number {self._attempt_number}",
+                        message=f"Retrying write attempt number {self._attempt_number} for operation `{self._operation}` with id {self._operation_id}",
```
ShaneHarvey, line 2840: Let's use the logging format:
```python
                    _debug_log(
                        _COMMAND_LOGGER,
                        message=f"Retrying write attempt number {self._attempt_number}",
                        clientId=...,
                        commandName=self._operation,
                        operationId=self._operation_id
```

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #2258, opened 2025-04-02
PYTHON-5252 Add dependency on pymongocrypt 1.13

blink1073:


sleepyStick (dismissed):

uv.lock
```diff
@@ -997,7 +998,6 @@ sdist = { url = "https://files.pythonhosted.org/packages/07/e9/ae44ea7d7605df9e5
 
 [[package]]
 name = "pymongo"
-version = "4.12.0.dev0"
```
sleepyStick, line 1000: was this an intentional deletion? *curious*

blink1073: It was not, but is a welcome change because imo the local package version shouldn't be part of the lock file.

sleepyStick: Ah i see, in one of my previous tickets it was auto-deleted and i kept adding it back because I thought we needed it there HAHA

sleepyStick: Ah i see, in one of my previous tickets it was auto-deleted and i kept adding it back because I thought we needed it there HAHA

sleepyStick (approved):

PR #2259, opened 2025-04-02
PYTHON-5263 Convert s3-related functions to generated config

blink1073:


NoahStapp (commented):
Can you schedule a patch build that ensures all of these functions don't behave differently than expected?

blink1073:
Mo Artifacts and upload coverage:

image


Download and merge coverage:

image

We don't actually use the upload working dir anymore.

NoahStapp (approved):

NoahStapp:
Can we remove the upload working dir then?

blink1073:
> Can we remove the upload working dir then?

Yeah makes sense.

PR #2261, opened 2025-04-03
PYTHON-5126 & PYTHON-5280 Addresses issues raised in DRIVERS-3097 and DRIVERS-3123 

caseyclements:
- See PYTHON-5126 and DRIVERS-3123 for details.

### DIFF
--- a/bson/binary.py
+++ b/bson/binary.py
@@ -462,6 +462,10 @@ class Binary(bytes):
                 raise ValueError(f"{padding=}. It must be in [0,1, ..7].")
             if padding and not vector:
                 raise ValueError("Empty vector with non-zero padding.")
+            if padding and not (vector[-1] & ((1 << padding) - 1)) == 0:
+                raise ValueError(
+                    "If padding p is provided, all bits in the final byte lower than p must be 0."
+                )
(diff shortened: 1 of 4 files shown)


bson/binary.py
```diff
@@ -513,6 +522,12 @@ def as_vector(self) -> BinaryVector:
             dtype_format = "B"
             format_string = f"<{n_values}{dtype_format}"
             unpacked_uint8s = list(struct.unpack_from(format_string, self, position))
+            if padding and not n_values:
+                raise ValueError("Corrupt data. Vector has a padding P, but no data.")
+            if padding and n_values and not (unpacked_uint8s[-1] & ((1 << padding) - 1)) == 0:
+                raise ValueError(
```
ShaneHarvey, line 528: Are these breaking changes for existing data?

caseyclements: They are, yeah. But only for a case that vector search doesn't use. I feel  confident that it won't affect anyone now.

blink1073 (approved):
LGTM!

bson/binary.py
ShaneHarvey: Please add a changelog entry to explain this because it's a breaking change.

PR #2262, opened 2025-04-03
PYTHON-1752 bulk_write should be able to accept a generator

sleepyStick:
Notes: 
- I only did this for collection.bulk_write, if we like how it was done, i'd be happy to replicate the change for client.bulk_write, but didn't wanna make the change until I knew we liked how it was done here
- if a user passes in a generator and doesn't care about ordering, i currently still consume the generator to sort the group the requests by type...the user can pass in a generator and insist on original ordering which wouldn't consume the generator. I was thinking of maybe changing this so that we only consume part of the generator (like maybe up until max_batch_size instead of the whole generator but this seemed to require creating an _AsyncBulk class for each batch which felt...wrong? I was probably going about this incorrectly? Not sure. 
- the test that i picked to duplicate to check generators was a bit arbitrary, but it felt like a good one to pick. If we want more, I'm happy to do so and open to suggestions for which tests specifically.

### DIFF
--- a/pymongo/asynchronous/bulk.py
+++ b/pymongo/asynchronous/bulk.py
@@ -26,6 +26,7 @@ from itertools import islice
 from typing import (
     TYPE_CHECKING,
     Any,
+    Generator,
     Iterator,
     Mapping,
     Optional,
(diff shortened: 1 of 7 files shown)


pymongo/asynchronous/bulk.py
```diff
@@ -214,28 +215,45 @@ def add_delete(
...
         self.ops.append((_DELETE, cmd))
 
-    def gen_ordered(self) -> Iterator[Optional[_Run]]:
+    def gen_ordered(self, requests) -> Iterator[Optional[_Run]]:
         """Generate batches of operations, batched by type of
         operation, in the order **provided**.
         """
         run = None
-        for idx, (op_type, operation) in enumerate(self.ops):
+        for idx, request in enumerate(requests):
```
ShaneHarvey, line 223: I think the goal of this ticket is to avoid inflating the whole generator upfront and only iterate requests as they are needed at the encoding step. For example:
```python
coll.bulk_write((InsertOne({'x': 'large'*1024*1024}) for _ in range(1_000_000))
```

If we inflate all at once like we do here, then that code will need to allocate all 1 million documents at once.

PR #2269, opened 2025-04-09
PYTHON-4324 CSOT avoid connection churn when operations timeout

ShaneHarvey:
[Draft] PYTHON-4324 CSOT avoid connection churn when operations timeout


ShaneHarvey:
Closing. The spec has changed so this work will need to be redone if we decide to go this route.

PR #2272, opened 2025-04-09
PYTHON-5288: SRV hostname validation fails when resolver and resolved hostnames are identical with three domain levels

aclark4life:


pymongo/asynchronous/srv_resolver.py
```diff
@@ -137,7 +137,7 @@ async def _get_srv_response_and_hosts(
 
         # Validate hosts
         for node in nodes:
-            if self.__fqdn == node[0].lower():
+            if self.__fqdn == node[0].lower() and len(node[0].split(".")) < 3:
```
ShaneHarvey, line 140: Is this no spec tests for this requirement? If not we need to open a ticket to add them. If so, we need to sync/implement them. Either way this PR should add a test.

ShaneHarvey (changes requested):
Could you please update the PR title to be more descriptive (and include the jira ticket) and add a changelog entry?

pymongo/asynchronous/srv_resolver.py
aclark4life: Test added for spec test `3. Throw when return address is identical to SRV hostname`. Of the 4 prose tests, only the first one tests for success, the rest test for errors. I added a test case for success in the third prose test, which arguably belongs in the first prose test case which already tests for successes. I could update the first test case to include mock results and move the new test there if need be. Not sure if there is a clear enough argument to be made for a new spec test _ticket_, but maybe.

ShaneHarvey: > Not sure if there is a clear enough argument to be made for a new spec test ticket, but maybe.

We 100% need a drivers ticket to add a new spec test for this case.

pymongo/synchronous/srv_resolver.py
```diff
@@ -137,7 +137,8 @@ def _get_srv_response_and_hosts(
 
         # Validate hosts
         for node in nodes:
-            if self.__fqdn == node[0].lower():
+            srv_host = node[0].lower()
+            if self.__fqdn == srv_host and len(srv_host.split(".")) < 3:
```
ShaneHarvey, line 141: Can we use `self.__slen` instead of `srv_host.split(".")`? We could also remain `__slen` for clarify.

test/asynchronous/test_dns.py
```diff
@@ -264,6 +267,12 @@ async def test_3_throw_when_return_address_is_identical_to_srv_hostname(self):
                 "mock_target": "mongo.local",
                 "expected_error": "Invalid SRV host",
             },
+            # When the SRV hostname has three or more dot-separated parts
+            # it is valid for the returned hostnames to be identical.
+            {
+                "query": "_mongodb._tcp.blogs.mongodb.com",
+                "mock_target": "blogs.mongodb.com",
+            },
```
ShaneHarvey, line 275: I suggest moving this to a new test since we're testing success here and the test name is "test_3_throw_when_return_address_is_identical_to_srv_hostname".

ShaneHarvey (changes requested):

pymongo/synchronous/srv_resolver.py
ShaneHarvey: And I don't know why this class is used `__` attributes, this is a private class anyway so we can rename `self.__fqdn` -> `self.fqdn` and `self.__slen` -> `self.nparts` or something.

But up to you, I'm less concerned about the naming in this PR. My bigger point is that the check here should be based on the length of FQDN, not the length of the response, to match the spec's description.

pymongo/asynchronous/srv_resolver.py
aclark4life: In addition to [this still-open one](https://jira.mongodb.org/browse/DRIVERS-2922) right ?

pymongo/asynchronous/srv_resolver.py
```diff
@@ -96,6 +96,7 @@ def __init__(
         except Exception:
             raise ConfigurationError(_INVALID_HOST_MSG % (fqdn,)) from None
         self.__slen = len(self.__plist)
+        self.nparts = len(self.__fqdn.split("."))
```
ShaneHarvey, line 99: We can use `split_fqdn` here instead of splitting again.

ShaneHarvey (changes requested):

pymongo/asynchronous/srv_resolver.py
ShaneHarvey: Yes in addition.

ShaneHarvey:
Looks good, let's add the changelog entry.

ShaneHarvey (approved):

pymongo/asynchronous/srv_resolver.py
aclark4life: https://jira.mongodb.org/browse/DRIVERS-3162

PR #2273, opened 2025-04-09
PYTHON-5297 - AsyncMongoClient connection error causes UnboundLocalError

NoahStapp:


pymongo/pool_shared.py
```diff
@@ -346,12 +346,10 @@ async def _configured_protocol_interface(
             ssl=ssl_context,
         )
     except _CertificateError:
-        transport.abort()
         # Raise _CertificateError directly like we do after match_hostname
         # below.
         raise
     except (OSError, SSLError) as exc:
```
ShaneHarvey, line 352: We still call transport.abort() when match_hostname fails below. What happens after abort is called? Don't we need to reap the background tasks somehow to avoid "Task was never awaited" warnings??

NoahStapp: There aren't any background tasks associated with the `transport, protocol` pair here. Are you referring to the monitor background tasks?

ShaneHarvey: Oh right, I forgot it's callback based.

ShaneHarvey (dismissed):

ShaneHarvey:
Good to add a changelog entry here referencing the UnboundLocalError bugfix.

ShaneHarvey (approved):

blink1073 (approved):
LGTM!

PR #2275, opened 2025-04-10
PYTHON-5304 Create standard non-linux tests

blink1073:
Patch build: https://spruce.mongodb.com/version/67f7dc90424300000778a187/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Before: 384 tasks across 8 build variants
After: 32 tasks across 4 build variants

image

image

### DIFF
--- a/.evergreen/scripts/generate_config_utils.py
+++ b/.evergreen/scripts/generate_config_utils.py
@@ -44,6 +44,7 @@ DISPLAY_LOOKUP = dict(
         standalone="Standalone", replica_set="Replica Set", sharded_cluster="Sharded Cluster"
     ),
     test_suites=dict(default="Sync", default_async="Async"),
+    sync={"sync": "Sync", "async": "Async"},
     coverage={"1": "cov"},
     no_ext={"1": "No C"},
 )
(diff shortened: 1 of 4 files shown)


blink1073:
The PYTHON_VERSION handling isn't working for Win32

blink1073:
Fixed: Using Python 3.9.13 environment at: C:\python\32\Python39.

I also updated to test all three topologies for each MongoDB version, since only one felt insufficient.

.evergreen/scripts/generate_config_utils.py
```diff
@@ -36,14 +36,15 @@
     ".replica_set .noauth .ssl",
     ".standalone .noauth .nossl",
 ]
-SYNCS = ["sync", "async", "sync_async"]
+SYNCS = ["sync", "async"]
```
NoahStapp, line 39: Is `sync_async` implied by default now?

NoahStapp (changes requested):

.evergreen/scripts/generate_config_utils.py
blink1073: Yes, it is the default test type.

NoahStapp (approved):

PR #2276, opened 2025-04-10
PYTHON-5288: [v4.12] SRV hostname validation fails when resolver and resolved hostnames are identical with three domain levels

blink1073:


aclark4life (approved):
LGTM, thanks!

PR #2282, opened 2025-04-14
PYTHON-5311 Create standard linux evergreen tasks

blink1073:
Patch build: https://spruce.mongodb.com/version/67fd13625ee31c0007ce6dec/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Before: 35 tasks across 3 build variants
After: 78 tasks across 3 build variants

### DIFF
--- a/.evergreen/scripts/generate_config.py
+++ b/.evergreen/scripts/generate_config.py
@@ -205,13 +205,15 @@ def create_load_balancer_variants():
 
 
 def create_compression_variants():
-    # Compression tests - standalone versions of each server, across python versions.
+    # Compression tests - use the standard linux tests.
     host = DEFAULT_HOST
-    base_task = ".compression"
     variants = []
     for compressor in "snappy", "zlib", "zstd":
         expansions = dict(COMPRESSOR=compressor)
-        tasks = [base_task] if compressor != "zstd" else [f"{base_task} !.4.0"]
+        if compressor == "zstd":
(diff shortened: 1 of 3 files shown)


.evergreen/scripts/generate_config.py
```diff
@@ -607,6 +609,30 @@ def create_server_version_tasks():
     return tasks
 
 
+def create_standard_linux_tasks():
+    tasks = []
+
+    for (version, topology), python in zip_cycle(
+        list(product(ALL_VERSIONS, TOPOLOGIES)), ALL_PYTHONS
+    ):
+        auth = "auth" if topology == "sharded_cluster" else "noauth"
```
NoahStapp, line 618: Do we have coverage of every auth-ssl-topology combo elsewhere? Always running set combinations here seems potentially risky.

NoahStapp (changes requested):

.evergreen/scripts/generate_config.py
blink1073: We do on the `* MongoDB` tasks.  This is essentially the same combination we have now for encryption tests.

NoahStapp: What differentiates these tasks from the `* MongoDB` tasks?

blink1073: It is trimmed down to not have every combination of topology/auth/ssl, since it is unlikely that it will have a general affect.

PR #2283, opened 2025-04-14
PYTHON-5310 - Fix uri_parser AttributeError when used directly

NoahStapp:


ShaneHarvey (changes requested):

test/test_default_exports.py
```diff
@@ -209,6 +209,11 @@ def test_pymongo_imports(self):
         )
         from pymongo.write_concern import WriteConcern, validate_boolean
 
+    def test_pymongo_submodule_attributes(self):
+        import pymongo
+
+        self.assertTrue(hasattr(pymongo, "uri_parser"))
```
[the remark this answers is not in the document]
NoahStapp: Yes, this assertion fails before this fix.

pymongo/__init__.py
```diff
@@ -45,6 +45,7 @@
     "WriteConcern",
     "has_c",
     "timeout",
+    "uri_parser",
```
ShaneHarvey, line 48: Is this the right fix? Back in 4.7 uri_parser wasn't part of `__all__` and this bug didn't exist. Perhaps we shouldn't include uri_parser here and instead we should `import pymongo.uri_parser` like we did before.

ShaneHarvey (changes requested):

pymongo/__init__.py
ShaneHarvey: I'm worried about unintentional side effects of including it in `__all__`.

NoahStapp: Good catch, we don't actually need to include `uri_parser` in `__all__`.

ShaneHarvey (approved):

PR #2285, opened 2025-04-15
PYTHON-5306 - Fix use of public MongoClient attributes before connection

NoahStapp:
The public attributes of `MongoClient` are `topology_description`, `nodes`, `address`, `primary`, `secondaries`, and `arbiters`. Of these, only `topology_description` and `nodes` do not already perform network I/O.

### DIFF
--- a/pymongo/asynchronous/mongo_client.py
+++ b/pymongo/asynchronous/mongo_client.py
@@ -109,6 +109,7 @@ from pymongo.operations import (
 )
 from pymongo.read_preferences import ReadPreference, _ServerMode
 from pymongo.results import ClientBulkWriteResult
+from pymongo.server_description import ServerDescription
 from pymongo.server_selectors import writable_server_selector
 from pymongo.server_type import SERVER_TYPE
 from pymongo.topology_description import TOPOLOGY_TYPE, TopologyDescription
@@ -779,7 +780,7 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
         keyword_opts["document_class"] = doc_class
         self._resolve_srv_info: dict[str, Any] = {"keyword_opts": keyword_opts}
 
-        seeds = set()
+        self._seeds = set()
         is_srv = False
         username = None
         password = None
@@ -804,18 +805,18 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
                     srv_max_hosts=srv_max_hosts,
                 )
                 is_srv = entity.startswith(SRV_SCHEME)
-                seeds.update(res["nodelist"])
+                self._seeds.update(res["nodelist"])
                 username = res["username"] or username
                 password = res["password"] or password
                 dbase = res["database"] or dbase
                 opts = res["options"]
                 fqdn = res["fqdn"]
             else:
-                seeds.update(split_hosts(entity, self._port))
-        if not seeds:
+                self._seeds.update(split_hosts(entity, self._port))
+        if not self._seeds:
             raise ConfigurationError("need to specify at least one host")
 
-        for hostname in [node[0] for node in seeds]:
+        for hostname in [node[0] for node in self._seeds]:
             if _detect_external_db(hostname):
                 break
 
@@ -838,7 +839,7 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
             srv_service_name = opts.get("srvServiceName", common.SRV_SERVICE_NAME)
 
         srv_max_hosts = srv_max_hosts or opts.get("srvmaxhosts")
-        opts = self._normalize_and_validate_options(opts, seeds)
+        opts = self._normalize_and_validate_options(opts, self._seeds)
 
         # Username and password passed as kwargs override user info in URI.
         username = opts.get("username", username)
@@ -857,7 +858,7 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
                 "username": username,
                 "password": password,
                 "dbase": dbase,
-                "seeds": seeds,
+                "seeds": self._seeds,
                 "fqdn": fqdn,
                 "srv_service_name": srv_service_name,
                 "pool_class": pool_class,
@@ -874,7 +875,7 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
         )
 
         if not is_srv:
-            self._init_based_on_options(seeds, srv_max_hosts, srv_service_name)
+            self._init_based_on_options(self._seeds, srv_max_hosts, srv_service_name)
 
         self._opened = False
         self._closed = False
@@ -1205,6 +1206,18 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
 
         .. versionadded:: 4.0
         """
+        if self._topology is None:
+            servers = {
+                (host, self._port): ServerDescription((host, self._port)) for host in self._seeds
+            }
+            return TopologyDescription(
+                TOPOLOGY_TYPE.Unknown,
+                servers,
+                None,
+                None,
+                None,
+                TopologySettings(),
+            )
         return self._topology.description
 
     @property
@@ -1218,6 +1231,8 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
           to any servers, or a network partition causes it to lose connection
           to all servers.
         """
+        if self._topology is None:
+            return frozenset()
         description = self._topology.description
         return frozenset(s.address for s in description.known_servers)
 
@@ -1576,6 +1591,8 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
 
         .. versionadded:: 3.0
         """
+        if self._topology is None:
+            await self._get_topology()
         topology_type = self._topology._description.topology_type
         if (
             topology_type == TOPOLOGY_TYPE.Sharded
@@ -1598,6 +1615,8 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
         .. versionadded:: 3.0
            AsyncMongoClient gained this property in version 3.0.
         """
+        if self._topology is None:
+            await self._get_topology()
         return await self._topology.get_primary()  # type: ignore[return-value]
 
     @property
@@ -1611,6 +1630,8 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
         .. versionadded:: 3.0
            AsyncMongoClient gained this property in version 3.0.
         """
+        if self._topology is None:
+            await self._get_topology()
         return await self._topology.get_secondaries()
 
     @property
@@ -1621,6 +1642,8 @@ class AsyncMongoClient(common.BaseObject, Generic[_DocumentType]):
         connected to a replica set, there are no arbiters, or this client was
         created without the `replicaSet` option.
         """
+        if self._topology is None:
+            await self._get_topology()
         return await self._topology.get_arbiters()
 
     @property
--- a/pymongo/synchronous/mongo_client.py
+++ b/pymongo/synchronous/mongo_client.py
@@ -101,6 +101,7 @@ from pymongo.operations import (
 )
 from pymongo.read_preferences import ReadPreference, _ServerMode
 from pymongo.results import ClientBulkWriteResult
+from pymongo.server_description import ServerDescription
 from pymongo.server_selectors import writable_server_selector
 from pymongo.server_type import SERVER_TYPE
 from pymongo.synchronous import client_session, database, uri_parser
@@ -777,7 +778,7 @@ class MongoClient(common.BaseObject, Generic[_DocumentType]):
         keyword_opts["document_class"] = doc_class
         self._resolve_srv_info: dict[str, Any] = {"keyword_opts": keyword_opts}
 
-        seeds = set()
+        self._seeds = set()
         is_srv = False
         username = None
         password = None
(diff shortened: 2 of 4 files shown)


pymongo/asynchronous/mongo_client.py
```diff
@@ -1205,6 +1206,18 @@ def topology_description(self) -> TopologyDescription:
 
         .. versionadded:: 4.0
         """
+        if self._topology is None:
```
NoahStapp, line 1209: We return a `TopologyDescription` with placeholder servers derived from the seeds passed to the client's constructor.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1218,6 +1231,8 @@ def nodes(self) -> FrozenSet[_Address]:
           to any servers, or a network partition causes it to lose connection
           to all servers.
         """
+        if self._topology is None:
```
NoahStapp, line 1234: As the docstring defines, we expect `nodes` to return an empty set if the client has not yet connected.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: We should make sure the returned TopologyDescription has the same _topology_id. Otherwise it could be confusing because _topology_id will change.

test/asynchronous/test_client.py
```diff
@@ -2170,6 +2194,18 @@ async def test_uuid_queries(self):
...
+        client = self.simple_client("mongodb+srv://test1.test.build.10gen.cc/", connect=False)
+        self.assertEqual(client.nodes, frozenset())
+        topology_description = client.topology_description
+        self.assertEqual(topology_description.topology_type, TOPOLOGY_TYPE.Unknown)
+        self.assertEqual(
+            topology_description.server_descriptions(),
+            {("unknown", None): ServerDescription(("unknown", None))},
+        )
+        await client.aconnect()
+        self.assertEqual(await client.address, None)
```
ShaneHarvey, line 2207: Can we add a tests for `address` and all the helper helpers without calling `aconnect`?

test/asynchronous/test_client.py
```diff
@@ -816,6 +816,30 @@ async def test_constants(self):
...
+        self.assertIsNotNone(c._topology)
+        c = await self.async_rs_or_single_client(connect=False)
+        # secondaries causes client to block until connected
+        await c.secondaries
+        self.assertIsNotNone(c._topology)
+        c = await self.async_rs_or_single_client(connect=False)
+        # arbiters causes client to block until connected
+        await c.arbiters
+        self.assertIsNotNone(c._topology)
+        c = await self.async_rs_or_single_client(connect=False)
```
ShaneHarvey, line 842: In this kind of test it's a good idea to close each client once it's no longer needed to avoid building up unclosed clients.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: Also can you give an example of what topology_description looks like in the connect=False srv case?

ShaneHarvey (changes requested):

test/asynchronous/test_client.py
NoahStapp: This test was mistakenly included: our test SRV URIs aren't real servers, so we can't actually connect to them.

test/asynchronous/test_client.py
NoahStapp: `async_rs_or_single_client` already automatically closes all clients at the end of the method.

pymongo/asynchronous/mongo_client.py
NoahStapp: `<TopologyDescription id: 67ffb68a6e12ca176991f521, topology_type: Unknown, servers: [<ServerDescription (('test1.test.build.10gen.cc', None), 27017) server_type: Unknown, rtt: None>]>`

test/asynchronous/test_client.py
NoahStapp: This is actually a larger restriction: the new test needs to be run against a real SRV server, or have accurate responses mocked out, to ensure correct behavior.

test/asynchronous/test_client.py
ShaneHarvey: I know but I'm saying it's better to close them explicitly to avoid building up unclosed clients.

test/asynchronous/test_client.py
NoahStapp: Ah, I forgot that SRV implies replica + SSL, thanks!

test/asynchronous/test_client.py
NoahStapp: We don't currently use that pattern anywhere, relying on the helper client creation methods to clean up clients for us. If it's better to avoid relying on those, should we rethink our approach?

pymongo/synchronous/mongo_client.py
```diff
@@ -1203,6 +1206,18 @@ def topology_description(self) -> TopologyDescription:
...
         """
+        if self._topology is None:
+            servers = {(host, port): ServerDescription((host, port)) for host, port in self._seeds}
+            td = TopologyDescription(
+                TOPOLOGY_TYPE.Unknown,
+                servers,
+                None,
+                None,
+                None,
+                TopologySettings(),
```
ShaneHarvey, line 1217: This is missing certain settings that are publicly accessible, like TopologyDescription.heartbeat_frequency. 

In the ticket I suggested making the MongoClient constructor always create the Topology instance, and then mutate it later once the SRV info is resolved. Did you consider that approach? It would avoid the problem of attempting to mock out these attributes and avoid the type-ignores since everything would always be non-None.

(It could be that this approach is a better fit, I just want to know if an alternative was considered.)

test/asynchronous/test_client.py
```diff
@@ -849,6 +849,54 @@ async def test_init_disconnected_with_auth(self):
         with self.assertRaises(ConnectionFailure):
             await c.pymongo_test.test.find_one()
 
+    @async_client_context.require_replica_set
```
ShaneHarvey, line 852: I believe we can remove require_replica_set.

test/asynchronous/test_client.py
ShaneHarvey: I'm saying for tests that specifically open multiple/many clients, it's good practice to add an explicit call to close().

pymongo/asynchronous/mongo_client.py
```diff
@@ -767,6 +768,7 @@ def __init__(
         self._timeout: float | None = None
         self._topology_settings: TopologySettings = None  # type: ignore[assignment]
         self._event_listeners: _EventListeners | None = None
+        self._initial_topology_id: Optional[ObjectId] = None  # type: ignore[assignment]
```
ShaneHarvey, line 771: Could we remove the type ignore here?

ShaneHarvey (changes requested):

test/asynchronous/test_client.py
NoahStapp: Not trying to pushback stubbornly here, just want to keep our codebase consistent where possible since we don't currently do that. Do we have a measurable improvement or benefit from explicitly closing clients that will be cleaned up by the test?

pymongo/synchronous/mongo_client.py
NoahStapp: I started with the approach you suggested, but switched to the current implementation for a few reasons:

1. `topology_description` is the only public attribute that requires changes of meaningful size. The rest either already do IO or are `node`, which has a trivial workaround. 
2. We don't currently mutate `Topology` instances or `TopologySettings` instances directly. To do so, we'd need to make more code changes and be confident that we haven't overlooked any problematic edge cases.
3. I foresee more potential bugs/confusion caused by introducing a new state specifically for SRV Topologies where we know that the Topology we currently have is not actually correct, and will be mutated upon SRV resolution.

This approach has fewer changes that could cause errors, especially since it isolates changes to resolving the specific issue at hand.

test/asynchronous/test_client.py
NoahStapp: Removing it causes failures on any non-replica topology: https://spruce.mongodb.com/task/mongo_python_driver_mongodb_latest_test_python3.9_auth_ssl_standalone_cov_patch_846b1fc25ced5979ade934bb27d6ffbb60f68267_68010865cc419c000710589f_25_04_17_13_56_05?execution=0&sortBy=STATUS&sortDir=ASC

I assume this is due to SRV requiring SSL + replica to be enabled.

test/asynchronous/test_client.py
ShaneHarvey: It's not strictly a performance issue. It's also about easier debugging. Leaving open clients around unnecessarily like this makes debugging harder because each client has many threads/tasks causing noise.

test/asynchronous/test_client.py
ShaneHarvey: Oh I see. "mongodb+srv://test1.test.build.10gen.cc" returns 2 hosts so it can't connect to standalone. It can connect to Mongos though so we should run this test on both replica set and sharded.

NoahStapp: `require_no_standalone`, then?

ShaneHarvey: Load balancers probably dont work here so it would need to be `@require_no_standalone+@require_no_load_balancer+@require_no_load_balancer`

test/asynchronous/test_client.py
NoahStapp: Ah, makes sense! I'll open a ticket to make sure we follow the same pattern elsewhere.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: Please include an example of this new/old behavior in the Jira ticket when closing.

pymongo/synchronous/mongo_client.py
ShaneHarvey: Thanks, could you update `TopologySettings()` here to be more accurate then?

ShaneHarvey (changes requested):

pymongo/synchronous/mongo_client.py
NoahStapp: `TopologySettings` sets defaults for each attribute, is passing no arguments to it not valid?

ShaneHarvey: Like I said above, TopologyDescription.heartbeat_frequency is one example of an attribute we need to set here to avoid heartbeat_frequency appearing to change from one value to another. There may be others.

ShaneHarvey: For example:
```python
>>> client = MongoClient(..., connect=False, heartbeatFrequencyMS=99999)
>>> client.topology_description.heartbeat_frequency
10
>>> client.admin.command("ping")
...
>>> client.topology_description.heartbeat_frequency
99.999

```

NoahStapp: Sorry, I misunderstood what you meant. The example makes it clear, thanks!

NoahStapp: Now we always initialize a `TopologySettings` in the constructor. If SRV is enabled, we create a new one after resolution but keep the existing topology id.

ShaneHarvey (approved):
Let's add a changelog entry and call it a day!

sleepyStick (dismissed):

doc/changelog.rst
```diff
@@ -10,6 +10,7 @@ Version 4.12.1 is a bug fix release.
 - Fixed a bug that could raise ``UnboundLocalError`` when creating asynchronous connections over SSL.
 - Fixed a bug causing SRV hostname validation to fail when resolver and resolved hostnames are identical with three domain levels.
 - Fixed a bug that caused direct use of ``pymongo.uri_parser`` to raise an ``AttributeError``.
+- Fixed a bug that could cause public ``pymongo.MongoClient`` and ``pymongo.AsyncMongoClient`` attributes to raise errors when accessed before a connection was established.
```
ShaneHarvey, line 13: Let's describe it a bit more specifically, something like: 

```
Fixed a bug where clients created with connect=False and a "mongodb+srv://" connection string
could cause public ``pymongo.MongoClient`` and ``pymongo.AsyncMongoClient`` attributes (topology_description,
nodes, address, primary, secondaries, arbiters) to incorrectly return a Database, leading to type
errors such as: "NotImplementedError: Database objects do not implement truth value testing or bool()".
```

doc/changelog.rst
```diff
@@ -10,10 +10,13 @@ Version 4.12.1 is a bug fix release.
 - Fixed a bug that could raise ``UnboundLocalError`` when creating asynchronous connections over SSL.
 - Fixed a bug causing SRV hostname validation to fail when resolver and resolved hostnames are identical with three domain levels.
 - Fixed a bug that caused direct use of ``pymongo.uri_parser`` to raise an ``AttributeError``.
+- Fixed a bug where clients created with connect=False and a "mongodb+srv://" connection string
+  could cause public ``pymongo.MongoClient`` and ``pymongo.AsyncMongoClient`` attributes (topology_description,
+  nodes, address, primary, secondaries, arbiters) to incorrectly return a Database, leading to type
+  errors such as: "NotImplementedError: Database objects do not implement truth value testing or bool().
```
ShaneHarvey, line 16: Missing trailing `"` for "NotImplementedError.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #2286, opened 2025-04-15
PYTHON-5309 Ensure AsyncMongoClient doesn't use PyOpenSSL

sleepyStick:
jira: https://jira.mongodb.org/browse/PYTHON-5309
problem: if a user had pyopenssl installed on their machine, the async client wouldnt configure ssl properly
brief description of my changes:
- ssl_context will try to load both stdlib ssl and pyopenssl 
- the driver will now dynamically determine which ssl to use (if the user has both) depending on async vs sync context
- to get this to work, `AutoEncryptionOpts` delays the resolution of `_kms_ssl_contexts`


CONTRIBUTING.md
```diff
@@ -420,3 +420,21 @@ partially-converted asynchronous version of the same name to the `test/asynchron
 Use this generated file as a starting point for the completed conversion.
 
 The script is used like so: `python tools/convert_test_to_async.py [test_file.py]`
+
+## Running PyMongo with SSL
```
ShaneHarvey, line 424: What's the motivation for adding this section?

pymongo/encryption_options.py
```diff
@@ -58,6 +58,7 @@ def __init__(
         bypass_query_analysis: bool = False,
         encrypted_fields_map: Optional[Mapping[str, Any]] = None,
         key_expiration_ms: Optional[int] = None,
+        is_sync: bool = True,
```
ShaneHarvey, line 61: This is a public api so we can't add "is_sync" here. What's the motivation for this?

ShaneHarvey (changes requested):

.evergreen/generated_configs/variants.yml
```diff
@@ -740,8 +740,8 @@ buildvariants:
       PYTHON_BINARY: /Library/Frameworks/Python.Framework/Versions/3.9/bin/python3
   - name: pyopenssl-rhel8-python3.10
     tasks:
-      - name: .replica_set .auth .ssl .sync
-      - name: .7.0 .auth .ssl .sync
+      - name: .replica_set .auth .ssl .sync_async
+      - name: .7.0 .auth .ssl .sync_async
```
ShaneHarvey, line 744: If we're now running all the async tests when pyopenssl is installed, then I think we can remove the new test_pyopenssl_ignored_in_async test added in test_ssl.

ShaneHarvey (changes requested):

CONTRIBUTING.md
sleepyStick: Oh uh, I was struggling to get it working and asked Noah a bunch of questions, so he suggested I add a section here.

pymongo/encryption_options.py
sleepyStick: Oh, I added `is_sync` as a param to a few functions because in `ssl_support.get_ssl_context()`, we'll need to know which version of ssl should be used,, is there another way to go about this that is preferred?

.evergreen/generated_configs/variants.yml
sleepyStick: hmm, i changed it so that *some* of the pyopenssl variants run async, not all. Should I make them all run async tests and remove that test in test_ssl?

pymongo/encryption_options.py
ShaneHarvey: Thanks, we still can not add "is_sync" here so we need to find another way.

ShaneHarvey: One way would be to lazily init the async SSLContext in kms_request().

sleepyStick: I think I did it? Not sure if this is what you meant, but what I have now would leave the params to AutoEncryptOpts unchanged. 
I basically delayed the parse_kms_tls_options because that's where is_sync was needed. I believe i appropriately delayed the definition of kms_tls_options but let me know if I missed something. I'm not super familiar with how encryption in the driver works >.<

NoahStapp:
Can you schedule a run of all the pyopenssl variants/tasks?

sleepyStick:
> Can you schedule a run of all the pyopenssl variants/tasks?

oops yeah, sorry i'll add it to the latest run with the updated evergreen patch changes from steve's PR

.evergreen/generated_configs/variants.yml
```diff
@@ -675,6 +680,7 @@ buildvariants:
       - rhel87-small
     batchtime: 10080
     expansions:
+      TEST_NAME: default
```
NoahStapp, line 683: Do we need to set `TEST_NAME=default`?

pymongo/network_layer.py
```diff
@@ -113,21 +118,23 @@ def _is_ready(fut: Future) -> None:
                 # Check for closed socket.
                 if fd == -1:
                     raise SSLError("Underlying socket has been closed") from None
-                if isinstance(exc, BLOCKING_IO_READ_ERROR):
+                if isinstance(exc, (BLOCKING_IO_READ_ERROR, PYBLOCKING_IO_READ_ERROR)):
```
NoahStapp, line 121: `PYBLOCKING_IO_<READ/WRITE/LOOKUP>_ERROR` should never be thrown in an async context.

pymongo/network_layer.py
```diff
@@ -46,22 +46,21 @@
     _HAVE_SSL = False
 
 try:
-    from pymongo.pyopenssl_context import (
-        BLOCKING_IO_LOOKUP_ERROR,
-        BLOCKING_IO_READ_ERROR,
-        BLOCKING_IO_WRITE_ERROR,
-        _sslConn,
-    )
+    from pymongo.pyopenssl_context import _sslConn as _pysslConn
```
NoahStapp, line 49: What's the motivation behind the renaming to `_pysslConn` here?

pymongo/ssl_support.py
```diff
@@ -57,6 +60,20 @@
     BLOCKING_IO_WRITE_ERROR = _ssl.BLOCKING_IO_WRITE_ERROR
     BLOCKING_IO_LOOKUP_ERROR = BLOCKING_IO_READ_ERROR
 
+    if HAVE_PYSSL:
```
NoahStapp, line 63: Can we use `Union` types here to remove the need for two separate exports of everything? Something like:

`BLOCKING_IO_ERRORS = _ssl.BLOCKING_IO_ERRORS | _pyssl.BLOCKING_IO_ERRORS`

Are there any situations where we specifically care if an exception type is from PyOpenSSL or stdlib SSL?

test/asynchronous/test_encryption.py
```diff
@@ -119,7 +119,9 @@ async def test_crypt_shared(self):
         # Test that we can pick up crypt_shared lib automatically
         self.simple_client(
             auto_encryption_opts=AutoEncryptionOpts(
-                KMS_PROVIDERS, "keyvault.datakeys", crypt_shared_lib_required=True
+                KMS_PROVIDERS,
```
NoahStapp, line 122: Are these whitespace changes intended?

.evergreen/generated_configs/variants.yml
sleepyStick: For now, I think we do? @blink1073 correct me if i'm wrong but this will be changed with the future test infrastructure changes?

blink1073: Yes, it will be cleaned up by [PYTHON-5333](https://jira.mongodb.org/browse/PYTHON-5333).

pymongo/network_layer.py
sleepyStick: When i was making these changes, I started interpreting py* to mean it would refer to pyopenssl (if it existed, and stdlib ssl if it did not) whereas ssl related vars without py in the front meant it was strictly referring to stdlib ssl. I'm not strongly attached to the renaming and wouldn't mind undoing it, but it helped my brain read the code.

test/asynchronous/test_encryption.py
sleepyStick: no, oops sorry

pymongo/ssl_support.py
sleepyStick: ooo good call out. I don't think we particularly care if its pyopenssl vs stdlib ssl error.

sleepyStick: I think python3.9 didn't love the union types? so i just made them tuples. But its a similar idea. 
I couldn't apply this to SSLError though because sometimes we'd raise SSLError so it needed to be one specific type.

pymongo/ssl_support.py
```diff
@@ -49,12 +52,24 @@
     import ssl as _stdlibssl  # noqa: F401
     from ssl import CERT_NONE, CERT_REQUIRED
 
-    HAS_SNI = _ssl.HAS_SNI
     IPADDR_SAFE = True
+
+    if HAVE_PYSSL:
+        HAS_SNI = _pyssl.HAS_SNI | _ssl.HAS_SNI
```
ShaneHarvey, line 58: This logic for HAS_SNI is problematic since this happens at import time but the actual ssl impl at runtime (stdlib vs pyopenssl) may or may not support SNI. I believe we need something like:
```python
        # We have to pass hostname / ip address to wrap_socket
        # to use SSLContext.check_hostname.
        if ssl_context.has_sni:
            ...
```

test/asynchronous/test_encryption.py
```diff
@@ -147,17 +147,25 @@ def test_init(self):
     def test_init_spawn_args(self):
         # User can override idleShutdownTimeoutSecs
         opts = AutoEncryptionOpts(
-            {}, "keyvault.datakeys", mongocryptd_spawn_args=["--idleShutdownTimeoutSecs=88"]
+            {},
+            "keyvault.datakeys",
+            mongocryptd_spawn_args=["--idleShutdownTimeoutSecs=88"],
```
ShaneHarvey, line 152: Could we revert these format-only changes?

pymongo/encryption_options.py
```diff
@@ -236,10 +237,14 @@ def __init__(
...
             self._mongocryptd_spawn_args.append("--idleShutdownTimeoutSecs=60")
         # Maps KMS provider name to a SSLContext.
-        self._kms_ssl_contexts = _parse_kms_tls_options(kms_tls_options)
+        self._kms_tls_options = kms_tls_options
+        self._kms_ssl_contexts: dict[str, SSLContext] = {}
         self._bypass_query_analysis = bypass_query_analysis
         self._key_expiration_ms = key_expiration_ms
 
+    def _parse_kms_tls_options(self, is_sync: bool) -> None:
+        self._kms_ssl_contexts = _parse_kms_tls_options(self._kms_tls_options, is_sync)
```
ShaneHarvey, line 246: This is problematic because a single AutoEncryptionOpts can be shared between multiple clients (even sync + async) at the same time. Currently the `_kms_ssl_contexts` will clash.

Another problem is that _parse_kms_tls_options is expensive and should only be called at most once for sync and async.

What if we do this at construction time?:
```
        # Maps KMS provider name to a SSLContext.
        self._kms_ssl_contexts = _parse_kms_tls_options(kms_tls_options, True)
        if pyopenssl_enabled:
            self._async_kms_ssl_contexts = _parse_kms_tls_options(kms_tls_options, False)
        else:
            self._async_kms_ssl_contexts = self._kms_ssl_contexts
```


Hmm actually, I thought of a better idea. _EncryptionIO can parse and store the contexts:
```python

class _EncryptionIO(AsyncMongoCryptCallback):
    def __init__(
        self,
        client: Optional[AsyncMongoClient[_DocumentTypeArg]],
        key_vault_coll: AsyncCollection[_DocumentTypeArg],
        mongocryptd_client: Optional[AsyncMongoClient[_DocumentTypeArg]],
        opts: AutoEncryptionOpts,
    ):
...
        self.opts = opts
        self._kms_ssl_contexts = _parse_kms_tls_options(opts._kms_tls_options, IS_SYNC)
...
```

pymongo/ssl_support.py
```diff
@@ -65,10 +80,15 @@ def get_ssl_context(
...
         allow_invalid_hostnames: bool,
         disable_ocsp_endpoint_check: bool,
-    ) -> _ssl.SSLContext:
+        is_sync: bool,
+    ) -> Union[_pyssl.SSLContext, _ssl.SSLContext]:  # type: ignore[name-defined]
         """Create and return an SSLContext object."""
+        if is_sync and HAVE_PYSSL:
+            ssl_in_use: types.ModuleType = _pyssl
+        else:
+            ssl_in_use = _ssl
```
ShaneHarvey, line 89: Can we rename `ssl_in_use` -> `_ssl` that way there's less code churn.

ShaneHarvey (changes requested):

pymongo/network_layer.py
ShaneHarvey: I'd prefer the old name (_sslConn). A larger refactor could be a good idea but I'd prefer this PR be as minimal as possible to avoid regressions because it's going into a bugfix release.

pymongo/ssl_support.py
sleepyStick: I don't think so? because _ssl is referring to stdlib ssl and i don't think we'd want it to be replaced by pyopenssl just because pyopenssl is installed?

pymongo/ssl_support.py
sleepyStick: Hmm I don't think I understand the problem? HAS_SNI was previously defined in the same spot. Did my changes introduce this problem? Or did the previous code also have the same problem?

pymongo/ssl_support.py
ShaneHarvey: It's a local variable assignment so what's the issue?
```python
        if is_sync and HAVE_PYSSL:
            _ssl: types.ModuleType = _pyssl
        else:
            _ssl = _ssl
```

pymongo/ssl_support.py
ShaneHarvey: The new logic is incorrect. If pyopenssl is install and `_pyssl.HAS_SNI=False` and `_ssl.HAS_SNI=True`, the sync code will incorrectly think that pyopenssl supports SNI.

doc/changelog.rst
```diff
@@ -12,6 +12,7 @@ Version 4.12.1 is a bug fix release.
 - Fixed a bug that caused direct use of ``pymongo.uri_parser`` to raise an ``AttributeError``.
 - Removed Eventlet testing against Python versions newer than 3.9 since
   Eventlet is actively being sunset by its maintainers and has compatibility issues with PyMongo's dnspython dependency.
+- Fixed a bug that would try to use PyOpenSSL with AsyncMongoClient, causing the client to fail.
```
NoahStapp, line 15: Make this more specific to the exact error reported, something like:

`Fixed a bug that would cause AsyncMongoClient to attempt to use PyOpenSSL when available, resulting in errors such as "pymongo.errors.ServerSelectionTimeoutError: 'SSLContext' object has no attribute 'wrap_bio'"`

pymongo/pool_shared.py
```diff
@@ -138,7 +138,7 @@ def _raise_connection_failure(
         msg += format_timeout_details(timeout_details)
     if isinstance(error, socket.timeout):
         raise NetworkTimeout(msg) from error
-    elif isinstance(error, SSLError) and "timed out" in str(error):
+    elif isinstance(error, (SSLError, PYSSLError)) and "timed out" in str(error):
```
NoahStapp, line 141: Do these need to be separate `SSLError` and `PYSSLError` types, or can we export a shared one from `ssl_support` to reduce code churn?

pymongo/ssl_support.py
sleepyStick: Ah, no you're right -- I've been conditioned to think that shadowing a global var with a local var is generally bad for code readability(?) so my brain forgot that it was something we could do.

pymongo/pool_shared.py
NoahStapp: The combined type could be here in `pool_shared.py`, as that's where it would get used.

NoahStapp: We only explicitly raise `SSLError` in `network_layer` for our socket-based Windows SSL I/O. Those will go away if we do [PYTHON-5215](https://jira.mongodb.org/browse/PYTHON-5215).

sleepyStick: Ahh i see, makes sense!

pymongo/ssl_support.py
sleepyStick: Er, am i doing it incorrectly? I'm seeing the error "UnboundLocalError: local variable '_ssl' referenced before assignment"

pymongo/ssl_support.py
sleepyStick: Ah, okay i understand how the logic is incorrect now (I previously thought that the two would be the same value regardless of pyopenssl vs stdlibssl) 
would you mind elaborating on your suggested fix though? I'm not sure I understand how that would resolve pyopenssl's has_sni and stdlib ssl's has_sni being two different values and how to navigate having those two potentially different values in the rest of the code?

pymongo/ssl_support.py
ShaneHarvey: Uh, I swore I've done this before. I guess not. The only way around that is like this:
```python
        if is_sync and HAVE_PYSSL:
            _ssl: types.ModuleType = _pyssl
        else:
            _ssl = globals()["_ssl"]
```

Or by renaming the global "_ssl".

sleepyStick: I actually kinda wanted to rename the global `_ssl` except i think there were some other modules that imported `_ssl` from `ssl_support` and then i wasn't sure if i was allowed to change it...

ShaneHarvey: You know what, I think it's simpler to go back to `ssl_in_use` or just `ssl`?

sleepyStick: All good, went with `ssl` because its shorter.

blink1073:
drivers-pr-bot please backport to v4.12

pymongo/ssl_support.py
sleepyStick: Not sure if I did it the way you envisioned it, but I created a helper function, `has_sni(is_sync) -> bool`, and call that everywhere `HAS_SNI` was being used.

ShaneHarvey (approved):
LGTM! Could you schedule the all the mac, windows, PyOpenSSL, OCSP, and encryption tasks to check for regressions before merging?

sleepyStick:
Opened a ticket for the encryption test failures as they seem to be unrelated to the changes I've made
https://jira.mongodb.org/projects/PYTHON/issues/PYTHON-5350?filter=allissues

doc/changelog.rst
```diff
@@ -18,6 +18,8 @@ Version 4.12.1 is a bug fix release.
   Eventlet is actively being sunset by its maintainers and has compatibility issues with PyMongo's dnspython dependency.
 - Fixed a bug where MongoDB cluster topology changes could cause asynchronous operations to take much longer to complete
   due to holding the Topology lock while closing stale connections.
+- Fixed a bug that would cause AsyncMongoClient to attempt to use PyOpenSSL when available, resulting in errors such as
+  "pymongo.errors.ServerSelectionTimeoutError: 'SSLContext' object has no attribute 'wrap_bio'"
```
NoahStapp, line 22:
```suggestion
  "pymongo.errors.ServerSelectionTimeoutError: 'SSLContext' object has no attribute 'wrap_bio'".
```

pymongo/asynchronous/pool.py
```diff
@@ -638,7 +638,7 @@ async def _raise_connection_failure(self, error: BaseException) -> NoReturn:
             reason = ConnectionClosedReason.ERROR
         await self.close_conn(reason)
         # SSLError from PyOpenSSL inherits directly from Exception.
-        if isinstance(error, (IOError, OSError, SSLError)):
+        if isinstance(error, (IOError, OSError, SSLError, PYSSLError)):
```
NoahStapp, line 641:
```suggestion
        if isinstance(error, (IOError, OSError, *SSLErrors)):
```

pymongo/asynchronous/pool.py
```diff
@@ -1052,7 +1052,7 @@ async def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> A
                     reason=_verbose_connection_error_reason(ConnectionClosedReason.ERROR),
                     error=ConnectionClosedReason.ERROR,
                 )
-            if isinstance(error, (IOError, OSError, SSLError)):
+            if isinstance(error, (IOError, OSError, SSLError, PYSSLError)):
```
NoahStapp, line 1055:
```suggestion
            if isinstance(error, (IOError, OSError, *SSLErrors)):
```

pymongo/asynchronous/pool.py
NoahStapp: We can use `SSLErrors` from `pool_shared.py` here.

pymongo/network_layer.py
```diff
@@ -71,7 +67,11 @@
 _UNPACK_COMPRESSION_HEADER = struct.Struct("<iiB").unpack
 _POLL_TIMEOUT = 0.5
 # Errors raised by sockets (and TLS sockets) when in non-blocking mode.
-BLOCKING_IO_ERRORS = (BlockingIOError, BLOCKING_IO_LOOKUP_ERROR, *ssl_support.BLOCKING_IO_ERRORS)
+BLOCKING_IO_ERRORS = (
```
NoahStapp, line 70: Intentional formatting change?

pymongo/ssl_support.py
NoahStapp: Can you expand on why the union types didn't work? We use `Union` elsewhere in our type hints.

NoahStapp (changes requested):

pymongo/ssl_support.py
sleepyStick: hmm okay not sure what i did incorrectly last time, but it works now! sorry about that >.<
(I'm going to guess i handled `BLOCKING_IO_ERRORS` incorrectly cuz i believe that's already a tuple of types and the union of two different types of tuples was not good?)

pymongo/network_layer.py
sleepyStick: nope, sorry it should be fixed now!

pymongo/ssl_support.py
blink1073: My guess is that you tried to use the pipe `|` operator

sleepyStick: That would be correct HAHA I just did it again:
```pymongo/ssl_support.py:60: error: Unsupported left operand type for |
Installing missing stub packages:
("type[WantReadError]")  [operator]
            BLOCKING_IO_READ_ERROR: Any = _pyssl.BLOCKING_IO_READ_ERROR | ...
                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...
pymongo/ssl_support.py:61: error: Unsupported left operand type for |
("type[WantWriteError]")  [operator]
            BLOCKING_IO_WRITE_ERROR: Any = _pyssl.BLOCKING_IO_WRITE_ERROR ...
                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...
```

NoahStapp: That error is from trying to use the pipe `|` operator, right? What happens when you use `Union[x, y]` instead?

sleepyStick: Ah, it works! thanks!

pymongo/ssl_support.py
```diff
@@ -56,20 +56,18 @@
...
-        BLOCKING_IO_READ_ERROR: tuple = (
-            _pyssl.BLOCKING_IO_READ_ERROR,
-            _ssl.BLOCKING_IO_READ_ERROR,
-        )
-        BLOCKING_IO_WRITE_ERROR: tuple = (
-            _pyssl.BLOCKING_IO_WRITE_ERROR,
-            _ssl.BLOCKING_IO_WRITE_ERROR,
-        )
+        BLOCKING_IO_ERRORS: tuple = _ssl.BLOCKING_IO_ERRORS + _pyssl.BLOCKING_IO_ERRORS
+        BLOCKING_IO_READ_ERROR: Any = Union[
```
ShaneHarvey, line 60: We cannot use Union at runtime. These should also be tuples.

pymongo/ssl_support.py
sleepyStick: yeah, it didn't immediately error (as a runtime object) so i thought it was just something I didn't know about how python works HAHA but okay, changing it back to a tuple now xD

pymongo/ssl_support.py
sleepyStick: Yeah (see convo chain above HAHA) they're back to tuples now :)

NoahStapp (dismissed):
🥳

CONTRIBUTING.md
ShaneHarvey: I'm concerned about people seeing this section and misinterpreting it to mean that they should be using tlsAllowInvalidCertificates. Can we remove it or make it very clear that this is only for local testing (using our test certificates) and not production?

ShaneHarvey (changes requested):

sleepyStick:
That's fair, I've removed it and added it to our onboarding google doc for now. :)

NoahStapp:
> That's fair, I've removed it and added it to our onboarding google doc for now. :)

I would rather keep the section in contributing.md to help any external contributors. Having a warning that this is purely for local testing and should never be done in production seems sufficient to me.

ShaneHarvey:
> I would rather keep the section in contributing.md to help any external contributors. Having a warning that this is purely for local testing and should never be done in production seems sufficient to me.

That's fine but it's unrelated to this PR so should be done in a different ticket.

ShaneHarvey:
@sleepyStick could you also update the PR title to describe the bug fix?

test/test_encryption.py
```diff
@@ -167,36 +168,48 @@ def test_init_spawn_args(self):
     @unittest.skipUnless(_HAVE_PYMONGOCRYPT, "pymongocrypt is not installed")
     def test_init_kms_tls_options(self):
         # Error cases:
+        opts = AutoEncryptionOpts({}, "k.d", kms_tls_options={"kmip": 1})
         with self.assertRaisesRegex(TypeError, r'kms_tls_options\["kmip"\] must be a dict'):
-            AutoEncryptionOpts({}, "k.d", kms_tls_options={"kmip": 1})
+            client = self.rs_or_single_client(auto_encryption_opts=opts)
+            client.db.coll.insert_one({"encrypted": "test"})
```
ShaneHarvey, line 174: These tests should raise from the AsyncMongoClient constructor, without even attempting to connect. We could even use AsyncMongoClient() directly:
```python
        with self.assertRaisesRegex(TypeError, r'kms_tls_options\["kmip"\] must be a dict'):
            AsyncMongoClient(auto_encryption_opts=opts)
```

Could you confirm?

sleepyStick: Ahhh yeah, i see now. Thanks for popping over to further explain it, and for working thru the soln with me!

pymongo/asynchronous/encryption.py
```diff
@@ -398,6 +398,7 @@ def __init__(self, client: AsyncMongoClient[_DocumentTypeArg], opts: AutoEncrypt
             encrypted_fields_map = _dict_to_bson(opts._encrypted_fields_map, False, _DATA_KEY_OPTS)
         self._bypass_auto_encryption = opts._bypass_auto_encryption
         self._internal_client = None
+        opts._kms_ssl_contexts(_IS_SYNC)
```
ShaneHarvey, line 401: It's worth adding a comment here to explain that we call this here so that parsing errors can be raised before creating internal clients.

ShaneHarvey (approved):

pymongo/encryption_options.py
```diff
@@ -236,10 +238,22 @@ def __init__(
         if not any("idleShutdownTimeoutSecs" in s for s in self._mongocryptd_spawn_args):
             self._mongocryptd_spawn_args.append("--idleShutdownTimeoutSecs=60")
         # Maps KMS provider name to a SSLContext.
-        self._kms_ssl_contexts = _parse_kms_tls_options(kms_tls_options)
+        self._kms_tls_options = kms_tls_options
+        self._sync_kms_ssl_contexts: dict[str, SSLContext] = None  # type:ignore[assignment]
+        self._async_kms_ssl_contexts: dict[str, SSLContext] = None  # type:ignore[assignment]
```
ShaneHarvey, line 243: Please remove `type:ignore[assignment]`. Those are bug factories and should only be used in exceptional circumstances.

ShaneHarvey (changes requested):

pymongo/asynchronous/encryption.py
sleepyStick: makes sense, added :)

pymongo/encryption_options.py
sleepyStick: and fixed

ShaneHarvey (approved):

PR #2296, opened 2025-04-17
PYTHON-5292 - Debug logs should only print on failed tests

NoahStapp:
Note: this PR does not turn on debug logs by default in our Evergreen tests. That will be tracked in a separate, follow-up ticket.


pymongo/logger.py
```diff
@@ -15,6 +15,7 @@
 
 import enum
 import logging
+import logging.handlers
```
blink1073, line 18: Is this import used?

NoahStapp: No, good catch!

blink1073 (approved):
LGTM!

PR #2301, opened 2025-04-21
PYTHON-5212 - Do not hold Topology lock while resetting pool

NoahStapp:


pymongo/asynchronous/pool.py
```diff
@@ -860,8 +861,13 @@ async def _reset(
         # PoolClosedEvent but that reset() SHOULD close sockets *after*
         # publishing the PoolClearedEvent.
         if close:
-            for conn in sockets:
-                await conn.close_conn(ConnectionClosedReason.POOL_CLOSED)
+            if not _IS_SYNC:
+                await asyncio.gather(
```
ShaneHarvey, line 865: We probably want to use return_exceptions=True here to ensure all tasks complete.

pymongo/asynchronous/pool.py
```diff
@@ -938,8 +949,13 @@ async def remove_stale_sockets(self, reference_generation: int) -> None:
                     and self.conns[-1].idle_time_seconds() > self.opts.max_idle_time_seconds
                 ):
                     close_conns.append(self.conns.pop())
-            for conn in close_conns:
-                await conn.close_conn(ConnectionClosedReason.IDLE)
+            if not _IS_SYNC:
+                await asyncio.gather(
```
ShaneHarvey, line 953: same here.

pymongo/asynchronous/topology.py
```diff
@@ -557,6 +551,11 @@ async def on_change(
             # that didn't include this server.
             if self._opened and self._description.has_server(server_description.address):
                 await self._process_change(server_description, reset_pool, interrupt_connections)
+        # Clear the pool from a failed heartbeat, done outside the lock to avoid blocking on connection close.
+        if self._opened and self._description.has_server(server_description.address) and reset_pool:
+            server = self._servers.get(server_description.address)
```
ShaneHarvey, line 556: The has_server -> _servers.get pattern is not safe to do here (https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use)

We also don't need to check for closed/opened because pool.reset is safe to call even after close().

Instead we can do:
```python
if reset_pool:
    server = self._servers.get(server_description.address)
    if server:
        await server.pool.reset(interrupt_connections=interrupt_connections)
```

ShaneHarvey (changes requested):
Is there any regression test we can add for this?

test/asynchronous/test_discovery_and_monitoring.py
```diff
@@ -370,6 +375,69 @@ async def test_pool_unpause(self):
...
+                    "appName": "SDAMConnectionCloseTest",
+                },
+            }
+            async with self.fail_point(fail_hello):
+                # Wait for server heartbeat to fail
+                await listener.async_wait_for_event(monitoring.ServerHeartbeatFailedEvent, 1)
+            # Wait until all idle connections are closed to simulate real-world conditions
+            await listener.async_wait_for_event(monitoring.ConnectionClosedEvent, 10)
+            # No operation latency should not significantly exceed close_delay
+            self.assertLessEqual(max(latencies), close_delay * 2.0)
```
ShaneHarvey, line 436: I worry this test will be flaky. A single op can easily take >100 ms in our CI depending on the host (mac/windows). Could you increase the close_delay to 0.1 and increase this line to `close_delay * 5`? Since there are 10 connections to close, `* 5` should still catch the regression right?

test/asynchronous/test_discovery_and_monitoring.py
```diff
@@ -370,6 +375,69 @@ async def test_pool_unpause(self):
...
+    async def test_connection_close_does_not_block_other_operations(self):
+        listener = CMAPHeartbeatListener()
+        client = await self.async_single_client(
+            appName="SDAMConnectionCloseTest",
+            event_listeners=[listener],
+            heartbeatFrequencyMS=500,
+            minPoolSize=10,
+        )
+        server = await (await client._get_topology()).select_server(
+            readable_server_selector, _Op.TEST
```
ShaneHarvey, line 390: readable_server_selector -> writeable_server_selector. The test is using primary read preference so we should wait for 10 connections to the primary node.

test/asynchronous/test_discovery_and_monitoring.py
NoahStapp: I would expect `close_delay * 5` to catch regressions consistently, yeah.

test/asynchronous/test_discovery_and_monitoring.py
```diff
@@ -370,6 +375,69 @@ async def test_pool_unpause(self):
...
+        latencies = []
+
+        async def run_task():
+            while True:
+                start_time = time.monotonic()
+                await client.db.test.find_one({})
+                elapsed = time.monotonic() - start_time
+                latencies.append(elapsed)
+                if elapsed >= close_delay:
+                    break
```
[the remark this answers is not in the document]
NoahStapp: The classic list-as-mutable-state technique!

test/test_client.py
```diff
@@ -1864,6 +1864,7 @@ def test_direct_connection(self):
             MongoClient(["host1", "host2"], directConnection=True)
 
     @unittest.skipIf("PyPy" in sys.version, "PYTHON-2927 fails often on PyPy")
+    @skipIf(os.environ.get("DEBUG_LOG"), "Enabling debug logs breaks this test")
```
ShaneHarvey, line 1867: Was this intentionally added?

test/asynchronous/test_discovery_and_monitoring.py
```diff
@@ -370,6 +375,71 @@ async def test_pool_unpause(self):
...
+                    "errorCode": 91,
+                    "appName": "SDAMConnectionCloseTest",
+                },
+            }
+            async with self.fail_point(fail_hello):
+                # Wait for server heartbeat to fail
+                await listener.async_wait_for_event(monitoring.ServerHeartbeatFailedEvent, 1)
+            # Wait until all idle connections are closed to simulate real-world conditions
+            await listener.async_wait_for_event(monitoring.ConnectionClosedEvent, 10)
+            should_exit.append(True)
```
ShaneHarvey, line 436: The test should join the task here after `should_exit.append(True)` otherwise we may miss a latency. Also one more thing:
```python
            # Wait until all idle connections are closed to simulate real-world conditions
            await listener.async_wait_for_event(monitoring.ConnectionClosedEvent, 10)
            # Wait for one more find to complete, then shutdown the task.
            n = len(latencies)
            await async_wait_until(lambda: len(latencies) >= n + 1, "run one more find")
            should_exit.append(True)
            await task.join()
            
```

test/test_client.py
NoahStapp: Whoops this was for testing purposes. Intended to be done in a separate ticket.

ShaneHarvey (approved):

PR #2325, opened 2025-04-30
PYTHON-5356 - Init unified test client SDAM for all unified tests

NoahStapp:
Explicitly connect clients for unified CSOT tests before the test begins to minimize flakiness caused by SDAM creation and startup.

Patch build with everything run: https://spruce.mongodb.com/version/6812632b5456710007edc380/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/test/asynchronous/unified_format.py
+++ b/test/asynchronous/unified_format.py
@@ -257,7 +257,7 @@ class EntityMapUtil:
                 current[key] = self._handle_placeholders(spec, value, subpath)
         return current
 
-    async def _create_entity(self, entity_spec, uri=None, init_client=False):
+    async def _create_entity(self, entity_spec, uri=None):
         if len(entity_spec) != 1:
             self.test.fail(f"Entity spec {entity_spec} did not contain exactly one top-level key")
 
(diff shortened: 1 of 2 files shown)


test/asynchronous/unified_format.py
```diff
@@ -303,6 +303,8 @@ async def _create_entity(self, entity_spec, uri=None):
             if uri:
                 kwargs["h"] = uri
             client = await self.test.async_rs_or_single_client(**kwargs)
+            if init_client:
+                await client.aconnect()
```
ShaneHarvey, line 307: We can remove init_client and always call aconnect() here, since aconnect() doesn't run any commands it's always safe to start SDAM tasks. If we added an explicit `client.admin.command()` then we would need the init_client flag.

ShaneHarvey (changes requested):

test/asynchronous/unified_format.py
NoahStapp: Do we actually want to call `client.admin.command()` instead? That would ensure we actually connect with a non-SDAM connection before the test.

ShaneHarvey: I suspect that will cause problems with event monitoring. We plan to add "ping" explicitly to the unified tests themselves to fix that.

ShaneHarvey (approved):

PR #2326, opened 2025-05-01
PYTHON-5369 - Re-raise socket.timeout errors if the deadline has alre…

NoahStapp:
…ady been execeeded


ShaneHarvey:
On second thought, I intentionally added the non-blocking read to avoid spurious reconnects on SDAM connections when using the driver in FaaS services that pause/resume the process, see https://jira.mongodb.org/browse/PYTHON-3186.

I'm not sure we can accept this change.

sleepyStick:
> On second thought, I intentionally added the non-blocking read to avoid spurious reconnects on SDAM connections when using the driver in FaaS services that pause/resume the process, see [jira.mongodb.org/browse/PYTHON-3186](https://jira.mongodb.org/browse/PYTHON-3186).
> 
> I'm not sure we can accept this change. Although, if the test_sigstop_sigcont regression test still passes, perhaps we can?

Looks like that regression test isn't passing...
https://spruce.mongodb.com/task/mongo_python_driver_test_macos_arm64_test_standard_v6.0_python3.13_sync_auth_ssl_sharded_cluster_patch_0ec57781d14160531887f61e4edf083acbef7686_6813d280e5d2f8000770ece5_25_05_01_19_58_57?execution=0&sortBy=STATUS&sortDir=ASC

blink1073:
Can we not use the FaaS detection logic and preserve the current behavior in that case?

NoahStapp:
Since `test_sigstop_sigcont` was added specifically for the FaaS fix, can we limit it to only run on FaaS test variants?

blink1073:
> Since test_sigstop_sigcont was added specifically for the FaaS fix, can we limit it to only run on FaaS test variants?

My vote is yes

sleepyStick (dismissed):

test/test_client.py
```diff
@@ -1967,6 +1967,7 @@ def test_srv_max_hosts_kwarg(self):
         "loadBalanced clients do not run SDAM",
     )
     @unittest.skipIf(sys.platform == "win32", "Windows does not support SIGSTOP")
+    @unittest.skipUnless(_is_faas(), "Non-FaaS environments raise timeouts faster")
```
ShaneHarvey, line 1970: We can't skip this test like this because we don't run the test suite on faas platforms. We want this test to run locally.

blink1073: Perhaps instead we can force the metadata to be faas through monkeypatching

NoahStapp:
> Did we determine that this change actually fixes the test flakes?

I'm running some patches right now to confirm, but so far it appears to fix the flakes on non-PyPy variants. PyPy still has this issue, so I propose we skip `test.test_transactions_unified.TestUnifiedErrorLabelsBlockConnection.test_add_RetryableWriteError_and_UnknownTransactionCommitResult_labels_to_connection_errors` on it.

NoahStapp:
> Did we determine that this change actually fixes the test flakes?

I have not seen any failures of these tests flaking on sync + CPython with this fix.

test/asynchronous/test_client.py
```diff
@@ -117,7 +117,7 @@
     WriteConcernError,
 )
 from pymongo.monitoring import ServerHeartbeatListener, ServerHeartbeatStartedEvent
-from pymongo.pool_options import _MAX_METADATA_SIZE, _METADATA, ENV_VAR_K8S, PoolOptions
+from pymongo.pool_options import _MAX_METADATA_SIZE, _METADATA, ENV_VAR_K8S, PoolOptions, _is_faas
```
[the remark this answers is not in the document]
blink1073: Bump

pymongo/network_layer.py
```diff
@@ -357,7 +358,12 @@ def receive_data(conn: Connection, length: int, deadline: Optional[float]) -> me
             except socket.timeout:
                 if conn.cancel_context.cancelled:
                     raise _OperationCancelled("operation cancelled") from None
-                if _PYPY:
+                if (
+                    _PYPY
+                    or not _is_faas()
+                    and deadline is not None
+                    and deadline - time.monotonic() < 0
+                ):
```
ShaneHarvey, line 366: I think we want sigstop/sigcont to work the same regardless of what env variables are defined. What if we take an alternative approach here and replace `not _is_faas()` with `not conn.is_sdam`? IE only do the extra non-blocking read on SDAM connections.

ShaneHarvey (changes requested):

blink1073 (approved):
LGTM!

pymongo/network_layer.py
ShaneHarvey: Are the flaky tests still fixed after the is_sdam change?

NoahStapp: They appear to be, yes: https://spruce.mongodb.com/version/681e08d75668e00007df8859/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=failed&taskName=%5E%28%28%3F%21pypy%29.%29*%24&variant=%5E%28%28%3F%21pypy%29.%29*%24. Filtering out PyPy failures, none of the flaky tests that should be fixed by this change are present.

ShaneHarvey (approved):

sleepyStick (approved):

PR #2329, opened 2025-05-02
PYTHON-5371 - Pass repr(ServerDescription) to logging

NoahStapp:


test/asynchronous/test_client.py
```diff
@@ -1906,6 +1906,7 @@ async def test_direct_connection(self):
         with self.assertRaises(ConfigurationError):
             AsyncMongoClient(["host1", "host2"], directConnection=True)
 
+    @unittest.skipIf(os.environ.get("DEBUG_LOG"), "Enabling debug logs breaks this test")
```
ShaneHarvey, line 1909: Why does this test break? I’m concerned about this pattern since it makes it harder to debug the test if it gets skipped. Could we leave a ticket open to come back to this test and make it work in debug mode?

NoahStapp: We were planning to replace this test with something less hacky in the future anyway. I can investigate deeper as to why this is breaking if we're planning to keep this test around instead.

NoahStapp:
Patch build with debug logs enabled without any failures of test_continuous_network_errors: https://spruce.mongodb.com/version/6814e3d321f66a0007561099/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=failed&taskName=%5E%28%28%3F%21pypy%29.%29%2A%24&variant=%5E%28%28%3F%21pypy%29.%29%2A%24

pymongo/asynchronous/topology.py
```diff
@@ -154,8 +154,8 @@ def __init__(self, topology_settings: TopologySettings):
                 _SDAM_LOGGER,
                 message=_SDAMStatusMessage.TOPOLOGY_CHANGE,
                 topologyId=self._topology_id,
-                previousDescription=initial_td,
-                newDescription=self._description,
+                previousDescription=initial_td.__repr__(),
+                newDescription=self._description.__repr__(),
```
ShaneHarvey, line 158: Could you update these to use `repr()` instead of `__repr__`?

ShaneHarvey (changes requested):

PR #2331, opened 2025-05-02
PYTHON-5339 Clean up GitHub PR definitions in Evergreen Project

blink1073:
Use the "pr" tag in the generated config as our only patch definition.  Produces 128 tasks.

Server version tests:

image

Standard tests:

image

Non standard tests:

image


blink1073:
evergreen retry

CONTRIBUTING.md
```diff
@@ -225,6 +225,7 @@ the pages will re-render and the browser will automatically refresh.
 - Run `just run-tests` to run the tests in an appropriate Python environment.
 - When done, run `just teardown-tests` to clean up and `just stop-server` to stop the server.
 
+foo
```
NoahStapp, line 228: Unintentionally included I assume.

NoahStapp (changes requested):

CONTRIBUTING.md
blink1073: Fixed

NoahStapp (commented):
Are these the tasks that will run on every PR?

blink1073:
Total time taken is already 20 hours 18 min, is there something specific you think we're lacking?

NoahStapp:
> Total time taken is already 20 hours 18 min, is there something specific you think we're lacking?

Having a task for each extra we have would make sense to me, to ensure we don't break anything unintentionally.

blink1073:
| Having a task for each extra we have would make sense to me, to ensure we don't break anything unintentionally.

We're already covering the ones that use the "non-standard" tasks the other ones don't do that because they rely on VMs or Atlas clusters.

blink1073:
I think mod_wsgi is the only outlier that would manually add the pr tag to, I'll do that.

blink1073:
That brings us to 128 tasks.

NoahStapp (approved):
Looks great!

PR #2332, opened 2025-05-02
PYTHON-5033 Use PyModule_Add on >= 3.13

aclark4life:
PYTHON-5033

### DIFF
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -3229,7 +3229,6 @@ _cbson_exec(PyObject *m)
 
 #if PY_VERSION_HEX >= 0x030D0000
     if (PyModule_Add(m, "_C_API", c_api_object) < 0) {
-        Py_DECREF(c_api_object);
         Py_DECREF(m);
         INITERROR;
     }
--- a/bson/_cbsonmodule.c
+++ b/bson/_cbsonmodule.c
@@ -3230,7 +3230,6 @@ _cbson_exec(PyObject *m)
 #if PY_VERSION_HEX >= 0x030D0000
     if (PyModule_Add(m, "_C_API", c_api_object) < 0) {
         Py_DECREF(c_api_object);
-        Py_DECREF(m);
         INITERROR;
     }
 # else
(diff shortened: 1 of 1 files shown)


bson/_cbsonmodule.c
```diff
@@ -3227,11 +3227,19 @@ _cbson_exec(PyObject *m)
         INITERROR;
     }
 
+#if PY_VERSION_HEX >= 0x030D0000
+    if (PyModule_Add(m, "_C_API", c_api_object) < 0) {
+        Py_DECREF(c_api_object);
```
ShaneHarvey, line 3232: We need to remove this decref, the new API always steals the reference: https://docs.python.org/3/c-api/module.html#c.PyModule_Add

ShaneHarvey: You removed the wrong line.

aclark4life: 😂

ShaneHarvey: Once again, you removed the wrong line.

aclark4life: 🤦 In 75233bb65a8db178931352f558b610fe9d6d1369 I only added two DECREFs … identical to the DECREFs in the `else` clause.  Are you saying we need to remove a DECREF from somewhere else, unrelated to this PR?

ShaneHarvey: We need to decref `m` here, not `c_api_object`.

aclark4life: OK but that was already done in 9c768f5f25c4dd365cf374cee670ebbb50a87bb5 and you said that I removed the wrong one … 🤔

ShaneHarvey: No it was not done in https://github.com/mongodb/mongo-python-driver/commit/9c768f5f25c4dd365cf374cee670ebbb50a87bb5 .

That commit removes code in a completely unrelated code block.

ShaneHarvey: Thanks. FYI you should only resolve your own comments, not other people's. That way the reviewers can double check their comment is resolved.

ShaneHarvey (approved):
LGTM once the tests pass

PR #2335, opened 2025-05-07
PYTHON-4542 Improved sessions API

aclark4life:
- Via context variable.

### DIFF
--- a/pymongo/asynchronous/client_session.py
+++ b/pymongo/asynchronous/client_session.py
@@ -139,6 +139,7 @@ import collections
 import time
 import uuid
 from collections.abc import Mapping as _Mapping
+from contextvars import ContextVar
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -204,6 +205,7 @@ class SessionOptions:
         causal_consistency: Optional[bool] = None,
         default_transaction_options: Optional[TransactionOptions] = None,
         snapshot: Optional[bool] = False,
+        bind: Optional[bool] = False,
     ) -> None:
         if snapshot:
             if causal_consistency:
@@ -222,6 +224,7 @@ class SessionOptions:
                 )
         self._default_transaction_options = default_transaction_options
         self._snapshot = snapshot
+        self._bind = bind
 
     @property
     def causal_consistency(self) -> bool:
@@ -1063,6 +1066,9 @@ class AsyncClientSession:
 
     def __copy__(self) -> NoReturn:
         raise TypeError("A AsyncClientSession cannot be copied, create a new session instead")
+
+
+SESSION: ContextVar[Optional[AsyncClientSession]] = ContextVar("SESSION", default=None)
 
 
 class _EmptyServerSession:
--- a/pymongo/asynchronous/client_session.py
+++ b/pymongo/asynchronous/client_session.py
@@ -517,6 +517,7 @@ class AsyncClientSession:
         # Is this an implicitly created session?
         self._implicit = implicit
         self._transaction = _Transaction(None, client)
+        self._token = None
 
     async def end_session(self) -> None:
         """Finish this session. If a transaction has started, abort it.
@@ -555,6 +556,7 @@ class AsyncClientSession:
     async def __aexit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
         if self._token:
             _SESSION.reset(self._token)
+            self._token = None
         await self._end_session(lock=True)
 
     @property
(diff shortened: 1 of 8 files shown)


pymongo/asynchronous/client_session.py
```diff
@@ -1065,6 +1068,9 @@ def __copy__(self) -> NoReturn:
         raise TypeError("A AsyncClientSession cannot be copied, create a new session instead")
 
 
+SESSION: ContextVar[Optional[AsyncClientSession]] = ContextVar("SESSION", default=None)
```
ShaneHarvey, line 1071: SESSION -> `_SESSION`. We don't want to make this public.

pymongo/asynchronous/client_session.py
```diff
@@ -222,6 +224,7 @@ def __init__(
                 )
         self._default_transaction_options = default_transaction_options
         self._snapshot = snapshot
+        self._bind = bind
```
ShaneHarvey, line 227: We have to bind/unbind the session in `ClientSession.__enter__/__exit__`. That way the stack of sessions is managed correctly (ie we call `_SESSION.reset(token)`). Think about how nested cases will work:
```python
session1 = client.start_session(bind=True)
with session1:
    session2 = client.start_session(bind=True)
    with session2:
        coll.find_one() # uses session2
    coll.find_one() # uses session1
coll.find_one() # uses implicit session
```

ShaneHarvey (changes requested):

test/asynchronous/test_session.py
```diff
@@ -394,12 +394,22 @@ async def test_cursor_clone(self):
...
-            self.assertTrue(clone.session is s)
+
+        # Nested sessions.
+        session1 = self.client.start_session(bind=True)
+        with session1:
+            session2 = self.client.start_session(bind=True)
+            with session2:
+                coll.find_one()  # uses session2
+            coll.find_one()  # uses session1
+        coll.find_one()  # uses implicit session
```
ShaneHarvey, line 412: This test has to actually verify the correct sessions are used via command monitoring events.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1355,13 +1355,18 @@ def _close_cursor_soon(
     def _start_session(self, implicit: bool, **kwargs: Any) -> AsyncClientSession:
         server_session = _EmptyServerSession()
         opts = client_session.SessionOptions(**kwargs)
-        return client_session.AsyncClientSession(self, server_session, opts, implicit)
+        bind = opts._bind
+        session = client_session.AsyncClientSession(self, server_session, opts, implicit)
+        if bind:
+            _SESSION.set(session)
```
ShaneHarvey, line 1361: This needs to be removed. We should only bind in `__enter__`.

pymongo/asynchronous/client_session.py
```diff
@@ -545,9 +548,12 @@ def _check_ended(self) -> None:
             raise InvalidOperation("Cannot use ended session")
 
     async def __aenter__(self) -> AsyncClientSession:
+        self._token = _SESSION.set(self)
```
ShaneHarvey, line 551: We should only do this if the bind option is set to True.

ShaneHarvey (changes requested):

pymongo/asynchronous/mongo_client.py
aclark4life: Done in 3c68a70970e96917b0c418900203928bd7678a00

pymongo/asynchronous/client_session.py
aclark4life: How about checking the cursor session ? E.g. c36e9df0f67a84bcbe8e7eb4062f9d5b7fc10005

test/asynchronous/test_session.py
aclark4life: How about comparing with the cursor session ? E.g. c36e9df0f67a84bcbe8e7eb4062f9d5b7fc10005

pymongo/asynchronous/mongo_client.py
aclark4life: @ShaneHarvey OK to resolve this one?

test/asynchronous/test_session.py
aclark4life: @ShaneHarvey I'm assuming that commit is wrong, I should be using command monitoring via event listeners right?

pymongo/asynchronous/client_session.py
aclark4life: Done in 3c68a70970e96917b0c418900203928bd7678a00

pymongo/asynchronous/client_session.py
aclark4life: Done in 71eca084a19daf10b1c73ce639397f6f1b77020f

PR #2336, opened 2025-05-08
PYTHON-5342 Fix test_dns_failures test

blink1073:


NoahStapp (approved):

test/asynchronous/test_srv_polling.py
```diff
@@ -186,6 +184,9 @@ def dns_resolver_response():
             ):
                 await assertion_method(expected_response, client)
 
+            # Close the client early to avoid affecting the next scenario run.
+            await client.close()
```
ShaneHarvey, line 188: Why does leaving the client open have side effects?

ShaneHarvey (approved):

test/asynchronous/test_srv_polling.py
blink1073: We discussed in office hours: the thought is that the patching was affecting the previous client(s) that were not yet garbage collected.

PR #2337, opened 2025-05-09
Revert "PYTHON-5126 & PYTHON-5280 Addresses issues raised in DRIVERS-3097 and DRIVERS-3123 "

blink1073:
Reverts mongodb/mongo-python-driver#2261


caseyclements:
We cannot simply revert this PR as it fixed two issues. One was a new feature, about what to do with ignored bits. The changes to this are quite small. 

The other though, was to update our code to pass new standardized tests added  for failing cases, where what was stored was invalid. This is the changes in the test runner. 

How would you like to proceed?

caseyclements (approved):
Let's revert for now, and I'll reapply changes for DRIVERS-3097.

PR #2350, opened 2025-05-23
PYTHON-5392 Better test assertions for comparisons

sleepyStick:
Another subtask PR for [PYTHON-5262](https://jira.mongodb.org/browse/PYTHON-5262), this one is for comparisons:
`self.assertTrue(... > ...)` -> `self.assertGreater(..., ...)`
`self.assertTrue(... >= ...)` -> `self.assertGreaterEqual(..., ...)`
`self.assertTrue(... < ...)` -> `self.assertLess(..., ...)`
`self.assertTrue(... <= ...)` -> `self.assertLessEqual(..., ...)`
`self.assertFalse(... >= ...)` -> `self.assertLess(..., ...)`
`self.assertFalse(... > ...)` -> `self.assertLessEqual(..., ...)`
`self.assertFalse(... <= ...)` -> `self.assertGreater(..., ...)`
`self.assertFalse(... < ...)` -> `self.assertGreaterEqual(..., ...)`
`self.assertTrue(... == ...)` -> `self.assertEqual(..., ...)`
`self.assertTrue(... != ...)` -> `self.assertNotEqual(..., ...)`
`self.assertFalse(... == ...)` -> `self.assertNotEqual(..., ...)`


test/test_bson.py
```diff
@@ -1045,6 +1045,7 @@ def test_exception_wrapping(self):
 
     def test_minkey_maxkey_comparison(self):
         # MinKey's <, <=, >, >=, !=, and ==.
+        # These tests should be kept as assertTrue as opposed to using unittest's build in comparison assertions.
```
NoahStapp: Can you add the reason why?

test/test_bson.py
```diff
@@ -1045,6 +1045,8 @@ def test_exception_wrapping(self):
 
     def test_minkey_maxkey_comparison(self):
         # MinKey's <, <=, >, >=, !=, and ==.
+        # These tests should be kept as assertTrue as opposed to using unittest's build in comparison assertions because
```
NoahStapp:
```suggestion
        # These tests should be kept as assertTrue as opposed to using unittest's built-in comparison assertions because
```

test/test_bson.py
```diff
@@ -1045,6 +1045,8 @@ def test_exception_wrapping(self):
...
+        # MinKey and MaxKey define their own __ge__, __le__, etc to be different and we want to explicitly test that.
```
NoahStapp:
```suggestion
        # MinKey and MaxKey define their own __ge__, __le__, and other comparison attributes, and we want to explicitly test that.
```

PR #2358, opened 2025-06-02
PYTHON-5344 and PYTHON-5403 Allow Instantiated MongoClients to Send Client Metadata On-Demand

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-5344 
https://jira.mongodb.org/browse/PYTHON-5403 (ticket for async tests)

Notes:
- I used mockupDB to capture the metadata that the client sends in the handshake
- i put the tests in the normal tests folder (as opposed to the `test/mockupdb`) to utilize synchro for the async -> sync tests.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -7,6 +7,9 @@ PyMongo 4.14 brings a number of changes including:
 
 - Added :attr:`bson.codec_options.TypeRegistry.codecs` and :attr:`bson.codec_options.TypeRegistry.fallback_encoder` properties
   to allow users to directly access the type codecs and fallback encoder for a given :class:`bson.codec_options.TypeRegistry`.
+- Added :meth:`pymongo.asynchronous.mongo_client.AsyncMongoClient.append_metadata` and
+  :meth:`pymongo.mongo_client.MongoClient.append_metadata` to allow instantiated MongoClients to send client metadata
+  on-demand
 
 Changes in Version 4.13.2 (2025/06/17)
 --------------------------------------


pymongo/asynchronous/mongo_client.py
```diff
@@ -1040,6 +1041,20 @@ async def target() -> bool:
         self._kill_cursors_executor = executor
         self._opened = False
 
+    def _append_metadata(self, driver_info: DriverInfo) -> None:
```
blink1073: I think this needs to be public: "Specifies that drivers MUST provide an API to allow updating client metadata (name, version, platform) after initialization"

sleepyStick: oh good catch!

test/asynchronous/test_client_metadata.py
```diff
@@ -37,6 +40,17 @@
...
+# Location of JSON test specifications.
+if _IS_SYNC:
+    _TEST_PATH = os.path.join(pathlib.Path(__file__).resolve().parent, "handshake", "unified")
+else:
+    _TEST_PATH = os.path.join(
+        pathlib.Path(__file__).resolve().parent.parent, "handshake", "unified"
+    )
+
+# Generate unified tests.
+globals().update(generate_test_classes(_TEST_PATH, module=__name__, RUN_ON_SERVERLESS=True))
```
blink1073: We don't test on serverless anymore, so you can remove `RUN_ON_SERVERLESS`.

test/asynchronous/unified_format.py
```diff
@@ -703,6 +704,7 @@ async def _collectionOperation_createChangeStream(self, target, *args, **kwargs)
     async def _databaseOperation_runCommand(self, target, **kwargs):
         self.__raise_if_unsupported("runCommand", target, AsyncDatabase)
         # Ensure the first key is the command name.
+        print(kwargs)
```
blink1073: there's several print statements left in this file

test/asynchronous/test_client_metadata.py
```diff
@@ -0,0 +1,224 @@
...
+    async def check_metadata_added(
+        self,
+        client: AsyncMongoClient,
+        add_name: str,
+        add_version: Optional[str],
+        add_platform: Optional[str],
+    ) -> None:
+        # send initial metadata
+        name, version, platform, metadata = await self.send_ping_and_get_metadata(client, True)
+        await asyncio.sleep(0.005)
```
blink1073, line 103: Why this specific sleep?  Is this a potential source of flakiness?

blink1073 (changes requested):

test/asynchronous/unified_format.py
sleepyStick: Oops, that's embarrassing...fixed now

test/asynchronous/test_client_metadata.py
sleepyStick: oh, good to know, thanks!

test/asynchronous/test_client_metadata.py
sleepyStick: We sleep to wait for the connection to become idle -- I don't think this would be a source of flakeness given that the client's are configured to have a timeout of 1ms, and we wait 5ms, but maybe there's something else I'm not considering? what do you think?

blink1073: Can you add a comment to that affect in case we do see some flakiness later?

sleepyStick: Just added "wait for connection to become idle", is that enough? (genuinely not sure)

blink1073: That works for me, thanks!

blink1073 (approved):
LGTM!

pymongo/asynchronous/mongo_client.py
```diff
@@ -1040,6 +1041,23 @@ async def target() -> bool:
         self._kill_cursors_executor = executor
         self._opened = False
 
+    def append_metadata(self, driver_info: DriverInfo) -> None:
+        """
+        Appends the given metadata to existing driver metadata.
+        """
+        metadata = self._options.pool_options.metadata
+        for k, v in driver_info._asdict().items():
```
ShaneHarvey: This needs to only allow the "name", "version", "platform" fields otherwise the user can inject arbitrary fields into the driver metadata here.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1040,6 +1041,23 @@ async def target() -> bool:
...
+            if v is None:
+                continue
+            if k in metadata:
```
ShaneHarvey: What case is this if-statement handling?

pymongo/asynchronous/mongo_client.py
```diff
@@ -1040,6 +1041,23 @@ async def target() -> bool:
...
+                metadata[k] = f"{metadata[k]}|{v}"
+            elif k in metadata["driver"]:
+                metadata["driver"][k] = "{}|{}".format(
```
ShaneHarvey: Could we move this logic into PoolOptions and refactor it with the existing code?:
```
        if driver:
            if driver.name:
                self.__metadata["driver"]["name"] = "{}|{}".format(
                    self.__metadata["driver"]["name"],
                    driver.name,
                )
            if driver.version:
                self.__metadata["driver"]["version"] = "{}|{}".format(
                    _METADATA["driver"]["version"],
                    driver.version,
                )
            if driver.platform:
                self.__metadata["platform"] = "{}|{}".format(_METADATA["platform"], driver.platform)
```

pymongo/asynchronous/mongo_client.py
```diff
@@ -1040,6 +1041,23 @@ async def target() -> bool:
         self._kill_cursors_executor = executor
         self._opened = False
 
+    def append_metadata(self, driver_info: DriverInfo) -> None:
+        """
+        Appends the given metadata to existing driver metadata.
```
ShaneHarvey: Could you update this docstring to follow our standard format (params, versionadded, etc...)?

ShaneHarvey (changes requested):

test/asynchronous/test_client_metadata.py
```diff
@@ -0,0 +1,225 @@
...
+import time
+import unittest
+from test.asynchronous import AsyncIntegrationTest
+from test.asynchronous.unified_format import generate_test_classes
+from test.utils_shared import CMAPListener
+from typing import Any, Optional
+
+import pytest
+
+from pymongo import AsyncMongoClient, MongoClient
```
NoahStapp:
```suggestion
from pymongo import AsyncMongoClient
```

test/asynchronous/test_client_metadata.py
```diff
@@ -0,0 +1,225 @@
...
+    async def asyncSetUp(self):
+        await super().asyncSetUp()
+        self.server = MockupDB()
+        self.handshake_req = None
+
+        def respond(r):
+            if "ismaster" in r:
+                # then this is a handshake request
+                self.handshake_req = r
+            return r.reply(OpMsgReply(minWireVersion=0, maxWireVersion=13))
```
NoahStapp: Is `maxWireVersion=13` specified by the spec or by MockupDB?

pymongo/asynchronous/mongo_client.py
sleepyStick: Would checking that an instance of DriverInfo was passed in satisfy this?

test/asynchronous/test_client_metadata.py
```diff
@@ -0,0 +1,225 @@
...
+        self,
+        client: AsyncMongoClient,
+        add_name: str,
+        add_version: Optional[str],
+        add_platform: Optional[str],
+    ) -> None:
+        # send initial metadata
+        name, version, platform, metadata = await self.send_ping_and_get_metadata(client, True)
+        # wait for connection to become idle
+        await asyncio.sleep(0.005)
```
NoahStapp, line 103: Can we refactor this into an `async_wait_until` with some predicate?

pymongo/asynchronous/mongo_client.py
sleepyStick: (I'm assuming you're referring to the `If v is None` statement here) Based on the spec/tests, it seemed like it was possible for  "name", "version" or "platform" to be None and therefore I didn't want do to anything special / add `None` into the metadata.

test/asynchronous/test_client_metadata.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+
+        metadata.pop("driver")
+        metadata.pop("platform")
+        new_metadata.pop("driver")
+        new_metadata.pop("platform")
+        self.assertEqual(metadata, new_metadata)
+
+    async def test_append_metadata(self):
+        client = AsyncMongoClient(
```
NoahStapp: All of these clients should be created with one of our `PyMongoTestCase` helpers, not raw `AsyncMongoClient()`.

test/asynchronous/unified_format.py
```diff
@@ -925,7 +931,6 @@ async def run_entity_operation(self, spec):
             )
         else:
             arguments = {}
-
```
NoahStapp: Whitespace removed?

NoahStapp (changes requested):

pymongo/asynchronous/mongo_client.py
ShaneHarvey: No because a user can create a subclass of DriverInfo and add more fields.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: I'm refering to `if k in metadata:`.

pymongo/asynchronous/mongo_client.py
sleepyStick: Good call, done!

pymongo/asynchronous/mongo_client.py
sleepyStick: I made my best attempt at it, but lmk if I was off / am still missing something / have it incorrectly formatted >.<

pymongo/pool_options.py
```diff
@@ -376,25 +376,31 @@ def __init__(
...
         env = _metadata_env()
         if env:
             self.__metadata["env"] = env
 
         _truncate_metadata(self.__metadata)
 
+    def _update_metadata(self, driver: DriverInfo):
+        """Updates the client's metadata"""
+        if driver.name:
+            self.__metadata["driver"]["name"] = "{}|{}".format(
```
ShaneHarvey: Could we make this method atomic? As in, first create a copy of `__metadata`, then mutate, then update? That way the driver handshake will never serialize half of the metadata.

test/asynchronous/test_client_metadata.py
sleepyStick: I think its needed so that MockupDB would work(?) without specification, the `maxWireVersion` is 6, and PyMongo raises a configuration error: `pymongo.errors.ConfigurationError: Server at localhost:56617 reports wire version 0, but this version of PyMongo requires at least 7 (MongoDB 4.0).`

test/asynchronous/test_client_metadata.py
sleepyStick: The spec specifically stated to wait 5ms

test/asynchronous/unified_format.py
sleepyStick: oops accident!

ShaneHarvey (approved):
API changes LGTM. Could you also add the changelog entry?

NoahStapp (approved):
LGTM! Great work!

PR #2359, opened 2025-06-05
PYTHON-5321 Remove Serverless testing

blink1073:
Removes all usages in Python files and generated Evergreen config.
We do not update any of the spec files, since those were not changed upstream.


aclark4life (approved):
Much less serverless, LGTM!

PR #2360, opened 2025-06-05
PYTHON-5406 - AsyncPeriodicExecutor must reset CSOT contextvars befor…

NoahStapp:
…e executing

### DIFF
--- a/pymongo/_asyncio_task.py
+++ b/pymongo/_asyncio_task.py
@@ -43,6 +43,7 @@ class _Task(asyncio.Task):
         return self._cancel_requests
 
 
+# We can directly pass an empty Context() object to create_task in Python >= 3.11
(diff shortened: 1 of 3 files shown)


pymongo/_asyncio_task.py
```diff
@@ -43,6 +43,7 @@ def cancelling(self) -> int:
         return self._cancel_requests
 
 
+# We can directly pass an empty Context() object to create_task in Python >= 3.11
```
ShaneHarvey: I'm not sure we'd want to do that. The user might be have contextvars that they want to propagate to the background tasks.

NoahStapp: A better idea I had was to explicitly remove our CSOT contextvars and leave the rest alone.

ShaneHarvey: Then, should we remove this comment?

ShaneHarvey (approved):

PR #2363, opened 2025-06-09
PYTHON-5405 Use legacy wait_for_read cancellation approach on Windows

blink1073:


pymongo/network_layer.py
```diff
@@ -286,6 +287,7 @@ async def _async_socket_receive(
 
 
 _PYPY = "PyPy" in sys.version
+_WINDOWS = os.name == "nt"
```
ShaneHarvey: Our usual windows check is `sys.platform == "win32"`

ShaneHarvey (changes requested):

pymongo/network_layer.py
blink1073: Updated

ShaneHarvey (approved):

PR #2376, opened 2025-06-10
PYTHON-5409 Update test_session.py comment

aclark4life:


blink1073:
The PR has no changes in it

PR #2383, opened 2025-06-12
PYTHON 5212 - Use asyncio.loop.sock_connect in _async_create_connection

NoahStapp:


pymongo/pool_shared.py
```diff
@@ -244,7 +244,8 @@ async def _async_create_connection(address: _Address, options: PoolOptions) -> s
             sock.settimeout(timeout)
             sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, True)
             _set_keepalive_times(sock)
-            sock.connect(sa)
+            sock.setblocking(False)
+            await asyncio.get_running_loop().sock_connect(sock, sa)
```
ShaneHarvey: We need to do the same above for unix sockets as well.

ShaneHarvey (changes requested):

pymongo/pool_shared.py
ShaneHarvey: Also don't we need to add more code to emulate the connect_timeout behavior now that we're using `sock.setblocking(False)`? Or is the async connect timeout behavior handled at a higher level?

NoahStapp: ~~Yes: we'll have to move all socket timeout behavior out of the `socket` itself and into the `AsyncNetworkInterface`. In the middle of that now.~~ We already do this.

NoahStapp: ~~Sorry, specifically for encryption `_async_configured_socket` raw sockets. We manage `AsyncNetworkInterface` timeouts at a higher level already.~~ Incorrect, we already handle raw async socket timeouts correctly to account for this change. The socket only needs to be nonblocking for creation.

ShaneHarvey: The problem is that `_async_create_connection` is called before AsyncNetworkInterface is ever created. So the timeout needs to be enforced here.

NoahStapp: I misunderstood what you meant. My next commit already addresses this with an `asyncio.wait_for`.

NoahStapp:
Scheduled every Evergreen task for maximum confidence in the correctness of the change.

NoahStapp:
Encryption test failures are due to `ssl_context.wrap_socket` not allowing non-blocking sockets when `do_handshake_on_connect=True`, which is the default. Will resolve tomorrow.

NoahStapp:
All test failures are known failures or persistent TLS PyPy issues.

pymongo/pool_shared.py
```diff
@@ -241,14 +242,21 @@ async def _async_create_connection(address: _Address, options: PoolOptions) -> s
...
+            # Socket needs to be non-blocking during connection to not block the event loop
+            sock.setblocking(False)
+            await asyncio.wait_for(
+                asyncio.get_running_loop().sock_connect(sock, sa), timeout=timeout
+            )
+            sock.settimeout(timeout)
             return sock
+        except asyncio.TimeoutError as e:
+            sock.close()
+            raise socket.timeout("timed out") from e
```
ShaneHarvey: Should we store the exception and carry on like we do in the sync OSError case?:
```python
        except asyncio.TimeoutError as e:
            sock.close()
            err = socket.timeout("timed out")
            err.__cause__ = e
```

If not, I worry about subtle behavior differences between sync and async.

NoahStapp: This is a slick workaround to the lack of `from e` without `raise`.

test/asynchronous/test_async_loop_unblocked.py
```diff
@@ -0,0 +1,56 @@
...
+        with self.assertRaisesRegex(ServerSelectionTimeoutError, "No servers found yet"):
+            await client.admin.command("ping")
+
+        t.cancel()
+        with self.assertRaises(asyncio.CancelledError):
+            await t
+
+        self.assertLessEqual(
+            sorted(latencies, reverse=True)[0],
+            0.2,
```
ShaneHarvey: I worry only adding 100ms will be flaky, can we make this 1 full second?

NoahStapp: Good call, I was slightly too optimistic about the EG host performance.

ShaneHarvey (approved):

PR #2393, opened 2025-06-18
PYTHON-5126 - Implemented new test cases for Binary Vector

caseyclements:
This PR simply adds new invalid cases that @qingyang-hu had added. This had been done long ago, but we're splitting the python changes out from those specifically related to ignored bit in packed_bit dtype.


blink1073 (approved):
LGTM!

PR #2395, opened 2025-06-20
PYTHON-5413 Handle flaky tests

blink1073:
The goal was to get the following to pass on multiple restarts:

- MongoDB 8.0 with pypy
- All of the Windows and MacOS standard tests
- Encryption crypt_shared with pypy
- Free-threading builds

Since these are the most flaky builds in the waterfall currently.  The current build has passed three times for the above variants.


test/asynchronous/test_client_bulk_write.py
```diff
@@ -619,16 +620,17 @@ async def test_15_unacknowledged_write_across_batches(self):
...
         await super().asyncSetUp()
         self.max_write_batch_size = await async_client_context.max_write_batch_size
         self.max_bson_object_size = await async_client_context.max_bson_size
         self.max_message_size_bytes = await async_client_context.max_message_size_bytes
 
     @async_client_context.require_version_min(8, 0, 0, -24)
     @async_client_context.require_failCommand_fail_point
+    @flaky(reason="PYTHON-5290", max_runs=3, affects_cpython_linux=True)
     async def test_timeout_in_multi_batch_bulk_write(self):
+        if sys.platform != "linux":
```
ShaneHarvey: Could we do this only in Evergreen CI?

ShaneHarvey (changes requested):
Could we only enable flakey in CI and not locally? I worry having it run by default locally will make it more difficult to debug test failures.

test/asynchronous/test_client_bulk_write.py
blink1073: Updated this and added logic/docs around enabling and disabling the flaky decorator.

ShaneHarvey (approved):

PR #2397, opened 2025-06-23
PYTHON-5289 Validate ignored bits are 0 on write for bson.BinaryVector

caseyclements:
This is the implementation of the spec changes applied in [specifications/pull/1812](https://github.com/mongodb/specifications/pull/1812).

### DIFF
--- a/bson/binary.py
+++ b/bson/binary.py
@@ -14,6 +14,7 @@
 from __future__ import annotations
 
 import struct
+import warnings
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Optional, Sequence, Tuple, Type, Union, overload
 from uuid import UUID
@@ -471,6 +472,10 @@ class Binary(bytes):
 
         metadata = struct.pack("<sB", dtype.value, padding)
         data = struct.pack(f"<{len(vector)}{format_str}", *vector)  # type: ignore
+        if padding and len(vector) and not (data[-1] & ((1 << padding) - 1)) == 0:
+            raise ValueError(
+                "Vector has a padding P, but bits in the final byte lower than P are non-zero. They must be zero."
+            )
         return cls(metadata + data, subtype=VECTOR_SUBTYPE)
 
     def as_vector(self) -> BinaryVector:
@@ -522,6 +527,12 @@ class Binary(bytes):
             dtype_format = "B"
             format_string = f"<{n_values}{dtype_format}"
             unpacked_uint8s = list(struct.unpack_from(format_string, self, position))
+            if padding and n_values and unpacked_uint8s[-1] & (1 << padding) - 1 != 0:
+                warnings.warn(
+                    "Vector has a padding P, but bits in the final byte lower than P are non-zero. In the next major version, they must be zero.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
             return BinaryVector(unpacked_uint8s, dtype, padding)
 
         else:
--- a/test/test_bson.py
+++ b/test/test_bson.py
@@ -739,7 +739,7 @@ class TestBSON(unittest.TestCase):
         """Tests of subtype 9"""
         # We start with valid cases, across the 3 dtypes implemented.
         # Work with a simple vector that can be interpreted as int8, float32, or ubyte
-        list_vector = [127, 7]
+        list_vector = [127, 8]
         # As INT8, vector has length 2
         binary_vector = Binary.from_vector(list_vector, BinaryVectorDtype.INT8)
         vector = binary_vector.as_vector()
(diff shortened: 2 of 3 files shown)


caseyclements:
Failing tests are unrelated.

bson/binary.py
```diff
@@ -522,6 +530,12 @@ def as_vector(self) -> BinaryVector:
             dtype_format = "B"
             format_string = f"<{n_values}{dtype_format}"
             unpacked_uint8s = list(struct.unpack_from(format_string, self, position))
+            if padding and n_values and unpacked_uint8s[-1] & (1 << padding) - 1 != 0:
+                warnings.warn(
+                    "Vector has a padding P, but bits in the final byte lower than P are non-zero. In the next major version, they must be zero.",
```
NoahStapp: So in the next major version, this warning will become an error to match the behavior in `from_vector`?

caseyclements: Yes.  We make the significant change  (exception not warning) in the next major change and document in changelog and api.  The ticket for that is [PYTHON-5280](https://jira.mongodb.org/browse/PYTHON-5280) and has 5.0 as its fix version.

test/test_bson.py
```diff
@@ -791,8 +791,15 @@ def test_vector(self):
...
                 self.fail("Failed to raise an exception.")
 
-        # Test form of Binary.from_vector(BinaryVector)
+        # Test one must pass zeros for all ignored bits
+        try:
+            Binary.from_vector([255], BinaryVectorDtype.PACKED_BIT, padding=7)
+        except Exception as exc:
+            self.assertIsInstance(exc, ValueError)
+        else:
+            self.fail("Failed to raise an exception.")
```
ShaneHarvey: This assertion should use assertRaises.

bson/binary.py
ShaneHarvey: Could you add "pymongo version 5.0" rather than "the next major version". This helps show the user the version (5.0) and that the warning is coming from pymongo.

ShaneHarvey (dismissed):
Any docs or changelog we need to update here?

test/test_bson.py
caseyclements: Yeah. You got me. This looked funny when I reviewed it last week. I'll update to follow best practice.

bson/binary.py
caseyclements: Sure can.

doc/changelog.rst
```diff
@@ -51,6 +51,9 @@ PyMongo 4.13 brings a number of changes including:
   or the `migration guide <https://www.mongodb.com/docs/languages/python/pymongo-driver/current/reference/migration/>`_ for more information.
 - Fixed a bug where :class:`pymongo.write_concern.WriteConcern` repr was not eval-able
   when using ``w="majority"``.
+- Ignored bits in a BSON BinaryVector of PACKED_BIT dtype should be set to zero.
+  On writes, this is enforced and is a breaking change.
+  Reads from the database will not fail, however a warning will be triggered.
```
ShaneHarvey: Suggest referring to BSON encoding vs decoding instead of writes vs reads.

doc/changelog.rst
```diff
@@ -51,6 +51,9 @@ PyMongo 4.13 brings a number of changes including:
   or the `migration guide <https://www.mongodb.com/docs/languages/python/pymongo-driver/current/reference/migration/>`_ for more information.
 - Fixed a bug where :class:`pymongo.write_concern.WriteConcern` repr was not eval-able
   when using ``w="majority"``.
+- Ignored bits in a BSON BinaryVector of PACKED_BIT dtype should be set to zero.
```
ShaneHarvey: "Ignored bits" refers to the padding bits right? Should we say "padding" to avoid introducing a new term?

caseyclements: Will update to clarify.

doc/changelog.rst
caseyclements: Ok. Updated with nod to the motivation.

doc/changelog.rst
```diff
@@ -51,6 +51,9 @@ PyMongo 4.13 brings a number of changes including:
   or the `migration guide <https://www.mongodb.com/docs/languages/python/pymongo-driver/current/reference/migration/>`_ for more information.
 - Fixed a bug where :class:`pymongo.write_concern.WriteConcern` repr was not eval-able
   when using ``w="majority"``.
+- When padding is set, ignored bits in a BSON BinaryVector of PACKED_BIT dtype should be set to zero.
+  When encoding, this is enforced and is a breaking change.
+  It is not enforced when decoding, so reading from the database will not fail, however a warning will be triggered.
```
NoahStapp: Should this state that in PyMongo >= 5.0, it will be enforced?

test/test_bson.py
```diff
@@ -784,15 +784,14 @@ def test_vector(self):
...
-            else:
-                self.fail("Failed to raise an exception.")
 
-        # Test form of Binary.from_vector(BinaryVector)
+        # Test one must pass zeros for all ignored bits
+        with self.assertRaises(ValueError):
+            Binary.from_vector([255], BinaryVectorDtype.PACKED_BIT, padding=7)
 
+        # Test form of Binary.from_vector(BinaryVector)
         assert padded_vec == Binary.from_vector(
```
NoahStapp, line 800: Should we test that the warning is raised as well?

caseyclements: Definitely. Thanks! Added.

PR #2402, opened 2025-06-25
PYTHON-5404 - Add docs + justfile target for profiling execution

NoahStapp:


tools/generate_flamegraph.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -eu
+
+sudo py-spy record -o ${1:-profile}.svg -r ${2:-2000} -- python $3
```
ShaneHarvey: Looks like you want output_name and sample_rate to be optional but positional arguments can't work like that. If one does `just flamegraph script.py` then $1=script.py and 2 and 3 will be empty.

CONTRIBUTING.md
```diff
@@ -460,3 +460,17 @@ partially-converted asynchronous version of the same name to the `test/asynchron
...
+1. Install `py-spy` if you haven't already:
+   ```bash
+   pip install py-spy
+   ```
+2. Inside your test script, perform any required setup and then loop over the code you want to profile for improved sampling
+3. Run the `flamegraph` justfile target to generate a `.svg` file containing the flame graph:
+  ```bash
+  just flamegraph <output_name=profile> <sample_rate=2000> <path/to/script>
+  ```
+4. Profiling should be done on a Linux system, as macOS and Windows do not support the `--native` option of `py-spy`.
```
ShaneHarvey: Suggest: "If you need to include native code (eg the C extensions), profiling should ..."

tools/generate_flamegraph.sh
NoahStapp: Good catch, went a little too fast and loose here.

tools/generate_flamegraph.py
```diff
@@ -0,0 +1,58 @@
+from __future__ import annotations
+
+import argparse
+import subprocess
+import sys
+
+
+def main():
```
ShaneHarvey: Is it worth having this file at all? We're adding 60 lines of code to change this:
```
pip install py-spy
py-spy record -o output.svg -- <path/to/script>
```
into:
```
pip install py-spy
just flamegraph <--output_name=profile> <--sample_rate=2000> <path/to/script>
```

Can we just pass all the args to `py-spy record` and call it a day? It already has a default value with sample_rate so all we really save here is the `-o profile.svg` which I'm not sure is worth this extra file. The `just flamegraph` approach is also inflexible since it doesn't allow passing arbitrary options to `py-spy record`.

CONTRIBUTING.md
```diff
@@ -460,3 +460,17 @@ partially-converted asynchronous version of the same name to the `test/asynchron
...
+## Generating a flame graph using py-spy
+To profile a test script and generate a flame graph, follow these steps:
+1. Install `py-spy` if you haven't already:
+   ```bash
+   pip install py-spy
+   ```
+2. Inside your test script, perform any required setup and then loop over the code you want to profile for improved sampling
+3. Run the `flamegraph` justfile target to generate a `.svg` file containing the flame graph:
+  ```bash
+  just flamegraph <output_name=profile> <sample_rate=2000> <path/to/script>
```
ShaneHarvey: Should we even add this to `just`? Since we already need to install `py-spy` manually why not document the py-spy command to run without `just`.

tools/generate_flamegraph.py
NoahStapp: Good point. I like the conciseness of `just flamegraph`, but we're already requiring enough args and manual steps that we don't save much at all by having it. Just including docs on how to run py-spy should be sufficient here.

ShaneHarvey (approved):

ShaneHarvey:
Thanks!

PR #2410, opened 2025-06-26
PYTHON-5382 - Add a test with min dependencies

NoahStapp:


.github/workflows/test-python.yml
```diff
@@ -220,3 +220,28 @@ jobs:
...
+          uses: supercharge/mongodb-github-action@90004df786821b6308fb02299e5835d0dae05d0d # 1.12.0
+          with:
+            mongodb-version: 6.0
+        - name: Run connect test
+          shell: bash
+          run: |
+            uv venv
+            source .venv/bin/activate
+            uv pip install -e ".[test]" --resolution=lowest-direct
+            pytest -v test/test_srv_polling.py
```
ShaneHarvey, line 248: Sorry I just noticed this was test_srv_polling but test_dns would probably be a better choice. Or both? And should we do both sync and async?

NoahStapp: Almost all of the `test_dns` tests require a sharded or replica cluster. We can add it here anyway, but it'll only add a handful of tests. And good catch, we should run sync and async tests here.

ShaneHarvey: Ah okay then, could you add a comment explaining why we use test_srv_polling only and not test_dns?

ShaneHarvey: Looks like test_dns's mocking is not compatible with dnspython 1.0. Perhaps it's not worth it to add in that case.

NoahStapp: And unfortunately there's no `resolver.query` method for the async resolver. I'll add a comment that the async `test_dns` is not compatible with dnspython 1.x, which is consistent with our docs for the async API.

NoahStapp: Adding a separate test for dnspython 2.0 for async support.

NoahStapp: Ended up having to use dnspython 2.1.0, as 2.0.0 does not include the `lifetime` kwarg we pass to our `resolve` calls, but only for the async API.

ShaneHarvey: Does that mean we have to update our docs and error message to say >=2.1 is required for async?

NoahStapp: Yes, I'll file a follow-up ticket for those changes.

NoahStapp: https://jira.mongodb.org/browse/PYTHON-5426

ShaneHarvey (approved):

PR #2412, opened 2025-06-30
PYTHON-5423 Always use subprocess.run instead of subprocess.check_call or subprocess.call

sleepyStick:
Didn't find any instances of `subprocess.call` and all (three) instances of `subprocess.check_call(...)` are now `subprocess.run(..., check=True)`


blink1073 (approved):
LGTM!

PR #2413, opened 2025-06-30
PYTHON-5421 Make parse_uri() return "options" as a dict rather than _CaseInsensitiveDictionary

sleepyStick:
- Added a method to _CaseInsensitiveDictionary to return itself as a plain dict
- parse_uri now returns "options" as a plain dict: 
```
>>> from pymongo.uri_parser import parse_uri
>>> type(parse_uri("mongodb://cluster0.example.mongodb.net/myDatabase?retryWrit\
es=true&retryReads=true").get("options"))
<class 'dict'>
```

### DIFF
--- a/pymongo/asynchronous/uri_parser.py
+++ b/pymongo/asynchronous/uri_parser.py
@@ -184,5 +184,5 @@ async def _parse_srv(
 
     return {
         "nodelist": nodes,
-        "options": options,
+        "options": options.as_dict(),
     }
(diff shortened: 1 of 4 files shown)


aclark4life (approved):
LGTM, thanks!

ShaneHarvey (commented):
Please add a changelog entry and a `..versionchanged` to parse_uri to describe the breaking change. Also, do we have a test that verifies options is returned as a dict and not a custom class?

sleepyStick:
> Please add a changelog entry and a `..versionchanged` to parse_uri to describe the breaking change. Also, do we have a test that verifies options is returned as a dict and not a custom class?

Going to make a follow up PR for these changes!

PR #2414, opened 2025-07-01
PYTHON-4779 Use slots for Selection and TopologyDescription

blink1073:
I ran a perf build: https://spruce.mongodb.com/task/mongo_python_driver_performance_benchmarks_perf_8.0_standalone_patch_578c6c2ad2559c4c939c682151eeeaf3b847ab3e_68643055c8c88400072d3e1b_25_07_01_19_00_56/trend-charts?execution=0

I did not see an appreciable change in FindOneByID8Threads or SmallDocInsertOne, unfortunately.

### DIFF
--- a/pymongo/server_selectors.py
+++ b/pymongo/server_selectors.py
@@ -32,6 +32,8 @@ TagSets = Sequence[TagSet]
 class Selection:
     """Input or output of a server selector function."""
 
+    __slots__ = ("totopology_description", "server_descriptions", "primary", "common_wire_version")
+
     @classmethod
     def from_topology_description(cls, topology_description: TopologyDescription) -> Selection:
         known_servers = topology_description.known_servers
--- a/pymongo/server_selectors.py
+++ b/pymongo/server_selectors.py
@@ -32,7 +32,7 @@ TagSets = Sequence[TagSet]
 class Selection:
     """Input or output of a server selector function."""
 
-    __slots__ = ("totopology_description", "server_descriptions", "primary", "common_wire_version")
+    __slots__ = ("topology_description", "server_descriptions", "primary", "common_wire_version")
 
     @classmethod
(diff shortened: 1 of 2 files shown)


ShaneHarvey (commented):
Could you add something like this to the changelog?
```
- Classes :class:`~bson.int64.Int64`, :class:`~bson.min_key.MinKey`,
  :class:`~bson.max_key.MaxKey`, :class:`~bson.timestamp.Timestamp`,
  :class:`~bson.regex.Regex`, and :class:`~bson.dbref.DBRef` all implement
  ``__slots__`` now. This means that their attributes are fixed, and new
  attributes cannot be added to them at runtime.
```

Also is there any measurable benefit at all to this change? Like what if we have 1 thread just doing client._select_server? Or 100 threads?

blink1073:
I see no significant differences with this script between this PR and `master`:

```python
from concurrent.futures import ThreadPoolExecutor, wait

client = MongoClient()
read_preference = ReadPreference.PRIMARY


def test():
    for i in range(100):
        client._select_server(read_preference, session, "testOperation")



def test2_target():
    client._select_server(read_preference, session, "testOperation")


def test2():
    futures = []
    with ThreadPoolExecutor(max_workers=100) as pool:
        for i in range(100):
            futures.append(pool.submit(test2_target))
        wait(futures)
    

if __name__ == '__main__':
    import timeit
    print(timeit.timeit("test()", setup="from __main__ import test, client, read_preference, session", number=1000))
    print(timeit.timeit("test2()", setup="from __main__ import test2, client, read_preference, session", number=1000))
```

blink1073:
Given that there is no measurable improvement, and Selection isn't even represented in our API docs, I think we  should close this PR and ticket.

ShaneHarvey:
Closing sounds good to me. We should only do this if we can show it leads to a measurable improvement.

PR #2416, opened 2025-07-03
PYTHON-4780 Implement fast path for server selection with Primary

blink1073:
Tested with the following script:

```python
from concurrent.futures import ThreadPoolExecutor, wait

client = MongoClient()
read_preference = ReadPreference.PRIMARY


def test():
    for i in range(100):
        client._select_server(read_preference, session, "testOperation")



def test2_target():
    client._select_server(read_preference, session, "testOperation")


def test2():
    futures = []
    with ThreadPoolExecutor(max_workers=100) as pool:
        for i in range(100):
            futures.append(pool.submit(test2_target))
        wait(futures)
    

if __name__ == '__main__':
    import timeit
    print(timeit.timeit("test()", setup="from __main__ import test, client, read_preference, session", number=1000))
    print(timeit.timeit("test2()", setup="from __main__ import test2, client, read_preference, session", number=1000))
```

On `master`:
0.4806627919897437
1.5110677920019953

With this change:
0.17988054199668113
1.233099208009662

### DIFF
--- a/pymongo/topology_description.py
+++ b/pymongo/topology_description.py
@@ -325,9 +325,7 @@ class TopologyDescription:
             return [description] if description else []
 
         # Primary selection fast path.
-        if self.topology_type == TOPOLOGY_TYPE.ReplicaSetWithPrimary and isinstance(
-            selector, Primary
-        ):
+        if self.topology_type == TOPOLOGY_TYPE.ReplicaSetWithPrimary and type(selector) is Primary:
             for sd in self._server_descriptions.values():
                 if sd.server_type == SERVER_TYPE.RSPrimary:
                     return [sd]
--- a/pymongo/topology_description.py
+++ b/pymongo/topology_description.py
@@ -324,6 +324,14 @@ class TopologyDescription:
             description = self.server_descriptions().get(address)
             return [description] if description else []
 
+        # Primary selection fast path.
+        if self.topology_type == TOPOLOGY_TYPE.ReplicaSetWithPrimary and isinstance(
+            selector, TOPOLOGY_TYPE.Primary
+        ):
+            for sd in self._server_descriptions.values():
+                if sd.server_type == SERVER_TYPE.RSPrimary:
+                    return [sd]
+
         selection = Selection.from_topology_description(self)
         # Ignore read preference for sharded clusters.
         if self.topology_type != TOPOLOGY_TYPE.Sharded:
--- a/pymongo/topology_description.py
+++ b/pymongo/topology_description.py
@@ -34,7 +34,7 @@ from bson.min_key import MinKey
 from bson.objectid import ObjectId
 from pymongo import common
 from pymongo.errors import ConfigurationError, PyMongoError
-from pymongo.read_preferences import ReadPreference, _AggWritePref, _ServerMode
+from pymongo.read_preferences import Primary, ReadPreference, _AggWritePref, _ServerMode
 from pymongo.server_description import ServerDescription
 from pymongo.server_selectors import Selection
 from pymongo.server_type import SERVER_TYPE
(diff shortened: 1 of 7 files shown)


ShaneHarvey:
Minor note, for the benchmark thread dispatching could be dominating `test2`. Could you change it to `pool.submit(test2)` and just remove `test2_target`? That would probably show a larger perf increase for threaded.

pymongo/topology_description.py
```diff
@@ -324,6 +324,14 @@ def apply_selector(
             description = self.server_descriptions().get(address)
             return [description] if description else []
 
+        # Primary selection fast path.
+        if self.topology_type == TOPOLOGY_TYPE.ReplicaSetWithPrimary and isinstance(
+            selector, Primary
+        ):
+            for sd in self._server_descriptions.values():
+                if sd.server_type == SERVER_TYPE.RSPrimary:
+                    return [sd]
```
ShaneHarvey: should this block also have a fast path for returning `[]` when there is no primary?

blink1073:
With your suggested benchmark changes:

On `master`:
0.4840990829980001
0.5537169170274865

With this PR:
0.1950377920002211
0.260104083979968

pymongo/topology_description.py
blink1073: Done

pymongo/topology_description.py
```diff
@@ -324,6 +324,16 @@ def apply_selector(
             description = self.server_descriptions().get(address)
             return [description] if description else []
 
+        # Primary selection fast path.
+        if self.topology_type == TOPOLOGY_TYPE.ReplicaSetWithPrimary and isinstance(
+            selector, Primary
```
ShaneHarvey: My one concern is that technically this could be a breaking change if a user is subclassing Primary and overriding `__call__` to do something special during server selection. One simple way to avoid that risk is doing `type(selector) is Primary` rather than isinstance. Or we could document the potential breaking change.

ShaneHarvey (commented):
Nice work! Could you add a changelog saying "Improved performance of selecting a server with the Primary selector."?

pymongo/topology_description.py
blink1073: I went with the code modification option

pymongo/topology_description.py
```diff
@@ -324,6 +324,14 @@ def apply_selector(
             description = self.server_descriptions().get(address)
             return [description] if description else []
 
+        # Primary selection fast path.
+        if self.topology_type == TOPOLOGY_TYPE.ReplicaSetWithPrimary and type(selector) is Primary:
+            for sd in self._server_descriptions.values():
+                if sd.server_type == SERVER_TYPE.RSPrimary:
+                    return [sd]
```
ShaneHarvey: Oh one more thing, don't we have to call the `custom_selector` here if one was provided?

ShaneHarvey: And add a test for it?

blink1073: I'll have to defer for now, working on a HELP ticket

blink1073: Done

pymongo/topology_description.py
```diff
@@ -324,6 +324,21 @@ def apply_selector(
             description = self.server_descriptions().get(address)
             return [description] if description else []
 
+        # Primary selection fast path.
+        if (
+            custom_selector is None
```
ShaneHarvey: Is the `custom_selector is None` check supposed to be here?

test/test_topology.py
```diff
@@ -123,6 +126,25 @@ def test_timeout_configuration(self):
...
+        self.assertEqual(description.apply_selector(Primary()), [])
+
+        # If we set a primary server, we should get it back.
+        sd = list(description._server_descriptions.values())[0]
+        sd._server_type = SERVER_TYPE.RSPrimary
+        self.assertEqual(description.apply_selector(Primary()), [sd])
+
+        # If there is a custom selector, it should be applied.
+        def custom_selector(servers):
+            return []
```
ShaneHarvey, line 144: Neat trick.

pymongo/topology_description.py
blink1073: No, I forgot to back that change out

PR #2420, opened 2025-07-03
PYTHON-5421 continued - update changelog, update docstring, and add testing

sleepyStick:
follow up PR to https://github.com/mongodb/mongo-python-driver/pull/2413

### DIFF
--- a/pymongo/asynchronous/uri_parser.py
+++ b/pymongo/asynchronous/uri_parser.py
@@ -81,7 +81,7 @@ async def parse_uri(
     :param srv_service_name: A custom SRV service name
 
     .. versionchanged:: 4.14
-       'options' is now type ``dict`` as opposed to a ``_CaseInsensitiveDictionary``.
+       ``options`` is now type ``dict`` as opposed to a ``_CaseInsensitiveDictionary``.
 
     .. versionchanged:: 4.6
        The delimiting slash (``/``) between hosts and connection options is now optional.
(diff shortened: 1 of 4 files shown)


aclark4life (changes requested):
It may just be me but I would do ``options`` instead of 'options' what do you think? Also I'd say "``parse_uri``'s ``options`` parameter" instead of ``parse_uri's`` ``options``. ``</nit>``

pymongo/asynchronous/uri_parser.py
```diff
@@ -80,6 +80,9 @@ async def parse_uri(
           wait for a response from the DNS server.
     :param srv_service_name: A custom SRV service name
 
+    .. versionchanged:: 4.14
+       'options' is now type ``dict`` as opposed to a ``_CaseInsensitiveDictionary``.
```
aclark4life: Here too!

pymongo/synchronous/uri_parser.py
```diff
@@ -80,6 +80,9 @@ def parse_uri(
           wait for a response from the DNS server.
     :param srv_service_name: A custom SRV service name
 
+    .. versionchanged:: 4.14
+       'options' is now type ``dict`` as opposed to a ``_CaseInsensitiveDictionary``.
```
aclark4life: And here

pymongo/asynchronous/uri_parser.py
sleepyStick: Sorry missed this, good catch!
just ``options`` right?

PR #2424, opened 2025-07-09
PYTHON-5315 Mark test_recover_from_initially_erroring_seedlist as flaky

blink1073:
This was missed in #2395.

Also corrects the less common spelling of "flakey" to ensure search-ability.


sleepyStick (approved):
thanks for fixing the typos!

PR #2431, opened 2025-07-15
PYTHON-4931 Add spec tests for GridFS rename

sleepyStick:
- removed associated patch files and lines in `remove-unimplemented-tests.sh`
- Based on the spec test changes, I'm guessing that `CorruptGridFile` is now a client error as opposed to just a generic error
- maybe random note: When i ran the gridfs tests locally, I saw that we're skipping test `test_download_when_an_intermediate_chunk_is_the_wrong_size` pending PYTHON-4598 but I cant see that ticket on jira? (it says its either been deleted or I don't have permissions?)


blink1073 (approved):
LGTM!

sleepyStick:
> * [PYTHON-4598](https://jira.mongodb.org/browse/PYTHON-4598)

Ticket Opened! Thanks for the clarification!

PR #2438, opened 2025-07-16
PYTHON-4019 Infinite loop in generic transactional provider due to dup keys

blink1073:
Copies the recommended warning from the [spec change](https://github.com/mongodb/specifications/commit/13117d601fb6ea177e485880175b5be1651a7e46).

### DIFF
--- a/pymongo/asynchronous/client_session.py
+++ b/pymongo/asynchronous/client_session.py
@@ -663,7 +663,8 @@ class AsyncClientSession:
         The ``callback`` MUST NOT silently handle command errors
         without allowing such errors to propagate. Command errors may abort the
         transaction on the server, and an attempt to commit the transaction will
-        be rejected with a ``NoSuchTransaction`` error.
+        be rejected with a ``NoSuchTransaction`` error.  For more information see
+        the `transactions specification`_.
 
         When :meth:`~AsyncClientSession.commit_transaction` raises an exception with
         the ``"UnknownTransactionCommitResult"`` error label,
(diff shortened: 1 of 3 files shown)


pymongo/asynchronous/client_session.py
```diff
@@ -660,6 +660,11 @@ async def callback(session, custom_arg, custom_kwarg=None):
         ``with_transaction`` starts a new transaction and re-executes
         the ``callback``.
 
+        The ``callback`` MUST NOT silently handle command errors
+        without allowing such errors to propagate. Command errors may abort the
+        transaction on the server, and an attempt to commit the transaction will
+        be rejected with a ``NoSuchTransaction`` error.
+
```
caseyclements, line 668: I suggest adding more of what's suggested in the spec change below. It explains how one could get into an infinite loop, and also how to ensure it doesn't happen.
```
For example, ``DuplicateKeyError`` is an error that aborts a transaction on the
server. If the callback catches ``DuplicateKeyError`` and does not re-throw it,
the driver will attempt to commit the transaction. The server will reject the
commit attempt with ``NoSuchTransaction`` error. This error has the
"TransientTransactionError" label and the driver will retry the commit. This
will result in an infinite loop.

Drivers MUST recommend that the callback re-throw command errors if they
need to be handled inside the callback. Drivers SHOULD also recommend using
Core Transaction API if a user wants to handle errors in a custom way.
```

caseyclements (changes requested):

pymongo/asynchronous/client_session.py
blink1073: That feels too verbose for a docstring, I added a link to the specification instead, which we do in other places.

NoahStapp (approved):

PR #2439, opened 2025-07-17
PYTHON-4677 Specify how maxTimeMS can be set for explain helpers

blink1073:


sleepyStick (approved):
LGTM

test/asynchronous/test_cursor.py
```diff
@@ -362,6 +362,24 @@ async def test_explain_with_read_concern(self):
         self.assertEqual(len(started), 1)
         self.assertNotIn("readConcern", started[0].command)
 
+    # https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.md#14-explain-helpers-allow-users-to-specify-maxtimems
+    async def test_explain_csot(self):
+        # Create a MongoClient with command monitoring enabled (referred to as client).
+        listener = AllowListEventListener("explain")
+        client = await self.rs_or_single_client(event_listeners=[listener])
```
sleepyStick: i think this should be `self.async_rs_or_single_client`

sleepyStick (changes requested):
wait JK just saw the test failures HAHA

test/asynchronous/test_cursor.py
blink1073: Thanks!  I accidentally created the sync version first, then hand-ported it.

PR #2443, opened 2025-07-21
PYTHON-5253 Automated Spec Resync Quick Followup/Fix

sleepyStick:
Accidentally left out the name of the flag when calling one of the scripts! 

But now it definitely works -- ran a patch build to confirm and it generated: https://github.com/mongodb/mongo-python-driver/pull/2442

Let that be the spec resync pr that should've ran this morning :)


caseyclements (approved):
LGTM

PR #2456, opened 2025-08-04
PYTHON-5257 - Turn on mypy disallow_any_generics

NoahStapp:


pymongo/network_layer.py
```diff
@@ -613,7 +613,7 @@ def buffer_updated(self, nbytes: int) -> None:
                 result = self._pending_messages.popleft()
             else:
                 result = asyncio.get_running_loop().create_future()
-            # Future has been cancelled, close this connection
+            # Future[Any] has been cancelled, close this connection
```
sleepyStick: is this an accident?

NoahStapp: Yup, good catch! Fixed.

sleepyStick (approved):

pymongo/synchronous/collection.py
```diff
@@ -3107,7 +3107,7 @@ def distinct(
         comment: Optional[Any] = None,
         hint: Optional[_IndexKeyHint] = None,
         **kwargs: Any,
-    ) -> list:
+    ) -> list[str]:
```
gablank, line 3110: This is not necessarily correct: 

```python

import pymongo as pm
cl = pm.MongoClient()
co = cl["test"]["test"]
co.insert_one({"x": 1})
type(next(iter(co.distinct("x"))))  # int
```

PR #2457, opened 2025-08-05
Fix In-Use Encryption examples

kevinAlbs:
This PR is intended to make small fixes and improvements to the [In-Use Encryption](https://pymongo.readthedocs.io/en/stable/examples/encryption.html) examples:

- Remove incorrect collection names in the [Automatic Queryable Encryption](https://pymongo.readthedocs.io/en/stable/examples/encryption.html#queryable-encryption):
    ```
    pymongo.errors.EncryptionError: Encrypted State Collection name should follow enxcol_.<collection>.esc naming pattern
    ```
    DRIVERS-2565 describes the server change that required the metadata collection names to fit this pattern. libmongocrypt [applies the collection names](https://github.com/mongodb/libmongocrypt/blob/b36115cacad817de9dda8cef3b07b4b35e4d01c6/src/mc-schema-broker.c#L647-L659), so it is not necessary to specify. This PR removes the names from the examples.

- Remove incorrect tuple wrapping in the Azure [CSFLE on-demand credentials](https://pymongo.readthedocs.io/en/stable/examples/encryption.html#csfle-on-demand-credentials):
    ```
    ValueError: kms_providers must be a dict
    ```

- Remove unused imports.

### DIFF
--- a/doc/examples/encryption.rst
+++ b/doc/examples/encryption.rst
@@ -647,7 +647,7 @@ as demonstrated by the following example:
   import os
   from bson.codec_options import CodecOptions
   from pymongo import MongoClient
-  from pymongo.encryption import Algorithm, ClientEncryption, QueryType
+  from pymongo.encryption import ClientEncryption
   from pymongo.encryption_options import AutoEncryptionOpts
 
   local_master_key = os.urandom(96)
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thanks!

PR #2459, opened 2025-08-06
PYTHON-4431 - Remove ReadTheDocs Documentation in Favor of Official Docs

NoahStapp:


blink1073:
We need to remove the migration guide from the index as well

NoahStapp:
> We need to remove the migration guide from the index as well

Is that guide being removed entirely? The official docs link back to the RTD page: https://www.mongodb.com/docs/languages/python/pymongo-driver/current/reference/upgrade/#overview

blink1073:
It links back to the v4.8 tag specifically, so we can remove it from future versions.

blink1073:
Actually, the README should be updated to point to the main docs site as well as the API docs and the Changelog as separate links, and perhaps be closer to the top of the readme.

blink1073 (approved):
Thanks!

PR #2460, opened 2025-08-06
PYTHON-5215 Add an asyncio.Protocol implementation for KMS

blink1073:
See benchmark [gist](https://gist.github.com/blink1073/dbaab17d6486b01f1d590eb2731a49b2).

Benchmark Results:

Before:  `4.93s, 5.26s`
After: `4.93s`, `5.05s`

Depends on https://github.com/mongodb-labs/drivers-evergreen-tools/pull/679

### DIFF
--- a/pymongo/synchronous/pool.py
+++ b/pymongo/synchronous/pool.py
@@ -55,6 +55,7 @@ from pymongo.errors import (  # type:ignore[attr-defined]
     WaitQueueTimeoutError,
 )
 from pymongo.hello import Hello, HelloCompat
+from pymongo.helpers_shared import _get_timeout_details, format_timeout_details
 from pymongo.lock import (
     _cond_wait,
     _create_condition,
@@ -76,9 +77,7 @@ from pymongo.pool_shared import (
     SSLErrors,
     _CancellationContext,
     _configured_socket_interface,
-    _get_timeout_details,
     _raise_connection_failure,
-    format_timeout_details,
 )
 from pymongo.read_preferences import ReadPreference
 from pymongo.server_api import _add_to_command
--- a/pymongo/asynchronous/encryption.py
+++ b/pymongo/asynchronous/encryption.py
@@ -76,12 +76,12 @@ from pymongo.errors import (
     NetworkTimeout,
     ServerSelectionTimeoutError,
 )
+from pymongo.helpers_shared import _get_timeout_details
 from pymongo.network_layer import PyMongoKMSProtocol, async_receive_kms, async_sendall
 from pymongo.operations import UpdateOne
 from pymongo.pool_options import PoolOptions
 from pymongo.pool_shared import (
     _configured_protocol_interface,
-    _get_timeout_details,
     _raise_connection_failure,
 )
 from pymongo.read_concern import ReadConcern
--- a/pymongo/asynchronous/pool.py
+++ b/pymongo/asynchronous/pool.py
@@ -58,6 +58,7 @@ from pymongo.errors import (  # type:ignore[attr-defined]
     WaitQueueTimeoutError,
 )
 from pymongo.hello import Hello, HelloCompat
+from pymongo.helpers_shared import _get_timeout_details, format_timeout_details
 from pymongo.lock import (
     _async_cond_wait,
     _async_create_condition,
@@ -79,9 +80,7 @@ from pymongo.pool_shared import (
     SSLErrors,
     _CancellationContext,
     _configured_protocol_interface,
-    _get_timeout_details,
     _raise_connection_failure,
-    format_timeout_details,
 )
 from pymongo.read_preferences import ReadPreference
 from pymongo.server_api import _add_to_command
--- a/pymongo/asynchronous/server.py
+++ b/pymongo/asynchronous/server.py
@@ -38,7 +38,6 @@ from pymongo.logger import (
     _SDAMStatusMessage,
 )
 from pymongo.message import _convert_exception, _GetMore, _OpMsg, _Query
-from pymongo.pool_shared import _get_timeout_details, format_timeout_details
 from pymongo.response import PinnedResponse, Response
 
 if TYPE_CHECKING:
@@ -225,11 +224,7 @@ class Server:
             if use_cmd:
                 first = docs[0]
                 await operation.client._process_response(first, operation.session)  # type: ignore[misc, arg-type]
-                # Append timeout details to MaxTimeMSExpired responses.
-                if first.get("code") == 50:
-                    timeout_details = _get_timeout_details(conn.opts)  # type:ignore[has-type]
-                    first["errmsg"] += format_timeout_details(timeout_details)  # type:ignore[index]
-                _check_command_response(first, conn.max_wire_version)
+                _check_command_response(first, conn.max_wire_version, pool_opts=conn.opts)  # type:ignore[has-type]
         except Exception as exc:
             duration = datetime.now() - start
             if isinstance(exc, (NotPrimaryError, OperationFailure)):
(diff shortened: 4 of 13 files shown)


blink1073:
I realized the benchmark test wasn't actually triggering the protocol -> I'm tweaking things locally

blink1073:
This has a couple commits from #2467

blink1073:
I still need to make a PR for the fix to the kms mock server's 404 response.

pymongo/network_layer.py
```diff
@@ -670,34 +476,145 @@ def _resolve_pending_messages(self, exc: Optional[Exception] = None) -> None:
...
+    def get_buffer(self, sizehint: int) -> memoryview:
+        """Called to allocate a new receive buffer.
+        The asyncio loop calls this method expecting to receive a non-empty buffer to fill with data.
+        If any data does not fit into the returned buffer, this method will be called again until
+        either no data remains or an empty buffer is returned.
+        """
+        # Reuse the active buffer if it has space.
+        if len(self._buffers):
+            buffer = self._buffers[-1]
+            if len(buffer.buffer) - buffer.end_index > sizehint:
```
NoahStapp: If `sizehint = -1`, which signals that the buffer size can be arbitrary, this check will always succeed, potentially returning an empty buffer, which is an error. We need to check that sizehint is a positive number as well.

pymongo/network_layer.py
```diff
@@ -670,34 +476,145 @@ def _resolve_pending_messages(self, exc: Optional[Exception] = None) -> None:
...
+        self._pending_reads: collections.deque[int] = collections.deque()
+        self._pending_listeners: collections.deque[Future[Any]] = collections.deque()
+
+    def connection_made(self, transport: BaseTransport) -> None:
+        """Called exactly once when a connection is made.
+        The transport argument is the transport representing the write side of the connection.
+        """
+        self.transport = transport  # type: ignore[assignment]
+
+    async def read(self, bytes_needed: int) -> bytes:
```
NoahStapp, line 508: To make sure I understand the intended flow here, is this example correct?

1. We call `kms_request` and enter the `while kms_context.bytes_needed > 0:` loop.
2. The first chunk of data, say 16 bytes worth is written into an existing buffer that still has space.
3. `PyMongoKMSProtocol.read()` is called and immediately returns those 10 bytes.
4. `kms_context.bytes_needed` updates to need 84 more bytes for a total of 100.
5. We call `PyMongoKMSProtocol.read()` again and wait on the `_pending_listeners` Future we create.
6. The second chunk of data, the remaining 84 bytes, requires a new buffer since the active buffer is full.
7. The Future is resolved with those bytes, which we return and feed into `kms_context` to complete the operation.

blink1073: - We call kms_request and enter the while kms_context.bytes_needed > 0: loop.
- The first chunk of data, say 16 bytes worth is written into an existing buffer that still has space.
- PyMongoKMSProtocol.read() is called and immediately returns those 10 bytes, pushing the start_index up by 10
- kms_context.bytes_needed updates to need 84 more bytes
- We call PyMongoKMSProtocol.read() again and wait on the _pending_listeners Future we created.
- The second chunk of data, the remaining 84 bytes, *may* require a new buffer
- If any bytes are available, we read in up to the newly requested 84 bytes from the active buffer(s), advancing `start_index` and exhausting buffers as appropriate
- Otherwise, we wait on the future to be resolved, which will contain up to the requested bytes.

pymongo/network_layer.py
blink1073: Updated

pymongo/network_layer.py
```diff
@@ -670,34 +476,146 @@ def _resolve_pending_messages(self, exc: Optional[Exception] = None) -> None:
...
+        self._bytes_ready += nbytes
+
+        # Update the length of the current buffer.
+        self._buffers[-1].end_index += nbytes
+
+        if not len(self._pending_reads):
+            return
 
-async def async_sendall(conn: PyMongoProtocol, buf: bytes) -> None:
+        bytes_needed = self._pending_reads.popleft()
```
NoahStapp: What happens if we need more bytes than we have? We've already popped the waiter and set it's result to `data`, which can only read up to `self._bytes_ready` bytes. Are we relying on the `kms_context.bytes_needed` loop to call the protocol `read()` method again and create a new waiter?

NoahStapp (changes requested):

pymongo/network_layer.py
blink1073: Right, we give the partial result back to the kms context, and let it ask for more.

NoahStapp: _Maybe_ we could get better performance by doing more of the looping inside the Protocol, but KMS requests won't be a significant part of runtime anyway so not worth spending more time on it. Can you add a comment to this effect somewhere saying that we rely on the looping behavior for this to function correctly?

blink1073: It's not really a question of perf, but the fact that the kms_request is blind until it knows the Content-Length, and we don't know what state it is in.

blink1073: I added a comment.

test/asynchronous/test_collection.py
```diff
@@ -335,6 +335,8 @@ async def test_create_index(self):
         await db.test.create_index(["hello", ("world", DESCENDING)])
         await db.test.create_index({"hello": 1}.items())  # type:ignore[arg-type]
 
+    # TODO: PYTHON-5491 - remove version max
```
NoahStapp: This change should be in a separate PR.

pymongo/network_layer.py
NoahStapp: If we're setting `sizehint` to be at least `16384` always, is this check worth doing in the first place? I'd expect us to rarely reuse the active buffer since we'll usually have a buffer of size `16384` and a sizehint of `16384`.

blink1073:
> If we're setting sizehint to be at least 16384 always, is this check worth doing in the first place? I'd expect us to rarely reuse the active buffer since we'll usually have a buffer of size 16384 and a sizehint of 16384.

The actual sizehint in practice was on the order of the bytes being read from the buffer (typically less than 1000).  Using the buffered protocol at all here is a bit of a mismatch imho.

test/asynchronous/test_collection.py
blink1073: Yeah it was, I just updated this branch.

blink1073:
> How long would refactoring to not use buffered take? No reason to use the lower-level API if we don't need to.

It's actually dead simple, I did it along the way when I was debugging a race condition.

blink1073:
> It's actually dead simple, I did it along the way when I was debugging a race condition.

I'll push a commit in the morning for comparison, we can always revert.

blink1073:
I'm happy with the simplification.  The tests are passing locally, this is ready for another look.

NoahStapp (commented):
Can you schedule a full Evergreen run?

blink1073:
Full patch build: https://spruce.mongodb.com/version/689f5483e112170007b0ce9f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I updated the timings in the PR description, there no significant change.

pymongo/synchronous/pool.py
```diff
@@ -124,7 +124,89 @@ def _set_non_inheritable_non_atomic(fd: int) -> None:  # noqa: ARG001
 _IS_SYNC = True
 
 
-class Connection:
+class BaseConnection:
```
NoahStapp, line 126: We don't utilize any of the `BaseConnection` abstractions for the sync API, correct? The change is purely for compatibility with the async API?

blink1073: No it is used by both sync and async KMS, it is what gets returned by  the `_connect_kms` helper function.

NoahStapp: I meant that we don't have a separate KMS networking interface for sync. Just wanted to confirm my understanding.

blink1073:
I had to fix merge conflicts

NoahStapp (approved):

PR #2461, opened 2025-08-06
PYTHON-5488 `append_metadata` should not add duplicates

sleepyStick:
(not sure if this was urgent or not, but it was a relatively quick fix and i was worried this would be blocking the langchain PR)

### DIFF
--- a/.evergreen/scripts/install-dependencies.sh
+++ b/.evergreen/scripts/install-dependencies.sh
@@ -48,9 +48,12 @@ if ! command -v just &>/dev/null; then
     _TARGET="--target x86_64-pc-windows-msvc"
   fi
   _BIN_DIR=$PYMONGO_BIN_DIR
+  mkdir -p ${_BIN_DIR}
   echo "Installing just..."
   mkdir -p "$_BIN_DIR" 2>/dev/null || true
   curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- $_TARGET --to "$_BIN_DIR" || {
+    # Remove just file if it exists (can be created if there was an install error).
+    rm -f ${_BIN_DIR}/just
     _pip_install rust-just just
   }
   echo "Installing just... done."
@@ -59,6 +62,7 @@ fi
 # Ensure uv is installed.
 if ! command -v uv &>/dev/null; then
   _BIN_DIR=$PYMONGO_BIN_DIR
+  mkdir -p ${_BIN_DIR}
   echo "Installing uv..."
   # On most systems we can install directly.
   curl -LsSf https://astral.sh/uv/install.sh | env UV_INSTALL_DIR="$_BIN_DIR" INSTALLER_NO_MODIFY_PATH=1 sh || {
--- a/.evergreen/scripts/kms_tester.py
+++ b/.evergreen/scripts/kms_tester.py
@@ -30,7 +30,7 @@ def _setup_azure_vm(base_env: dict[str, str]) -> None:
     env["AZUREKMS_CMD"] = "tar xf mongo-python-driver.tgz"
     run_command(f"{azure_dir}/run-command.sh", env=env)
 
-    env["AZUREKMS_CMD"] = "bash .evergreen/just.sh setup-tests kms azure-remote"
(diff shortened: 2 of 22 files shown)


pymongo/pool_options.py
```diff
@@ -386,6 +386,8 @@ def __init__(
 
     def _update_metadata(self, driver: DriverInfo) -> None:
         """Updates the client's metadata"""
+        if driver.name in self.__metadata["driver"]["name"].split("|"):
+            return
```
Jibola: I'm thinking we should be case-insensitive. "LangChain" vs. "langchain" should not make a difference. 

I cannot remember if we already lowercase everything in name (I don't think we do).

Jibola (changes requested):

pymongo/pool_options.py
sleepyStick: good call, changed!

blink1073:
Another thought, should we be adding a `|` if there is no version or platform, so that the fields always align?

pymongo/pool_options.py
```diff
@@ -386,6 +386,8 @@ def __init__(
 
     def _update_metadata(self, driver: DriverInfo) -> None:
         """Updates the client's metadata"""
+        if driver.name.lower() in self.__metadata["driver"]["name"].lower().split("|"):
```
NoahStapp: Need to check for `driver.name` being `None` before calling `lower()`.

test/asynchronous/test_client_metadata.py
```diff
@@ -210,6 +210,46 @@ async def test_doesnt_update_established_connections(self):
         self.assertIsNone(self.handshake_req)
         self.assertEqual(listener.event_count(ConnectionClosedEvent), 0)
 
+    async def test_duplicate_driver_name_no_op(self):
```
NoahStapp, line 218: Can we refactor `check_metadata_added` to support checking that the new metadata is the same as the old metadata? Most of this method is just duplicate code from `check_metadata_added`.

NoahStapp:
> If I add DriverInfo with no `version`, and then add one with `version`, how can we distinguish which `name` the `version` belongs to?

Good point. Always adding "|" to each field, even if it's not present, seems like a good fix.

pymongo/pool_options.py
sleepyStick: the type of DriverInfo.name shouldn't be none. Would the check here be enforcing that the caller is following that rule?

NoahStapp: We check on line 393 that `driver.name` exists, so extending the same check here makes sense. We don't explicitly prevent `driver.name` from being None, so I'd prefer to err on the safe side to prevent errors being thrown by misbehaving users.

pymongo/pool_options.py
```diff
@@ -386,20 +386,26 @@ def __init__(
...
-            )
-        if driver.version:
-            metadata["driver"]["version"] = "{}|{}".format(
-                metadata["driver"]["version"],
-                driver.version,
-            )
-        if driver.platform:
-            metadata["platform"] = "{}|{}".format(metadata["platform"], driver.platform)
+
+        name = driver.name if driver.name else ""
```
blink1073: These can be simplified to `driver.xxx or ""`

blink1073:
drivers-pr-bot please backport to v4.14

blink1073 (approved):
LGTM!

pymongo/pool_options.py
```diff
@@ -386,20 +386,26 @@ def __init__(
 
     def _update_metadata(self, driver: DriverInfo) -> None:
         """Updates the client's metadata"""
+        if driver.name and driver.name.lower() in self.__metadata["driver"]["name"].lower().split(
```
NoahStapp, line 389: Minor nit: we don't need the `split()` call here. Does having it improve readability?

sleepyStick: I was mainly worried about potential substrings? like if driver.name is a substring of an existing name, its not the same and therefore still should be added?

NoahStapp: Ah good point! Great find on that edge case, better to be safe here and use the split!

PR #2477, opened 2025-08-15
PYTHON-5492 Fix handling of MaxTimeMSExpired responses

blink1073:
It turns out we did have a race condition as to whether we'd get a client side or server side timeout.  This ensures we get the same information either way.

Also fixes a sporadic bug when installing `just` by ensuring that the bin dir exists before attempting to install into it.

### DIFF
--- a/pymongo/asynchronous/server.py
+++ b/pymongo/asynchronous/server.py
@@ -38,6 +38,7 @@ from pymongo.logger import (
     _SDAMStatusMessage,
 )
 from pymongo.message import _convert_exception, _GetMore, _OpMsg, _Query
+from pymongo.pool_shared import _get_timeout_details, format_timeout_details
 from pymongo.response import PinnedResponse, Response
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


blink1073:
drivers-pr-bot please backport to v4.14

sleepyStick (approved):

pymongo/asynchronous/server.py
```diff
@@ -224,6 +225,10 @@ async def run_operation(
             if use_cmd:
                 first = docs[0]
                 await operation.client._process_response(first, operation.session)  # type: ignore[misc, arg-type]
+                # Append timeout details to MaxTimeMSExpired responses.
+                if first.get("code") == 50:
+                    timeout_details = _get_timeout_details(conn.opts)  # type:ignore[has-type]
+                    first["errmsg"] += format_timeout_details(timeout_details)  # type:ignore[index]
```
ShaneHarvey, line 231: This will fail if the document_class is RawBSONDocument:
```python
>>> from bson.raw_bson import *
>>> c = MongoClient(document_class=RawBSONDocument)
>>> _ = c.t.t.insert_many([{'x':i} for i in range(1000000)])
>>> c.t.t.find_one({'x': -1}, maxTimeMS=1)
...
  File "/Users/shane/git/mongo-python-driver/pymongo/synchronous/server.py", line 231, in run_operation
    first["errmsg"] += format_timeout_details(timeout_details)  # type:ignore[index]
    ~~~~~^^^^^^^^^^
TypeError: 'RawBSONDocument' object does not support item assignment
```

PR #2488, opened 2025-08-19
PYTHON-5143 Support auto encryption in unified tests

blink1073:
- [x] Sync spec tests
- [x] Implement unified test format version 1.23
- [x] Support autoEncryptOpts
- [x] Drop QE state collections (_enxcol.<coll>.esc and _enxcol.<coll>.ecoc)

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">=3.9"
 resolution-markers = [
     "python_full_version == '3.14.*'",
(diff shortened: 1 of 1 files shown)


test/asynchronous/unified_format.py
```diff
@@ -488,6 +515,7 @@ def tearDownClass(cls) -> None:
     async def asyncSetUp(self):
         # super call creates internal client cls.client
         await super().asyncSetUp()
+
```
NoahStapp: Unintended whitespace?

test/unified_format_shared.py
```diff
@@ -118,10 +119,22 @@
     ("kmip", KMIP_CREDS),
     ("kmip:name1", KMIP_CREDS),
 ]:
+    # Use the temp aws creds for autoEncryptOpts.
```
NoahStapp, line 122: Can you explain why we do this specifically for AWS?

aclark4life (approved):
LGTM and looks helpful for https://github.com/mongodb/django-mongodb-backend/pull/329

test/asynchronous/helpers.py
```diff
@@ -115,6 +120,20 @@
     "privateKey": os.environ.get("FLE_GCP_PRIVATEKEY", ""),
 }
 KMIP_CREDS = {"endpoint": os.environ.get("FLE_KMIP_ENDPOINT", "localhost:5698")}
+AWS_TEMP_CREDS = {
+    "accessKeyId": os.environ.get("CSFLE_AWS_TEMP_ACCESS_KEY_ID", ""),
+    "secretAccessKey": os.environ.get("CSFLE_AWS_TEMP_SECRET_ACCESS_KEY", ""),
+    "sessionToken": os.environ.get("CSFLE_AWS_TEMP_SESSION_TOKEN", ""),
+}
+
+ALL_KMS_PROVIDERS = dict(
```
NoahStapp: Do these need to be here where they'll be duplicated by synchro? Or can they go in a shared file in `test`?

NoahStapp (changes requested):

test/asynchronous/helpers.py
blink1073: Are you suggesting we move the rest of the constants as well?

test/unified_format_shared.py
blink1073: For clientEncryptionOpts, the tests are using the non-temp credentials, and for autoEncryptOpts they're using the temp credentials.  Ideally they'd be the same, but other teams have already implemented.

test/asynchronous/unified_format.py
blink1073: Updated

test/asynchronous/helpers.py
NoahStapp: Any constant (or function, value, anything really) that doesn't change based on async/sync should live outside of the synchro'd directories, yeah.

test/unified_format_shared.py
NoahStapp: For every provider? Or only for AWS tests?

blink1073: AWS is the only provider that has temp credentials as an option

test/asynchronous/helpers.py
blink1073: Okay, I'll refactor a bit

test/unified_format_shared.py
NoahStapp: Got it, makes sense.

test/asynchronous/helpers.py
blink1073: Updated, let's see how the tests do

blink1073:
Okay, this is ready for review again

PR #2491, opened 2025-08-19
PYTHON-5395 - Add convert_decimal to CodecOptions

NoahStapp:
### Context
We currently reject Python `decimal.Decimal` instances and require the user to manually convert to Decimal128.

This PR adds a new `convert_decimal` kwarg to CodecOptions that enables automatic conversion of `decimal.Decimal` to `Decimal128` during BSON encoding.

### Summary and example of changes
Current behavior and new behavior when `convert_decimal=False` (default)
```
>>> bson.encode({"d": decimal.Decimal('1.0')})
  File "<python-input-7>", line 1, in <module>
    encode({"d": decimal.Decimal('1.0')})
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 1053, in encode
bson.errors.InvalidDocument: Invalid document {'d': Decimal('1.0')} | cannot encode object: Decimal('1.0'), of type: <class 'decimal.Decimal'>
```

New behavior when `convert_decimal=True`
```
>>> bson.encode({"d": decimal.Decimal('1.0')}, codec_options=CodecOptions(convert_decimal=True))
b'\x18\x00\x00\x00\x13d\x00\n\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00>0\x00'
```
### Testing
`test/test_bson.py` has a new test `TestCodecOptions.test_convert_decimal` to verify correctness.

A changelog entry will be added in this PR if we determine that this is the correct approach to move ahead with.

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -919,6 +919,7 @@ for _typ in _ENCODERS:
     if hasattr(_typ, "_type_marker"):
         _MARKERS[_typ._type_marker] = _ENCODERS[_typ]
 
+
 _BUILT_IN_TYPES = tuple(t for t in _ENCODERS)
 
 
@@ -1030,8 +1031,8 @@ def _dict_to_bson(
     return _PACK_INT(len(encoded) + 5) + encoded + b"\x00"
 
 
-# if _USE_C:
-#     _dict_to_bson = _cbson._dict_to_bson
+if _USE_C:
+    _dict_to_bson = _cbson._dict_to_bson
 
 
 _CODEC_OPTIONS_TYPE_ERROR = TypeError("codec_options must be an instance of CodecOptions")


NoahStapp:
Will update the tests to reflect the change later today.

bson/codec_options.py
```diff
@@ -362,6 +365,9 @@ def __init__(self, *args, **kwargs):
                 return DatetimeMS objects when the underlying datetime is
                 out-of-range and 'datetime_clamp' to clamp to the minimum and
                 maximum possible datetimes. Defaults to 'datetime'.
+            :param convert_decimal: If ``True``, instances of :class:`~decimal.Decimal` will
+            be automatically converted to :class:`~bson.decimal128.Decimal128` when encoding to BSON.
+            Defaults to ``False``.
```
blink1073: This needs a versionadded entry

NoahStapp: Added for 4.15.

bson/__init__.py
```diff
@@ -1017,8 +1030,8 @@ def _dict_to_bson(
     return _PACK_INT(len(encoded) + 5) + encoded + b"\x00"
 
 
-if _USE_C:
-    _dict_to_bson = _cbson._dict_to_bson
+# if _USE_C:
```
sleepyStick: is this comment intentionally left here?

bson/__init__.py
```diff
@@ -908,7 +919,6 @@ def _encode_maxkey(name: bytes, dummy0: Any, dummy1: Any, dummy2: Any) -> bytes:
     if hasattr(_typ, "_type_marker"):
         _MARKERS[_typ._type_marker] = _ENCODERS[_typ]
 
-
```
sleepyStick: minor comment but deleted new line?

bson/__init__.py
NoahStapp: Good catch, holdover from local testing, fixed.

bson/__init__.py
```diff
@@ -858,6 +859,16 @@ def _encode_decimal128(name: bytes, value: Decimal128, dummy0: Any, dummy1: Any)
     return b"\x13" + name + value.bid
 
 
+def _encode_python_decimal(
+    name: bytes, value: decimal.Decimal, dummy0: Any, opts: CodecOptions[Any]
```
aclark4life, line 863: `dummy0` is here because … why is `dummy0` here? I see that it is in a lot of places but I'm not sure why.
```
(.venv) alex.clark@M-FD7GQ2QTFM django-mongodb-cli % grep -r dummy0 src/mongo-python-driver                                                                                      1 ↵
src/mongo-python-driver/bson/__init__.py:    data: Any, _view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
src/mongo-python-driver/bson/__init__.py:    data: Any, _view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
src/mongo-python-driver/bson/__init__.py:    data: Any, _view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
src/mongo-python-driver/bson/__init__.py:    data: Any, _view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
src/mongo-python-driver/bson/__init__.py:    data: Any, _view: Any, position: int, dummy0: int, opts: CodecOptions[Any], dummy1: Any
src/mongo-python-driver/bson/__init__.py:    data: Any, view: Any, position: int, dummy0: Any, opts: CodecOptions[Any], dummy1: Any
src/mongo-python-driver/bson/__init__.py:    data: Any, _view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
src/mongo-python-driver/bson/__init__.py:    data: Any, _view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
src/mongo-python-driver/bson/__init__.py:    data: Any, _view: Any, position: int, dummy0: Any, dummy1: Any, dummy2: Any
src/mongo-python-driver/bson/__init__.py:def _encode_float(name: bytes, value: float, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_bytes(name: bytes, value: bytes, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_text(name: bytes, value: str, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_binary(name: bytes, value: Binary, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_bool(name: bytes, value: bool, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_datetime(name: bytes, value: datetime.datetime, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_datetime_ms(name: bytes, value: DatetimeMS, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_none(name: bytes, dummy0: Any, dummy1: Any, dummy2: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_regex(name: bytes, value: Regex[Any], dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_int(name: bytes, value: int, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_timestamp(name: bytes, value: Any, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_long(name: bytes, value: Any, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_decimal128(name: bytes, value: Decimal128, dummy0: Any, dummy1: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_minkey(name: bytes, dummy0: Any, dummy1: Any, dummy2: Any) -> bytes:
src/mongo-python-driver/bson/__init__.py:def _encode_maxkey(name: bytes, dummy0: Any, dummy1: Any, dummy2: Any) -> bytes:
src/mongo-python-driver/bson/json_util.py:def _parse_legacy_regex(doc: Any, dummy0: Any) -> Any:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_oid(doc: Any, dummy0: Any) -> ObjectId:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_symbol(doc: Any, dummy0: Any) -> str:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_code(doc: Any, dummy0: Any) -> Code:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_regex(doc: Any, dummy0: Any) -> Regex[str]:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_dbref(doc: Any, dummy0: Any) -> Any:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_dbpointer(doc: Any, dummy0: Any) -> Any:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_int32(doc: Any, dummy0: Any) -> int:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_int64(doc: Any, dummy0: Any) -> Int64:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_double(doc: Any, dummy0: Any) -> float:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_decimal128(doc: Any, dummy0: Any) -> Decimal128:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_minkey(doc: Any, dummy0: Any) -> MinKey:
src/mongo-python-driver/bson/json_util.py:def _parse_canonical_maxkey(doc: Any, dummy0: Any) -> MaxKey:
src/mongo-python-driver/bson/json_util.py:def _parse_timestamp(doc: Any, dummy0: Any) -> Timestamp:
src/mongo-python-driver/bson/json_util.py:def _encode_noop(obj: Any, dummy0: Any) -> Any:
src/mongo-python-driver/bson/json_util.py:def _encode_objectid(obj: ObjectId, dummy0: Any) -> dict:  # type: ignore[type-arg]
src/mongo-python-driver/bson/json_util.py:def _encode_timestamp(obj: Timestamp, dummy0: Any) -> dict:  # type: ignore[type-arg]
src/mongo-python-driver/bson/json_util.py:def _encode_decimal128(obj: Timestamp, dummy0: Any) -> dict:  # type: ignore[type-arg]
src/mongo-python-driver/bson/json_util.py:def _encode_minkey(dummy0: Any, dummy1: Any) -> dict:  # type: ignore[type-arg]
src/mongo-python-driver/bson/json_util.py:def _encode_maxkey(dummy0: Any, dummy1: Any) -> dict:  # type: ignore[type-arg]
src/mongo-python-driver/pymongo/message.py:        self, dummy0: Any, conn: _AgnosticConnection, use_cmd: bool = False
```

bson/__init__.py
aclark4life: intentional?

bson/__init__.py
NoahStapp: `dummy0` and `dummy1` are used to ensure consistency of function parameter shape across groups of methods that need to have identical signatures even if they don't need all of the arguments.  The `_encode_...` BSON methods all use this same pattern.

bson/__init__.py
NoahStapp: Resolved.

bson/__init__.py
aclark4life: Ah, thanks!

bson/_cbsonmodule.c
```diff
@@ -791,14 +812,15 @@ int convert_codec_options(PyObject* self, PyObject* options_obj, codec_options_t
 
     options->unicode_decode_error_handler = NULL;
 
-    if (!PyArg_ParseTuple(options_obj, "ObbzOOb",
+    if (!PyArg_ParseTuple(options_obj, "ObbzOObb",
```
aclark4life, line 815: I stared at this for a while but then found: https://docs.python.org/3/c-api/arg.html. So, just confirming that's a format string for the args below.

sleepyStick: lol i was sitting on this one too, googled it, found that same site, was still confused / overwhelemed by words so then i asked mongo-gpt HAHA

bson/_cbsonmodule.c
```diff
@@ -1436,6 +1462,16 @@ static int _write_element_to_buffer(PyObject* self, buffer_t buffer,
                                           in_fallback_call);
         Py_DECREF(binary_value);
         return result;
+    } else if (options->convert_decimal && _check_decimal(value)) {
```
aclark4life, line 1465: So after case 100 or so and if case 19 wasn't the case, auto-convert ?

bson/_cbsonmodule.c
NoahStapp: Yup this is a format string.

bson/_cbsonmodule.c
NoahStapp: For consistency with how the rest of `_write_element_to_buffer` works, we check the converison edge cases such as this after the native BSON `case` statement.

bson/__init__.py
```diff
@@ -858,6 +859,16 @@ def _encode_decimal128(name: bytes, value: Decimal128, dummy0: Any, dummy1: Any)
...
+) -> bytes:
+    if opts.convert_decimal:
+        converted = Decimal128(value)
+        return b"\x13" + name + converted.bid
+    else:
+        raise InvalidDocument("decimal.Decimal must be converted to bson.decimal128.Decimal128.")
```
ShaneHarvey, line 869: Suggest mentioning the convert_decimal option in this error message.

ShaneHarvey (changes requested):
This PR doesn't add new functionality since users can already configure a CodecOptions that automatically encodes Decimal, for example:
```python
class DecimalEncoder(TypeEncoder):
    @property
    def python_type(self):
        return Decimal

    def transform_python(self, value):
        return Decimal128(value)


DECIMAL_CODECOPTS = CodecOptions(type_registry=TypeRegistry([DecimalEncoder()]))
```
TypeRegistry can also be configured to auto-decode Decimal128 to decimal by implementing TypeDecoder.

Given this, is it necessary to add this `convert_decimal` flag?

NoahStapp:
> This PR doesn't add new functionality since users can already configure a CodecOptions that automatically encodes Decimal, for example:
> 
> ```python
> class DecimalEncoder(TypeEncoder):
>     @property
>     def python_type(self):
>         return Decimal
> 
>     def transform_python(self, value):
>         return Decimal128(value)
> 
> 
> DECIMAL_CODECOPTS = CodecOptions(type_registry=TypeRegistry([DecimalEncoder()]))
> ```
> 
> TypeRegistry can also be configured to auto-decode Decimal128 to decimal by implementing TypeDecoder.
> 
> Given this, is it necessary to add this `convert_decimal` flag?

The `convert_decimal` flag is significantly less friction than configuring their own CodecOptions with an Encoder. Especially since practically every custom TypeEncoder to do so will be identical. If that's the case, we should do the conversion automatically ourselves.

I'm fine to make this auto-conversion the default instead of adding a flag, but then we would explicitly disallow users from making their own custom Encoders for `decimal.Decimal`.

PR #2497, opened 2025-08-19
PYTHON-5505 Prototype system overload retry loop for all operations

ShaneHarvey:
PYTHON-5505 Prototype system overload retry loop for all operations

Description:
All operations that fail with the "Retryable" error label will be retried up to 3 times. When the last failure includes the "SystemOverloaded" error label we apply exponential backoff with jitter before attempting a retry.

PR Tasks:
- [X] Validate command path
- [X] Validate retryable write path
- [X] Validate retryable write path with an otherwise non-retryable operation (eg update_many).
- [X] Validate retryable read path
- [x] Validate cursor getMore path

Passing patch build: https://spruce.mongodb.com/version/68a64f28268fea00074a72a4/tasks

### DIFF
--- a/pymongo/asynchronous/mongo_client.py
+++ b/pymongo/asynchronous/mongo_client.py
@@ -2915,7 +2915,7 @@ class _ClientConnectionRetryable(Generic[T]):
                     and conn.supports_sessions
                 )
                 is_mongos = conn.is_mongos
-                if not sessions_supported:
+                if not self._always_retryable and not sessions_supported:
                     # A retry is not possible because this server does
                     # not support sessions raise the last error.
                     self._check_last_error()
--- a/pymongo/synchronous/mongo_client.py
+++ b/pymongo/synchronous/mongo_client.py
@@ -2905,7 +2905,7 @@ class _ClientConnectionRetryable(Generic[T]):
                     and conn.supports_sessions
                 )
                 is_mongos = conn.is_mongos
-                if not sessions_supported:
+                if not self._always_retryable and not sessions_supported:
                     # A retry is not possible because this server does
                     # not support sessions raise the last error.
                     self._check_last_error()


test/asynchronous/test_backpressure.py
```diff
@@ -0,0 +1,155 @@
...
+        "errorLabels": ["Retryable"],
+    },
+}
+
+
+class TestBackpressure(AsyncIntegrationTest):
+    RUN_ON_LOAD_BALANCER = True
+
+    @async_client_context.require_failCommand_appName
+    async def test_retry_overload_error_command(self):
```
ShaneHarvey, line 45: Note these scenarios will be later turned into unified spec tests.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2783,14 +2786,19 @@ async def run(self) -> T:
                 # most likely be a waste of time.
                 raise
             except PyMongoError as exc:
+                overload = False
```
NoahStapp: For my own understanding: what's the purpose of this `overload` value? It's `True` when the error is an overload error?

sleepyStick: that's my understanding of it -- and then when `overload` is true and we've reached `max_retries`, we raise the error, otherwise we `backoff` and try again. (I think) 

And after this initial attempt, we only ever set `overload` to `exc.error.has_error_label("Retryable")`, is there a reason why we cant just set it to that here? 

(side note: I like how `_overload` and `_retrying` are the same length, but I think `_overloaded` is a slightly more accurate var? i don't think this matters that much though since I feel like `_overload` does get the point across for me)

NoahStapp: I think `_overload` is only set when the error is a result of the server being overloaded. If so, `_overloaded` is a clearer variable name.

ShaneHarvey: That's all correct. I've renamed it `overloaded`.

ShaneHarvey:
test_retry_overload_error_update_many fails on actions but passed locally. I'm investigating...
```python
____________ TestBackpressure.test_retry_overload_error_update_many ____________

self = <test.asynchronous.test_backpressure.TestBackpressure testMethod=test_retry_overload_error_update_many>

    @async_client_context.require_failCommand_appName
    async def test_retry_overload_error_update_many(self):
        # Even though update_many is not a retryable write operation, it will
        # still be retried via the "Retryable" error label.
        await self.db.t.insert_one({"x": 1})
    
        # Ensure command is retried on overload error.
        fail_once = mock_overload_error.copy()
        fail_once["mode"] = {"times": _MAX_RETRIES}
        async with self.fail_point(fail_once):
>           await self.db.t.update_many({}, {"$set": {"x": 2}})
```
https://github.com/mongodb/mongo-python-driver/actions/runs/17106265661/job/48515851763?pr=2497

pymongo/synchronous/mongo_client.py
```diff
@@ -2773,14 +2776,19 @@ def run(self) -> T:
...
+                overloaded = False
                 # Execute specialized catch on read
                 if self._is_read:
                     if isinstance(exc, (ConnectionFailure, OperationFailure)):
                         # ConnectionFailures do not supply a code property
                         exc_code = getattr(exc, "code", None)
-                        if self._is_not_eligible_for_retry() or (
-                            isinstance(exc, OperationFailure)
-                            and exc_code not in helpers_shared._RETRYABLE_ERROR_CODES
+                        overloaded = exc.has_error_label("Retryable")
```
gracicot: Wouldn't it be more appropriate to mark overloaded when the error label `SystemOverloadedError` is present? See [this line](https://github.com/10gen/mongo/blob/8536c690f50947f2fbc41049d00fe64e39e982c3/src/mongo/db/error_labels.h#L55).Today our rate limiter is the only place where we use `Retryable` error label, but there might be more use cases in the future. I think `SystemOverloadedError` would convey the meaning better.

ShaneHarvey: IIRC last I saw the design we only had "Retryable" and not "SystemOverloadedError". I see we have both labels now. So the correct logic is to retry IFF "Retryable" is present and apply backoff IFF "SystemOverloadedError" is present?

Note: this PR doesn't address connection establishment errors due to the ingress connection rate limiter. That will come later.

gracicot: Ah! Yes. We do add this error label to provide a way for the server to report operations as being idempotent. The `SystemOverloadedError` was actually added in [SPM-3784](https://jira.mongodb.org/browse/SPM-3784) and implemented in the [SERVER-96975](https://jira.mongodb.org/browse/SERVER-96975) ticket.

> So the correct logic is to retry IFF "Retryable" is present and apply backoff IFF "SystemOverloadedError" is present?

Yes. In the design document of SPM-4003, it defines both DefaultRetryStrategy and AdaptiveRetryStrategy to use SystemOverloadedError for overloading specific behaviour. The request rate limiter (not to be confused with the ingress connection rate limiter) already return an error with this label.

ShaneHarvey: Thanks. Fixed.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: Update: this was renamed "always_retryable". "overloaded" is now based on the "SystemOverloaded" error label.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2801,18 +2811,24 @@ async def run(self) -> T:
...
                 # Specialized catch on write operation
                 if not self._is_read:
-                    if not self._retryable:
-                        raise
+                    retryable_write_label = False
                     if isinstance(exc, ClientBulkWriteException) and exc.error:
-                        retryable_write_error_exc = isinstance(
-                            exc.error, PyMongoError
-                        ) and exc.error.has_error_label("RetryableWriteError")
+                        if isinstance(exc.error, PyMongoError):
```
NoahStapp: Do we need this `if-else` block? Both branches are identical.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2783,14 +2786,21 @@ async def run(self) -> T:
                 # most likely be a waste of time.
                 raise
             except PyMongoError as exc:
+                always_retryable = False
```
NoahStapp, line 2789: `always_retryable` represents what exactly? That an error is retryable under all conditions?

NoahStapp (changes requested):

pymongo/asynchronous/mongo_client.py
ShaneHarvey: Yes it means it's safe to retry even if the operation is otherwise non-retryable, like an update_many. There's probably a better name for it.

pymongo/asynchronous/mongo_client.py
ShaneHarvey: Yeah this can be cleaned up. I this it's a problem with ClientBulkWriteException violating Liskov substitution principle. Perhaps ClientBulkWriteException should override has_error_label() to behave correctly. I suspect the NoWritesPerformed check below is buggy because of this issue too.

ShaneHarvey: I opened: https://jira.mongodb.org/browse/PYTHON-5511

ShaneHarvey: Refactored to remove the duplicate label checks. I decided not to do PYTHON-5511 yet to reduce the scope of this PR.

pymongo/asynchronous/mongo_client.py
NoahStapp: `overloaded` still takes precedence over `always_retryable`, correct?

NoahStapp:
> Found the bug causing the update_many test to fail. This was only happening on standalone nodes which don't support retryable writes. The fix is to skip the check for `sessions_supported` when attempting to retry after a system overload error. I believe this is safe.
> 
> Passing patch build: [spruce.mongodb.com/version/68a64f28268fea00074a72a4/tasks](https://spruce.mongodb.com/version/68a64f28268fea00074a72a4/tasks)

We can't retry writes on standalone anyway, so how does skipping the `sessions_supported` check fix that? Wouldn't that cause us to attempt a retry when we can't?

ShaneHarvey:
> We can't retry writes on standalone anyway, so how does skipping the sessions_supported check fix that? Wouldn't that cause us to attempt a retry when we can't?

This feature adds the ability to retry all commands that fail with the "Retryable" error lable, even on standalone. The server only adds this label when it's safe to retry (eg the command was rejected before doing any work).

PR #2499, opened 2025-08-20
PYTHON-5508 - Add built-in DecimalEncoder and DecimalDecoder

NoahStapp:


bson/decimal128.py
```diff
@@ -58,6 +61,38 @@
 _VALUE_OPTIONS = Union[decimal.Decimal, float, str, Tuple[int, Sequence[int], int]]
 
 
+class DecimalEncoder(TypeEncoder):
+    """Converts Python :class:`decimal.Decimal` to BSON :class:`Decimal128`.
+
+    .. warning:: When converting BSON data types to and from built-in data types,
+       the possibility of data loss is always present due to mismatches in underlying implementations.
```
ShaneHarvey: "possibility of data loss is always present" is extremely alarming. We have tests that do auto-convert and they don't have data loss issues so I think the language is too strong. Can we identify and explain exactly where the risk is and when it would occur?

doc/changelog.rst
```diff
@@ -1,5 +1,15 @@
 Changelog
 =========
+Changes in Version 4.15.0 (XXXX/XX/XX)
+--------------------------------------
+.. warning:: When converting BSON data types to and from built-in data types, the possibility of data loss is always present
+ due to mismatches in underlying implementations.
```
ShaneHarvey: I'm not sure we need this warning here too.

bson/decimal128.py
```diff
@@ -58,6 +61,38 @@
...
+
+    @property
+    def python_type(self) -> Type[Decimal]:
+        return Decimal
+
+    def transform_python(self, value: Any) -> Decimal128:
+        return Decimal128(value)
+
+
+class DecimalDecoder(TypeDecoder):
```
ShaneHarvey, line 82: Could we add a docstring example showing out to use these classes?

sleepyStick (approved):
LGTM!

bson/decimal128.py
NoahStapp: Decimal128 supports up to 34 decimal digits of precision, decimal.Decimal by default supports up to 28 digit during arithmetic operations. When creating a new decimal.Decimal value, that 28 digit limit does not exist, so I don't think we need any warning here.

ShaneHarvey: Yeah it seems like the truncation issue will be caught on the encoding side:
```python
>>> Decimal128(decimal.Decimal('1'*34))
Decimal128('1111111111111111111111111111111111')
>>> Decimal128(decimal.Decimal('1'*35))
  File "<python-input-6>", line 1, in <module>
    Decimal128(decimal.Decimal('1'*35))
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shane/git/mongo-python-driver/bson/decimal128.py", line 218, in __init__
    self.__high, self.__low = _decimal_to_128(value)
                              ~~~~~~~~~~~~~~~^^^^^^^
  File "/Users/shane/git/mongo-python-driver/bson/decimal128.py", line 76, in _decimal_to_128
    value = ctx.create_decimal(value)
decimal.Inexact: [<class 'decimal.Inexact'>]
```

sleepyStick (approved):

PR #2501, opened 2025-08-21
PYTHON-5506 Prototype adaptive token bucket retry

ShaneHarvey:
PYTHON-5506 Prototype adaptive token bucket retry

Description:
Add adaptive token bucket based retry policy.
Successfully completed commands deposit 0.1 token. 
Failed retry attempts consume 1 token. 
A retry is only permitted if there is an available token.
For simplicity, this POC adds a retry bucket per MongoClient. In the future we will consider adding a retry bucket per server.

Patch with perf benchmarks: https://spruce.mongodb.com/version/68a7b69b74baee000728de31/tasks

### DIFF
--- a/pymongo/asynchronous/helpers.py
+++ b/pymongo/asynchronous/helpers.py
@@ -141,7 +141,7 @@ class _RetryPolicy:
         self.backoff_initial = backoff_initial
         self.backoff_max = backoff_max
 
-    async def record_success(self, retry: bool):
+    async def record_success(self, retry: bool) -> None:
         """Record a successful operation."""
         await self.token_bucket.deposit(retry)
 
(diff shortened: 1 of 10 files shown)


pymongo/asynchronous/helpers.py
```diff
@@ -89,23 +93,95 @@ async def _backoff(
...
+        """Record a successful operation."""
+        await self.token_bucket.deposit(retry)
+
+    async def backoff(self, attempt: int) -> None:
+        """Return the backoff duration for the given ."""
+        await _backoff(max(0, attempt - 1), self.backoff_initial, self.backoff_max)
+
+    async def should_retry(self, attempt: int) -> bool:
+        """Return if we have budget to retry and how long to backoff."""
+        # TODO: Check CSOT deadline here.
```
blink1073: Is this for a follow up PR?

ShaneHarvey: Implemented it now.

PR #2502, opened 2025-08-22
PYTHON-5514 Specific assertions for "is" and "is not None"

sleepyStick:
(grouped them together because there wasn't a lot of each, but i can seperate the two into two different PRs if that's preferred) 

`self.assertTrue(x is y)` -> `self.assertIs(x, y)`
`self.assertTrue(x is not None)` -> `self.assertIsNotNone(x)`

### DIFF
--- a/test/asynchronous/test_collection.py
+++ b/test/asynchronous/test_collection.py
@@ -1319,7 +1319,7 @@ class AsyncTestCollection(AsyncIntegrationTest):
             self.assertIn(exc.code, (9, 10147, 16840, 17009))
             # Just check that we set the error document. Fields
             # vary by MongoDB version.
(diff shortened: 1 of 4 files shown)


aclark4life (approved):
LGTM! The only minor issue (nit, even) I notice is that "better" is subjective and what's actually changing is the assertions, from general (True/False) to specific (Is/IsNot) and that is objectively better 😂

sleepyStick:
> LGTM! The only minor issue (nit, even) I notice is that "better" is subjective and what's actually changing is the assertions, from general (True/False) to specific (Is/IsNot) and that is objectively better 😂

Good point! thanks for the name change!

PR #2504, opened 2025-08-25
fix deprecated functions

celaraze:
datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes.


ShaneHarvey:
This was already fixed in PyMongo 4.4 in [PYTHON-3702](https://jira.mongodb.org/browse/PYTHON-3702).

PR #2509, opened 2025-08-26
DRIVERS-3218 Avoid clearing the connection pool when the server connection rate limiter triggers

blink1073:


pymongo/asynchronous/pool.py
```diff
@@ -1395,6 +1412,20 @@ async def checkin(self, conn: AsyncConnection) -> None:
...
+                        # returns any other error, the connection is not usable, so return false.
+                        # If it doesn't return an error and actually reads data, the connection is
+                        # also not usable, so return false.
+                        conn.conn.get_conn.settimeout(0.001)
+                        close_conn = True
+                        try:
+                            conn.conn.get_conn.read(1)
+                        except Exception as _:
+                            # TODO: verify the exception
+                            close_conn = False
```
ShaneHarvey: 2 comments: 
1) I believe this logic needs to move to connection checkout. Here in connection check in we already know the connection is useable because we're checking it back in after a successful command.
2) Instead of a 1ms read can we reuse the existing _perished() + conn_closed() methods?

ShaneHarvey (commented):
Nice work!

pymongo/asynchronous/pool.py
blink1073: Done

pymongo/asynchronous/pool.py
```diff
@@ -1094,15 +1105,25 @@ async def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> A
...
+        except BaseException as e:
             async with self.lock:
                 self.active_contexts.discard(conn.cancel_context)
+            # Handle system overload condition.  When the base AutoReconnect is
+            # raised and we are not an sdam pool, add to backoff and add the
+            # appropriate error label.
+            if not self.is_sdam and type(e) == AutoReconnect:
+                self._backoff += 1
+                e._add_error_label("SystemOverloaded")
+                e._add_error_label("Retryable")
```
ShaneHarvey: We need to move this logic so that it covers the TCP+TLS handshake which happen up above.

blink1073: I set a breakpoint in the TCP+TLS handshake error handler and confirmed that handshakes are succeeding.  The error only occurs on hello/auth.

ShaneHarvey: Okay I'm actually surprised by this since the design [SPM-4319](https://jira.mongodb.org/browse/SPM-4319) indicates the rate limiter rejection happens before the TLS handshake.

pymongo/network_layer.py
```diff
@@ -338,8 +338,11 @@ async def read(self, request_id: Optional[int], max_message_size: int) -> tuple[
         if self._done_messages:
             message = await self._done_messages.popleft()
         else:
-            if self._closing_exception:
-                raise self._closing_exception
+            if self._closed.done():
```
NoahStapp, line 341: Is calling `is_closing` here better? It'll catch more edge cases in theory.

blink1073: Hmm let me try that.

blink1073: No, it is ambiguous as to whether `connection_lost` as been called yet.  Since `connection_lost` is synchronous, checking for `self._closed.done()` assures that we have actually lost the connection.

pymongo/asynchronous/pool.py
```diff
@@ -1022,6 +1029,20 @@ async def remove_stale_sockets(self, reference_generation: int) -> None:
...
+        # raised and we are not an sdam pool, add to backoff and add the
+        # appropriate error label.
+        if (
+            not self.is_sdam
+            and "connection reset by peer" in str(error).lower()
+            or ("connection closed" in str(error).lower() and self._backoff)
+        ):
+            self._backoff += 1
+            error._add_error_label("SystemOverloaded")
+            error._add_error_label("Retryable")
```
ShaneHarvey: Could you merge backpressure? Originally I added the incorrect labels here. It should be "SystemOverloadedError" and "RetryableError"

pymongo/asynchronous/pool.py
```diff
@@ -1022,6 +1029,20 @@ async def remove_stale_sockets(self, reference_generation: int) -> None:
...
+            print(f"Setting backoff in {phase}:", self._backoff)  # noqa: T201
```
ShaneHarvey: Instead of inspecting the error message after the fact, is it possible we can record some state to determine if the error happened during DNS+TCP or after? Like:
```
# Assume all non dns/tcp/timeout errors mean the server rejected the connection due to overload.
if not errorDuringDnsTcp and not timeoutError:
    error._add_error_label("SystemOverloadedError")
```

pymongo/asynchronous/pool.py
blink1073: Done

pymongo/asynchronous/pool.py
blink1073: DNS is already resolved by the time we make a pool as far as I can tell, and we can't distinguish between TCP connection and TLS handshake for async.

blink1073: I added some logic in 7548f7b that looks for a specific error attached to AutoReconnect.

blink1073: For discussion: we can still run into the condition were we hit this [line](https://github.com/mongodb/mongo-python-driver/blob/ef59602e39382e2d0fdf03858cf8dca458ef6553/pymongo/network_layer.py#L553), and there is no closing error, and we have not yet received any data on the protocol.  I verified this by setting a flag when buffer_updated is called.  We don't have a way to ascribe more semantic meaning to this condition as far as I can tell from the Protocol/Transport docs and the available properties on each.

blink1073: The problem got even worse when using gevent, since the error was something completely different, so I reverted any extra handling of the connection error.

blink1073: After today's discussion I added 6a47fc6b914940c5b2e4b44516d2f0c771ef31f1 to only add the labels if we have completed DNS+TCP connection.

pymongo/asynchronous/pool.py
```diff
@@ -1022,6 +1029,19 @@ async def remove_stale_sockets(self, reference_generation: int) -> None:
...
+        # Look for an AutoReconnect error raised from a ConnectionResetError with
+        # errno == errno.ECONNRESET or raised from an OSError that we've created due to
+        # a closed connection.
+        # If found, set backoff and add error labels.
+        if self.is_sdam or type(error) != AutoReconnect:
+            return
+        self._backoff += 1
+        error._add_error_label("SystemOverloadedError")
+        error._add_error_label("RetryableError")
+        print(f"Setting backoff in {phase}:", self._backoff)  # noqa: T201
```
ShaneHarvey: Could we remove this print now?

pymongo/synchronous/pool.py
```diff
@@ -1069,10 +1098,11 @@ def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> Connect
                     reason=_verbose_connection_error_reason(ConnectionClosedReason.ERROR),
                     error=ConnectionClosedReason.ERROR,
                 )
+            if context["has_created_socket"]:
+                self._handle_connection_error(error, "handshake")
             if isinstance(error, (IOError, OSError, *SSLErrors)):
                 details = _get_timeout_details(self.opts)
                 _raise_connection_failure(self.address, error, timeout_details=details)
```
ShaneHarvey, line 1116: _raise_connection_failure will raise a new exception in which case the error labels won't be propagated. Is that a bug? I guess it's not because  _raise_connection_failure only handles `IOError, OSError, *SSLErrors` while _handle_connection_error only handles `AutoReconnect`, right?

pymongo/asynchronous/pool.py
blink1073: Do we want to log it somehow?  In the Server Tech Design they call out "A message will be logged and the function will be returned from."

pymongo/synchronous/pool.py
blink1073: Correct, they are orthogonal

pymongo/asynchronous/pool.py
ShaneHarvey: Yes we'll add a CMAP log message and/or an a new event. Feel free to add a log message now or we can defer to the spec PR.

blink1073: Added

PR #2511, opened 2025-08-27
PYTHON-5520 Add windows arm64 wheel support

finnagin:
[PYTHON-5520](https://jira.mongodb.org/browse/PYTHON-5520) Add windows arm64 wheel support

This would add builds to create windows arm64 wheels.


blink1073 (approved):
Thank you!

PR #2521, opened 2025-09-09
PYTHON-5328 CRUD Support in Driver for Prefix/Suffix/Substring Indexes

blink1073:
Depends on https://github.com/mongodb/libmongocrypt/pull/1065

- [x] Update libmongocrypt to 1.15.1.  Done in [PYTHON-5489](https://jira.mongodb.org/browse/PYTHON-5489)
- [x] Sync  spec tests for auto encryption.  Done in [PYTHON-5143](https://jira.mongodb.org/browse/PYTHON-5143).
- [x] Expose textPreview options. Pass them to libmongcrypt with mongocrypt_ctx_setopt_algorithm_text.  Done in    [PYTHON-5531](https://jira.mongodb.org/browse/PYTHON-5531) and this PR.
- [x] Implement [prose tests](https://github.com/mongodb/specifications/commit/25ebca16d46dd62e9fa491302751ca7760ddf80c) for explicit encryption.  Done in this PR.

### DIFF
--- a/pymongo/asynchronous/encryption.py
+++ b/pymongo/asynchronous/encryption.py
@@ -67,7 +67,7 @@ from pymongo.asynchronous.mongo_client import AsyncMongoClient
 from pymongo.asynchronous.pool import AsyncBaseConnection
 from pymongo.common import CONNECT_TIMEOUT
 from pymongo.daemon import _spawn_daemon
-from pymongo.encryption_options import AutoEncryptionOpts, RangeOpts
+from pymongo.encryption_options import AutoEncryptionOpts, RangeOpts, TextOpts
 from pymongo.errors import (
     ConfigurationError,
     EncryptedCollectionError,
@@ -516,6 +516,11 @@ class Algorithm(str, enum.Enum):
 
     .. versionadded:: 4.4
     """
+    TEXTPREVIEW = "TextPreview"
+    """**BETA** - TextPreview.
+
+    .. versionadded:: 4.15
+    """
 
 
 class QueryType(str, enum.Enum):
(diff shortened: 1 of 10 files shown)


pymongo/asynchronous/encryption.py
```diff
@@ -516,6 +516,11 @@ class Algorithm(str, enum.Enum):
 
     .. versionadded:: 4.4
     """
+    TEXTPREVIEW = "TextPreview"
```
aclark4life, line 519: Are we required to call this `TEXTPREVIEW` instead of calling it `TEXT` and designating "feature preview"?

blink1073: Yes, following the pattern we used for RANGEPREVIEW, to ensure that if the semantics change it won't do so silently.

ShaneHarvey:
We usually only bump the min version if the new pymongo version is incompatible. If that's not the case I'd say the version should remain the same.

test/asynchronous/test_encryption.py
```diff
@@ -3443,6 +3443,261 @@ async def test_collection_name_collision(self):
...
+            case_sensitive=True,
+            diacritic_sensitive=True,
+            prefix=dict(strMaxQueryLength=10, strMinQueryLength=2),
+            suffix=dict(strMaxQueryLength=10, strMinQueryLength=2),
+        )
+        encrypted_value = await self.client_encryption.encrypt(
+            "foobarbaz",
+            key_id=self.key1_id,
+            algorithm=Algorithm.TEXTPREVIEW,
+            contention_factor=0,
```
aclark4life, line 3503: 0 is the default right ? And if so do we need to include it here?

test/asynchronous/test_encryption.py
```diff
@@ -3443,6 +3443,261 @@ async def test_collection_name_collision(self):
...
+        )
+        with self.assertRaises(EncryptionError) as ctx:
+            await self.client_encryption.encrypt(
+                "foo",
+                key_id=self.key1_id,
+                algorithm=Algorithm.TEXTPREVIEW,
+                query_type=QueryType.PREFIXPREVIEW,
+                text_opts=text_opts,
+            )
+        # Expect an error from libmongocrypt with a message containing the string: "contention factor is required for textPreview algorithm".
```
aclark4life, line 3694: Asked and answered!

aclark4life (approved):
LGTM

pymongo/asynchronous/encryption.py
```diff
@@ -905,6 +935,7 @@ async def _encrypt_helper(
                 contention_factor=contention_factor,
                 range_opts=range_opts_bytes,
                 is_expression=is_expression,
+                text_opts=text_opts_bytes,
```
ShaneHarvey, line 938: Passing `text_opts` even when it's None will fail on old versions of pymongocrypt. So we need to bump the min version or avoid passing text_opts by default.

PR #2524, opened 2025-09-10
PYTHON-5537 Update typing dependencies

blink1073:
Account for the `--strict-bytes` parameter added in [mypy 1.15](https://mypy-lang.blogspot.com/2025/02/mypy-115-released.html), which is enabled by our use of `strict=true` without explicitly opting out of the behavior.

The only runtime behavior I made was having `PyMongoKMSProtocol.read()` return bytes, which simplified the code and matched the behavior of `PyMongoProtocol.read()`.

### DIFF
--- a/test/test_typing.py
+++ b/test/test_typing.py
@@ -475,7 +475,7 @@ class TestDocumentType(PyMongoTestCase):
         # This should fail because the output is a Movie.
         assert out["foo"]  # type:ignore[typeddict-item]
         # pyright gives reportTypedDictNotRequiredAccess for the following:
-        assert out["_id"]  # type:ignore
+        assert out["_id"]  # type:ignore[unused-ignore]
 
     @only_type_check
     def test_typeddict_empty_document_type(self) -> None:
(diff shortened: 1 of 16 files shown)


test/test_typing.py
```diff
@@ -475,7 +475,7 @@ def test_typeddict_not_required_document_type(self) -> None:
         # This should fail because the output is a Movie.
         assert out["foo"]  # type:ignore[typeddict-item]
         # pyright gives reportTypedDictNotRequiredAccess for the following:
-        assert out["_id"]  # type:ignore
+        assert out["_id"]  # type:ignore[unused-ignore]
```
NoahStapp, line 478: What's the purpose of this? `# type:ignore[unused-ignore]` sounds like an oxymoron.

NoahStapp (changes requested):

test/test_typing.py
blink1073: mypy wants to ignore it, pyright does not.  💀

NoahStapp: classic lol

blink1073:
drivers-pr-bot please backport to v4.15

PR #2528, opened 2025-09-11
PYTHON-5543 PyMongoBaseProtocol should inherit from asyncio.BaseProtocol

aclark4life:
- Import/use `BaseProtocol` instead of `Protocol` in `PyMongoBaseProtocol` implementation added in PYTHON-5215.
- Resolves issue with instance type check in uvloop that prevents correct buffered config from being set.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -7,8 +7,7 @@ Changes in Version 4.15.1 (XXXX/XX/XX)
 Version 4.15.1 is a bug fix release.
 
 - Fixed a bug in ``AsyncMongoClient`` that caused a
-  ``ServerSelectionTimeoutError`` when used with `` with ``uvicorn`` and
-  ``uvloop``.
+  ``ServerSelectionTimeoutError`` when used with ``uvicorn``, ``FastAPI``, or ``uvloop``.
 
 Changes in Version 4.15.0 (2025/09/10)
 --------------------------------------


doc/changelog.rst
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Changes in Version 4.15.1 (XXXX/XX/XX)
+--------------------------------------
+
+Version 4.15.1 is a bug fix release.
+
+- Fixed a bug in ``MongoClient.network_layer`` and ``AsyncMongoClient.network_layer``
+  that could cause an ``TimeoutError`` when connecting to servers.
```
NoahStapp: We don't need to call out that the bug was in `network_layer`. We do need to call out that it was a bug with `AsyncMongoClient`, and only with `uvicorn` and `uvloop`.

aclark4life: OK I'll fix, what about the difference between PYTHON-5543 and PYTHON-5544? I couldn't reproduce for PYTHON-5543 because the reproducer was incomplete. Should we assume some use of `uvicorn` with FastAPI?

ShaneHarvey: I closed PYTHON-5543 as a duplicate. Let me know if that's incorrect.

NoahStapp: `uvicorn` defaults to `uvloop` under the hood.

doc/changelog.rst
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Changes in Version 4.15.1 (XXXX/XX/XX)
```
ShaneHarvey, line 4: Should we add the date `2025/09/11` here now?

NoahStapp: Yes.

aclark4life: I'll probably add it in a separate "Prep release" PR

doc/changelog.rst
```diff
@@ -1,6 +1,15 @@
...
+--------------------------------------
+
+Version 4.15.1 is a bug fix release.
+
+- Fixed a bug in ``AsyncMongoClient`` that caused a
+  ``ServerSelectionTimeoutError`` when used with `` with ``uvicorn`` and
+  ``uvloop``.
```
NoahStapp:
```suggestion
- Fixed a bug in ``AsyncMongoClient`` that caused a
  ``ServerSelectionTimeoutError`` when used with ``uvicorn``, ``FastAPI``, or ``uvloop``.
```

PR #2530, opened 2025-09-11
Prep 4.15.1

aclark4life:


.github/workflows/codeql.yml
```diff
@@ -42,11 +42,11 @@ jobs:
       with:
         ref: ${{ inputs.ref }}
         persist-credentials: false
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@v6
```
ShaneHarvey: Do we actually want to backport these changes? I'd say 4.15.1 should only have PYTHON-5540 and PYTHON-5543.

doc/changelog.rst
```diff
@@ -10,6 +10,9 @@ Version 4.15.1 is a bug fix release.
    that would cause a ``TypeError`` when using ``pymongocrypt<1.16`` by passing an unsupported ``type_opts`` parameter even if
    Queryable Encryption text queries beta was not used.
 
+- Fixed a bug in ``AsyncMongoClient`` that caused a
```
ShaneHarvey: Could you also fix the bullet point above? ` - Fixed a bug in` has an extra space before the `-`, hence it renders incorrectly. The same issue is in the 4.14.1 section too.

.github/workflows/codeql.yml
ShaneHarvey: Thanks. Could you revert the `v6` change here too?

PR #2539, opened 2025-09-16
PYTHON-5449 - Do not attach invalid document in exception message

NoahStapp:
This PR removes the full document from the error message and stores it in the `document` property on the error instead.

### DIFF
--- a/bson/errors.py
+++ b/bson/errors.py
@@ -44,9 +44,6 @@ class InvalidDocument(BSONError):
         ..versionadded:: 4.16"""
         return self._document
 
-    def _set_document(self, value: Any) -> None:
-        self._document = value
-
 
 class InvalidId(BSONError):
     """Raised when trying to create an ObjectId from invalid data."""


bson/errors.py
```diff
@@ -31,6 +33,19 @@ class InvalidStringData(BSONError):
 class InvalidDocument(BSONError):
     """Raised when trying to create a BSON object from an invalid document."""
 
+    def __init__(self, message: str, document: Optional[Any] = None) -> None:
+        super().__init__(message)
+        self._document = document
+
+    @property
+    def document(self) -> Any:
+        """The invalid document that caused the error."""
```
ShaneHarvey: Please add a versionadded and changelog reference.

bson/errors.py
```diff
@@ -31,6 +33,19 @@ class InvalidStringData(BSONError):
...
+        return self._document
+
+    @document.setter
+    def document(self, value: Any) -> None:
```
ShaneHarvey: Does this need to be public?

bson/_cbsonmodule.c
```diff
@@ -1659,20 +1659,13 @@ void handle_invalid_doc_error(PyObject* dict) {
...
-                goto cleanup;
-            }
+            // Add doc to the error class as a property.
+            PyObject_SetAttrString(InvalidDocument, "document", dict);
             const char * msg_utf8 = PyUnicode_AsUTF8(msg);
             if (msg_utf8 == NULL) {
                 goto cleanup;
             }
-            PyObject *new_msg = PyUnicode_FromFormat("Invalid document %s | %s", dict_str_utf8, msg_utf8);
+            PyObject *new_msg = PyUnicode_FromFormat("Invalid document: %s", msg_utf8);
```
ShaneHarvey, line 1666: For an invalid field in a nested doc, does this yield errors like "Invalid document: Invalid document: Invalid document: Invalid document: Invalid document: ...."?

bson/__init__.py
```diff
@@ -1009,7 +1009,7 @@ def _dict_to_bson(
                 try:
                     elements.append(_element_to_bson(key, value, check_keys, opts))
                 except InvalidDocument as err:
-                    raise InvalidDocument(f"Invalid document {doc} | {err}") from err
+                    raise InvalidDocument(f"Invalid document: {err}", doc) from err
```
ShaneHarvey, line 1012: Shouldn't this InvalidDocument except clause include the encoding of `_id` too?

bson/errors.py
NoahStapp: I can see this being a useful behavior for users, so making this public seems reasonable.

bson/__init__.py
NoahStapp: That would be a pre-existing bug, no? Happy to fix if we identify why the code currently looks like this. Maybe we assume that `_id` will be valid if it reaches this far?

bson/_cbsonmodule.c
NoahStapp: No, it works as expected:

```
encode({"t": {"f": Wrapper(1)}})

# Produces
Invalid document: cannot encode object: 1, of type: <class 'test.test_bson.TestBSON.test_doc_in_invalid_document_error_as_property.<locals>.Wrapper'>
```

bson/__init__.py
ShaneHarvey: Correct it would be a pre-existing bug. Could you open a ticket for it?
```python
>>>> encode({'_id': set()})
  File "<python-input-6>", line 1, in <module>
    encode({'_id': set()})
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 1053, in encode
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 1006, in _dict_to_bson
    elements.append(_name_value_to_bson(b"_id\x00", doc["_id"], check_keys, opts))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 980, in _name_value_to_bson
    raise InvalidDocument(f"cannot encode object: {value!r}, of type: {type(value)!r}")
bson.errors.InvalidDocument: cannot encode object: set(), of type: <class 'set'>
```
vs the expected behavior:
```python
>>>> encode({'_id': 1, 'foo': set()})
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 1010, in _dict_to_bson
    elements.append(_element_to_bson(key, value, check_keys, opts))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 994, in _element_to_bson
    return _name_value_to_bson(name, value, check_keys, opts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 980, in _name_value_to_bson
    raise InvalidDocument(f"cannot encode object: {value!r}, of type: {type(value)!r}")
bson.errors.InvalidDocument: cannot encode object: set(), of type: <class 'set'>

The above exception was the direct cause of the following exception:

  File "<python-input-7>", line 1, in <module>
    encode({'_id': 1, 'foo': set()})
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 1053, in encode
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shane/git/mongo-python-driver/bson/__init__.py", line 1012, in _dict_to_bson
    raise InvalidDocument(f"Invalid document {doc} | {err}") from err
bson.errors.InvalidDocument: Invalid document {'_id': 1, 'foo': set()} | cannot encode object: set(), of type: <class 'set'>
```

bson/errors.py
ShaneHarvey: Sorry, I was suggesting we make it private, not public.

NoahStapp: What's the reasoning for making it private? Security concerns if users misuse it? I find having the full document very useful for debugging and I imagine users do too.

bson/__init__.py
NoahStapp: Opened [PYTHON-5560](https://jira.mongodb.org/browse/PYTHON-5560) to track.

bson/errors.py
ShaneHarvey: I agree the getter should be public but when would we expect a user to set the document? We try to be conservative with public apis and only add ones that have a strong justification.

NoahStapp: Oh whoops sorry I misread which method this was about. Totally agree, fixing.

bson/_cbsonmodule.c
```diff
@@ -1659,20 +1659,13 @@ void handle_invalid_doc_error(PyObject* dict) {
...
-            PyObject *dict_str = PyObject_Str(dict);
-            if (dict_str == NULL) {
-                goto cleanup;
-            }
-            const char * dict_str_utf8 = PyUnicode_AsUTF8(dict_str);
-            if (dict_str_utf8 == NULL) {
-                goto cleanup;
-            }
+            // Add doc to the error class as a property.
+            PyObject_SetAttrString(InvalidDocument, "document", dict);
```
ShaneHarvey: Does this need to be "_document" now?

ShaneHarvey (changes requested):

bson/_cbsonmodule.c
ShaneHarvey: This also highlights that this PR needs to add a test to ensure `.document` is the expected dict.

NoahStapp: Interestingly the tests fail if this is `_document`.

NoahStapp: Are the two modified tests in this PR not sufficient to ensure that `.document` behaves as expected?

ShaneHarvey: This line is setting a class attribute on the InvalidDocument class so the logic is incorrect, eg:
```python
InvalidDocument.document = dict
```

NoahStapp: Ah sorry I misunderstood what you meant. You're right that `_error` returns a class, not an instance.

NoahStapp: Fixed. Had to do some C API delving to figure this out 😅

bson/_cbsonmodule.c
```diff
@@ -1659,19 +1659,19 @@ void handle_invalid_doc_error(PyObject* dict) {
...
         if (msg) {
-            // Add doc to the error class as a property.
-            PyObject_SetAttrString(InvalidDocument, "document", dict);
             const char * msg_utf8 = PyUnicode_AsUTF8(msg);
             if (msg_utf8 == NULL) {
                 goto cleanup;
             }
             PyObject *new_msg = PyUnicode_FromFormat("Invalid document: %s", msg_utf8);
+            // Add doc to the error instance as a property.
+            PyObject *new_evalue = PyObject_CallFunctionObjArgs(InvalidDocument, new_msg, dict, NULL);
```
ShaneHarvey, line 1671: Need to check the `new_msg == NULL` case before we call PyObject_CallFunctionObjArgs here.

ShaneHarvey (changes requested):

bson/errors.py
```diff
@@ -31,6 +33,20 @@ class InvalidStringData(BSONError):
...
+        self._document = document
+
+    @property
+    def document(self) -> Any:
+        """The invalid document that caused the error.
+
+        ..versionadded:: 4.16"""
+        return self._document
+
+    def _set_document(self, value: Any) -> None:
```
ShaneHarvey: `_set_document` can be removed.

PR #2540, opened 2025-09-16
PYTHON-5555 Fix AWS Lambda build

blink1073:
Updated to use Python 3.10, which solved a dns resolve error on lambda.

I did not find the root cause, but since 3.9 is EOL next month I left it at this solution.

I also updated the errors raised by srv_resolver to include the original error, which was useful in debugging the problem.

Passing build: https://spruce.mongodb.com/task/mongo_python_driver_faas_lambda_test_aws_lambda_deployed_patch_eca38b730b8227c52cd5f0c655f617e969743ee1_68c9e481e31406000770e31e_25_09_16_22_28_25/logs?execution=0


blink1073:
Thanks for tracking that down!

PR #2543, opened 2025-09-17
PYTHON-5550 Add a test that uses uvloop as the event loop

blink1073:


.github/workflows/test-python.yml
```diff
@@ -163,6 +166,36 @@ jobs:
...
+    name: Integration Tests
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+      - name: Install uv
+        uses: astral-sh/setup-uv@557e51de59eb14aaaba2ed9621916900a91d50c6 # v5
+        with:
+          enable-cache: true
+          python-version: "3.9"
```
aclark4life: 3.9 until end of October right?

blink1073: Oh, good catch, we already dropped 3.9 support, updating

PR #2544, opened 2025-09-17
PYTHON-2390 - Retryable reads use the same implicit session

NoahStapp:


test/test_retryable_reads.py
```diff
@@ -216,6 +216,38 @@ def test_retryable_reads_are_retried_on_the_same_mongos_when_no_others_are_avail
...
+        }
+
+        listener = OvertCommandListener()
+        client = self.rs_or_single_client(
+            directConnection=False,
+            event_listeners=[listener],
+            retryReads=True,
+        )
+
+        set_fail_point(client, fail_command)
```
ShaneHarvey: Let's use self.fail_point() here.

test/test_retryable_reads.py
```diff
@@ -216,6 +216,38 @@ def test_retryable_reads_are_retried_on_the_same_mongos_when_no_others_are_avail
...
+
+        client.t.t.estimated_document_count()
```
ShaneHarvey: Can we extend this test to cover a few other operations as well?

test/test_retryable_reads.py
```diff
@@ -216,6 +216,38 @@ def test_retryable_reads_are_retried_on_the_same_mongos_when_no_others_are_avail
...
+        set_fail_point(client, fail_command)
+
+        #  Assert that both events occurred on the same session.
+        lsids = [
+            event.command["lsid"]
+            for event in listener.started_events
+            if event.command_name == "count"
+        ]
+        self.assertEqual(len(lsids), 2)
+        self.assertEqual(lsids[0], lsids[1])
```
ShaneHarvey: Should we fix [PYTHON-2391](https://jira.mongodb.org/browse/PYTHON-2391) first? Otherwise this test doesn't prove the fix works correctly since `first.command` is the same dict instance as `second.command`.

NoahStapp: Yes, we should merge the PR for [PYTHON-2391](https://jira.mongodb.org/browse/PYTHON-2391) (https://github.com/mongodb/mongo-python-driver/pull/2545) first to verify this works correctly.

pymongo/synchronous/client_session.py
```diff
@@ -587,6 +589,32 @@ def operation_time(self) -> Optional[Timestamp]:
...
+    def leave_alive(self) -> bool:
+        """Whether to leave this session alive when it is
+        no longer in use.
+        Typically used for implicit sessions that are used for multiple operations within a single larger operation.
+        """
+        return self._leave_alive
+
+    @leave_alive.setter
+    def leave_alive(self, value: bool) -> None:
+        self._leave_alive = value
```
ShaneHarvey: None of these new apis should be public since implicit sessions are only used internally by pymongo.

pymongo/synchronous/change_stream.py
```diff
@@ -258,8 +255,10 @@ def _run_aggregation_cmd(
         )
 
     def _create_cursor(self) -> CommandCursor:  # type: ignore[type-arg]
-        with self._client._tmp_session(self._session, close=False) as s:
-            return self._run_aggregation_cmd(session=s, explicit_session=self._session is not None)
+        with self._client._tmp_session(self._session) as s:
+            if s:
+                s.leave_alive = True
+            return self._run_aggregation_cmd(session=s)
```
ShaneHarvey, line 259: Wouldn't leaving the cursor alive here leak a session every time the change stream cursor is closed?

pymongo/synchronous/client_session.py
```diff
@@ -587,6 +589,32 @@ def operation_time(self) -> Optional[Timestamp]:
         """
         return self._operation_time
 
+    @property
+    def implicit(self) -> bool:
+        """Whether this session was implicitly created by the driver."""
+        return self._implicit
+
+    @property
+    def attached_to_cursor(self) -> bool:
```
ShaneHarvey: Do we need two attributes to track ownership?

NoahStapp: Can you be more specific? As in we need a second attribute to track a different axis of ownership?

pymongo/synchronous/change_stream.py
NoahStapp: Good catch. This shouldn't have `leave_alive` set.

pymongo/synchronous/client_session.py
ShaneHarvey: Eh I was thinking tmp_session would work via the existing implicit=True/False attribute + a new "owner" attribute but your attached_to_cursor+leave_alive implementation seems simpler.

ShaneHarvey:
Looks like there's one test failure to fix:
```python
__________________________ TestSession.test_database ___________________________

self = <test.asynchronous.test_session.TestSession testMethod=test_database>

    async def asyncTearDown(self):
        monitoring._SENSITIVE_COMMANDS.update(self.sensitive_commands)
        await self.client.drop_database("pymongo_test")
        used_lsids = self.initial_lsids.copy()
        for event in self.session_checker_listener.started_events:
            if "lsid" in event.command:
                used_lsids.add(event.command["lsid"]["id"])
    
        current_lsids = {s["id"] for s in session_ids(self.client)}
>       self.assertLessEqual(used_lsids, current_lsids)
E       AssertionError: {Binary(b'\x92\xf1~\xfa\\\xbdI\xed\x96m\xba\x12\xdd\xba\x92\xbf', 4), Binary(b'\xbd\x86\xc1\xbc\xd35L\xb6\x80"\xbc\xb1D\x0f:a', 4)} not less than or equal to {Binary(b'\xbd\x86\xc1\xbc\xd35L\xb6\x80"\xbc\xb1D\x0f:a', 4)}

test/asynchronous/test_session.py:116: AssertionError
```

pymongo/asynchronous/client_session.py
```diff
@@ -588,6 +590,32 @@ def operation_time(self) -> Optional[Timestamp]:
         """
         return self._operation_time
 
+    @property
+    def _is_implicit(self) -> bool:
+        """Whether this session was implicitly created by the driver."""
+        return self._implicit
```
ShaneHarvey: This is personal preference but do we really need these `@property` helpers? Usually we just access the private attribute directly, eg:
```
if session._implicit:...
if session._attached_to_cursor:...
```
This way there's less indirection and boilerplate code.

ShaneHarvey (changes requested):

pymongo/asynchronous/client_session.py
NoahStapp: For internal attributes it makes more sense to not have `@property`, agreed.

pymongo/asynchronous/client_bulk.py
```diff
@@ -538,6 +537,7 @@ async def _execute_command(
                     session._start_retryable_write()
                     self.started_retryable_write = True
                 session._apply_to(cmd, retryable, ReadPreference.PRIMARY, conn)
+                session._leave_alive = True
```
ShaneHarvey: Would it make more sense to move this to `_process_results_cursor` where we actually create a cursor? It seems out of place here.

NoahStapp: Agreed.

ShaneHarvey (approved):
LGTM! Glad to see this bug finally fixed after 6 years.

PR #2547, opened 2025-09-17
 PYTHON-5565 Add minimum version test for Encryption

blink1073:
- Ensured lower bounds for all encryption deps.  I chose `certifi>=2023.7.22` due to https://github.com/advisories/GHSA-xqr8-7jwr-rhp7.
- I also had to add min deps for gevent and simplejson in order to avoid build errors.  For gevent I chose the version that added Python 3.9 support from 2020.  For simplejson I chose a version from 2019.
- Added a new `--test-min-deps` (`TEST_MIN_DEPS`) option that enables `--resolution=lowest-direct`.  I brought in the env var override code from drivers-orchestration to minimize adding more ad-hoc environment var handling.
- Added instructions for running tests with min dependencies
- Adds 6 new tasks for encryption on Linux.  I could not test on Windows yet until we have the rest of the min dependencies due to a uv error.

Passing build: https://spruce.mongodb.com/version/68d52c900f310c00074ad4fd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


.evergreen/scripts/generate_config.py
```diff
@@ -528,22 +528,20 @@ def create_aws_lambda_variants():
...
     for python, sync in product(ALL_PYTHONS, SYNCS):
-        task_input = ("sharded_cluster", "auth", "ssl", sync, python)
-        if task_input not in task_inputs:
-            task_inputs.append(task_input)
+        task_combos.add(("sharded_cluster", "auth", "ssl", sync, python))
 
     # Assemble the tasks.
     seen = set()
-    for topology, auth, ssl, sync, python in task_inputs:
+    for topology, auth, ssl, sync, python in sorted(task_combos):
```
NoahStapp, line 544: What's the use of sorting here?

blink1073: Otherwise the order is nondeterministic and can change every time you run the script

blink1073: (Resulting in a pre-commit failure)

PR #2548, opened 2025-09-18
Remove Python 2 compatibility SON methods

adamchainz:
Python 3.0 [removed `dict.has_key()`](https://docs.python.org/3/whatsnew/3.0.html#views-and-iterators-instead-of-lists:~:text=Removed%2E%20dict%2Ehas%5Fkey%28%29) and [the `dict.iter*()` methods](https://docs.python.org/3/whatsnew/3.0.html#views-and-iterators-instead-of-lists:~:text=Also%2C%20the%20dict%2Eiterkeys%28%29%2C%20dict%2Eiteritems%28%29%20and%20dict%2Eitervalues%28%29%20methods%20are%20no%20longer%20supported).

fcedc510e1039ab47b1c2d52f1392f298feaf81b removed `SON.iteritems()` to align with its removal in Python 3. This commit cleans up by removing the remaining compatibility methods: `has_key()`, `iterkeys()`, and `itervalues()`.

### DIFF
--- a/bson/son.py
+++ b/bson/son.py
@@ -98,17 +98,6 @@ class SON(Dict[_Key, _Value]):
     def __iter__(self) -> Iterator[_Key]:
         yield from self.__keys
 
-    def has_key(self, key: _Key) -> bool:
-        return key in self.__keys
-
-    def iterkeys(self) -> Iterator[_Key]:
-        return self.__iter__()
-
-    # fourth level uses definitions from lower levels
-    def itervalues(self) -> Iterator[_Value]:
-        for _, v in self.items():
-            yield v
-
     def values(self) -> list[_Value]:  # type: ignore[override]
         return [v for _, v in self.items()]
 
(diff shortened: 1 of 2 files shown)


blink1073:
Merge commits are fine, we squash on merge.

ShaneHarvey:
> I totally understand if you want to make that a major version bump, in which case I can work on making SON inherit from dict in a single, larger PR

We aren't planning on doing a 5.0 release anytime soon so I don't think it's worthwhile to work on this at this time. Next steps would be to consider changing mongoengine to replace SON with dict (https://github.com/MongoEngine/mongoengine/issues/2898), updating our SON docs to discourage new applications from using it, and maybe opening a new ticket to track revisiting SON->dict change in our next major release.

ShaneHarvey:
Also if it's possible to optimize SON without any breaking changes that could be worthwhile.

PR #2555, opened 2025-09-23
PYTHON-5570 Do not freeze the lockfile

blink1073:
- Avoid using UV_FROZEN and `--freeze`
- Use a custom pre-commit hook to capture all input files for lock file changes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -132,4 +132,4 @@ repos:
     pass_filenames: false
     fail_fast: true
     additional_dependencies:
-      - uv
+      - "uv>=0.8.4"


.pre-commit-config.yaml
```diff
@@ -128,3 +122,14 @@ repos:
...
     require_serial: true
     additional_dependencies: ["shrub.py>=3.10.0", "pyyaml>=6.0.2"]
+
+  - id: uv-lock
+    name: uv-lock
+    entry: uv lock
+    language: python
+    require_serial: true
+    files: ^(uv\.lock|pyproject\.toml|requirements.txt|requirements/.*\.txt)$
+    pass_filenames: false
```
sleepyStick, line 132: oh i remember you were looking for this flag during our call on monday!

blink1073: Yeah it actually autocompleted in VSCode!  I'm not even sure which extension enabled that

PR #2559, opened 2025-09-25
PYTHON-5573 Require dnspython 2.6.1+

blink1073:


blink1073:
> Thanks, could we update this changelog entry to mention that rationale again?

Done

PR #2561, opened 2025-09-25
PYTHON-5577 - Drop support for OpenSSL 1.0.2

NoahStapp:


blink1073 (approved):
I love me an all-deletion PR!

PR #2577, opened 2025-10-06
PYTHON-5449 & PYTHON-5571 [v4.15] Do not attach invalid document in exception message and fix memory leak

blink1073:


bson/errors.py
```diff
@@ -31,6 +33,17 @@ class InvalidStringData(BSONError):
...
+    def __init__(self, message: str, document: Optional[Any] = None) -> None:
+        super().__init__(message)
+        self._document = document
+
+    @property
+    def document(self) -> Any:
+        """The invalid document that caused the error.
+
+        ..versionadded:: 4.16"""
+        return self._document
```
ShaneHarvey, line 45: I don't think we should backport PYTHON-5449 since it has api changes. Is it possible to only backport the memory leak fix?

blink1073: Perhaps, it would have to be basically a new PR.

PR #2580, opened 2025-10-06
PYTHON-5598 Add generate_config method to ensure auth is tested on free-threaded python 3.14t

caseyclements:
Creates new method create_test_standard_auth_tasks that creates tasks "test-standard-auth-*" that only include sharded-cluster. Because the list of clusters is one, we get a task that contains auth and python 3.14t.

### DIFF
--- a/.evergreen/scripts/generate_config.py
+++ b/.evergreen/scripts/generate_config.py
@@ -202,12 +202,12 @@ def create_enterprise_auth_variants():
     for host in ["rhel8", "macos", "win64"]:
         expansions = dict(TEST_NAME="enterprise_auth", AUTH="auth")
         display_name = get_variant_name("Auth Enterprise", host)
-        tasks = [".test-non-standard .auth !.free-threaded"]
+        tasks = [".test-standard-auth .auth !.free-threaded"]
         # https://jira.mongodb.org/browse/PYTHON-5586
         if host == "macos":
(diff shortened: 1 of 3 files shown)


.evergreen/scripts/generate_config.py
```diff
@@ -602,6 +602,46 @@ def create_test_non_standard_tasks():
     return tasks
 
 
+def create_test_standard_auth_tasks():
```
NoahStapp, line 605: We need to update the auth variant generation functions so their tasks use `test-standard-auth` instead of `test-non-standard`.

NoahStapp (changes requested):

.evergreen/scripts/generate_config.py
caseyclements: "All auth variant generation methods should have their tasks updated from .test-non-standard to .test-standard-auth to use these new tasks." This line wasn't / isn't clear to me. It appeared like it would be a natural result of the new create_test_standard_auth_tasks().

Do I have to manually edit variants.yml?

blink1073: No, you edit the variant function(s) to use the new task tag

caseyclements:
> Can you run a patch build with all of the auth variants?

https://github.com/mongodb/mongo-python-driver/pull/2580#pullrequestreview-3311181806

(This is more than just the auth variants.)

PR #2581, opened 2025-10-06
PYTHON-5588 Fix python binary used in FIPS tests

blink1073:
Passing build: https://spruce.mongodb.com/version/68e42c782b320700072d0c64/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/scripts/generate_config.py
+++ b/.evergreen/scripts/generate_config.py
@@ -469,7 +469,8 @@ def create_alternative_hosts_variants():
 
     version = "latest"
     for host_name in OTHER_HOSTS:
-        expansions = dict(VERSION="latest")
+        # Use explicit Python 3.11 binary on the host since the default python3 is 3.9.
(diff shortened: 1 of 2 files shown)


.evergreen/scripts/generate_config.py
```diff
@@ -469,7 +469,8 @@ def create_alternative_hosts_variants():
 
     version = "latest"
     for host_name in OTHER_HOSTS:
-        expansions = dict(VERSION="latest")
+        # Use explicit Python 3.11 binary on the host since the default python3 is 3.9.
+        expansions = dict(VERSION="latest", PYTHON_BINARY="/usr/bin/python3.11")
```
sleepyStick, line 473: for my own curiosity, why 3.11 and not 3.10?

blink1073: It's what is already installed by the OS on that host

PR #2751, opened 2026-04-13
PYTHON-5791 - test_list_database_names should not check ordering

NoahStapp:
[PYTHON-5791](https://jira.mongodb.org/browse/PYTHON-5791)

## Changes in this PR

Latest server version changed the order of databases returned by test_list_database_names. Our test explicitly checks for order despite the spec not requiring it to do so. Requiring identical ordering is both fragile and does not affect correctness.

## Test Plan

Changed test to verify contents only and not ordering.

### Checklist for Author
- ~[] Did you update the changelog (if necessary)?~
- [X] Is there test coverage?
- [X] Is any followup work tracked in a JIRA ticket? If so, add link(s).

### Checklist for Reviewer


aclark4life (approved):
Nice!

PR #2770, opened 2026-04-23
PYTHON-5779 Coverage increase for `compression_support.py`

aclark4life:
PYTHON-5779

## Changes in this PR
Adds `test/test_compression_support.py` with unit tests for `pymongo/compression_support.py`:

- **`validate_compressors()`** — string vs. iterable input, comma-separated parsing, order preservation, empty input, unknown-name warning, and per-compressor "library unavailable" warnings (snappy / zlib / zstd, including the Py3.14 message split). Uses `patch("pymongo.compression_support._have_xxx", ...)` so results are deterministic regardless of which optional libraries are installed locally.
- **`validate_zlib_compression_level()`** — bounds (-1 / 9), out-of-range `ValueError`, non-integer `TypeError`, and string-to-int coercion.
- **`CompressionSettings.get_compression_context()`** — returns the right `*Context` subclass per compressor name, picks the first compressor when several are listed, and returns `None` for empty / unknown input.
- **`ZlibContext`** — compress/decompress roundtrip.
- **`decompress()`** — zlib branch (bytes and `memoryview` inputs), snappy and zstd roundtrips (gated on `_have_snappy()` / `_have_zstd()`), and the unknown-`compressorId` `ValueError` path.

zlib-dependent tests are gated on `_have_zlib()` and import `zlib` lazily so the module loads on Python builds without zlib.

## Test Plan

```
uv run --extra test python -m pytest test/test_compression_support.py -v
```

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/test/test_compression_support.py
+++ b/test/test_compression_support.py
@@ -83,8 +83,10 @@ class TestValidateCompressors(unittest.TestCase):
         self.assertEqual(result, ["zlib"])
 
     def test_string_input_comma_separated(self):
-        result = validate_compressors(None, "zlib,snappy")
+        with patch("pymongo.compression_support._have_snappy", return_value=True):
+            result = validate_compressors(None, "zlib,snappy")
         self.assertIn("zlib", result)
+        self.assertIn("snappy", result)
 
     def test_iterable_input(self):
         result = validate_compressors(None, ["zlib"])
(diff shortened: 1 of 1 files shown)


test/test_compression_support.py
```diff
@@ -0,0 +1,238 @@
...
+    _have_snappy,
+    _have_zlib,
+    _have_zstd,
+    decompress,
+    validate_compressors,
+    validate_zlib_compression_level,
+)
+
+
+class TestValidateCompressors(unittest.TestCase):
```
sleepyStick, line 40: throughout this file, a lot of the test cases within a class all feel _very_ similar -- is there perhaps a way to re-write them to reduce code duplication?

sleepyStick (approved):

PR #2771, opened 2026-04-23
PYTHON-5784 Coverage increase for `periodic_executor.py`

aclark4life:
PYTHON-5784

## Changes in this PR
Adds `test/asynchronous/test_periodic_executor.py` (synchro'd to `test/test_periodic_executor.py`) with unit tests for `pymongo/periodic_executor.py`, covering both `PeriodicExecutor` and `AsyncPeriodicExecutor`:

- `join()` before `open()` is a safe no-op.
- Target returning `False` stops the executor.
- Target raising an exception stops the executor (verified by asserting target is called exactly once).
- `skip_sleep()` causes the next iteration to run without waiting the full interval.
- `wake()` interrupts the current sleep and triggers an early run.
- `open()` after the target returned `False` restarts the executor.

## Test Plan
Added unit tests using `asyncio.Event` / `threading.Event` for deterministic synchronization. The exception test swaps `threading.excepthook` to a no-op so the intentional `RuntimeError` in the worker thread is not surfaced by pytest's threadexception plugin (no-op for the async executor since no thread is involved).

```
uv run --extra test python -m pytest test/asynchronous/test_periodic_executor.py test/test_periodic_executor.py -v
```

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer


test/test_periodic_executor.py
```diff
@@ -0,0 +1,425 @@
...
+
+    return PeriodicExecutor(interval=interval, min_interval=min_interval, target=target, name=name)
+
+
+def _make_async(interval=30.0, min_interval=0.01, target=None, name="test"):
+    async def _default_target():
+        return True
+
+    if target is None:
+        target = _default_target
```
NoahStapp: Does this need to differ from the way `_make_sync` above does it?

test/test_periodic_executor.py
```diff
@@ -0,0 +1,425 @@
...
+    return AsyncPeriodicExecutor(
+        interval=interval, min_interval=min_interval, target=target, name=name
+    )
+
+
+def _run(coroutine):
+    return asyncio.run(coroutine)
```
NoahStapp: The individual tests that use asynchronous methods should use `AsyncUnitTest` or another appropriate test class from `test/asynchronous/__init__.py` instead of this. Each `asyncio.run()` call adds a significant amount of runtime overhead to setup and teardown the event loop.

test/test_periodic_executor.py
```diff
@@ -0,0 +1,425 @@
...
+
+
+# ---------------------------------------------------------------------------
+# PeriodicExecutor (sync / threading)
+# ---------------------------------------------------------------------------
+
+
+class TestPeriodicExecutorRepr(unittest.TestCase):
+    def test_repr_contains_class_and_name(self):
+        ex = _make_sync(name="myexec")
```
NoahStapp: We can move all the `_make_sync()` calls into an `asyncSetUp/setUp` method on a base class shared by all of the test classes here.

test/test_periodic_executor.py
```diff
@@ -0,0 +1,425 @@
...
+        ex.open()
+        try:
+            self.assertIsNotNone(ex._thread)
+            # Give thread a moment to start.
+            time.sleep(0.05)
+            self.assertTrue(ex._thread.is_alive())
+        finally:
+            ex.close()
+            ex.join(timeout=2)
+
```
NoahStapp, line 39: Similarly, safe closing of each executor can be done in an `asyncTearDown/tearDown` method on a base class.

test/test_periodic_executor.py
```diff
@@ -0,0 +1,425 @@
...
+        pe_module._EXECUTORS.clear()
+        _shutdown_executors()  # Should not raise.
+
+
+# ---------------------------------------------------------------------------
+# AsyncPeriodicExecutor
+# ---------------------------------------------------------------------------
+
+
+class TestAsyncPeriodicExecutorRepr(unittest.TestCase):
```
NoahStapp: Can we move the `AsyncPeriodicExecutor` tests into `test/asynchronous/` and generate the synchronous ones using synchro? That would half the amount of code we need to maintain.

NoahStapp (changes requested):

test/test_periodic_executor.py
aclark4life: Should #2774 get the same treatment?

NoahStapp: Yes. The only tests that should not go through synchro are tests that don't need modification between async and sync, as well as tests against common modules shared between them.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+class AsyncPeriodicExecutorTestBase(AsyncUnitTest):
+    async def asyncSetUp(self):
+        self.executor = _make_executor()
+
+    async def asyncTearDown(self):
+        self.executor.close()
+        await self.executor.join(timeout=2)
+
+
+class TestAsyncPeriodicExecutorRepr(AsyncUnitTest):
```
NoahStapp: This class should also inherit from `AsyncPeriodicExecutorTestBase` and use its setup/teardown methods.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+    async def test_repr_contains_class_and_name(self):
+        executor = _make_executor(name="exec")
+        executor_repr = repr(executor)
+        self.assertIn("AsyncPeriodicExecutor", executor_repr)
+        self.assertIn("exec", executor_repr)
+
+
+class TestAsyncPeriodicExecutorBasic(AsyncPeriodicExecutorTestBase):
```
NoahStapp: What's the intended purpose of having so many separate classes, each with a couple of tests? We generally only separate tests inside files into separate classes if the groupings need different infrastructure (unit vs integration) or specific setup/configuration common to them.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+    async def test_update_interval(self):
```
NoahStapp: This test and the following don't test behavior but implementation.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+        self.assertEqual(self.executor._interval, 60)
+
+    async def test_skip_sleep(self):
+        self.assertFalse(self.executor._skip_sleep)
+        self.executor.skip_sleep()
+        self.assertTrue(self.executor._skip_sleep)
+
+
+class TestAsyncPeriodicExecutorLifecycle(AsyncPeriodicExecutorTestBase):
+    async def test_open_starts_worker(self):
```
NoahStapp: Same here, this is testing implementation, not behavior.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+        self.executor.open()
+        if _IS_SYNC:
+            self.assertIsNotNone(self.executor._thread)
+            self.assertTrue(self.executor._thread.is_alive())
+        else:
+            self.assertIsNotNone(self.executor._task)
+
+    async def test_close_sets_stopped(self):
```
NoahStapp: Same here.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+        self.executor.open()
+        self.executor.close()
+        self.assertTrue(self.executor._stopped)
+        await self.executor.join(timeout=1)
+
+    async def test_join_without_open_is_safe(self):
+        await self.executor.join(timeout=0.01)
```
NoahStapp: What about something more explicit like this:

```
try:
    await self.executor.join(timeout=0.01)
except Exception as e:
    self.fail(f"join() raised unexpected Exception {e}")
```

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+            ran = asyncio.Event()
+
+        async def target():
+            ran.set()
+            return False
+
+        self.executor = _make_executor(target=target)
+        self.executor.open()
+        if _IS_SYNC:
+            self.assertTrue(ran.wait(timeout=2), "target never ran")
```
NoahStapp: Is this needed if we're calling `join()` with the same timeout immediately after? We could check that `ran` is set after the join, which would mean that the executor stopped running or else the join would timeout.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+        else:
+            await asyncio.wait_for(ran.wait(), timeout=2)
+        await self.executor.join(timeout=2)
+        self.assertTrue(self.executor._stopped)
```
NoahStapp: We can remove this if my above comment applies.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+            threading.excepthook = _capture_excepthook
+            try:
+
+                def target():
+                    ran.set()
+                    raise RuntimeError("error")
+
+                self.executor = _make_executor(target=target)
+                self.executor.open()
+                self.assertTrue(ran.wait(timeout=2), "target never ran")
```
NoahStapp: Same here as above.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+        else:
+            ran = asyncio.Event()
+
+            async def target():
+                ran.set()
+                raise RuntimeError("error")
+
+            self.executor = _make_executor(target=target)
+            self.executor.open()
+            await asyncio.wait_for(ran.wait(), timeout=2)
```
NoahStapp: And here.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,285 @@
...
+            await self.executor.join(timeout=2)
+            self.assertTrue(self.executor._stopped)
+            if self.executor._task is not None and self.executor._task.done():
+                self.executor._task.exception()
+
+    async def test_skip_sleep_flag_skips_interval(self):
+        call_times = []
+
+        async def target():
+            call_times.append(time.monotonic() if _IS_SYNC else asyncio.get_running_loop().time())
```
NoahStapp: Why use `asyncio.get_running_loop().time()` over `time.monotonic()`? Asyncio's default event loop returns `time.monotonic()` anyway: https://github.com/python/cpython/blob/b8ebd078f90007d48fcab85effadb33769cd080c/Lib/asyncio/base_events.py#L771.

NoahStapp (changes requested):

test/asynchronous/test_periodic_executor.py
NoahStapp: This question still stands @aclark4life.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,234 @@
...
+
+        self.executor = _make_executor(target=target)
+        self.executor.open()
+        await self.executor.join(timeout=2)
+        self.executor.open()
+        await self.executor.join(timeout=2)
+        self.assertGreaterEqual(called[0], 2)
+
+
+class TestShouldStop(AsyncUnitTest):
```
NoahStapp: Both of these tests are again testing implementation, not behavior. Do they add confidence in the behavior of `PeriodicExecutor` when used?

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,234 @@
...
+            self.assertFalse(executor._thread_will_exit)
+
+        def test_returns_true_and_sets_thread_will_exit(self):
+            executor = _make_executor()
+            executor._stopped = True
+            self.assertTrue(executor._should_stop())
+            self.assertTrue(executor._thread_will_exit)
+
+
+class TestRegisterExecutor(AsyncUnitTest):
```
NoahStapp: Does this test meaningfully add coverage that our existing`test_cleanup_executors_on_client_del` in `test/test_monitor.py` does not?

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,234 @@
...
+
+        async def target():
+            return True
+
+    return AsyncPeriodicExecutor(
+        interval=interval, min_interval=min_interval, target=target, name=name
+    )
+
+
+class AsyncPeriodicExecutorTestBase(AsyncUnitTest):
```
NoahStapp: Every test except for `test_join_without_open_is_safe` now explicitly calls `_make_executor()` and `.join()` as part of execution, making this base class largely unused.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,234 @@
...
+                    raise RuntimeError("error")
+
+                self.executor = _make_executor(target=target)
+                self.executor.open()
+                self.executor.join(timeout=2)
+                self.assertTrue(ran.is_set(), "target never ran")
+            finally:
+                threading.excepthook = orig_excepthook
+            self.assertEqual(len(captured_exc), 1)
+            self.assertIsInstance(captured_exc[0], RuntimeError)
```
NoahStapp: Can we instead check that `target` is called only once due to an exception being thrown? That would be equivalent here and require less code.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,234 @@
...
+
+        self.executor = _make_executor(interval=30.0, min_interval=0.001, target=target)
+        self.executor.skip_sleep()
+        self.executor.open()
+        await self.executor.join(timeout=3)
+        self.assertGreaterEqual(len(call_times), 2)
+        self.assertLess(call_times[1] - call_times[0], 5.0)
+
+    async def test_wake_causes_early_run(self):
+        call_count = [0]
```
NoahStapp: This can be a plain integer if we use `nonlocal` in `target`.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,234 @@
...
+        else:
+            woken = asyncio.Event()
+
+        async def target():
+            call_count[0] += 1
+            if call_count[0] == 1:
+                woken.set()
+            if call_count[0] >= 2:
+                return False
+            return True
```
NoahStapp:
```suggestion
            nonlocal call_count
            call_count += 1
            if call_count == 1:
                woken.set()
            return call_count < 2
```

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,234 @@
...
+
+        self.executor = _make_executor(interval=30.0, min_interval=0.01, target=target)
+        self.executor.open()
+        if _IS_SYNC:
+            woken.wait(timeout=2)
+        else:
+            await asyncio.wait_for(woken.wait(), timeout=2)
+        self.executor.wake()
+        await self.executor.join(timeout=3)
+        self.assertGreaterEqual(call_count[0], 2)
```
NoahStapp:
```suggestion
        self.assertGreaterEqual(call_count, 2)
```

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,234 @@
...
+
+    async def test_open_after_target_returns_false(self):
```
NoahStapp, line 132: Same here as above, prefer `nonlocal` over the list trick.

NoahStapp (dismissed):

test/asynchronous/test_periodic_executor.py
aclark4life: > This question still stands @aclark4life.

No good reason, especially in the context of removing non-behavior-checking tests. I'll push a single class (plus base class) unless we come across a reason to add more.

test/asynchronous/test_periodic_executor.py
aclark4life: TIL `nonlocal` exists.

test/asynchronous/test_periodic_executor.py
```diff
@@ -0,0 +1,182 @@
...
+            woken.wait(timeout=2)
+        else:
+            assert isinstance(woken, asyncio.Event)
+            await asyncio.wait_for(woken.wait(), timeout=2)
+        executor.wake()
+        await executor.join(timeout=3)
+        self.assertGreaterEqual(call_count, 2)
+
+    async def test_update_interval_changes_next_wait(self):
+        call_times = []
```
blink1073, line 115: This call site needs to be updated to use `nonlocal` as well

blink1073 (changes requested):

blink1073 (approved):
LGTM

PR #2772, opened 2026-04-23
PYTHON-5782 Coverage increase, remove old code in `message.py`

aclark4life:
PYTHON-5782

## Changes in this PR
Adds `test/test_message.py` with unit tests for `pymongo/message.py`.

Also removes a dead code branch in `_convert_write_result()` — a legacy compatibility shim for MongoDB < 2.6 that manually reconstructed the upserted `_id` when it wasn't an ObjectId. MongoDB 2.6+ always returns the `upserted` field directly, so this branch was unreachable.

Also fixes a bug in `_raise_document_too_large()` where the error message used `{operation!r}` (repr formatting), producing `'update' command document too large` instead of `update command document too large`. This mismatched the C extension's `%s`-formatted equivalent and broke the new test.

Tests cover:

**Result / error conversion**
- `_convert_exception()` — converts an `Exception` into an `errmsg`/`errtype` failure document for event publishing
- `_convert_client_bulk_exception()` — same as above but also captures the error `code`, used by the client-level bulk write API
- `_convert_write_result()` — converts a legacy GLE (Get Last Error) write result into write-command format, handling insert/update/delete/upsert, wtimeout, and errInfo branches
- `_raise_document_too_large()` — raises `DocumentTooLarge` with a size-aware message for inserts or a generic message for other operations

**Wire-format message construction**
- `_op_msg()` — builds an OP_MSG wire message, injecting `$db` and `$readPreference`, and temporarily popping document-sequence fields before encoding

**Command builders**
- `_gen_find_command()` — builds a `find` command document from a query spec, applying projections, skip/limit, batch size, read concern, collation, and cursor options
- `_gen_get_more_command()` — builds a `getMore` command document, conditionally including batch size, `maxTimeMS`, and `comment` (gated on wire version ≥ 9)

**Read preference**
- `_maybe_add_read_preference()` — injects `$readPreference` into a query spec for non-primary modes, skipping plain `secondaryPreferred` when no tags or maxStalenessSeconds are set

## Test Plan
Added unit tests using a real `ZlibContext` for compression paths and a `MagicMock` connection for `_gen_get_more_command`. Focuses on the pure-Python code paths without requiring a live MongoDB server.

```
hatch run test:test test/test_message.py -v
```

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/pymongo/message.py
+++ b/pymongo/message.py
@@ -652,7 +652,7 @@ def _raise_document_too_large(operation: str, doc_size: int, max_size: int) -> N
     else:
         # There's nothing intelligent we can say
         # about size for update and delete
-        raise DocumentTooLarge(f"{operation!r} command document too large")
+        raise DocumentTooLarge(f"{operation} command document too large")
 
 
 # From the Client Side Encryption spec:
--- a/pymongo/message.py
+++ b/pymongo/message.py
@@ -652,7 +652,7 @@ def _raise_document_too_large(operation: str, doc_size: int, max_size: int) -> N
     else:
         # There's nothing intelligent we can say
         # about size for update and delete
-        raise DocumentTooLarge(f"{operation} command document too large")
+        raise DocumentTooLarge(f"{operation!r} command document too large")
 
 
 # From the Client Side Encryption spec:
--- a/pymongo/message.py
+++ b/pymongo/message.py
@@ -171,14 +171,6 @@ def _convert_write_result(
     elif operation == "update":
         if "upserted" in result:
             res["upserted"] = [{"index": 0, "_id": result["upserted"]}]
-        # Versions of MongoDB before 2.6 don't return the _id for an
-        # upsert if _id is not an ObjectId.
-        elif result.get("updatedExisting") is False and affected == 1:
-            # If _id is in both the update document *and* the query spec
-            # the update document _id takes precedence.
-            update = command["updates"][0]
-            _id = update["u"].get("_id", update["q"].get("_id"))
-            res["upserted"] = [{"index": 0, "_id": _id}]
     return res
 
 
--- a/test/test_message.py
+++ b/test/test_message.py
@@ -1,0 +1,586 @@
+# Copyright 2026-present MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for message.py."""
+
+from __future__ import annotations
+
+import struct
+import sys
+
+sys.path[0:0] = [""]
+
+from test import unittest
+
+from bson import CodecOptions
+from pymongo.errors import DocumentTooLarge
+from pymongo.message import (
+    MAX_INT32,
+    MIN_INT32,
+    _compress,
+    _convert_client_bulk_exception,
+    _convert_exception,
+    _convert_write_result,
+    _gen_find_command,
+    _gen_get_more_command,
+    _get_more,
+    _get_more_compressed,
+    _get_more_impl,
+    _get_more_uncompressed,
+    _maybe_add_read_preference,
+    _op_msg,
+    _op_msg_no_header,
+    _op_msg_uncompressed,
+    _query,
+    _query_compressed,
+    _query_impl,
+    _query_uncompressed,
+    _raise_document_too_large,
+    _randint,
+)
+from pymongo.read_preferences import Primary, ReadPreference, Secondary, SecondaryPreferred
+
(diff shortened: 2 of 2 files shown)


aclark4life:
Test failures flaky:

```
  Error response from daemon: No such image: quay.io/pypa/manylinux_2_28_aarch64:2026.03.20-1
  Error response from daemon: Get "https://quay.io/v2/": context deadline exceeded
```

```
(api/gridfs/asynchronous/grid_file: line   10) timeout   http://dochub.mongodb.org/core/gridfsspec - HTTPSConnectionPool(host='dochub.mongodb.org', port=443): Read timed out. (read timeout=60)
```

test/test_message.py
```diff
@@ -0,0 +1,316 @@
...
+    def test_secondary_preferred_no_tags_does_not_add(self):
+        spec: dict = {"find": "col"}
+        result = _maybe_add_read_preference(spec, ReadPreference.SECONDARY_PREFERRED)
+        self.assertNotIn("$readPreference", result)
+
+    def test_secondary_preferred_with_tags_adds_read_preference(self):
+        pref = SecondaryPreferred(tag_sets=[{"dc": "east"}])
+        spec: dict = {"find": "col"}
+        result = _maybe_add_read_preference(spec, pref)
+        self.assertIn("$readPreference", result)
```
NoahStapp, line 78: This should make the same assertions as `test_secondary_adds_read_preference` for consistency.

test/test_message.py
```diff
@@ -0,0 +1,316 @@
...
+        self.assertNotIn("upserted", result)
+
+    def test_update_with_upserted_id(self):
+        cmd = {"updates": [{"q": {}, "u": {"_id": 42}}]}
+        result = _convert_write_result("update", cmd, {"n": 1, "upserted": 42})
+        self.assertIn("upserted", result)
+        self.assertEqual(result["upserted"][0]["_id"], 42)
+
+    def test_update_legacy_upsert_id_from_update_doc(self):
+        # Pre-2.6 servers omit "upserted"; _id is extracted from the update doc (takes
```
NoahStapp: We don't support servers <4.2, so no need to test them. Any code that still makes a distinction should be removed.

NoahStapp (changes requested):

pymongo/message.py
```diff
@@ -171,14 +171,6 @@ def _convert_write_result(
...
             res["upserted"] = [{"index": 0, "_id": result["upserted"]}]
-        # Versions of MongoDB before 2.6 don't return the _id for an
-        # upsert if _id is not an ObjectId.
-        elif result.get("updatedExisting") is False and affected == 1:
-            # If _id is in both the update document *and* the query spec
-            # the update document _id takes precedence.
-            update = command["updates"][0]
-            _id = update["u"].get("_id", update["q"].get("_id"))
-            res["upserted"] = [{"index": 0, "_id": _id}]
     return res
```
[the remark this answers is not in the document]
NoahStapp: Agreed that the PR should be updated to reflect that there's this minor refactor here.

test/test_message.py
```diff
@@ -0,0 +1,303 @@
...
+        self.assertIn("1000000", msg)
+
+    def test_raise_document_too_large_update_generic_message(self):
+        with self.assertRaises(DocumentTooLarge) as ctx:
+            _raise_document_too_large("update", 2_000_000, 1_000_000)
+        self.assertIn("update", str(ctx.exception))
+
+    # _gen_find_command
+
+    def test_find_basic(self):
```
NoahStapp, line 208: All of the calls to `_gen_find_command` and `_gen_get_more_command` should use keyword args for readability.

test/test_message.py
```diff
@@ -0,0 +1,303 @@
...
+        self.assertIn("errInfo", result["writeErrors"][0])
+
+    # _op_msg
+
+    def test_op_msg_max_doc_size_zero_without_docs(self):
+        max_doc_size = _op_msg(0, {"ping": 1}, "testdb", None, _OPTS)[3]
+        self.assertEqual(max_doc_size, 0)
+
+    def test_op_msg_max_doc_size_matches_largest_encoded_doc(self):
+        docs = [{"_id": 1, "x": 2}, {"_id": 3, "x": 4}]
```
NoahStapp: These encode to the same size, so the `assert` would pass regardless of how `max_doc_size` is set. One of these should be larger than the other to ensure correctness.

test/test_message.py
```diff
@@ -0,0 +1,303 @@
...
+        with self.assertRaises(DocumentTooLarge) as ctx:
+            _raise_document_too_large("insert", 2_000_000, 1_000_000)
+        msg = str(ctx.exception)
+        self.assertIn("2000000", msg)
+        self.assertIn("1000000", msg)
+
+    def test_raise_document_too_large_update_generic_message(self):
+        with self.assertRaises(DocumentTooLarge) as ctx:
+            _raise_document_too_large("update", 2_000_000, 1_000_000)
+        self.assertIn("update", str(ctx.exception))
```
NoahStapp:
```suggestion
        with self.assertRaisesRegex(DocumentTooLarge, "update command document too large"):
            _raise_document_too_large("update", 2_000_000, 1_000_000)
```

NoahStapp (changes requested):

NoahStapp:
Typecheck is failing.

NoahStapp (approved):

PR #2781, opened 2026-04-24
PYTHON-5806 [PoC] rename network_layer.py to network_layer_shared.py

aclark4life:
Rename pymongo/network_layer.py to pymongo/network_layer_shared.py to make explicit that this file is intentionally shared between async and sync code, and is not processed by synchro. Update all 9 import sites.

Add clarifying module docstrings to periodic_executor.py and lock.py explaining why they also live in the top-level package alongside both async and sync implementations.

PYTHON-5806

### Checklist for Author
- [ ] Is there test coverage?

### Checklist for Reviewer


aclark4life:
Shared means used by both async and sync but not async itself.

PR #2786, opened 2026-04-29
URL-encode client_id in Azure IMDS token request

CoffeePrism:
## Summary

Apply `urllib.parse.quote()` to the `client_id` parameter in `_get_azure_response()` before interpolating it into the Azure IMDS URL, consistent with the existing treatment of `resource` (which is already encoded via `quote()` in `_OIDCAzureCallback.__init__`).

## Motivation

The `resource` parameter is URL-encoded at the call site (`auth_oidc_shared.py:103`), but `client_id` is interpolated raw via f-string (`_azure_helpers.py:32`). This inconsistency means a `client_id` containing URL-special characters (`&`, `=`, `#`, etc.) could unintentionally alter the query string structure.

For comparison, the Node.js driver uses the safe `url.searchParams.append('client_id', username)` API for the same operation (`src/client-side-encryption/providers/azure.ts:127`), which automatically handles encoding.

## Changes

- `pymongo/_azure_helpers.py`: Apply `quote()` to `client_id` before URL interpolation
- `test/test_azure_helpers.py`: Add `test_client_id_is_url_encoded` to verify special characters are properly percent-encoded

## Test

All 13 tests in `test/test_azure_helpers.py` pass, including the new encoding test.

### DIFF
--- a/pymongo/_azure_helpers.py
+++ b/pymongo/_azure_helpers.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import json
 from typing import Any, Optional
+from urllib.parse import quote
 
 
 def _get_azure_response(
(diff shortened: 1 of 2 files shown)


CoffeePrism:
Closing — will resubmit from the correct account.

PR #2789, opened 2026-05-02
PYTHON-5676 Consolidate Command Execution Logic

aclark4life:
[PYTHON-5676](https://jira.mongodb.org/browse/PYTHON-5676)

## Changes in this PR

Consolidate the per-command execution path so that the standard command path (`network.command()`) and the cursor path (`Server.run_operation()` for `find`/`getMore`) share a single implementation, and centralize per-command debug logging behind a reusable context manager. Pure refactor — no public API changes.

**New private APIs**
- `pymongo/_telemetry.py` — `_CommandTelemetry`, a context manager that emits the command `STARTED`/`SUCCEEDED`/`FAILED` debug-log events, owns the duration clock, and exposes the computed `duration` and `failure` for reuse by APM events. It wires only the logging channel; APM publishing remains at the call sites. The shape (`handle_succeeded` / `handle_failed`, started in `__enter__`, failure in `__exit__`) lets the monitoring and OpenTelemetry channels be layered on top later without changing call sites.
- `pymongo/{a,}synchronous/network.py` — `_network_command_core()`, which performs send/receive, response unpacking, `_process_response`, `_check_command_response`, command logging + inline APM publishing, and optional decryption. Both `network.command()` and `Server.run_operation()` route through it.

**Refactors / removals**
- Remove the dead legacy OP_QUERY/OP_GET_MORE cursor path from `Server.run_operation()`. All supported servers have wire version ≥ 8 (`MIN_SUPPORTED_WIRE_VERSION`), so `operation.use_command()` always selects the OP_MSG command path; its result is now asserted rather than branched on.
- Simplify cursor / command-cursor response parsing to assume the command (OP_MSG) response format.
- `bulk.py` / `client_bulk.py` write paths use `_CommandTelemetry` for command logging in place of the duplicated inline `_debug_log` blocks.

**Behavior change**
- `unack_write()` emits the `SUCCEEDED` command log unconditionally, so command logging matches the APM success event that was already published in both the `result is not None` and `result is None` branches.

## Test Plan

Refactor that preserves existing behavior (other than the `unack_write` logging consistency noted above), so it relies on existing coverage rather than new tests:
- Command monitoring and command logging spec tests (sync and async).
- Bulk write, client bulk write, cursor, and CRUD suites (sync and async).

`mypy` and `ruff` (lint + format) are clean across the changed modules.

### Checklist for Author
- [ ] Did you update the changelog (if necessary)? — Not necessary; internal refactor with no public API change.
- [x] Is there test coverage? — Covered by the existing command monitoring/logging, bulk, cursor, and CRUD suites.
- [x] Is any followup work tracked in a JIRA ticket? If so, add link(s). — Foundational for [PYTHON-5745](https://jira.mongodb.org/browse/PYTHON-5745) (consolidate command logging + monitoring, #2822) and [PYTHON-5052](https://jira.mongodb.org/browse/PYTHON-5052) (OpenTelemetry support), which extend `_CommandTelemetry`.

### Checklist for Reviewer


aclark4life:
@blink1073 In case it helps, I refactored this PR to use a `_CommandTelemetry` context manager for logging so it could potentially be merged ahead of #2822 where you could then add telemetry.

blink1073:
Okay cool, I'll take a look

blink1073:
Based on a first pass, here are my thoughts:

- I like that this PR primarily focuses on `command()` and `server`.  It reduces the scope of the changes.
- I think the remaining tasks should be:
   - move more logic out of bulk and the bulk context into `network.py` so that all command execution/telemetry is centralized.  To enable a cleaner separation, remove the higher level commands from `AsyncConnection` and use free functions in `network.py`.
   - instead of relying on a `client` object passed to `AsyncConnection` to determine whether to log, explicitly pass a flag for whether to log the command.
   - only call `client._process_response` from `network.py`.

blink1073:
This is too hard to review now, I had wanted to review/merge the state it was in this morning.

aclark4life:
> This is too hard to review now, I had wanted to review/merge the state it was in this morning.

Fixed! I haven't reviewed the 8th commit yet.

blink1073:
I think the Phase 5 and Phase 6 commits need to be reverted

aclark4life:
> I think the remaining tasks should be:
> * move more logic out of bulk and the bulk context into `network.py` so that all command execution/telemetry is centralized.  To enable a cleaner separation, remove the higher level commands from `AsyncConnection` and use free functions in `network.py`.
> * instead of relying on a `client` object passed to `AsyncConnection` to determine whether to log, explicitly pass a flag for whether to log the command.
> * only call `client._process_response` from `network.py`.

OK back to 5 hours ago, do you want me to try to address these issues or leave them for #2822 ?

blink1073:
> OK back to 5 hours ago, do you want me to try to address these issues or leave them for https://github.com/mongodb/mongo-python-driver/pull/2822 ?

I think this PR as it stands should close #2822, and then a follow up PR should do the remaining command consolidation work.

blink1073:
Actually no, the next PR would close both tickets, because we need to move event logging into the telemetry class as well.

blink1073:
These failures look like they're caused by a different handling of exceptions:

```python
FAILED test/asynchronous/test_collection.py::AsyncTestCollection::test_write_large_document - pymongo.errors.DocumentTooLarge: BSON document too large (16777248 bytes) - the connected server supports BSON document sizes up to 16777216 bytes.
FAILED test/asynchronous/test_cursor.py::TestCursor::test_maxtime_ms_message - AssertionError: '(configured timeouts: connectTimeoutMS: 20000.0ms' not found in "operation exceeded time limit, full error: {'ok': 0.0, 'errmsg': 'operation exceeded time limit', 'code': 50, 'codeName': 'MaxTimeMSExpired'}"
```

blink1073 (approved):
LGTM!

PR #2804, opened 2026-05-15
PYTHON-5671: Use fine-grained locks in the connection pool to reduce contention

sophiayangDB:
PYTHON-5671

## Changes in this PR
Based on investigation in PYTHON-5007, using fine-grained locks improved thread efficiency by about 20% in a small read scenario. This PR seeks to implement the suggestions.

Previously, code in `pool.py` was protected by the sole lock `self.lock` and its two conditionals. Now, there are six individual locks that are the sole protectors of the following:
- `self._conns_lock`
Protects `self.conns` and `self.gen`
- `self._active_contexts_lock`
Protects `self.active_contexts` as well as the subsequent updates to `self.__pinned_sockets`, `self.active_sockets`, `next_connection_id`, `self.ncursors`, and `self.ntxns`
- `self.size_cond`
Protects `self.requests` and `self.gen`
- `self._max_connecting_cond`
Protects `self._pending`
- `self._operation_count_lock`
Protects `self.operation_count`
- `self.lock`
Protects `self.state`

Other strategies used to mitigate lock contention include:
- Moving lines that do not require a lock to be performed before lock acquisition
- Use of temporary, local variables to avoid nested lock contention
- Removing lock acquisition from `update_is_writable()` as its sole caller already has the lock
- Copying `self.active_contexts` to update its references when additional additions/deletions to the set should not be operated upon or their absence does not modify behavior, avoids errors that arise if calling on the original and it's length/size changes

## Test Plan
Tests already exist in `main/test/test_pooling.py` and `test/asynchronous/test_pooling.py`. Since this was only a code refactor, no additional tests were added.

### Checklist for Author
- [yes] Did you update the changelog (if necessary)?
- [yes] Is there test coverage?
- [n/a] Is any followup work tracked in a JIRA ticket? If so, add link(s).

### Checklist for Reviewer

### DIFF
--- a/doc/contributors.rst
+++ b/doc/contributors.rst
@@ -108,3 +108,4 @@ The following is a list of people who have contributed to
 - Steven Silvester (blink1073)
 - Noah Stapp (NoahStapp)
 - Cal Jacobson (cj81499)
+- Sophia Yang (sophiayangDB)
--- a/pymongo/asynchronous/pool.py
+++ b/pymongo/asynchronous/pool.py
@@ -802,8 +802,8 @@ class Pool:
     async def ready(self) -> None:
         # Take the lock to avoid the race condition described in PYTHON-2699.
         state_changed = False
-        if self.state != PoolState.READY:
-            async with self.lock:
+        async with self.lock:
+            if self.state != PoolState.READY:
                 self.state = PoolState.READY
                 state_changed = True
         if not state_changed:
@@ -832,13 +832,12 @@ class Pool:
         interrupt_connections: bool = False,
     ) -> None:
         old_state = self.state
-        if self.closed:
-            return
         is_fork = False
         async with self.lock:
+            if self.closed:
+                return
             if self.opts.pause_enabled and pause and not self.opts.load_balanced:
                 old_state, self.state = self.state, PoolState.PAUSED
-            self.gen.inc(service_id)
             newpid = os.getpid()
             if self.pid != newpid:
                 self.pid = newpid
@@ -848,19 +847,21 @@ class Pool:
                 self.active_sockets = 0
             async with self._operation_count_lock:
                 self.operation_count = 0
-        async with self._conns_lock:
-            if service_id is None:
+        if service_id is None:
+            async with self._conns_lock:
+                self.gen.inc(service_id)
                 sockets, self.conns = self.conns, collections.deque()
         if service_id is not None:
             discard: collections.deque = collections.deque()  # type: ignore[type-arg]
             keep: collections.deque = collections.deque()  # type: ignore[type-arg]
-            for conn in self.conns:
-                if conn.service_id == service_id:
-                    discard.append(conn)
-                else:
-                    keep.append(conn)
-            sockets = discard
             async with self._conns_lock:
+                self.gen.inc(service_id)
+                for conn in self.conns:
+                    if conn.service_id == service_id:
+                        discard.append(conn)
+                    else:
+                        keep.append(conn)
+                sockets = discard
                 self.conns = keep
 
         if close:
@@ -932,8 +933,9 @@ class Pool:
         Pool.
         """
         self.is_writable = is_writable
-        for _socket in self.conns:
-            _socket.update_is_writable(self.is_writable)  # type: ignore[arg-type]
+        async with self._conns_lock:
+            for _socket in self.conns:
+                _socket.update_is_writable(self.is_writable)  # type: ignore[arg-type]
 
     async def reset(
         self, service_id: Optional[ObjectId] = None, interrupt_connections: bool = False
@@ -964,7 +966,7 @@ class Pool:
 
         if self.opts.max_idle_time_seconds is not None:
             close_conns = []
-            async with self.lock:
+            async with self._conns_lock:
                 while (
                     self.conns
                     and self.conns[-1].idle_time_seconds() > self.opts.max_idle_time_seconds
@@ -1000,13 +1002,13 @@ class Pool:
                 close_conn = False
                 # Close connection and return if the pool was reset during
                 # socket creation or while acquiring the pool lock.
-                if self.gen.get_overall() != reference_generation:
-                    close_conn = True
-                if not close_conn:
-                    async with self._conns_lock:
+                async with self._conns_lock:
+                    if self.gen.get_overall() != reference_generation:
+                        close_conn = True
+                    else:
                         self.conns.appendleft(conn)
-                    async with self._active_contexts_lock:
-                        self.active_contexts.discard(conn.cancel_context)
+                async with self._active_contexts_lock:
+                    self.active_contexts.discard(conn.cancel_context)
                 if close_conn:
                     await conn.close_conn(ConnectionClosedReason.STALE)
                     return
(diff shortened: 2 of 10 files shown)


doc/contributors.rst
```diff
@@ -108,3 +108,4 @@ The following is a list of people who have contributed to
 - Steven Silvester (blink1073)
 - Noah Stapp (NoahStapp)
 - Cal Jacobson (cj81499)
+- Sophia Yang (sophiayangDB)
```
sleepyStick, line 111: yay!! first pymongo change!!

pymongo/asynchronous/pool.py
```diff
@@ -777,29 +784,41 @@ def __init__(
...
-                        _CONNECTION_LOGGER,
-                        message=_ConnectionStatusMessage.POOL_READY,
-                        clientId=self._client_id,
-                        serverHost=self.address[0],
-                        serverPort=self.address[1],
-                    )
-                if self.enabled_for_cmap:
-                    assert self.opts._event_listeners is not None
-                    self.opts._event_listeners.publish_pool_ready(self.address)
+                state_changed = True
```
NoahStapp: The `self.state != PoolState.READY` check now being outside the lock can lead to the following race:

```
# Two threads/concurrent tasks, A and B, both call Pool.ready() simultaneously. 
# The pool is not ready, so both A and B try to grab the lock.
# A wins and sets `self.state = PoolState.READY` and `state_changed = True`, then releases the lock before logging/emitting events.
# B gets the lock A releases and does the same, even though the pool has now already been marked as ready.
# As a result, logs and events are doubled.
```
This is the issue described by [PYTHON-2699](https://jira.mongodb.org/browse/PYTHON-2699) that we still need to avoid.

One option is to check again inside the lock:
```
async with self.lock:
    if self.state != PoolState.READY:
        self.state = PoolState.READY
        state_changed = True
```

pymongo/asynchronous/pool.py
```diff
@@ -813,39 +832,49 @@ async def _reset(
...
             newpid = os.getpid()
             if self.pid != newpid:
                 self.pid = newpid
+                is_fork = True
+        if is_fork:
+            async with self._active_contexts_lock:
                 self.active_sockets = 0
+            async with self._operation_count_lock:
                 self.operation_count = 0
+        async with self._conns_lock:
```
NoahStapp: We should only grab the lock here if `service_id` is None. Checking `service_id` does not require the lock.

pymongo/asynchronous/pool.py
```diff
@@ -813,39 +832,49 @@ async def _reset(
...
-                for conn in self.conns:
-                    if conn.service_id == service_id:
-                        discard.append(conn)
-                    else:
-                        keep.append(conn)
-                sockets = discard
+        if service_id is not None:
+            discard: collections.deque = collections.deque()  # type: ignore[type-arg]
+            keep: collections.deque = collections.deque()  # type: ignore[type-arg]
+            for conn in self.conns:
```
NoahStapp: We need to either hold a lock here or iterate over a copy of `self.conns`. Otherwise a concurrent `checkin` or `_get_conn` call, both of which modify `self.conns`, would trigger a `RuntimeError: deque mutated during iteration` error.

pymongo/asynchronous/pool.py
```diff
@@ -903,9 +932,8 @@ async def update_is_writable(self, is_writable: Optional[bool]) -> None:
         Pool.
         """
         self.is_writable = is_writable
-        async with self.lock:
-            for _socket in self.conns:
-                _socket.update_is_writable(self.is_writable)  # type: ignore[arg-type]
+        for _socket in self.conns:
```
NoahStapp: Same here as above, we need to hold a lock or iterate over a copy of self.conns to avoid a potential `RuntimeError: deque mutated during iteration`.

pymongo/asynchronous/pool.py
```diff
@@ -1379,28 +1410,29 @@ async def checkin(self, conn: AsyncConnection) -> None:
...
                 close_conn = False
-                async with self.lock:
-                    # Hold the lock to ensure this section does not race with
-                    # Pool.reset().
-                    if self.stale_generation(conn.generation, conn.service_id):
-                        close_conn = True
-                    else:
-                        conn.update_last_checkin_time()
-                        conn.update_is_writable(bool(self.is_writable))
+                if self.stale_generation(conn.generation, conn.service_id):
```
NoahStapp: Does the old code's comment of `# Hold the lock to ensure this section does not race with Pool.reset()` no longer apply here?

pymongo/asynchronous/pool.py
```diff
@@ -813,39 +832,49 @@ async def _reset(
         interrupt_connections: bool = False,
     ) -> None:
         old_state = self.state
-        async with self.size_cond:
-            if self.closed:
-                return
+        if self.closed:
```
NoahStapp: Checking `self.closed` outside of the lock can lead to a similar race as above in `ready()`, where two concurrent calls of `_reset()` end up double-publishing a `PoolClosedEvent`.

NoahStapp (changes requested):
Great start, welcome to the PyMongo codebase!

Since this is a performance-motivated change, can you add benchmark results to the PR description showing the reported 20% improvement in a small-read scenario?

I'd also be curious to see how these changes affect other scenarios, as several places now acquire multiple locks rather than just the one. For situations with low contention, this is likely slower.

pymongo/asynchronous/pool.py
sophiayangDB: Opting to hold the lock since copying and immediately sending to garbage collection for setting a single variable doesn't seem worth it

pymongo/asynchronous/pool.py
sophiayangDB: Good catch! Opting to hold the lock since this doesn't seem that expensive.
(Just noticed I accidentally dropped the `service_id is None` case, fixing that)

pymongo/asynchronous/pool.py
sophiayangDB: Guess who didn't read the comments 😅 Double-checking doesn't seem to be worth it, reverting back to the old version here.

pymongo/asynchronous/pool.py
sophiayangDB: Noted!

pymongo/asynchronous/pool.py
sophiayangDB: In theory, the new locks would protect all the necessary resources. In practice there does seem to be a race here. Still thinking it through 🤔

NoahStapp:
Have you run performance testing to get the actual improvement for small reads, as well as the impact these changes have on other workloads?

pymongo/asynchronous/pool.py
```diff
@@ -709,6 +709,11 @@ def __init__(
         :param options: a PoolOptions instance
         :param is_sdam: whether to call hello for each new AsyncConnection
         """
+        # Main lock only used to protect updating attributes.
+        # Avoid any additional work while holding the lock.
+        # If looping over an attribute, copy the container and do not take the lock.
+        self.lock = _async_create_lock()
```
NoahStapp: `self.lock` should be renamed to something more appropriate for its new, limited scope.

pymongo/asynchronous/pool.py
```diff
@@ -777,29 +784,41 @@ def __init__(
             )
         # Similar to active_sockets but includes threads in the wait queue.
         self.operation_count: int = 0
+        # This lock should be contended on every operation.
+        self._operation_count_lock = _async_create_lock()
+
+        self.active_contexts: set[_CancellationContext] = set()
+        self.active_sockets = 0
```
NoahStapp: These initializations don't need to move from their original places.

NoahStapp (changes requested):

sophiayangDB:
See performance metrics [here](https://docs.google.com/spreadsheets/d/1F8eRJGH-sov4IUc8MXClIMSRTsxO4YV7d-DnC853Ug8/edit?usp=sharing)

PR #2720, opened 2026-03-04
PYTHON-5744 [PoC] Client Side Support for OpenTelemetry

NoahStapp:
[PYTHON-5744](https://jira.mongodb.org/browse/PYTHON-5744)

## Changes in this PR

A PoC for implementing OpenTelemetry support in PyMongo. 

EDIT: I'm going to split the logging + monitoring refactor into a separate ticket. That maintains better separation of concerns and clarity. This PR will remain open as the artifact produced by the PoC work.

The bulk of the changes here are a refactor to consolidate our existing command logging and monitoring code as part of the addition of OpenTelemetry. The old logging and monitoring calls had a lot of duplication and reduced clarity for the code surrounding them. With this refactor, all handling of telemetry is centralized and streamlined, improving readability and maintainability.

## Test Plan

All current tests pass. The OpenTelemetry unified tests detailed in [DRIVERS-719](https://jira.mongodb.org/browse/DRIVERS-719) are not implemented here in the PoC.

An example OpenTelemetry span for an `insert` operation:
```
{
    "name": "insert",
    "context": {
        "trace_id": "0xe25a92db9a4128e88cb51d6a9b400733",
        "span_id": "0x86d36f6e2b948ea0",
        "trace_state": "[]"
    },
    "kind": "SpanKind.CLIENT",
    "parent_id": null,
    "start_time": "2026-03-04T21:07:37.254308Z",
    "end_time": "2026-03-04T21:07:37.286829Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "db.system": "mongodb",
        "db.namespace": "otel_test_db",
        "db.command.name": "insert",
        "db.query.summary": "insert otel_test_db.test_collection",
        "server.address": "localhost",
        "server.port": 27017,
        "network.transport": "tcp",
        "db.mongodb.driver_connection_id": 1,
        "db.collection.name": "test_collection",
        "db.mongodb.server_connection_id": 10
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "telemetry.sdk.language": "python",
            "telemetry.sdk.name": "opentelemetry",
            "telemetry.sdk.version": "1.39.1",
            "service.name": "unknown_service"
        },
        "schema_url": ""
    }
}
```

### Checklist for Author
- [ ] Did you update the changelog (if necessary)?
- [ ] Is there test coverage?
- [ ] Is any followup work tracked in a JIRA ticket? If so, add link(s).

### Checklist for Reviewer
[...]

PR #2822, opened 2026-05-27
PYTHON-5745 Consolidate command logging and monitoring logic

blink1073:
[PYTHON-5745]

## Changes in this PR

Introduces `pymongo/_telemetry.py` with a `_CommandTelemetry` context manager that unifies debug logging and APM event publishing into a single call site, eliminating duplicated telemetry blocks at each command call site. This is a prerequisite for PYTHON-5052 (OpenTelemetry support).

Based on the spike in #2720.

## Test Plan

- `just typing` passes
- `just lint` passes
- Full CI covers command monitoring and logging specs

### Checklist for Author
- [ ] Is there test coverage?

### Checklist for Reviewer


pymongo/asynchronous/bulk.py
```diff
@@ -337,82 +283,27 @@ async def unack_write(
...
-                serverHost=bwc.conn.address[0],
-                serverPort=bwc.conn.address[1],
-                serviceId=bwc.conn.service_id,
-            )
-        if bwc.publish:
-            cmd = bwc._start(cmd, request_id, docs)
-        try:
-            result = await bwc.conn.unack_write(msg, max_doc_size)  # type: ignore[func-returns-value, misc, override]
-            duration = datetime.datetime.now() - bwc.start_time
-            if result is not None:
```
blink1073, line 361: A note about this block and the equivalent in `client_bulk.py`: `unack_write` always returns `None` so there will never be a `result`.  The spec says that we always need to emit a log, so this logic was incorrect before.

PR #2852, opened 2026-06-04
PYTHON-5676 Consolidate command execution logic

aclark4life:
[PYTHON-5676](https://jira.mongodb.org/browse/PYTHON-5676)

## Changes in this PR

Consolidates command execution into a single code path — the ticket's definition of done ("all database operations are done from a central location").

Previously the same execution skeleton (start clock → publish STARTED event → send → receive → process response → check for errors → publish SUCCEEDED/FAILED event) was duplicated across multiple call sites. This PR introduces **`pymongo/asynchronous/command_runner.py`** (with a sync mirror auto-generated by synchro) whose private `_run_command()` owns that entire skeleton: APM event publishing, logging, the send/receive round trip, `$clusterTime` gossip, response processing, error checking, and auto-encryption decryption.

The public API surface of the new module is:

- **`run_command()`** — high-level entry point used by `pool.py`. Takes a raw `spec` dict, applies read concern, collation, `$clusterTime`, compression, auto-encryption, and CSOT, encodes it as OP_MSG, and delegates to `_run_command`.
- **`run_bulk_write_command()`** — entry point for collection and client bulk writes. Accepts a `_BulkWriteContextProto` (a Protocol capturing the fields needed from `_BulkWriteContextBase`) rather than individual connection/session/listener arguments, and passes bulk-specific flags to `_run_command`.
- **`run_cursor_command()`** — entry point for cursor-returning commands in `server.py`. Passes cursor-specific flags (exhaust `more_to_come`, `unpack_res`, `cursor_id`) to `_run_command`.

All call sites now route through these functions. No public API changes.

## Test Plan

No new tests: this is a behavior-preserving refactor. The existing APM/command-monitoring, command-logging, bulk write, and CRUD spec suites assert the exact event and log documents that any regression would change.

### Checklist for Author
- [ ] Did you update the changelog (if necessary)? — Not necessary; internal refactor with no user-facing behavior change.
- [x] Is there test coverage? — Covered by existing spec suites (behavior preserved).

### Checklist for Reviewer


aclark4life:
Failures related to #2853

pymongo/synchronous/client_bulk.py
```diff
@@ -236,78 +232,32 @@ def write_command(
...
-                driverConnectionId=bwc.conn.id,
-                serverConnectionId=bwc.conn.server_connection_id,
-                serverHost=bwc.conn.address[0],
-                serverPort=bwc.conn.address[1],
-                serviceId=bwc.conn.service_id,
-            )
-        if bwc.publish:
-            bwc._start(cmd, request_id, op_docs, ns_docs)
         try:
-            reply = bwc.conn.write_command(request_id, msg, bwc.codec)  # type: ignore[misc, arg-type]
```
NoahStapp, line 265: `AsyncConnection.write_command` doesn't appear to have any remaining callers, so it (and its sync counterpart) are dead code that should be removed.

pymongo/asynchronous/bulk.py
```diff
@@ -336,83 +287,34 @@ async def unack_write(
...
-                serverPort=bwc.conn.address[1],
-                serviceId=bwc.conn.service_id,
-            )
-        if bwc.publish:
-            cmd = bwc._start(cmd, request_id, docs)
-        try:
-            result = await bwc.conn.unack_write(msg, max_doc_size)  # type: ignore[func-returns-value, misc, override]
-            duration = datetime.datetime.now() - bwc.start_time
-            if result is not None:
-                reply = _convert_write_result(bwc.name, cmd, result)  # type: ignore[arg-type]
```
NoahStapp, line 365: `_convert_write_result` is also dead code now.

pymongo/asynchronous/bulk.py
```diff
@@ -250,81 +246,36 @@ async def write_command(
...
-                requestId=request_id,
-                operationId=request_id,
-                driverConnectionId=bwc.conn.id,
-                serverConnectionId=bwc.conn.server_connection_id,
-                serverHost=bwc.conn.address[0],
-                serverPort=bwc.conn.address[1],
-                serviceId=bwc.conn.service_id,
-            )
-        if bwc.publish:
-            bwc._start(cmd, request_id, docs)
```
NoahStapp, line 273: `_BulkWriteContext._start, _BulkWriteContext._succeed, _BulkWriteContext._fail, and _ClientBulkWriteContext._start` are all dead code.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,333 @@
...
+    from pymongo.asynchronous.pool import AsyncConnection
+    from pymongo.message import _OpMsg, _OpReply
+    from pymongo.monitoring import _EventListeners
+    from pymongo.pool_options import PoolOptions
+    from pymongo.typings import _Address, _DocumentOut, _DocumentType
+
+_IS_SYNC = False
+
+
+async def run_command(
```
NoahStapp: This number of kwargs, several of which directly gate which network behavior to use, is extremely hard to parse and makes every call site much more complex than before. What if we split those different paths into three public methods that all wrap a private `_run_command`: one for acknowledged commands, one for unacknowledged commands, and one for cursor commands? That way we still have one unified command execution point, but multiple entry points that don't require a huge list of kwargs on every invocation.

NoahStapp (changes requested):

pymongo/asynchronous/bulk.py
```diff
@@ -250,81 +249,36 @@ async def write_command(
...
+                start=bwc.start_time,
+                codec_options=bwc.codec,
+                op_id=bwc.op_id,
+                command_name=bwc.name,
+                use_conn_transport=True,
+                process_response=False,
+                decrypt_reply=False,
+            )
+            reply = result_docs[0]
+            # Process the response from the server.
```
blink1073: We should pass `process_response=True` and remove this line.

pymongo/asynchronous/client_bulk.py
```diff
@@ -236,78 +235,32 @@ async def write_command(
...
+                start=bwc.start_time,
+                codec_options=bwc.codec,
+                op_id=bwc.op_id,
+                command_name=bwc.name,
+                use_conn_transport=True,
+                process_response=False,
+                decrypt_reply=False,
+            )
+            reply = result_docs[0]
             # Process the response from the server.
```
blink1073: Same here

pymongo/asynchronous/client_bulk.py
```diff
@@ -327,81 +280,34 @@ async def unack_write(
...
-        if bwc.publish:
-            cmd = bwc._start(cmd, request_id, op_docs, ns_docs)
+        """Send an unacknowledged client-level batch write command."""
+        # Historically the STARTED log omits the ops/nsInfo while the published
+        # CommandStartedEvent includes them, so log ``cmd`` but publish a copy
+        # carrying those fields.
+        published = dict(cmd)
+        published["ops"] = op_docs
+        published["nsInfo"] = ns_docs
+        reply: Mapping[str, Any] = {"ok": 1}
```
blink1073: We can use the response from `run_unacknowledged_command`, right?

pymongo/asynchronous/pool.py
blink1073: I think this method (and `_raise_if_not_writable`) should be moved to the command runner logic.

blink1073 (changes requested):

pymongo/asynchronous/bulk.py
```diff
@@ -250,81 +249,32 @@ async def write_command(
...
-                    serverHost=bwc.conn.address[0],
-                    serverPort=bwc.conn.address[1],
-                    serviceId=bwc.conn.service_id,
-                    isServerSideError=isinstance(exc, OperationFailure),
-                )
-
-            if bwc.publish:
-                bwc._fail(request_id, failure, duration)
-            # Process the response from the server.
             if isinstance(exc, (NotPrimaryError, OperationFailure)):
```
blink1073: Should this also be handled by `run_acknowledged_command`?

pymongo/asynchronous/client_bulk.py
```diff
@@ -236,81 +235,31 @@ async def write_command(
...
-                    serviceId=bwc.conn.service_id,
-                    isServerSideError=isinstance(exc, OperationFailure),
-                )
-
-            if bwc.publish:
-                bwc._fail(request_id, failure, duration)
             # Top-level error will be embedded in ClientBulkWriteException.
             reply = {"error": exc}
-            # Process the response from the server.
             if isinstance(exc, OperationFailure):
```
blink1073: Same question

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,538 @@
...
+            service_id=conn.service_id,
+        )
+
+    reply: Optional[Union[_OpReply, _OpMsg]]
+    try:
+        if more_to_come:
+            reply = await conn.receive_message(None)
+        elif unacknowledged:
+            if use_conn_transport:
+                if not conn.is_writable:
```
blink1073: Hmm, now this is defined twice.  Maybe it should be a public method on `conn`.

blink1073 (changes requested):

pymongo/asynchronous/client_bulk.py
```diff
@@ -236,85 +235,31 @@ async def write_command(
...
-                    serviceId=bwc.conn.service_id,
-                )
-            if bwc.publish:
-                bwc._succeed(request_id, reply, duration)  # type: ignore[arg-type]
-            # Process the response from the server.
-            await self.client._process_response(reply, bwc.session)  # type: ignore[arg-type]
+            result_docs, _, _ = await run_acknowledged_command(
+                bwc.conn,  # type: ignore[arg-type]
+                cmd,
+                bwc.db_name,
```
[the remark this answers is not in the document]
blink1073: These seem worth investigating.   We may need to put `_process_response` back at the call site, but hopefully not.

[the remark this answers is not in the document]
NoahStapp: Is it a big deal if we need to but then the telemetry consolidation will resolve it?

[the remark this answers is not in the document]
blink1073: I think they're orthogonal concerns.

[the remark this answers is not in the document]
blink1073: The last commit only added a comment, I don't think it addressed this feedback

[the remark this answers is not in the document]
aclark4life: IIUC, we either accept changing the ordering or we put _process_response back at the call site? Can we determine if changing the ordering matters? I would say let's go ahead, but admittedly I don't understand the full context.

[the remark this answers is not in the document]
blink1073: The ordering doesn't bother me as much as the other comment about `including the special error gossip of {} vs exc.details when swallowing exceptions`, given that the tests still pass.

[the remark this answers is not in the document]
aclark4life: > The ordering doesn't bother me as much as the other comment about `including the special error gossip of {} vs exc.details when swallowing exceptions`, given that the tests still pass.

@NoahStapp suggested we check the spec and the appropriate spec appears to be the driver sessions spec, [specifically](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.md?plain=1#L749-L753):

> Whenever a driver receives a cluster time from a server it MUST compare it to the current highest seen cluster time for
the deployment. If the new cluster time is higher than the highest seen cluster time it MUST become the new highest seen
cluster time. Two cluster times are compared using only the BsonTimestamp value of the `clusterTime` embedded field (be
sure to include both the timestamp and the increment of the BsonTimestamp in the comparison). The signature field does
not participate in the comparison.

Since `run_acknowledged_command` calls `_process_response(response_doc)` before it calls `_check_command_response`, in an `OperationFailure` (a server error response that carries `$clusterTime`), gossip already happened inside `run_acknowledged_command` before the exception propagated to the `except` block. By the time `client_bulk` catches it, the spec requirement  "MUST compare it to the current highest seen cluster time" is already satisfied. For non-server errors like network failures there's no response document at all, so there's no `$clusterTime` to gossip — the original `_process_response({})` was also a no-op in that case. So I think it's OK to leave it as is.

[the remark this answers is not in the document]
blink1073: Okay sounds good, I say merge and iterate.

blink1073:
There are some merge conflicts

aclark4life:
> There are some merge conflicts

rebased

pymongo/asynchronous/pool.py
```diff
@@ -426,6 +427,11 @@ async def command(
         except BaseException as error:
             await self._raise_connection_failure(error)
 
+    def _raise_if_not_writable(self) -> None:
```
blink1073: Can we move this back to its original location to avoid a confusing diff?

aclark4life: I don't think so, since `if unacknowledged` was added on 396.

blink1073: I mean if it were moved back to right after `send_message`

blink1073: Then at least we'd see the diff within the method, rather than it looking like a new method.

blink1073: Whoops, the diff was collapsed, it should be after `receive_message`

blink1073: Well, not as clean as I hoped, but as close as we're gonna get, thanks!

blink1073 (dismissed):

aclark4life:
@blink1073 Thanks! Can you merge? I'm blocked because I committed last … or something.

blink1073:
@NoahStapp has a requested change, are you good to merge?

pymongo/asynchronous/command_encoder.py
NoahStapp, line 1: Do we actually need this as a separate file from `command_runner.py`? It contains only one method, and that method just calls methods from `command_runner.py` after some processing.

pymongo/asynchronous/server.py
```diff
@@ -158,152 +154,60 @@ async def run_operation(
...
-                database_name=dbn,
-            )
-
-        # Decrypt response.
-        client = operation.client  # type: ignore[assignment]
-        if client and client._encrypter:
-            if use_cmd:
-                decrypted = await client._encrypter.decrypt(reply.raw_command_response())
-                docs = _decode_all_selective(decrypted, operation.codec_options, user_fields)
+                res["cursor"]["nextBatch"] = docs
```
NoahStapp: This block is a [regression from master](https://github.com/mongodb/mongo-python-driver/blob/93cbd3fd69e295529025644cd44b20b473681fc3/pymongo/asynchronous/server.py#L226-L227).

pymongo/message.py
NoahStapp: `self.publish` is dead code that we can remove.

NoahStapp: Dead code, can be removed.

pymongo/asynchronous/bulk.py
```diff
@@ -253,83 +248,26 @@ async def write_command(
...
+            client=client,
+            session=bwc.session,  # type: ignore[arg-type]
+            listeners=bwc.listeners,
+            address=bwc.conn.address,
+            start=bwc.start_time,
+            codec_options=bwc.codec,
+            op_id=bwc.op_id,
+            command_name=bwc.name,
+            use_conn_transport=True,
+            decrypt_reply=False,
```
NoahStapp: This should have `set_conn_more_to_come=False` for clarity. The default value for `run_acknowledged_command` is `True`.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,536 @@
...
+    unacknowledged: bool = False,
+    speculative_hello: bool = False,
+    ensure_db: bool = False,
+    process_response: bool = True,
+    decrypt_reply: bool = True,
+    use_conn_transport: bool = False,
+    max_doc_size: int = 0,
+    more_to_come: bool = False,
+    set_conn_more_to_come: bool = True,
+    is_command_response: bool = True,
```
NoahStapp: Dead code now that [PYTHON-5713](https://jira.mongodb.org/browse/PYTHON-5713) is done.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,536 @@
...
+
+        if reply is not None:
+            if set_conn_more_to_come:
+                conn.more_to_come = reply.more_to_come
+            if unpack_res is not None:
+                docs = unpack_res(
+                    reply,
+                    cursor_id,
+                    codec_options,
+                    legacy_response=not is_command_response,
```
NoahStapp: [PYTHON-5713](https://jira.mongodb.org/browse/PYTHON-5713) also makes `legacy_response` dead code.

NoahStapp (changes requested):

pymongo/asynchronous/command_encoder.py
aclark4life: Possibly not … I combined them so we'll find out.

pymongo/synchronous/command_runner.py
NoahStapp, line 1: This file shouldn't exist since it's the synchronous version of the deleted `pymongo/asynchronous/network.py`.

pymongo/asynchronous/server.py
```diff
@@ -158,152 +154,47 @@ async def run_operation(
...
-
-        # Decrypt response.
-        client = operation.client  # type: ignore[assignment]
-        if client and client._encrypter:
-            if use_cmd:
-                decrypted = await client._encrypter.decrypt(reply.raw_command_response())
-                docs = _decode_all_selective(decrypted, operation.codec_options, user_fields)
+        user_fields = _CURSOR_DOC_FIELDS if use_cmd else None
+
+        def _build_reply_doc(docs: list[dict[str, Any]], reply: Optional[_OpMsg]) -> _DocumentOut:  # noqa: ARG001
```
NoahStapp: This function can be removed entirely: I don't see another value passed to `reply_doc_builder`, can't we just inline the docs[0] itself within `run_cursor_command`?

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,675 @@
...
+    if client and client._encrypter and reply and decrypt_reply:
+        decrypted = await client._encrypter.decrypt(reply.raw_command_response())
+        docs = cast(
+            "list[dict[str, Any]]", _decode_all_selective(decrypted, codec_options, user_fields)
+        )
+
+    return docs, reply, duration
+
+
+async def run_acknowledged_command(
```
NoahStapp: This and `run_unacknowledged_command` are nearly identical besides a few kwargs and which fixed kwargs they pass through to `_run_command`. Will the logging/monitoring consolidation refactor allow us to either make these more distinct or reduce duplication? Or should we figure out a way to do that here in this PR?

aclark4life:
@NoahStapp @blink1073 GTG here I think …

NoahStapp (changes requested):

pymongo/synchronous/command_runner.py
blink1073: Did you mean to make this comment on `command_encoder.py`?  I would expect `network.py` to be deleted in both, and `command_runner.py` added to both.  If you agree, I'll also see about forcing git to recognize the link between the old and new files.

pymongo/asynchronous/command_runner.py
aclark4life: > This and `run_unacknowledged_command` are nearly identical besides a few kwargs and which fixed kwargs they pass through to `_run_command`. Will the logging/monitoring consolidation refactor allow us to either make these more distinct or reduce duplication? Or should we figure out a way to do that here in this PR?

Those were created in response to [this](https://github.com/mongodb/mongo-python-driver/pull/2852#discussion_r3363814768), so I'm fine with leaving them for now. Handing off to @blink1073 !

pymongo/synchronous/server.py
```diff
@@ -335,7 +226,7 @@ def run_operation(
                 duration=duration,
                 request_id=request_id,
                 from_command=use_cmd,
-                docs=docs,
+                docs=docs,  # type: ignore[arg-type]
             )
```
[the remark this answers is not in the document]
blink1073: the type ignore is actually needed

pymongo/asynchronous/server.py
```diff
@@ -335,7 +226,7 @@ async def run_operation(
                 duration=duration,
                 request_id=request_id,
                 from_command=use_cmd,
-                docs=docs,
+                docs=docs,  # type: ignore[arg-type]
             )
```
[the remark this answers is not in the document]
blink1073: the type ignore is actually needed

pymongo/asynchronous/command_runner.py
NoahStapp: The current code doesn't resolve my earlier comment that you linked: we still have a huge list of kwargs that's very difficult to parse and makes every call site much more complex than the pre-consolidation code.

pymongo/synchronous/command_runner.py
NoahStapp: Yes, sorry.

pymongo/synchronous/bulk.py
```diff
@@ -253,83 +248,27 @@ def write_command(
...
+            listeners=bwc.listeners,
+            address=bwc.conn.address,
+            start=bwc.start_time,
+            codec_options=bwc.codec,
+            op_id=bwc.op_id,
+            command_name=bwc.name,
+            use_conn_transport=True,
+            decrypt_reply=False,
+            set_conn_more_to_come=False,
+        )
```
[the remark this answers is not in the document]
blink1073: Updated

pymongo/asynchronous/bulk.py
```diff
@@ -253,83 +248,27 @@ async def write_command(
...
+            set_conn_more_to_come=False,
+        )
```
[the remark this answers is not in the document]
blink1073: Updated

pymongo/asynchronous/server.py
blink1073: Done

pymongo/synchronous/command_runner.py
blink1073: No, we'd have to do a rebase-merge to preserve the link between the files.

pymongo/asynchronous/command_runner.py
blink1073: I pushed a refactor, but I still want to look over it in context it before re-requesting review.

blink1073: Okay, I'm happy with the current state with `run_command`, `run_bulk_write_command` and `run_cursor_command`.  I've minimized the amount of kwargs needed as much as possible.  `run_command` inherits the large number of kwargs in `AsyncConnection.command`.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,566 @@
...
+
+"""Encoding and execution of commands over a connection.
+
+The public :func:`command` entry point applies read preference, read concern,
+collation, ``$clusterTime``, auto-encryption, and CSOT to a command spec,
+encodes it as an OP_MSG message, and then delegates to :func:`run_command`.
+
+Three public entry points each wrap the private :func:`_run_command`:
+
+- :func:`run_command` — standard network-transport commands (acknowledged or
```
NoahStapp: What's the difference between "network-transport" and "connection-transport"?

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,566 @@
...
+    from pymongo.pool_options import PoolOptions
+    from pymongo.read_concern import ReadConcern
+    from pymongo.read_preferences import _ServerMode
+    from pymongo.typings import _Address, _CollationIn, _DocumentOut, _DocumentType
+    from pymongo.write_concern import WriteConcern
+
+_IS_SYNC = False
+
+
+class _BulkWriteContextProto(Protocol):
```
NoahStapp: What does this do that the existing `_BulkWriteContext` does not?

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,566 @@
...
+    msg: bytes,
+    *,
+    client: Optional[AsyncMongoClient[Any]],
+    orig: Optional[MutableMapping[str, Any]] = None,
+    max_doc_size: int = 0,
+    unacknowledged: bool = False,
+) -> tuple[list[dict[str, Any]], Optional[_OpMsg], datetime.timedelta]:
+    """Send a bulk write batch and return ``(docs, reply, duration)``.
+
+    :param bwc: A bulk write context supplying the connection, session, listeners, etc.
```
NoahStapp: Missing several parameter docstrings here.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,566 @@
...
+    user_fields: Optional[Mapping[str, Any]] = None,
+    pool_opts: Optional[PoolOptions] = None,
+    max_doc_size: int = 0,
+    more_to_come: bool = False,
+    unpack_res: Optional[Callable[..., Any]] = None,
+    cursor_id: Optional[int] = None,
+) -> tuple[list[dict[str, Any]], Optional[_OpMsg], datetime.timedelta]:
+    """Run a cursor ``find``/``getMore`` operation over ``conn``.
+
+    :param more_to_come: Receive only, without sending (exhaust ``getMore``).
```
NoahStapp: Same here.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,566 @@
...
+    :param read_concern: The read concern for this command.
+    :param parse_write_concern_error: Whether to parse the ``writeConcernError``
+        field in the command response.
+    :param collation: The collation for this command.
+    :param compression_ctx: optional compression Context.
+    :param unacknowledged: True if this is an unacknowledged command.
+    :param user_fields: Response fields that should be decoded
+        using the TypeDecoders from codec_options, passed to
+        bson._decode_all_selective.
+    :param exhaust_allowed: True if we should enable OP_MSG exhaustAllowed.
```
NoahStapp, line 479: Missing a parameter docstring for `write_concern`.

NoahStapp (changes requested):

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,566 @@
...
+    pool_opts: Optional[PoolOptions] = None,
+    unacknowledged: bool = False,
+    speculative_hello: bool = False,
+    ensure_db: bool = False,
+    process_response: bool = True,
+    decrypt_reply: bool = True,
+    use_conn_transport: bool = False,
+    max_doc_size: int = 0,
+    more_to_come: bool = False,
+    set_conn_more_to_come: bool = True,
```
NoahStapp:
```suggestion
    set_conn_more_to_come: bool = False,
```

Two of the three call sites that in turn call `_run_command` pass `set_conn_more_to_come=False`, suggesting that it should be the default.

NoahStapp (changes requested):

pymongo/asynchronous/command_runner.py
blink1073: That was stale, I updated the docstring

pymongo/asynchronous/command_runner.py
blink1073: Ah, good call, we're already importing from `message.py`.  I removed the protocol.

pymongo/asynchronous/command_runner.py
blink1073: Updated

pymongo/asynchronous/command_runner.py
blink1073: Updated

pymongo/asynchronous/command_runner.py
blink1073: Done

pymongo/asynchronous/command_runner.py
blink1073: Done

pymongo/asynchronous/server.py
```diff
@@ -159,152 +155,42 @@ async def run_operation(
...
+            session=operation.session,  # type: ignore[arg-type]
+            listeners=listeners,
+            address=conn.address,
+            start=start,
+            codec_options=operation.codec_options,
+            user_fields=user_fields,
+            command_name=operation.name,
+            pool_opts=conn.opts,
+            max_doc_size=max_doc_size,
+            more_to_come=bool(more_to_come),
```
NoahStapp: I don't think we need the `bool` here, `more_to_come = operation.conn_mgr and operation.conn_mgr.more_to_come` is already a boolean value.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+    max_doc_size: int = 0,
+    more_to_come: bool = False,
+    set_conn_more_to_come: bool = False,
+    unpack_res: Optional[Callable[..., Any]] = None,
+    cursor_id: Optional[int] = None,
+) -> tuple[list[dict[str, Any]], Optional[_OpMsg], datetime.timedelta]:
+    """Send ``msg`` over ``conn`` and return ``(docs, reply, duration)``.
+
+    Private shared implementation. Use :func:`run_command`,
+    :func:`run_bulk_write_command`, or :func:`run_cursor_command` instead.
```
NoahStapp:
```suggestion
    Private shared implementation. Should not be called directly outside this module. Use :func:`run_command`, :func:`run_bulk_write_command`, or :func:`run_cursor_command` instead.
```

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+
+    It publishes the
```
NoahStapp:
```suggestion
    Publishes the
```

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+    """Send a bulk write batch and return ``(docs, reply, duration)``.
+
+    :param bwc: Bulk write context supplying the connection, session, listeners, etc.
+    :param cmd: The encoded command document.
+    :param request_id: The request id of the encoded message.
+    :param msg: The encoded bytes to send.
+    :param client: The AsyncMongoClient, for ``$clusterTime`` gossip and logging.
+    :param orig: The original command document for the APM ``STARTED`` event;
+        defaults to ``cmd``.
+    :param max_doc_size: The largest document size; passed to ``conn.send_message``.
```
NoahStapp:
```suggestion
    :param max_doc_size: The largest document size in the batch, passed to ``conn.send_message``.
```

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+    :param session: The session to update from the response.
+    :param listeners: The event listeners, or ``None`` to disable APM.
+    :param address: The (host, port) of ``conn`` for APM events.
+    :param start: The ``datetime`` the operation began, for duration timing.
+    :param codec_options: The CodecOptions used to decode the reply.
+    :param command_name: The command name for APM events.
+    :param user_fields: Response fields decoded with the codec's TypeDecoders.
+    :param pool_opts: PoolOptions forwarded to ``_check_command_response``.
+    :param max_doc_size: The largest document size, for ``conn.send_message``.
+    :param more_to_come: Receive only, without sending (exhaust ``getMore``).
```
NoahStapp:
```suggestion
    :param more_to_come: Receive only, without sending. Used for ``getMore`` on exhaust cursors.
```

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+    read_concern: Optional[ReadConcern] = None,
+    parse_write_concern_error: bool = False,
+    collation: Optional[_CollationIn] = None,
+    compression_ctx: Union[SnappyContext, ZlibContext, ZstdContext, None] = None,
+    unacknowledged: bool = False,
+    user_fields: Optional[Mapping[str, Any]] = None,
+    exhaust_allowed: bool = False,
+    write_concern: Optional[WriteConcern] = None,
+) -> _DocumentType:
+    """Encode and execute a command over ``conn``, or raise socket.error.
```
NoahStapp, line 453: Inconsistent docstring formatting, some parameters are missing periods and some aren't.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+    :param parse_write_concern_error: Whether to parse the ``writeConcernError``
+        field in the command response.
+    :param collation: The collation for this command.
+    :param compression_ctx: optional compression Context.
+    :param unacknowledged: True if this is an unacknowledged command.
+    :param user_fields: Response fields that should be decoded
+        using the TypeDecoders from codec_options, passed to
+        bson._decode_all_selective.
+    :param exhaust_allowed: True if we should enable OP_MSG exhaustAllowed.
+    :param write_concern: The write concern for this command; applied via CSOT.
```
NoahStapp:
```suggestion
    :param write_concern: The write concern for this command. Applied via CSOT.
```

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+        unpack_res=unpack_res,
+        cursor_id=cursor_id,
+    )
+
+
+async def run_command(
+    conn: AsyncConnection,
+    dbname: str,
+    spec: MutableMapping[str, Any],
+    is_mongos: bool,  # noqa: ARG001
```
NoahStapp: Dead argument, we don't use `is_mongos` at all.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+        elif use_conn_transport:
+            if session is not None and session._starting_transaction:
+                session._transaction.set_in_progress()
+            await conn.send_message(msg, max_doc_size)
+            reply = await conn.receive_message(request_id)
+        else:
+            await async_sendall(conn.conn.get_conn, msg)
+            reply = await async_receive_message(conn, request_id)
+
+        if reply is not None:
```
NoahStapp:
```suggestion
        if reply:
```

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+                    user_fields=user_fields,
+                )
+            else:
+                docs = reply.unpack_response(codec_options=codec_options, user_fields=user_fields)
+            response_doc = docs[0]
+            if not conn.ready:
+                cluster_time = response_doc.get("$clusterTime")
+                if cluster_time:
+                    conn._cluster_time = cluster_time
+            if process_response and client:
```
NoahStapp: Do we need `process_response` both here and as a kwarg at all? We already gate on `if reply is not None` above, which will always fail for an unacknowledged write.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+        codec_options=codec_options,
+        user_fields=user_fields,
+        orig=orig,
+        check=check,
+        allowable_errors=allowable_errors,
+        parse_write_concern_error=parse_write_concern_error,
+        speculative_hello=speculative_hello,
+        unacknowledged=unacknowledged,
+        process_response=not unacknowledged,
+        decrypt_reply=not unacknowledged,
```
NoahStapp: We don't need to set `decrypt_reply` explicitly here: the `if client and client._encrypter and reply and decrypt_reply` check already requires `reply` to be truthy.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+        start=bwc.start_time,
+        codec_options=bwc.codec,
+        op_id=bwc.op_id,
+        command_name=bwc.name,
+        orig=orig,
+        max_doc_size=max_doc_size,
+        unacknowledged=unacknowledged,
+        use_conn_transport=True,
+        decrypt_reply=False,
+        set_conn_more_to_come=False,
```
NoahStapp: This is the default value, no need to pass it explicitly.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+        start=start,
+        codec_options=codec_options,
+        user_fields=user_fields,
+        command_name=command_name,
+        pool_opts=pool_opts,
+        ensure_db=True,
+        use_conn_transport=True,
+        max_doc_size=max_doc_size,
+        more_to_come=more_to_come,
+        set_conn_more_to_come=False,
```
NoahStapp: This is the default value, no need to pass it explicitly.

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,572 @@
...
+            reply = None
+            docs: list[dict[str, Any]] = [{"ok": 1}]
+        elif use_conn_transport:
+            if session is not None and session._starting_transaction:
+                session._transaction.set_in_progress()
+            await conn.send_message(msg, max_doc_size)
+            reply = await conn.receive_message(request_id)
+        else:
+            await async_sendall(conn.conn.get_conn, msg)
+            reply = await async_receive_message(conn, request_id)
```
NoahStapp: This section is hard to parse with its mixing of "ack or unack" and "socket or transport" logic. First determining which network protocol to use and then if a reply is expected is more readable:

```
send, receive = (conn.send_message,
  conn.receive_message) if use_conn_transport else (async_sendall, async_receive_message)

if more_to_come:
    ...
elif unacknowledged:
   ...
```

NoahStapp (changes requested):

pymongo/asynchronous/command_runner.py
blink1073: Updated for consistency

pymongo/asynchronous/command_runner.py
blink1073: Applied locally to avoid switching back and forth between local and remote commits

pymongo/asynchronous/command_runner.py
blink1073: Removed

pymongo/asynchronous/command_runner.py
blink1073: Applied locally

pymongo/asynchronous/command_runner.py
blink1073: Good call, removed

pymongo/asynchronous/command_runner.py
blink1073: Removed

pymongo/asynchronous/command_runner.py
blink1073: Removed

pymongo/asynchronous/command_runner.py
blink1073: Removed

pymongo/asynchronous/command_runner.py
blink1073: Consolidated.  I'm pushing now but won't re-request review until I've got it green and have reviewed the changes in context

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,545 @@
...
+                conn.server_connection_id,
+                op_id,
+                service_id=conn.service_id,
+                database_name=dbname,
+            )
+        raise
+
+    duration = datetime.datetime.now() - start
+    published_reply: _DocumentOut
+    published_reply = docs[0]
```
NoahStapp: Does this need to be two separate lines? Do we even need the `_DocumentOut` type hint?

pymongo/asynchronous/command_runner.py
```diff
@@ -0,0 +1,545 @@
...
+            conn.server_connection_id,
+            op_id,
+            service_id=conn.service_id,
+            speculative_hello=speculative_hello,
+            database_name=dbname,
+        )
+
+    if client and client._encrypter and reply and decrypt_reply:
+        decrypted = await client._encrypter.decrypt(reply.raw_command_response())
+        docs = cast(
```
NoahStapp, line 314: Do we need this cast? `_decode_all_selective` already returns `list[_DocumentType]` which is bound to `Mapping[str, Any]`.

NoahStapp (changes requested):

pymongo/asynchronous/command_runner.py
blink1073: We don't need the first line at all, removed

pymongo/asynchronous/command_runner.py
blink1073: Yeah we need it, because mypy

NoahStapp: What do you mean? The following passing typechecks:

`docs = _decode_all_selective(decrypted, codec_options, user_fields) # type: ignore[arg-type]`

blink1073: Yeah, with an ignore.  Six of one, half dozen of the other.

NoahStapp (approved):

pymongo/asynchronous/command_runner.py
NoahStapp: Fair enough 😅 !

blink1073:
@aclark4life I'll let you merge and close the ticket

PR #2853, opened 2026-06-05
PYTHON-5861 Fix macOS ARM64 test regressions

blink1073:
PYTHON-5861

## Changes in this PR

Fix macOS ARM64 CI test regressions introduced after merging PYTHON-3689.

## Test Plan

Evergreen patch: https://spruce.corp.mongodb.com/version/6a2229b13ce7700007a16be9/tasks

### Checklist for Author
- [x] Is there test coverage? — changes are to tests only

### Checklist for Reviewer


test/asynchronous/test_pooling.py
```diff
@@ -568,8 +568,7 @@ async def test_max_pool_size(self):
 
         async def f():
             for _ in range(5):
-                await collection.find_one({"$where": delay(0.1)})
-                assert len(cx_pool.conns) <= max_pool_size
```
NoahStapp, line 572: Do we not need this assert anymore?

NoahStapp (changes requested):

test/asynchronous/test_pooling.py
blink1073: I didn't think it made sense with the test restructure.  I added back in, let's see if it passes

test/asynchronous/test_pooling.py
```diff
@@ -600,7 +600,7 @@ async def test_max_pool_size_none(self):
 
         async def f():
             for _ in range(5):
-                await collection.find_one({"$where": delay(0.1)})
+                await collection.find_one({})
```
NoahStapp: Does this test no longer need the delay?

NoahStapp (changes requested):

test/asynchronous/test_pooling.py
blink1073: Good call out, I reverted these changes and added the skips instead

test/asynchronous/test_concurrency.py
```diff
@@ -16,17 +16,24 @@
...
+from test.utils_shared import delay
 
 _IS_SYNC = False
 
 
 class TestAsyncConcurrency(AsyncIntegrationTest):
     async def _task(self, client):
-        await client.db.test.find_one({})
-        await asyncio.sleep(0.20)
+        await client.db.test.find_one({"$where": delay(0.20)})
```
NoahStapp, line 31: I'm a little confused by the changes in this file. We've added `$where` but then also skip the test now that `$where` is too slow on macOS CI?

NoahStapp (changes requested):

test/asynchronous/test_concurrency.py
blink1073: There was an actual race condition before.  We do need the `$where` clause in this test.

NoahStapp (approved):

PR #2855, opened 2026-06-05
PYTHON-5796: Reduce insert_one({}) collection-priming calls in test suite

sophiayangDB:
PYTHON-5796

## Changes in this PR
- Replace `insert_one({})` with explicit `create_collection()` calls where the intent is solely to test against an existing collection
- When possible, add collection step to setUp/asyncSetUp on relevant base classes

## Test Plan
Only modified test files.

### Checklist for Author
- [ ] Is there test coverage?

### Checklist for Reviewer


sophiayangDB:
MacOS test failures are unrelated

test/asynchronous/test_index_management.py
```diff
@@ -143,20 +142,16 @@ class TestSearchIndexProse(SearchIndexIntegrationBase):
     async def test_case_1(self):
         """Driver can successfully create and list search indexes."""
 
-        # Create a collection with the "create" command using a randomly generated name (referred to as ``coll0``).
```
sleepyStick, line 146: hmm i think we need to keep the randomly generated name cuz that's what the spec test says (https://github.com/mongodb/specifications/blob/master/source/index-management/tests/README.md#case-1-driver-can-successfully-create-and-list-search-indexes)

test/asynchronous/test_session.py
```diff
@@ -978,6 +978,12 @@ async def asyncSetUp(self):
         await super().asyncSetUp()
         self.listener = SessionTestListener()
         self.client = await self.async_rs_or_single_client(event_listeners=[self.listener])
+        await self.client.pymongo_test.drop_collection("test")
+        await self.client.pymongo_test.create_collection("test")
+
+    async def asyncTearDown(self):
+        await self.client.pymongo_test.drop_collection("test")
```
sleepyStick: nit pick / minor comment: does the drop collection need to be here if its also already done in the asyncSetup?

test/test_cursor.py
```diff
@@ -1623,6 +1620,7 @@ def test_get_item(self):
             self.db.test.find_raw_batches()[0]
 
     def test_collation(self):
+        self.db.test.insert_one({})
```
sleepyStick, line 1623: silly question: should this be a create collection?

sophiayangDB: We need at least one document since the next line calls `next()`, an empty collection would error.

test/asynchronous/test_session.py
sophiayangDB: Oh good point, I don't think it does.

test/asynchronous/test_index_management.py
sleepyStick: ah, i see its been moved to the setup,, can we either this comment up with the line of code maybe?

sleepyStick (approved):
nice work!

PR #2857, opened 2026-06-08
PYTHON-1894: Client-side errors in in-progress transactions must not change transaction state

sophiayangDB:
PYTHON-1894

## Changes in this PR
Transactions are only advanced from STARTING -> IN_PROGRESS after the first operation reaches the server response stage. This means that client-side errors do not advance state.

## Test Plan
Removed the temporary unified test skips and expected failures for this case.

### Checklist for Author
- [ ] Is there test coverage?

### Checklist for Reviewer


pymongo/asynchronous/client_session.py
```diff
@@ -433,6 +433,7 @@ def __init__(self, opts: Optional[TransactionOptions], client: AsyncMongoClient[
         self.attempt = 0
         self.client = client
         self.has_completed_command = False
+        self.has_sent_command = False
```
sleepyStick: silly question, but it seems like this is only ever set and never used? what purpose does it serve?

.evergreen/remove-unimplemented-tests.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 PYMONGO=$(dirname "$(cd "$(dirname "$0")" || exit; pwd)")
 
-rm $PYMONGO/test/transactions/legacy/errors-client.json  # PYTHON-1894
```
sleepyStick, line 4: did you run the resync spec script after removing this line? i don't think i see this file added?

pymongo/asynchronous/client_session.py
sophiayangDB: oops that's from an old commit

.evergreen/remove-unimplemented-tests.sh
sophiayangDB: oops

sophiayangDB: the file does not exist anymore

sleepyStick (approved):
good work!

PR #2858, opened 2026-06-08
PYTHON-5867 - Close sockets on interruption or cancellation during async connection creation

NoahStapp:
[PYTHON-5867](https://jira.mongodb.org/browse/PYTHON-5867)

## Changes in this PR

Close sockets opened as part of async connection creation when an interruption or cancellation occurs.

## Test Plan

Fix existing PyPy test failure.

### Checklist for Author
- [ ] Is there test coverage?

### Checklist for Reviewer


sophiayangDB (approved):
LGTM!

test/asynchronous/test_async_cancellation.py
```diff
@@ -16,14 +16,20 @@
 from __future__ import annotations
 
 import asyncio
+import functools
+import socket as _socket
+import ssl as _ssl
```
aclark4life: Do `_socket` and `_ssl` need to be private here ?

NoahStapp: Good catch, vestigial import naming from an earlier version that overrode them.

aclark4life:
How about a context manager to avoid repeated `BaseException` handling ? 

E.g. add something like:

```
@contextmanager
def _close_on_error(closer):
    try:
        yield
    except BaseException:
        closer()
        raise
```

So then:

```
try:
    ...
    return sock
except BaseException:
    sock.close()
    raise
```

would become something like:

```
with _close_on_error(sock.close):
    ...
    return sock
```

NoahStapp:
> How about a context manager to avoid repeated `BaseException` handling ?
> 
> E.g. add something like:
> 
> ```
> @contextmanager
> def _close_on_error(closer):
>     try:
>         yield
>     except BaseException:
>         closer()
>         raise
> ```
> 
> So then:
> 
> ```
> try:
>     ...
>     return sock
> except BaseException:
>     sock.close()
>     raise
> ```
> 
> would become something like:
> 
> ```
> with _close_on_error(sock.close):
>     ...
>     return sock
> ```

I like it, good cleanup!

pymongo/pool_shared.py
```diff
@@ -266,42 +279,37 @@ async def _async_configured_socket(
...
-        if _has_sni(False):
-            loop = asyncio.get_running_loop()
-            ssl_sock = await loop.run_in_executor(
-                None,
-                functools.partial(ssl_context.wrap_socket, sock, server_hostname=host),  # type: ignore[assignment, misc, unused-ignore]
-            )
-        else:
-            loop = asyncio.get_running_loop()
-            ssl_sock = await loop.run_in_executor(None, ssl_context.wrap_socket, sock)  # type: ignore[assignment, misc, unused-ignore]
+        with _cleanup_on_error(sock.close):
```
aclark4life, line 282: Hmmm, the exhaustive code review 🤖 found a "cross-thread fd-reuse hazard":

>  _async_configured_socket hands sock to a worker thread via `run_in_executor(wrap_socket, sock)` and asyncio can't stop that thread on cancel and `wrap_socket` calls sock.detach() before the slow handshake 

It suggests replacing the above `with _clean_on_error` block with

```
ssl_sock = await asyncio.shield(future)
```

and then after the `except (OSError, *SSLErrors) as exc` add another `except BaseException`:

```
except BaseException:
    future.add_done_callback(_close_executor_socket)
    raise
```

and add the proposed corresponding `_close_executor_socket` helper:

```
def _close_executor_socket(future):
    try:
        ssl_sock = future.result()
    except BaseException:
        pass
    else:
        ssl_sock.close()
```

aclark4life (changes requested):

NoahStapp:
This needs more iteration to determine the specific solution we want, closing for now.

PR #2864, opened 2026-06-10
PYTHON-5272 Implement TLS session resumption for connections

blink1073:
[PYTHON-5272](https://jira.mongodb.org/browse/PYTHON-5272)

## Changes in this PR

Added TLS session resumption to the connection pool, avoiding a full handshake on each new connection to the same server. Session reuse is active on the sync path unconditionally, and on the async path on Python 3.11 or later.  Most of the changes are the addition of unit tests to verify the cache behavior.

## Test Plan

Added unit and integration tests covering:
- `_SSLSessionCache` get/set behavior
- Pool creates a session cache when TLS is configured
- A session is stored after a successful connection
- The cached session is passed to `wrap_socket` on subsequent connections (sync, no server required)
- The cached session is set on the `SSLObject` before the handshake on Python 3.11+ (async, no server required)

### Checklist for Author
- [x] Is there test coverage? — Yes
- [ ] Is any followup work tracked in a JIRA ticket? If so, add link(s). — Async session reuse on Python < 3.11 deferred; no separate ticket yet

### Checklist for Reviewer


test/asynchronous/test_ssl.py
```diff
@@ -128,6 +128,298 @@ def test_config_ssl(self):
...
+        self.assertIsNone(cache[0])
+        cache[0] = "session"
+        self.assertEqual(cache[0], "session")
+        cache[0] = "new_session"
+        self.assertEqual(cache[0], "new_session")
+
+    @unittest.skipUnless(_IS_SYNC, "Tests sync wrap_socket path only")
+    def test_tls_session_reused_on_second_connection(self):
+        """Cached TLS session is passed to wrap_socket on subsequent connections."""
+        import unittest.mock as mock
```
sleepyStick: can imports be moved to the top of the file? -- there are a few repeating imports in each of these test cases >.<

sleepyStick (changes requested):

test/asynchronous/test_ssl.py
blink1073: Done!

test/asynchronous/test_ssl.py
```diff
@@ -140,8 +143,6 @@ def test_ssl_session_cache(self):
     @unittest.skipUnless(_IS_SYNC, "Tests sync wrap_socket path only")
     def test_tls_session_reused_on_second_connection(self):
         """Cached TLS session is passed to wrap_socket on subsequent connections."""
-        import unittest.mock as mock
-
         from pymongo.pool_shared import _configured_socket_interface
```
sleepyStick: wait is there a reason this one cant be moved up? (it looks like it was moved up in the sync file?)

sleepyStick (changes requested):
sorry one last question!

test/asynchronous/test_ssl.py
blink1073: Fixed

sleepyStick (approved):

PR #2877, opened 2026-06-16
PYTHON-5885 & PYTHON-5954 Client Backpressure with baseBackoffMS

NoahStapp:
[PYTHON-5885](https://jira.mongodb.org/browse/PYTHON-5885)

## Changes in this PR

Adds `baseBackoffMS` support for backoff calculation on retryable overload errors.

## Test Plan

Adds a new prose test `test_05_overload_errors_with_retryafterms_override_backoff` to verify the `baseBackoffMS` behavior. Note that this new test requires a server with `baseBackoffMS` support.

### Checklist for Author
- [X] Did you update the changelog (if necessary)?
- [X] Is there test coverage?

### Checklist for Reviewer


NoahStapp:
Note: [SERVER-130142](https://jira.mongodb.org/browse/SERVER-130142) will rename `retryAfterMS` to `baseBackoffMS`. That change will require the spec to update as well as this PR.

pymongo/errors.py
```diff
@@ -33,10 +33,16 @@
...
+    def __init__(
+        self,
+        message: str = "",
+        error_labels: Optional[Iterable[str]] = None,
+        base_backoff_ms: Optional[int] = None,
+    ) -> None:
         super().__init__(message)
         self._message = message
         self._error_labels = set(error_labels or [])
+        self._base_backoff_ms = base_backoff_ms
```
Jibola: Why does the base `PyMongoError` also need to initialize `self._base_backoff_ms`? It's only a viable label on `OperationFailure`.

Jibola (changes requested):

aclark4life:
`retryAfterMS` is now `baseBackoffMS`correct? Is so can you update the PR desc ?

pymongo/asynchronous/mongo_client.py
```diff
@@ -2843,26 +2844,29 @@ async def run(self) -> T:
...
+                    if isinstance(exc_to_check, (ConnectionFailure, OperationFailure)):
                         # ConnectionFailures do not supply a code property
-                        exc_code = getattr(exc, "code", None)
-                        overloaded = exc.has_error_label("SystemOverloadedError")
+                        exc_code = getattr(exc_to_check, "code", None)
+                        overloaded = exc_to_check.has_error_label("SystemOverloadedError")
                         if overloaded:
                             self._max_retries = self._client.options.max_adaptive_retries
-                        always_retryable = exc.has_error_label("RetryableError") and overloaded
+                            self._base_backoff_ms = exc_to_check._base_backoff_ms
```
aclark4life: 🤖 suggests:

```
self._base_backoff_ms = getattr(exc_to_check, "_base_backoff_ms", None)
``` 

here and [here](https://github.com/mongodb/mongo-python-driver/pull/2877/changes#diff-8dfbbdba94166c58103fd44781f9133a77767f13c9b816030a631f29d99aee5bR2896) because  `_base_backoff_ms` is a property defined only on `OperationFailure` but `exc_to_check` can be a `ConnectionFailure` and the `if overloaded:` guard only checks the error label not the type.

aclark4life (changes requested):

pymongo/asynchronous/mongo_client.py
NoahStapp: ~The server can only attach `SystemOverloadedError` to `OperationFailure` errors, so this can't ever occur currently. We could add the `getattr` as a safeguard in case that changes in the future, but it also makes that existing contract with the server less clear.~

Actually, never mind. Having to add a type override here instead of `getattr` is silly.

pymongo/asynchronous/helpers.py
```diff
@@ -103,9 +106,13 @@ def __init__(
         self.backoff_initial = backoff_initial
         self.backoff_max = backoff_max
 
-    def backoff(self, attempt: int) -> float:
-        """Return the backoff duration for the given attempt."""
-        return _backoff(max(0, attempt - 1), self.backoff_initial, self.backoff_max)
+    def backoff(self, attempt: int, base_backoff: Optional[float] = None) -> float:
+        """Return the actual backoff duration for the given attempt and base backoff."""
+        return _backoff(
+            max(0, attempt),
```
aclark4life, line 112: What is the actual backoff duration and how does changing from `attempt -1` to `attempt` help to implement ?

NoahStapp: `attempt - 1 -> attempt` is to comply with the latest spec, which conforms to the server's exponential backoff formula for consistency. The actual backoff duration for two different `backoff` calls with the same `attempt` argument can differ due to `base_backoff`.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2794,6 +2794,7 @@ def __init__(
         self._attempt_number = 0
         self._is_run_command = is_run_command
         self._is_aggregate_write = is_aggregate_write
+        self._base_backoff_ms: Optional[float] = None
```
aclark4life, line 2797: Should this be `int` as it is in `errors.py` or vice versa ?

NoahStapp: Good catch, `errors.py` should use a float for consistency.

test/asynchronous/test_client_backpressure.py
```diff
@@ -294,6 +294,70 @@ async def test_04_overload_retries_limited_configured(self):
...
+                # 9. Run the following command to disable `baseBackoffMS` on overload errors.
+                await client.admin.command("setParameter", 1, externalClientBaseBackoffMS=0)
+
+        # 10. Assert absolute bounds on each run's duration.
+        # A run can never be faster than the sum of its backoffs.
+        # With jitter pinned to 1, the default backoffs are 0.2 + 0.4 = 0.6s
+        # and the baseBackoffMS=50 backoffs are 0.1 + 0.2 = 0.3s.
+        self.assertGreaterEqual(exponential_backoff_time, 0.6)
+        self.assertGreaterEqual(with_base_backoff_ms_time, 0.3)
+        self.assertLess(with_base_backoff_ms_time, 0.6)
```
aclark4life, line 359: Do we care if unrelated perf issues push this past 0.6?

NoahStapp: Perf issues are an annoying thorn in our side for this type of test: it's more resilient to assert `with_base_backoff_ms_time < exponential_backoff_time`, but that doesn't ensure that we backed off for the correct amount of time required by `base_backoff_ms`. I could see preferring the less accurate but also less flakey assertion over the current one, thoughts?

pymongo/asynchronous/mongo_client.py
```diff
@@ -2932,7 +2937,10 @@ async def run(self) -> T:
 
                 self._always_retryable = always_retryable
                 if overloaded:
-                    delay = self._retry_policy.backoff(self._attempt_number)
+                    delay = self._retry_policy.backoff(
+                        self._attempt_number,
+                        self._base_backoff_ms / 1000 if self._base_backoff_ms else None,
+                    )
```
aclark4life, line 2943: Ye Olde `None` confusion: if `_base_backoff_ms` is truthy, divide by 1000 else `None` but if the server sends `0` (which can presumably never happen?) then `_base_backoff_ms` is `None` instead of zero. If you go down this rabbit hole, maybe we do this instead:
```
self._base_backoff_ms / 1000 if self._base_backoff_ms is not None else None
```

aclark4life (dismissed):
LGTM pending optional nit fix suggestions (if accepted)!

test/asynchronous/test_client_backpressure.py
aclark4life: Yeah, we have a way to deal with flaky tests (I think?) so maybe this just goes into that category, unless we do something crazy like < `exponential_backoff_time` if `exponential_backoff_time == base_backoff_ms` 😂 IOW some conditional to compare them before using `exponential_backoff_time` instead of `base_backoff_ms` but that's starting to sound worse than just accepting flakiness.

pymongo/asynchronous/mongo_client.py
NoahStapp: The server should never send `baseBackoffMS = 0`, as doing so would result in no backoff. I don't think we need to complicate this edge case here as a result.

aclark4life (approved):
LGTM

Jibola (approved):

PR #2881, opened 2026-06-17
PYTHON-5862 Remove usage of uv lock file

aclark4life:
PYTHON-5862

## Changes in this PR

Removes the `uv.lock` file and all machinery around it. Keeping a lock file for a driver library causes more churn than it prevents — Dependabot raises weekly PRs that require manual regeneration, and the lock is redundant with the pinning strategy already applied in `pyproject.toml`.

### What changed

| File | Change |
|---|---|
| `uv.lock` | Deleted |
| `.pre-commit-config.yaml` | Removed `uv-lock` hook |
| `.evergreen/run-tests.sh` | Removed post-test lock-file cleanup |
| `.evergreen/scripts/setup-dev-env.sh` | Export `UV_NO_LOCK=1` unconditionally (before the CI guard) so all `uv` invocations — including `just resync` on Evergreen — never create or update a lock file |
| `.evergreen/scripts/setup_tests.py` | Set `UV_NO_LOCK=1`; always write `UV_EXCLUDE_NEWER` to `test-env.sh` (using existing env value when set, computing a 7-day-ago default otherwise) so the value survives into new shell sessions |
| `.evergreen/combine-coverage.sh` | Export `UV_NO_LOCK` and `UV_EXCLUDE_NEWER` **before** sourcing `setup-dev-env.sh`, which calls `uv sync` internally |
| `.github/workflows/test-python.yml` | Set `UV_NO_LOCK=1` at workflow level; add a `Set UV_EXCLUDE_NEWER` step in each uv-using job using `${UV_EXCLUDE_NEWER:-$(date ...)}` so a pre-set value is never overwritten |
| `CONTRIBUTING.md` | Update Dependabot triage guidance: check out PR, run `just lint`, run `just typing` if needed, commit and push |

### Why `UV_EXCLUDE_NEWER`

Without a lock file, `uv` resolves the latest available versions on every run. `UV_EXCLUDE_NEWER` sets a cutoff date (7 days ago by default, overridable) so resolution is stable across CI runs on the same day and avoids picking up packages published within the last week.

## Test Plan

CI-infrastructure change only — no driver logic altered. Verified by running the Evergreen patch and confirming no unexpected lock files are created.

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer


sleepyStick (dismissed):

NoahStapp:
Superseded by https://github.com/mongodb/mongo-python-driver/pull/2902.

PR #2882, opened 2026-06-17
PYTHON-5879 Skip coverage on macOS/Win64 encryption tests

aclark4life:
PYTHON-5879

## Changes in this PR

Adds a `cov` tag to non-standard Evergreen tasks that enable coverage (`COVERAGE=1`), then excludes those tasks from the macOS and Win64 encryption variants using `!.cov`.

Both sync and async encryption test suites each take >30 min on macOS/Win64 with coverage enabled, which combined exceeds the 60-minute task timeout. Linux (RHEL8) encryption variants still run the coverage tasks, so encryption code paths remain covered in the combined Codecov report.

### Companion non-coverage tasks

The cycling algorithm in `create_test_non_standard_tasks()` happens to assign python 3.13/3.14t/3.14 to all three `latest` topology slots. Since none of those is `ALL_PYTHONS[0]` ("3.10"), every "latest" non-standard task was getting `COVERAGE=1` and the `cov` tag — meaning `!.cov` removed **all** latest tasks from macOS/Win64, leaving nothing to activate in patch builds.

Fix: for each coverage task emitted, also emit a non-coverage companion with the same python/topology/version but no `COVERAGE` expansion and no `cov` tag. This gives macOS/Win64 encryption variants three "latest" tasks to run without coverage overhead:

- `test-non-standard-latest-python3.13-noauth-nossl-standalone`
- `test-non-standard-latest-python3.14t-noauth-ssl-replica-set`
- `test-non-standard-latest-python3.14-auth-ssl-sharded-cluster`

## Test Plan

Config change only — verified the generated `tasks.yml` and `variants.yml` reflect the new tags, selectors, and companion tasks correctly.

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/.evergreen/scripts/generate_config.py
+++ b/.evergreen/scripts/generate_config.py
@@ -648,8 +648,9 @@ def create_test_non_standard_tasks():
     tasks = []
     task_combos = set()
     # For each version and topology, rotate through the CPythons.
+    # Only the sharded_cluster topology runs on PRs to keep patch build size manageable.
     for (version, topology), python in zip_cycle(list(product(ALL_VERSIONS, TOPOLOGIES)), CPYTHONS):
-        pr = version == "latest"
+        pr = version == "latest" and topology == "sharded_cluster"
         task_combos.add((version, topology, python, pr))
     # For each PyPy and topology, rotate through the MongoDB versions.
     for (python, topology), version in zip_cycle(list(product(PYPYS, TOPOLOGIES)), ALL_VERSIONS):
--- a/.evergreen/scripts/generate_config.py
+++ b/.evergreen/scripts/generate_config.py
@@ -132,7 +132,8 @@ def create_encryption_variants() -> list[BuildVariant]:
         display_name = get_variant_name(encryption, host, **expansions)
         tasks = [".test-non-standard"]
         if host != "rhel8":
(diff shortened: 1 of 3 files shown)


blink1073:
The result of this change is that we don't run on Windows or MacOS at all.  I think the generator script needs to be updated to add a different task target that doesn't include `cov`.

blink1073:
For PRs, we should probably only run the sharded variant to cut down the number of jobs that run on each commit.

.evergreen/scripts/generate_config.py
```diff
@@ -132,7 +132,11 @@ def get_encryption_expansions(encryption):
         display_name = get_variant_name(encryption, host, **expansions)
         tasks = [".test-non-standard"]
         if host != "rhel8":
-            tasks = [".test-non-standard !.pypy"]
+            # Exclude PyPy (no Evergreen toolchain on macOS/win64) and coverage tasks
```
blink1073: This comment isn't accurate, we have PyPy in our toolchain on all hosts.

blink1073 (changes requested):

blink1073 (approved):
LGTM

PR #2887, opened 2026-06-22
PYTHON-5890 Automate sync file generation and staging

blink1073:
PYTHON-5890

This PR improves the developer experience around synchro by making sync file generation automatic and transparent — developers should not need to think about when to run synchro or manually stage generated files.

## Changes in this PR

The pre-commit `synchro` hook now auto-stages generated sync files so developers no longer need to stage them manually after a commit triggers regeneration.

Running `just test` and `just run-tests` now ensures sync files are current before the test suite executes.

Lint problems in async source files are caught and reported before any sync file generation is attempted, giving a clearer error signal when the async sources themselves have issues.

In CI (`pre-commit --all-files`), the hook verifies that committed sync files match what would be generated from the current async sources, and prints the full list of affected files on failure.

## Usage

**Local commit** — modify an async source file, then `git commit`. The hook regenerates the corresponding sync files and stages them automatically alongside your changes.

**Local tests** — run `just test` or `just run-tests`. Sync files are regenerated first so the test suite always sees up-to-date code.

**CI** — `pre-commit --all-files` fails with a clear file list if any sync file is out of date with its async source.

## Test Plan

- Pre-commit hooks (`ruff`, `ruff-format`, `synchro`, `synchro-stage-files`) all pass on commit
- `just typing` passes with no errors
- Smoke test for synchro added in `tools/test_synchro.py`, run in the `static` CI job

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,25 +18,36 @@ repos:
     exclude: .patch
     exclude_types: [json]
 
-- repo: https://github.com/astral-sh/ruff-pre-commit
-  # Ruff version.
-  rev: v0.15.16
+- repo: local
   hooks:
     - id: ruff
-      args: ["--fix", "--show-fixes"]
+      name: ruff
+      entry: uv run --group lint ruff check --fix --show-fixes
+      language: system
+      types_or: [python, pyi]
+      require_serial: true
     - id: ruff-format
+      name: ruff-format
+      entry: uv run --group lint ruff format
+      language: system
+      types_or: [python, pyi]
+      require_serial: true
 
 - repo: local
   hooks:
     - id: synchro
       name: synchro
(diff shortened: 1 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -18,25 +18,36 @@ repos:
     exclude: .patch
     exclude_types: [json]
 
-- repo: https://github.com/astral-sh/ruff-pre-commit
-  # Ruff version.
-  rev: v0.15.16
+- repo: local
```
blink1073, line 21: I put `ruff` into a dependency-group so we can use a shared version here and in `synchro.py`.

sleepyStick (approved):

PR #2889, opened 2026-06-22
PYTHON-5890 Skip ruff in synchro when called from just test/run-tests

blink1073:
[PYTHON-5890](https://jira.mongodb.org/browse/PYTHON-5890) (follow up)

## Changes in this PR

Unrelated linting issues in async files should not block test runs. When `synchro.py` is invoked via `just test` or `just run-tests`, all ruff calls (pre-check on async sources, fix pass on generated files, format pass on generated files) are now skipped.

This is controlled by a `SYNCHRO_SKIP_RUFF` environment variable, set in the `run-synchro` justfile recipe. The pre-commit hook invokes `synchro.py` directly without that variable, so ruff continues to run during commits.

## Test Plan

Existing synchro smoke test continues to pass. Manual verification: `just test` and `just run-tests` no longer abort due to ruff errors in async source files; `git commit` still runs ruff via the pre-commit hook.

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -45,6 +45,17 @@ post:
   - func: "cleanup"
 
 tasks:
+  - name: linkcheck
+    commands:
+      - command: subprocess.exec
+        type: test
+        params:
+          binary: bash
+          args:
+            - .evergreen/just.sh
+            - docs-linkcheck
+          working_dir: src
+
   - name: resync_specs
     commands:
       - command: subprocess.exec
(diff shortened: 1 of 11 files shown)


sleepyStick (approved):

blink1073:
This isn't quite right.  I think we'll need a different approach that uses the git status for this to work.  I'll open a new PR to remove the `justfile` usage of synchro and open a different ticket for the test auto-sync.

PR #2891, opened 2026-06-24
PYTHON-5745 Consolidate command telemetry

blink1073:
[PYTHON-5745](https://jira.mongodb.org/browse/PYTHON-5745)

## Changes in this PR

Introduces `pymongo/_telemetry.py` with a `_CommandTelemetry` class that unifies structured logging and APM event publishing for command lifecycle events into a single internal API.  This refactor is a stepping stone to adding OpenTelemetry support in a `_telemetry.py`, so all event-related handling is in one centralized location.

## Test Plan

No new tests: this is a behavior-preserving refactor. The existing APM / command-monitoring and command-logging spec suites assert the exact event and log documents that any regression would change.

### Checklist for Author
- [ ] Did you update the changelog (if necessary)? — Not necessary; internal refactor with no user-facing behavior change.
- [x] Is there test coverage? — Covered by existing spec suites (behavior preserved).
- [x] Is any followup work tracked in a JIRA ticket? If so, add link(s). — [PYTHON-5846](https://jira.mongodb.org/browse/PYTHON-5846) will extend this pattern to cover the remaining logging and APM event types.

### Checklist for Reviewer

### DIFF
--- a/pymongo/_telemetry.py
+++ b/pymongo/_telemetry.py
@@ -124,13 +124,12 @@ class _CommandTelemetry:
         self._duration = datetime.datetime.now() - self._start
         if not self._should_log and not self._publish:
             return
-        duration = self._duration
         if self._should_log:
             _debug_log(
                 _COMMAND_LOGGER,
                 message=_CommandStatusMessage.SUCCEEDED,
                 clientId=self._topology_id,
-                durationMS=duration,
+                durationMS=self._duration,
                 reply=reply,
                 commandName=self._name,
                 databaseName=self._dbname,
@@ -146,7 +145,7 @@ class _CommandTelemetry:
         if self._publish:
             assert self._listeners is not None
             self._listeners.publish_command_success(
-                duration,
+                self._duration,
                 reply,
                 command_name,
                 self._request_id,
@@ -168,13 +167,12 @@ class _CommandTelemetry:
         self._duration = datetime.datetime.now() - self._start
         if not self._should_log and not self._publish:
             return
-        duration = self._duration
         if self._should_log:
             _debug_log(
                 _COMMAND_LOGGER,
                 message=_CommandStatusMessage.FAILED,
                 clientId=self._topology_id,
-                durationMS=duration,
+                durationMS=self._duration,
                 failure=failure,
                 commandName=self._name,
                 databaseName=self._dbname,
@@ -190,7 +188,7 @@ class _CommandTelemetry:
         if self._publish:
             assert self._listeners is not None
             self._listeners.publish_command_failure(
-                duration,
+                self._duration,
                 failure,
                 command_name,
                 self._request_id,
--- a/pymongo/_telemetry.py
+++ b/pymongo/_telemetry.py
@@ -39,6 +39,7 @@ class _CommandTelemetry:
     """
 
     __slots__ = (
+        "_active",
         "_cmd",
         "_conn",
         "_dbname",
@@ -66,6 +67,7 @@ class _CommandTelemetry:
         self._topology_id = topology_id
         self._should_log = topology_id is not None and _COMMAND_LOGGER.isEnabledFor(logging.DEBUG)
         self._publish = listeners is not None and listeners.enabled_for_commands
+        self._active = self._should_log or self._publish
         self._listeners = listeners
         self._conn = conn
         self._cmd = cmd
@@ -76,25 +78,30 @@ class _CommandTelemetry:
         self._start: datetime.datetime
         self._duration: datetime.timedelta
 
+    def _emit_log(self, message: _CommandStatusMessage, **extra: Any) -> None:
+        _debug_log(
+            _COMMAND_LOGGER,
+            message=message,
+            clientId=self._topology_id,
+            commandName=self._name,
+            databaseName=self._dbname,
+            requestId=self._request_id,
+            operationId=self._request_id,
+            driverConnectionId=self._conn.id,
+            serverConnectionId=self._conn.server_connection_id,
+            serverHost=self._conn.address[0],
+            serverPort=self._conn.address[1],
+            serviceId=self._conn.service_id,
+            **extra,
+        )
+
     def started(self, orig: MutableMapping[str, Any], ensure_db: bool) -> None:
         """Emit the STARTED log entry and APM event, and start the duration clock."""
         self._start = datetime.datetime.now()
+        if not self._active:
+            return
         if self._should_log:
-            _debug_log(
-                _COMMAND_LOGGER,
-                message=_CommandStatusMessage.STARTED,
-                clientId=self._topology_id,
-                command=self._cmd,
-                commandName=self._name,
-                databaseName=self._dbname,
-                requestId=self._request_id,
-                operationId=self._request_id,
-                driverConnectionId=self._conn.id,
-                serverConnectionId=self._conn.server_connection_id,
-                serverHost=self._conn.address[0],
-                serverPort=self._conn.address[1],
-                serviceId=self._conn.service_id,
-            )
+            self._emit_log(_CommandStatusMessage.STARTED, command=self._cmd)
         if self._publish:
             assert self._listeners is not None
             if ensure_db and "$db" not in orig:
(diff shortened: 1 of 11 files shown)


pymongo/_telemetry.py
```diff
@@ -0,0 +1,202 @@
...
+        self,
+        reply: _DocumentOut,
+        command_name: str,
+        speculative_hello: bool,
+    ) -> None:
+        """Emit the SUCCEEDED log entry and APM event."""
+        self._duration = datetime.datetime.now() - self._start
+        if not self._should_log and not self._publish:
+            return
+        duration = self._duration
```
NoahStapp: Do we need this assignment? `self._duration` should only get assigned once per `_CommandTelemetry` lifecycle.

pymongo/_telemetry.py
```diff
@@ -0,0 +1,202 @@
...
+        self,
+        failure: _DocumentOut,
+        command_name: str,
+        is_server_side_error: bool,
+    ) -> None:
+        """Emit the FAILED log entry and APM event."""
+        self._duration = datetime.datetime.now() - self._start
+        if not self._should_log and not self._publish:
+            return
+        duration = self._duration
```
NoahStapp: Same here.

pymongo/pool_shared.py
```diff
@@ -44,10 +45,20 @@
...
 
 
+class _ConnectionTelemetryInfo(Protocol):
+    """Protocol for connection fields consumed by :class:`~pymongo._telemetry._CommandTelemetry`."""
+
+    id: int
+    server_connection_id: Optional[int]
+    address: tuple[str, int]
+    service_id: Optional[ObjectId]
+
```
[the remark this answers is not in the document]
blink1073: Intentional: `AsyncConnection.__init__` takes `address: tuple[str, int]` with a non-optional port, and `_CommandTelemetry` passes `address[1]` directly to log calls and APM publish methods that expect `int`. Using `_Address` (`tuple[str, Optional[int]]`) here would require null-handling in telemetry that isn't needed for TCP connections.

pymongo/_telemetry.py
```diff
@@ -0,0 +1,202 @@
...
+        self,
+        topology_id: Optional[ObjectId],
+        conn: _ConnectionTelemetryInfo,
+        listeners: Optional[_EventListeners],
+        cmd: MutableMapping[str, Any],
+        dbname: str,
+        request_id: int,
+        op_id: Optional[int],
+    ) -> None:
+        self._topology_id = topology_id
```
NoahStapp, line 67: We construct the whole object even if `self.should_log = False`, does it make sense to have the constructor short-circuit out with that as the first check?

pymongo/_telemetry.py
```diff
@@ -0,0 +1,202 @@
...
+        self._conn = conn
+        self._cmd = cmd
+        self._name = next(iter(cmd))
+        self._dbname = dbname
+        self._request_id = request_id
+        self._op_id = op_id
+        self._start: datetime.datetime
+        self._duration: datetime.timedelta
+
+    def started(self, orig: MutableMapping[str, Any], ensure_db: bool) -> None:
```
NoahStapp, line 98: Could we refactor `started, succeeded, failed` (or a subset) to use a common internal helper like `_emit`? All three are very similar and effectively just pass through kwargs to `debug_log`.

pymongo/pool_shared.py
```diff
@@ -44,10 +45,20 @@
 
 SSLErrors = (PYSSLError, SSLError)
 if TYPE_CHECKING:
+    from bson.objectid import ObjectId
     from pymongo.pyopenssl_context import _sslConn
     from pymongo.typings import _Address
 
 
+class _ConnectionTelemetryInfo(Protocol):
```
NoahStapp, line 53: Is there a reason to use this over the existing `_AgnosticConnection`?

NoahStapp (changes requested):

pymongo/_telemetry.py
blink1073: Removed

pymongo/_telemetry.py
blink1073: Removed

pymongo/_telemetry.py
blink1073: I don't think a short-circuit makes sense in the constructor, but I added a short-circuit in the methods

pymongo/_telemetry.py
blink1073: Makes sense, done

pymongo/pool_shared.py
blink1073: Re [#discussion_r3475117295](https://github.com/mongodb/mongo-python-driver/pull/2891#discussion_r3475117295): Two reasons: (1) `_AgnosticConnection` is `Union["AsyncConnection", "Connection"]` — a type alias, not a class, so `AsyncConnection`/`Connection` can't inherit from it. (2) `_AgnosticConnection` is defined in `typings.py` and forward-references both pool classes, which would introduce a circular import when used from `pool_shared.py`. The Protocol also narrows the surface to exactly the four fields `_CommandTelemetry` needs, making the contract explicit.

blink1073: I didn't mean to send the raw response.  My main goal was to have a properties-only subset that would be used here and probably the pool telemetry as well.

pymongo/asynchronous/command_runner.py
```diff
@@ -341,6 +245,7 @@ async def run_bulk_write_command(
     :param max_doc_size: The largest document size in the batch, passed to ``conn.send_message``.
     :param unacknowledged: When ``True``, send only and fake an ``{"ok": 1}`` reply.
     """
+    topology_id = client._topology_settings._topology_id if client is not None else None
```
NoahStapp: Should we add a private `_topology_id` attribute on `MongoClient` at this point?

NoahStapp (changes requested):

pymongo/asynchronous/command_runner.py
blink1073: Makes sense, I had that in my previous PR at one point

pymongo/_telemetry.py
```diff
@@ -0,0 +1,185 @@
...
+            serverConnectionId=self._conn.server_connection_id,
+            serverHost=self._conn.address[0],
+            serverPort=self._conn.address[1],
+            serviceId=self._conn.service_id,
+            **extra,
+        )
+
+    def started(self, orig: MutableMapping[str, Any], ensure_db: bool) -> None:
+        """Emit the STARTED log entry and APM event, and start the duration clock."""
+        self._start = datetime.datetime.now()
```
NoahStapp, line 100: Nit: this (and the similar lines in the other methods) should be after the `if not self.active check`.

pymongo/pool_shared.py
NoahStapp: Is the idea to have a single, minimal listing of connection attributes that we need for telemetry? I don't think either of the AI responses are actually true in this context since `_ConnectionTelemetryInfo` is used purely as a type annotation.

NoahStapp (changes requested):

pymongo/pool_shared.py
blink1073: Yeah, ignore the AI response

pymongo/_telemetry.py
blink1073: `telemetry.duration` is accessed unconditionally at the end of `_run_command` (always needed for the `Response` object), so `_start` must be set before any early-return path. If we moved `self._start = datetime.datetime.now()` to after the `_active` check in `started()`, `telemetry.duration` would raise `AttributeError` on every non-APM/non-logging command. The same applies to the `_duration` assignments in `succeeded()`/`failed()`.

blink1073: What the heck, I have "Ask permissions" turned on, and this is twice now it just replied.  😠

blink1073: But yeah, even without logging we need duration upstream

NoahStapp: Can't trust the robots man

pymongo/pool_shared.py
NoahStapp: I'm fine with trying `_ConnectionTelemetryInfo` out, we can remove it in a future refactor or consolidation if it turns out to be less useful than expected here.

NoahStapp (approved):
LGTM!

PR #2868, opened 2026-06-11
PYTHON-5814 Configurable DNS domain validation for SRV records

sleepyStick:
[PYTHON-5814](https://jira.mongodb.org/browse/PYTHON-5814)

## Changes in this PR

- add `srvAllowedHostsSuffix` param

### Checklist for Author
- [ ] Did you update the changelog (if necessary)?
- [ ] Is there test coverage?
- [ ] Is any followup work tracked in a JIRA ticket? If so, add link(s).

### Checklist for Reviewer
- [ ] Does the title of the PR reference a JIRA Ticket?
- [ ] Do you fully understand the implementation? (Would you be comfortable explaining how this code works to someone else?)
- [ ] Is all relevant documentation (README or docstring) updated?

PR #2903, opened 2026-06-29
PYTHON-5814 Configurable DNS domain validation for SRV records

sleepyStick:
[PYTHON-5814](https://jira.mongodb.org/browse/PYTHON-5814)

## Changes in this PR

THIS IS THE SAME PR AS https://github.com/mongodb/mongo-python-driver/pull/2868 BUT THIS ONE MERGES TO A BRANCH OF MAIN AND NOT MAIN ITSELF!! THE OTHER PR IS FOR MAIN AND WILL BE ADDRESSED LATER
- add `srvAllowedHostsSuffix` param 

### Checklist for Author
- [ ] Is there test coverage?

### Checklist for Reviewer


.pre-commit-config.yaml
```diff
@@ -115,6 +115,7 @@ repos:
     # - test/versioned-api/crud-api-version-1-strict.json:514: nin ==> inn, min, bin, nine
     # - test/test_client.py:188: te ==> the, be, we, to
     args: ["-L", "fle,fo,infinit,isnt,nin,te,aks"]
+    exclude: ^pymongo/public_suffix_list\.dat$
```
sleepyStick, line 118: psl is full of mostly not words,,, checking it raises lots of "typos" but aren't actually typos lol

test/srv_seedlist/replica-set/srvAllowedHostsSuffix-mismatch.json
```diff
@@ -0,0 +1,6 @@
+{
+  "uri": "mongodb+srv://test12.test.build.10gen.cc/?srvAllowedHostsSuffix=test.build.10gen.cc",
```
Jibola, line 2: May be a dumb question; what's the mismatch here?

pymongo/_psl.py
```diff
@@ -0,0 +1,51 @@
...
+# permissions and limitations under the License.
+
+"""Public Suffix List lookup for srvAllowedHostsSuffix validation."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+
+def _load_public_suffixes() -> tuple[set[str], set[str], set[str]]:
```
Jibola, line 25: It could be a good idea to do a simple cache this information so that it doesn't do a full file read each time we check for a public suffix.

pymongo/asynchronous/srv_resolver.py
```diff
@@ -71,11 +72,29 @@ def __init__(
         connect_timeout: Optional[float],
         srv_service_name: str,
         srv_max_hosts: int = 0,
+        srv_allowed_hosts_suffix: Optional[str] = None,
     ):
-        self.__fqdn = fqdn
+        self.__fqdn = fqdn.lower()
```
Jibola, line 77: Is this `lower` call necessary? I don't see anywhere that mutates or calls `self.__fqdn` that has been changed by the newly introduced logic.

Jibola (changes requested):
Have a few comments, but things otherwise look good.

pymongo/asynchronous/mongo_client.py
```diff
@@ -450,6 +450,18 @@ def __init__(
...
+            hostname. When set, the driver accepts any returned host whose name ends
+            with this suffix (e.g. ``".atlas.mongodb.com"``). The value must contain
+            at least two labels and must not be a public suffix (per the Public Suffix
+            List). Only valid with ``mongodb+srv://`` URIs.
+
+            .. warning::
+
+               This option relaxes a built-in DNS spoofing safeguard. Use the most
+               specific suffix possible for your deployment rather than a broad
+               company-wide domain.
```
Jibola: Can you add an example of a good configuration just so the users can have an explicit understanding of what to do?

Jibola (changes requested):

test/srv_seedlist/replica-set/srvAllowedHostsSuffix-mismatch.json
sleepyStick: the comment on the equivalent yml file would say "# DNS record for test12.test.build.10gen.cc returns localhost.build.10gen.cc which would not match test.build.10gen.cc" (obv that's in review on the drivers PR but yeah)

pymongo/_psl.py
sleepyStick: oh good call, done in ([7902127](https://github.com/mongodb/mongo-python-driver/pull/2903/commits/790212733c2427235a1c30edcf51bdd1b648247e))

pymongo/asynchronous/srv_resolver.py
sleepyStick: errr not necessary, i think my bot(s) noticed it an existing bug -- because I added a test case about case insensitivity for the new parameter i'm introducing -- for example if a user passed in a URI like "mongodb+srv://test1.TEST.BUILD.10GEN.CC/" instead of "mongodb+srv://test1.test.build.10gen.cc/" the current logic (without the `lower`) would have this not connect. (i can undo the change if we wanted to tackle this problem separately or add a test to solidify it if we're okay with doing it here -- sorry it slipped my mind to explain it when i was putting up the PR)

pymongo/asynchronous/mongo_client.py
sleepyStick: done in [c8f8c9f](https://github.com/mongodb/mongo-python-driver/pull/2903/commits/c8f8c9f6c0714eace8daaae7c4ac80db07e0c58b)!

Jibola (approved):

PR #2907, opened 2026-07-01
PYTHON-5846 Consolidate CMAP, heartbeat, and SDAM telemetry into _telemetry.py

blink1073:
PYTHON-5846

## Changes in this PR

Consolidates all APM event publishing and structured logging into `pymongo/_telemetry.py`, giving OpenTelemetry support a single place to hook into. Adds four new telemetry classes and one helper following the `_CommandTelemetry` pattern from #2891:

- **`_CmapTelemetry`** — connection pool and connection lifecycle events, owned by `Pool`
- **`_HeartbeatTelemetry`** — server heartbeat events, owned per-check by `Monitor`
- **`_SdamTelemetry`** — topology and server description change events, owned by `Topology` and `Server`
- **`_ServerSelectionTelemetry`** — server selection log entries (log-only per spec), constructed per `select_server` call
- **`log_command_retry`** — retry log entries for retryable reads and writes

## Test Plan

Covered by existing test suites (`test_connection_monitoring.py`, `test_heartbeat_monitoring.py`, `test_sdam_monitoring_spec.py`, `test_connection_logging.py`, `test_server_selection_logging.py`, unified format spec tests). No new public API.

### Checklist for Author
- [ ] Is there test coverage?

### Checklist for Reviewer


pymongo/asynchronous/mongo_client.py
```diff
@@ -3043,12 +3042,12 @@ async def _read(self) -> T:
...
                 self._check_last_error()
             if self._retrying:
-                _debug_log(
-                    _COMMAND_LOGGER,
-                    message=f"Retrying read attempt number {self._attempt_number}",
-                    clientId=self._client._topology_settings._topology_id,
-                    commandName=self._operation,
-                    operationId=self._operation_id,
+                log_command_retry(
+                    self._client._topology_settings._topology_id,
```
NoahStapp:
```suggestion
                    self._client._topology_id,
```

pymongo/asynchronous/mongo_client.py
```diff
@@ -3014,12 +3013,12 @@ async def _write(self) -> T:
                     self._check_last_error()
                     self._retryable = False
                 if self._retrying:
-                    _debug_log(
-                        _COMMAND_LOGGER,
-                        message=f"Retrying write attempt number {self._attempt_number}",
-                        clientId=self._client._topology_id,
-                        commandName=self._operation,
-                        operationId=self._operation_id,
+                    log_command_retry(
```
NoahStapp: What do you think about a helper method here to avoid duplicating the entire `log_command_retry` call except for `is_write` between this and the next usage?

pymongo/asynchronous/monitor.py
```diff
@@ -18,18 +18,18 @@
...
 
 from pymongo import common, periodic_executor
 from pymongo._csot import MovingMinimum
+from pymongo._telemetry import _HeartbeatTelemetry
 from pymongo.asynchronous.srv_resolver import _SrvResolver
 from pymongo.errors import NetworkTimeout, _OperationCancelled
 from pymongo.hello import Hello
 from pymongo.lock import _async_create_lock
-from pymongo.logger import _SDAM_LOGGER, _debug_log, _SDAMStatusMessage
+from pymongo.logger import _SDAM_LOGGER, _debug_log
```
NoahStapp: Unused imports.

pymongo/asynchronous/pool.py
```diff
@@ -475,21 +466,7 @@ async def authenticate(self, reauthenticate: bool = False) -> None:
...
-                    clientId=self._client_id,
-                    serverHost=self.address[0],
-                    serverPort=self.address[1],
-                    driverConnectionId=self.id,
-                    durationMS=duration,
-                )
-            if self.enabled_for_cmap:
-                assert self.listeners is not None
-                self.listeners.publish_connection_ready(self.address, self.id, duration)
+            self._telemetry.connection_ready(self.id)
```
NoahStapp: Intentional removal of the comment here?

pymongo/asynchronous/pool.py
```diff
@@ -996,20 +904,7 @@ async def connect(self, handler: Optional[_MongoClientErrorHandler] = None) -> A
...
-                message=_ConnectionStatusMessage.CONN_CREATED,
-                clientId=self._client_id,
-                serverHost=self.address[0],
-                serverPort=self.address[1],
-                driverConnectionId=conn_id,
-            )
-        if self.enabled_for_cmap:
-            assert listeners is not None
-            listeners.publish_connection_created(self.address, conn_id)
+        self._telemetry.connection_created(conn_id)
```
NoahStapp, line 908: Same here, intentional comment removal?

pymongo/_telemetry.py
```diff
@@ -183,3 +197,492 @@ def failed(
...
+    def pool_ready(self) -> None:
+        # Log before publishing to prevent potential listener preemption in tests.
+        if self._should_log:
+            self._emit_log(_ConnectionStatusMessage.POOL_READY)
+        if self._should_publish:
+            assert self._listeners is not None
+            self._listeners.publish_pool_ready(self._address)
+
+    def pool_cleared(self, service_id: Optional[ObjectId], interrupt_connections: bool) -> None:
+        # Log before publishing to prevent potential listener preemption in tests.
```
NoahStapp, line 278: The comment here and in `pool_closed` are no longer visible at the location where they're needed.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,492 @@ def failed(
...
+                reason=_verbose_connection_error_reason(reason),
+                error=reason,
+            )
+
+    def checkout_started(self) -> None:
+        should_log = self._should_log
+        should_publish = self._should_publish
+        # Always record start time: logging or publishing may be enabled by the time
+        # checkout_succeeded or checkout_failed is called to compute the duration.
+        self._checkout_start = time.monotonic()
```
NoahStapp: This is shared across all connections in the pool and is called outside of a lock, introducing a possible race condition that can cause timing inconsistencies.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,492 @@ def failed(
...
+        if self._should_log:
+            self._emit_log(_ConnectionStatusMessage.POOL_CLOSED)
+        if self._should_publish:
+            assert self._listeners is not None
+            self._listeners.publish_pool_closed(self._address)
+
+    def connection_created(self, conn_id: int) -> None:
+        # Always record start time: logging or publishing may be enabled by the time
+        # connection_ready is called to compute the duration.
+        self._conn_created_start = time.monotonic()
```
NoahStapp: Same race condition here as above.

pymongo/asynchronous/topology.py
```diff
@@ -345,32 +296,13 @@ async def _select_servers_loop(
...
-                        _SERVER_SELECTION_LOGGER,
-                        message=_ServerSelectionStatusMessage.FAILED,
-                        selector=selector,
-                        operation=operation,
-                        operationId=operation_id,
-                        topologyDescription=self.description,
-                        clientId=self.description._topology_settings._topology_id,
-                        failure=self._error_message(selector),
-                    )
+                ss.failed(self._error_message(selector))
```
NoahStapp: This will log `self.description` at the time of creation instead of at the time of failure. If SDAM updates `self.description` while this loop is waiting and before it calls `ss.failed`, we log the old description instead of the new.

NoahStapp (changes requested):

pymongo/_telemetry.py
blink1073: I don't follow?  The caller doesn't need to know the order

pymongo/asynchronous/mongo_client.py
blink1073: Done

pymongo/asynchronous/monitor.py
blink1073: They were used, I missed one direct usage, now moved to `_telemetry.py`

pymongo/asynchronous/pool.py
blink1073: Yes, see other comment

pymongo/asynchronous/pool.py
blink1073: Ditto

pymongo/_telemetry.py
NoahStapp: I mean that looking at the code in `pool.py` there's no longer a comment explaining why it's called there. We changed what logging code was called, not why it's called.

pymongo/_telemetry.py
blink1073: Updated to keep the timing info in the pool class as it was before.

pymongo/_telemetry.py
blink1073: Ditto

pymongo/asynchronous/topology.py
blink1073: Fixed

pymongo/_telemetry.py
blink1073: I still don't follow what you mean.  Before the logging and publishing was handled inline, now it is moved elsewhere, and so did the comment.

NoahStapp: Oh sorry, I misread the comment 😅. You're right!

pymongo/asynchronous/pool.py
```diff
@@ -1156,34 +1012,22 @@ async def checkout(
         elif conn.active:
             await self.checkin(conn)
 
-    def _raise_if_not_ready(self, checkout_started_time: float, emit_event: bool) -> None:
+    def _raise_if_not_ready(self, emit_event: bool, checkout_start: float) -> None:
```
NoahStapp: intentional re-ordering + naming of `checkout_started_time`?

pymongo/asynchronous/pool.py
```diff
@@ -1156,34 +1012,22 @@ async def checkout(
...
+                )
 
             details = _get_timeout_details(self.opts)
             _raise_connection_failure(
                 self.address, AutoReconnect("connection pool paused"), timeout_details=details
             )
 
     async def _get_conn(
-        self, checkout_started_time: float, handler: Optional[_MongoClientErrorHandler] = None
+        self, handler: Optional[_MongoClientErrorHandler] = None, checkout_start: float = 0.0
```
NoahStapp: Same here. Also, does a default of 0 make sense here?

pymongo/asynchronous/pool.py
```diff
@@ -1259,8 +1091,8 @@ async def _get_conn(
                             if self.conns or self._pending < self._max_connecting:
                                 self._max_connecting_cond.notify()
                             emitted_event = True
-                            self._raise_wait_queue_timeout(checkout_started_time)
-                        self._raise_if_not_ready(checkout_started_time, emit_event=False)
+                            self._raise_wait_queue_timeout(checkout_start)
+                        self._raise_if_not_ready(emit_event=False, checkout_start=checkout_start)
```
NoahStapp: All these uses too.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,504 @@ def failed(
...
+        _debug_log(
+            _CONNECTION_LOGGER,
+            message=message,
+            clientId=self._client_id,
+            serverHost=self._address[0],
+            serverPort=self._address[1],
+            **extra,
+        )
+
+    def pool_created(self, non_default_options: dict[str, Any]) -> None:
```
NoahStapp, line 258: Need docstrings on these public methods for consistency with the other telemetry classes.

pymongo/asynchronous/topology.py
```diff
@@ -321,17 +279,10 @@ async def _select_servers_loop(
...
-            _debug_log(
-                _SERVER_SELECTION_LOGGER,
-                message=_ServerSelectionStatusMessage.STARTED,
-                selector=selector,
-                operation=operation,
-                operationId=operation_id,
-                topologyDescription=self.description,
-                clientId=self.description._topology_settings._topology_id,
-            )
+        ss = _ServerSelectionTelemetry(
```
NoahStapp, line 286: Can we pass this through as an argument from `select_server` so we don't have to make two `_ServerSelectionTelemetry` per call?

pymongo/asynchronous/server.py
```diff
@@ -74,6 +69,7 @@ def __init__(
         self._events = None
```
NoahStapp: Can we cut all three of these now that `_SdamTelemetry` exists?

pymongo/_telemetry.py
```diff
@@ -183,3 +197,504 @@ def failed(
...
+    def waiting(self, remaining_time_ms: int) -> None:
+        """Emit the server selection WAITING log entry."""
+        if self._should_log:
+            self._emit_log(
+                _ServerSelectionStatusMessage.WAITING,
+                self._topology_description,
+                remainingTimeMS=remaining_time_ms,
+            )
+
+    def failed(self, failure: str, topology_description: Any = None) -> None:
```
NoahStapp: `topology_description` appears to always be passed

NoahStapp (changes requested):

pymongo/asynchronous/pool.py
blink1073: Restored to original signature

pymongo/asynchronous/pool.py
blink1073: Restored to original signature

pymongo/asynchronous/pool.py
blink1073: Done

pymongo/_telemetry.py
blink1073: Added :allthedocstrings:

pymongo/asynchronous/topology.py
blink1073: Done

pymongo/asynchronous/server.py
blink1073: Done

pymongo/_telemetry.py
blink1073: Fixed

pymongo/asynchronous/topology.py
```diff
@@ -435,18 +367,7 @@ async def select_server(
...
-                message=_ServerSelectionStatusMessage.SUCCEEDED,
-                selector=selector,
-                operation=operation,
-                operationId=operation_id,
-                topologyDescription=self.description,
-                clientId=self.description._topology_settings._topology_id,
-                serverHost=server.description.address[0],
-                serverPort=server.description.address[1],
-            )
+        ss.succeeded(server.description.address[0], server.description.address[1])
```
NoahStapp: Similar here to the `ss.failed()` issue earlier, should we pass the current description to be logged instead of the potentially stale one saved to `ss` at the start of selection?

NoahStapp (changes requested):

pymongo/asynchronous/topology.py
blink1073: Done

pymongo/_telemetry.py
```diff
@@ -183,3 +197,519 @@ def failed(
...
+    ) -> None:
+        self._topology_id = topology_id
+        self._address = address
+        self._listeners = listeners
+        self._awaited = awaited
+        # Cached at construction: this object is short-lived (one heartbeat check) so
+        # listener registration and logging level are stable for its lifetime.
+        self._should_publish = listeners is not None and listeners.enabled_for_server_heartbeat
+        self._should_log = _SDAM_LOGGER.isEnabledFor(logging.DEBUG)
+
```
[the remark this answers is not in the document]
blink1073: Done

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+    def topology_opened(self) -> None:
+        """Emit the topology opened log entry and APM event."""
+        if self._should_log:
+            self._emit_log(_SDAMStatusMessage.START_TOPOLOGY)
+        if self._publish_tp:
+            assert self._listeners is not None
+            self._enqueue(self._listeners.publish_topology_opened, (self._topology_id,))
+
+    def topology_description_changed(self, old_td: Any, new_td: Any) -> None:
```
NoahStapp: These `Any` hints should be `TopologyDescription`, right?

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+                (old_td, new_td, self._topology_id),
+            )
+        if self._should_log:
+            self._emit_log(
+                _SDAMStatusMessage.TOPOLOGY_CHANGE,
+                previousDescription=repr(old_td),
+                newDescription=repr(new_td),
+            )
+
+    def topology_closed(self, old_td: Any, new_td: Any) -> None:
```
NoahStapp: Same here.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+            assert self._listeners is not None
+            self._enqueue(self._listeners.publish_server_opened, (address, self._topology_id))
+        if self._should_log:
+            self._emit_log(
+                _SDAMStatusMessage.START_SERVER,
+                serverHost=address[0],
+                serverPort=address[1],
+            )
+
+    def server_description_changed(self, sd_old: Any, sd_new: Any, address: _Address) -> None:
```
NoahStapp: Same here with `ServerDescription`.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+        "_operation_id",
+        "_selector",
+        "_should_log",
+        "_topology_description",
+        "_topology_id",
+    )
+
+    def __init__(
+        self,
+        topology_id: Any,
```
NoahStapp: `_HeartbeatTelemetry` hints this as `ObjectID`, we should be consistent across all of the telemetry classes.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+        selector: Any,
+        operation: str,
+        operation_id: Optional[int],
+        topology_description: Any,
```
NoahStapp: Same here for `TopologyDescription`.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+        self._selector = selector
+        self._operation = operation
+        self._operation_id = operation_id
+        self._topology_description = topology_description
+        # Cached at construction: this object is short-lived (one select_server call) so
+        # logging level is stable for its lifetime.
+        self._should_log = _SERVER_SELECTION_LOGGER.isEnabledFor(logging.DEBUG)
+
+    def _emit_log(
+        self, message: _ServerSelectionStatusMessage, topology_description: Any, **extra: Any
```
NoahStapp: And here.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+    def waiting(self, remaining_time_ms: int) -> None:
+        """Emit the server selection WAITING log entry."""
+        if self._should_log:
+            self._emit_log(
+                _ServerSelectionStatusMessage.WAITING,
+                self._topology_description,
+                remainingTimeMS=remaining_time_ms,
+            )
+
+    def failed(self, failure: str, topology_description: Any) -> None:
```
NoahStapp: Here.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+        """Emit the server selection FAILED log entry with the current topology description."""
+        if self._should_log:
+            self._emit_log(
+                _ServerSelectionStatusMessage.FAILED,
+                topology_description,
+                failure=failure,
+            )
+
+    def succeeded(
+        self, server_host: str, server_port: Optional[int], topology_description: Any
```
NoahStapp: Here.

pymongo/_telemetry.py
```diff
@@ -183,3 +197,520 @@ def failed(
...
+
+
+def log_srv_monitor_failure(failure: Exception) -> None:
+    """Emit a log entry when the SRV monitor fails to poll DNS records."""
+    if _SDAM_LOGGER.isEnabledFor(logging.DEBUG):
+        _debug_log(_SDAM_LOGGER, message="SRV monitor check failed", failure=repr(failure))
+
+
+def log_command_retry(
+    topology_id: Any,
```
NoahStapp: Here.

NoahStapp (changes requested):
Type-hinting cleanup, otherwise looks great!

pymongo/_telemetry.py
blink1073: Fixed

pymongo/_telemetry.py
blink1073: Done

pymongo/_telemetry.py
blink1073: Done

pymongo/_telemetry.py
blink1073: It still needs to be optional for command telemetry, but otherwise done

pymongo/_telemetry.py
blink1073: Done

pymongo/_telemetry.py
blink1073: Done

pymongo/_telemetry.py
blink1073: done

pymongo/_telemetry.py
blink1073: Done

pymongo/_telemetry.py
blink1073: Done (needs to be optional)

blink1073:
I'm waiting for #2894 to be merged since it was already approved.  I'll have some merge conflicts.

pymongo/_telemetry.py
```diff
@@ -183,3 +199,530 @@ def failed(
...
+        self,
+        topology_id: ObjectId,
+        address: _Address,
+        listeners: Optional[_EventListeners],
+        awaited: bool,
+    ) -> None:
+        self._topology_id = topology_id
+        self._address = address
+        self._listeners = listeners
+        self._awaited = awaited
```
NoahStapp: This should be calculated at emit rather than construction.

pymongo/_telemetry.py
```diff
@@ -183,3 +199,530 @@ def failed(
...
+        topology_description: TopologyDescription,
+    ) -> None:
+        self._topology_id = topology_id
+        self._selector = selector
+        self._operation = operation
+        self._operation_id = operation_id
+        self._topology_description = topology_description
+        # Cached at construction: this object is short-lived (one select_server call) so
+        # logging level is stable for its lifetime.
+        self._should_log = _SERVER_SELECTION_LOGGER.isEnabledFor(logging.DEBUG)
```
NoahStapp, line 655: A user could dynamically change the logging level after construction, while the driver is still inside the `_select_servers_loop` call. Edge case, so not sure if it's worth having to check on every emit call.

pymongo/_telemetry.py
```diff
@@ -183,3 +199,530 @@ def failed(
...
+            assert self._listeners is not None
+            self._listeners.publish_connection_created(self._address, conn_id)
+
+    def connection_ready(self, conn_id: int, creation_time: float) -> None:
+        """Emit the connection ready log entry and APM event."""
+        should_log = self._should_log
+        should_publish = self._should_publish
+        if not should_log and not should_publish:
+            return
+        duration = max(0.0, time.monotonic() - creation_time)
```
NoahStapp: Can all of these identical calls use our existing helper `_monotonic_duration()`, moved somehow more appropriate?

pymongo/asynchronous/server.py
```diff
@@ -59,11 +54,8 @@ def __init__(
         self._pool = pool
         self._monitor = monitor
         self._topology_id = topology_id
```
NoahStapp: No longer used.

pymongo/asynchronous/topology.py
```diff
@@ -435,18 +367,7 @@ async def select_server(
...
-                message=_ServerSelectionStatusMessage.SUCCEEDED,
-                selector=selector,
-                operation=operation,
-                operationId=operation_id,
-                topologyDescription=self.description,
-                clientId=self.description._topology_settings._topology_id,
-                serverHost=server.description.address[0],
-                serverPort=server.description.address[1],
-            )
+        ss.succeeded(server.description.address[0], server.description.address[1], self.description)
```
NoahStapp: `succeeded` doesn't use any state from `_ServerSelectionTelemetry` not already present here, should we refactor it into an independent method like `log_srv_monitor_failure`? That way we don't need to muddle the return values + types of `_select_servers_loop` and its callers.

NoahStapp (changes requested):

pymongo/_telemetry.py
blink1073: Done

pymongo/_telemetry.py
blink1073: I'd say no, either the whole transaction should be logged or not at all.

pymongo/_telemetry.py
blink1073: Moved and shared

pymongo/asynchronous/server.py
blink1073: Removed

pymongo/asynchronous/topology.py
blink1073: Good call

NoahStapp (approved):

PR #2925, opened 2026-07-09
PYTHON-5930 Fix SSL_CERT_FILE/SSL_CERT_DIR trust store handling

blink1073:
PYTHON-5930

## Changes in this PR
Fixes a test failure caused by `SSL_CERT_FILE`/`SSL_CERT_DIR` being merged with the OS/certifi certificate store instead of replacing it, on Windows and on macOS when using PyOpenSSL. The fix is scoped to just those platform/backend combinations, since elsewhere `load_default_certs()` already honors these env vars correctly (including falling back to the platform default for whichever of the two is left unset).

Also fixes an unrelated Evergreen config bug found while validating this change: the `pyopenssl-macos` variant never actually ran on macOS. `create_pyopenssl_variants()` wasn't passing `host=host` to `create_variant()`, so it silently fell back to a RHEL host. The `pyopenssl-win64` variant only worked by accident because of a separate win64 name-based special case. Without this fix, the macOS+PyOpenSSL half of this change could never have been verified in CI.

## Test Plan
Ran `just typing` and `just lint`; both pass. Added unit tests covering both the Windows bypass and the preserved fallback behavior elsewhere. Confirmed with an Evergreen patch build across `test-win64`, `pyopenssl-win64`, and `pyopenssl-macos` (now genuinely running on `macos-14`): all tasks passed — [patch #10613](https://spruce.corp.mongodb.com/version/6a50239632468c00072fb9ef).

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer


test/asynchronous/test_ssl.py
```diff
@@ -136,6 +136,42 @@ def test_config_ssl(self):
...
+        with (
+            mock.patch.dict(os.environ, env, clear=True),
+            mock.patch.object(sys, "platform", "win32"),
+            mock.patch("ssl.SSLContext.load_default_certs") as mock_default,
+            mock.patch("ssl.SSLContext.load_verify_locations") as mock_verify,
+        ):
+            get_ssl_context(None, None, None, None, False, False, False, False)
+            mock_verify.assert_called_once_with(cafile=CA_PEM, capath=None)
+            mock_default.assert_not_called()
+
```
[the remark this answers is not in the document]
blink1073: Added `test_ssl_cert_file_env_var_bypasses_default_certs_on_macos_pyopenssl` covering the darwin+PyOpenSSL bypass branch directly.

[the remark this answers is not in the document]
blink1073: Done

blink1073:
I'm investigating the ECS failure separately.  It is failing on the waterfall as well.

pymongo/ssl_support.py
```diff
@@ -133,7 +135,17 @@ def get_ssl_context(
         if ca_certs is not None:
             ctx.load_verify_locations(ca_certs)
         elif verify_mode != CERT_NONE:
-            ctx.load_default_certs()
+            cert_file = os.environ.get("SSL_CERT_FILE")
+            cert_dir = os.environ.get("SSL_CERT_DIR")
+            # load_default_certs() wrongly merges in the OS/certifi store on
+            # Windows, and on macOS with PyOpenSSL; bypass it only there.
```
NoahStapp:
```suggestion
            # load_default_certs() wrongly merges in the OS/certifi store on
            # Windows and on macOS with PyOpenSSL
```

test/asynchronous/test_ssl.py
```diff
@@ -136,6 +136,59 @@ def test_config_ssl(self):
     def test_use_pyopenssl_when_available(self):
         self.assertTrue(HAVE_PYSSL)
 
+    def test_ssl_cert_file_env_var_preserves_default_fallback_off_windows(self):
```
NoahStapp: This test has "windows" in the name but tests correct behavior for Linux? The comment is also unclear, I don't really follow.

test/asynchronous/test_ssl.py
```diff
@@ -136,6 +136,59 @@ def test_config_ssl(self):
...
+            mock.patch.object(sys, "platform", "linux"),
+            mock.patch("ssl.SSLContext.load_default_certs") as mock_default,
+            mock.patch("ssl.SSLContext.load_verify_locations") as mock_verify,
+        ):
+            get_ssl_context(None, None, None, None, False, False, False, False)
+            mock_default.assert_called_once()
+            mock_verify.assert_not_called()
+
+    def test_ssl_cert_file_env_var_bypasses_default_certs_on_windows(self):
+        # PYTHON-5930: on win32 (and macOS+PyOpenSSL), load_default_certs() merges
```
NoahStapp: I don't think this comment needs to mention macOS + PyOpenSSL since the following test does.

test/test_ssl.py
```diff
@@ -135,6 +135,59 @@ def test_config_ssl(self):
     def test_use_pyopenssl_when_available(self):
         self.assertTrue(HAVE_PYSSL)
 
+    def test_ssl_cert_file_env_var_preserves_default_fallback_off_windows(self):
```
NoahStapp: These tests don't need to be async/sync, can we consolidate into a single shared test file?

NoahStapp (changes requested):

test/test_ssl.py
blink1073: I think that should be deferred to the code consolidation work

test/asynchronous/test_ssl.py
blink1073: Fixed

test/asynchronous/test_ssl.py
blink1073: Fixed

test/asynchronous/test_ssl.py
blink1073: Renamed to `test_ssl_cert_file_env_var_uses_default_certs_on_linux` and simplified the comment.

test/asynchronous/test_ssl.py
blink1073: Dropped the macOS+PyOpenSSL mention from this comment, good catch.

test/test_ssl.py
NoahStapp: I'd prefer not to add code that we know is already duplicated at creation.

blink1073: Went further than just the 3 SSL_CERT_FILE tests: moved the entire `TestClientSSL` class (all pure unit tests, no async dependency) to a new non-mirrored `test/test_ssl_support.py`, following the existing `test_pyopenssl_context.py` precedent for standalone test files. Removed it from both `test/asynchronous/test_ssl.py` and the generated `test/test_ssl.py`.

blink1073: Fair enough, I moved all the unit tests to a new module

pymongo/ssl_support.py
```diff
@@ -133,7 +135,17 @@ def get_ssl_context(
         if ca_certs is not None:
             ctx.load_verify_locations(ca_certs)
         elif verify_mode != CERT_NONE:
-            ctx.load_default_certs()
+            cert_file = os.environ.get("SSL_CERT_FILE")
+            cert_dir = os.environ.get("SSL_CERT_DIR")
```
NoahStapp:
```suggestion
            cert_file = os.environ.get("SSL_CERT_FILE") or None
            cert_dir = os.environ.get("SSL_CERT_DIR") or None
```
Prevent the `SSL_CERT_FILE=""` case from failing `ctx.load_verify_locations` below.

test/test_ssl_support.py
```diff
@@ -0,0 +1,410 @@
...
+if HAVE_SSL:
+    import ssl
+
+CERT_PATH = os.path.join(pathlib.Path(__file__).resolve().parent, "certificates")
+CLIENT_PEM = os.path.join(CERT_PATH, "client.pem")
+CA_PEM = os.path.join(CERT_PATH, "ca.pem")
+CRL_PEM = os.path.join(CERT_PATH, "crl.pem")
+
+
+class TestClientSSL(unittest.TestCase):
```
NoahStapp: Shouldn't every test in this class (except the first) have `@unittest.skipUnless(HAVE_SSL)` or `@unittest.skipUnless(_HAVE_PYOPENSSL)`?

test/test_ssl_support.py
```diff
@@ -0,0 +1,410 @@
...
+except ImportError:
+    pass
+
+if HAVE_SSL:
+    import ssl
+
+CERT_PATH = os.path.join(pathlib.Path(__file__).resolve().parent, "certificates")
+CLIENT_PEM = os.path.join(CERT_PATH, "client.pem")
+CA_PEM = os.path.join(CERT_PATH, "ca.pem")
+CRL_PEM = os.path.join(CERT_PATH, "crl.pem")
```
NoahStapp: `CERT_PATH`, `CLIENT_PATH`, and `CA_PEM` are all already defined in `helpers_shared.py`.

test/test_ssl_support.py
```diff
@@ -0,0 +1,410 @@
...
+            mock.patch.dict(os.environ, env, clear=True),
+            mock.patch.object(sys, "platform", "darwin"),
+            mock.patch("pymongo.pyopenssl_context.SSLContext.load_default_certs") as mock_default,
+            mock.patch("pymongo.pyopenssl_context.SSLContext.load_verify_locations") as mock_verify,
+        ):
+            get_ssl_context(None, None, None, None, False, False, False, True)
+            mock_verify.assert_called_once_with(cafile=CA_PEM, capath=None)
+            mock_default.assert_not_called()
+
+    def test_ssl_session_cache(self):
```
NoahStapp: This test doesn't appear to test anything beside list behavior.

test/test_ssl_support.py
```diff
@@ -0,0 +1,410 @@
...
+CLIENT_PEM = os.path.join(CERT_PATH, "client.pem")
+CA_PEM = os.path.join(CERT_PATH, "ca.pem")
+CRL_PEM = os.path.join(CERT_PATH, "crl.pem")
+
+
+class TestClientSSL(unittest.TestCase):
+    @unittest.skipIf(HAVE_SSL, "The ssl module is available, can't test what happens without it.")
+    def test_no_ssl_module(self):
+        # Explicit
+        self.assertRaises(ConfigurationError, MongoClient, ssl=True)
```
NoahStapp: Prefer using `PyMongoTestCase.simple_client` here to avoid possible leaks if an assertion fails.

test/test_ssl_support.py
NoahStapp:
```suggestion
class TestClientSSL(PyMongoTestCase):
```

NoahStapp (changes requested):

blink1073:
Making a new file was a mistake, I'm going to revert it.  This is exactly the scope creep I was trying to avoid.

NoahStapp:
> Making a new file was a mistake, I'm going to revert it. This is exactly the scope creep I was trying to avoid.

Can we move just the new tests there and open a follow-up ticket to refactor the existing test suite? Avoids scope creep but doesn't add known tech debt here.

PR #2935, opened 2026-07-14
PYTHON-5724 Add integration test exercising compression logic

aclark4life:
[PYTHON-5724](https://jira.mongodb.org/browse/PYTHON-5724)

## Changes in this PR
Adds an integration test that sends real commands with each available wire-protocol compressor enabled, exercising the end-to-end compress/decompress path in `compression_support.py`.

- Adds `test_compression_commands` in `test/asynchronous/test_client.py` (sync mirror generated by `just synchro`).

## Test Plan
For each available compressor, the test handshakes so the server negotiates it, then asserts the pooled connection received the correct compressor context type:

| Compressor | Context class | `compressor_id` |
|------------|---------------|-----------------|
| `snappy`   | `SnappyContext` | 1 |
| `zlib`     | `ZlibContext`   | 2 |
| `zstd`     | `ZstdContext`   | 3 |

It then round-trips an insert/find of compressible data with both directions instrumented:

- **Send path:** spies on the connection's `compression_context.compress()` and asserts it was invoked.
- **Receive path:** wraps `pymongo.network_layer.decompress` (the symbol both the sync and async read paths call) and asserts it was invoked with the expected `compressor_id`. A plain equality check on the returned document would pass whether or not the reply was compressed, so the spy is what makes this a real assertion about decompression.

If the server negotiates no compression at all, the test reports as skipped rather than passing with nothing exercised. Compressors whose libraries are unavailable are not attempted; `zlib` is always tried since it is stdlib.

### Checklist for Author
- [ ] Did you update the changelog (if necessary)? N/A — test-only change.
- [x] Is there test coverage? This PR is test coverage for `compression_support.py`.
- [ ] Is any followup work tracked in a JIRA ticket? If so, add link(s). N/A.

### Checklist for Reviewer


test/asynchronous/test_client.py
```diff
@@ -1897,6 +1903,75 @@ def compression_settings(client):
...
+                    decompressed = []
+                    original_decompress = network_layer.decompress
+
+                    def decompress_spy(
+                        data, compressor_id, _original=original_decompress, _recorded=decompressed
+                    ):
+                        _recorded.append(compressor_id)
+                        return _original(data, compressor_id)
+
+                    # Round-trip a command large enough to compress.
```
NoahStapp: All non-sensitive commands are compressed, we don't have a size limit.

NoahStapp (changes requested):

NoahStapp (approved):

PR #2948, opened 2026-07-21
PYTHON-5786 Add test variant for the Secure Frontend Processor (monguard)

NoahStapp:
[PYTHON-5786](https://jira.mongodb.org/browse/PYTHON-5786)

## Changes in this PR

Adds a SFP (monguard) test variant in accordance with the SFP testing spec.

## Test Plan

Adds the tests described by the SFP testing spec. 

[Evergreen patch](https://spruce.corp.mongodb.com/version/6a5f8a86acf78a00077fccb3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

### Checklist for Author
- [X] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/.evergreen/scripts/setup_tests.py
+++ b/.evergreen/scripts/setup_tests.py
@@ -436,6 +436,19 @@ def handle_test_env() -> None:
         # We do not want the default client_context to be initialized.
         write_env("DISABLE_CONTEXT")
 
+    if test_name == "sfp":
+        secrets = get_secrets("drivers/sfp")
+
+        # Write file with SFP Atlas X509 client certificate:
+        decoded = base64.b64decode(secrets["SFP_ATLAS_X509_BASE64"]).decode("utf8")
+        cert_file = ROOT / ".evergreen/atlas_x509_sfp_client_certificate.pem"
+        with cert_file.open("w") as file:
+            file.write(decoded)
+        write_env("SFP_ATLAS_X509_CERT", str(cert_file))
+
+        # We do not want the default client_context to be initialized.
+        write_env("DISABLE_CONTEXT")
+
     if test_name == "numpy":
         UV_ARGS.append("--with numpy")
 
--- a/test/test_sfp.py
+++ b/test/test_sfp.py
@@ -1,0 +1,111 @@
+# Copyright 2026-present MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


.evergreen/scripts/setup_tests.py
```diff
@@ -436,6 +436,19 @@
...
         write_env("DISABLE_CONTEXT")
 
+    if test_name == "sfp":
+        secrets = get_secrets("drivers/sfp")
+
+        # Write file with SFP Atlas X509 client certificate:
+        decoded = base64.b64decode(secrets["SFP_ATLAS_X509_BASE64"]).decode("utf8")
+        cert_file = ROOT / ".evergreen/atlas_x509_sfp_client_certificate.pem"
+        with cert_file.open("w") as file:
+            file.write(decoded)
```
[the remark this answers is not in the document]
NoahStapp: This is written to an EG host, owned and controlled by us.

test/test_sfp.py
```diff
@@ -0,0 +1,111 @@
...
+    "compressor": {"compressors": "zlib"},
+    "server_api": {"server_api": ServerApi("1")},
+}
+
+
+def _require_env(name: str) -> str:
+    value = os.environ.get(name)
+    if not value:
+        raise Exception(f"Must set {name} env variable to test.")
+    return value
```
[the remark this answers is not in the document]
NoahStapp: I think we want to fail the suite here since this is about explicitly testing SFP, which requires the vars to be set.

test/test_sfp.py
```diff
@@ -0,0 +1,111 @@
...
+        client = self.simple_client(_require_env("SFP_ATLAS_URI"))
+        self.assert_ping(client)
+        self.assert_connection_status(client, authenticated=False)
+
+    def test_scram_sha_256(self):
+        uri = _require_env("SFP_ATLAS_URI")
+        username = _require_env("SFP_ATLAS_USER")
+        password = _require_env("SFP_ATLAS_PASSWORD")
+        for variation, kwargs in VARIATIONS.items():
+            with self.subTest(variation=variation):
```
blink1073, line 87: This requires [pytest 9.0+](https://docs.pytest.org/en/stable/changelog.html#id108).  We need to bump our min version in requirements and pytest.ini config

NoahStapp: This is actually [unittest.subTest](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest), not the `pytest` version. Unittest support was added in Python 3.4.

blink1073: "In addition, [unittest.TestCase.subTest()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest) is now also supported."

blink1073: I explicitly looked this up because I knew this was a feature gap, I've run into it before

NoahStapp: With pytest < 9.0, subtests still run, but if one fails it stops there instead of running the others.

blink1073: Cool, thanks for verifying

blink1073 (approved):
LGTM

PR #2961, opened 2026-07-26
PYTHON-5968 Reject extra fields in Extended JSON timestamps

winklemad:
PYTHON-5968

## Changes in this PR

`bson.json_util.loads` now rejects `$timestamp` value documents that contain anything other than the required `t` and `i` components. Previously, unexpected fields could be silently discarded while constructing `bson.timestamp.Timestamp`.

This adds regressions for both additional fields and same-length invalid key sets, and updates the changelog and contributor list.

AI assistance disclosure: I used OpenAI Codex to help investigate the existing behavior, search for duplicates, implement the change, and run validation. I reviewed and understand every changed line and can explain and maintain the contribution.

## Test Plan

- `python3 -m unittest test.test_json_util.TestJsonUtil.test_timestamp test.test_json_util.TestJsonUtil.test_timestamp_with_invalid_fields`
- `python3 -m unittest test.test_json_util test.test_bson_corpus`
- `ruff check bson/json_util.py test/test_json_util.py`
- `ruff format --check bson/json_util.py test/test_json_util.py`
- `python3 -m compileall -q bson/json_util.py test/test_json_util.py`
- `git diff --check`

All listed checks passed.

### Checklist for Author
- [x] Is there test coverage?
- [ ] Is any followup work tracked in a JIRA ticket? If so, add links.

No follow-up work is currently identified. As an external contributor without an existing JIRA ticket, I retained the template `PYTHON-XXXX` placeholder for a MongoDB employee to update.

### Checklist for Reviewer
- [x] Do you fully understand the implementation?
- [x] Is all relevant documentation updated?

### DIFF
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -788,7 +788,11 @@ def _parse_binary(doc: Any, json_options: JSONOptions) -> Union[Binary, uuid.UUI
 
 
 def _parse_timestamp(doc: Any, dummy0: Any) -> Timestamp:
+    if len(doc) != 1:
+        raise TypeError(f"Bad $timestamp, extra field(s): {doc}")
     tsp = doc["$timestamp"]
+    if not isinstance(tsp, Mapping):
+        raise TypeError(f'$timestamp value must be a document with "t" and "i" components: {doc}')
     if set(tsp) != {"t", "i"}:
         raise TypeError(f'$timestamp must include exactly "t" and "i" components: {doc}')
     return Timestamp(tsp["t"], tsp["i"])
--- a/bson/json_util.py
+++ b/bson/json_util.py
@@ -789,6 +789,8 @@ def _parse_binary(doc: Any, json_options: JSONOptions) -> Union[Binary, uuid.UUI
 
 def _parse_timestamp(doc: Any, dummy0: Any) -> Timestamp:
     tsp = doc["$timestamp"]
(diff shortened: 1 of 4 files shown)


blink1073:
Hi @winklemad, thank you for the fix!  I agree with the copilot suggestion above.  We should also add tests for the two new guards.

winklemad:
Thanks @blink1073 — agreed on both.

For the wrapper: I'll add the same `if len(doc) != 1` guard the other `_parse_*` helpers use, so `{"$timestamp": {"t": 1, "i": 2}, "extra": 1}` is rejected instead of silently dropping the `extra` field once `object_hook` matches on `$timestamp`. And I'll add a `Mapping` check on the inner value before `set(tsp)`, so something like `{"$timestamp": ["t", "i"]}` fails with a clear error instead of the confusing `list indices must be integers` it hits today.

I'll add regression tests for both new guards next to the existing one. Thanks for the review!

winklemad:
Pushed both guards plus the two regression tests in a9d62cc — the outer-wrapper `len(doc) != 1` check and the `Mapping` guard on the value. Ready for another look, thanks!

blink1073 (approved):
Thank you!

winklemad:
Thanks @blink1073 — really appreciate the quick review, and for tidying up the formatting and sorting the ticket on your end.

sleepyStick (approved):

PR #2964, opened 2026-07-28
PYTHON-5947 Add OpenTelemetry Operation and Transaction Span Support

blink1073:
PYTHON-5947

## Changes in this PR

Extends the OpenTelemetry support added in PYTHON-5945 with:

- **Operation-level spans**: one span per public API call, spanning all retry attempts, with each attempt's command span nested underneath. A cursor's whole lifetime is a single operation span, so `getMore` commands nest under the originating `find`/`aggregate` instead of producing spans of their own.
- **Transaction pseudo-spans**: a `"transaction"` span wrapping `start_transaction()` through `commit_transaction()`/`abort_transaction()`, with the operations inside nested under it. A retried `with_transaction()` produces one span covering all attempts.
- **Unified test format support**: `observeTracingMessages`/`expectTracingMessages`, plus the vendored spec test suite.

Failure paths stay conformant: an operation that fails before any command reaches the wire still carries `db.operation.name` and `db.operation.summary`, and `killCursors`/`endSessions` get operation spans rather than leaving their command spans unparented.

Opt-in, with no behavior change unless the `tracing` client option or `OTEL_PYTHON_INSTRUMENTATION_MONGODB_ENABLED` is set. When tracing is off, no telemetry object is constructed at all, following the fast path PYTHON-5977 introduced for command telemetry. Enablement is resolved once when the client is built rather than re-read from the environment on every command, which takes the check from 248 ns to 17 ns on PyPy and 532 ns to 82 ns on CPython.

## Reviewer's guide

Of ~9,300 added lines, most aren't hand-written:

| | Lines | Needs review? |
|---|---|---|
| Vendored spec fixtures (`test/open_telemetry/*.json`) | ~4,500 | No, `resync-specs.sh` output |
| Generated sync mirrors (`pymongo/synchronous/*`, mirrored `test/*`) | ~2,000 | No, `just synchro` output |
| Hand-written | ~2,700 | Yes, and ~1,400 of that is `test_otel.py` |

**Where to look.** `pymongo/_otel.py` is the core (~500 lines): span primitives plus all of the spec's naming and attribute rules. `pymongo/_telemetry.py` holds the lifecycle wrappers and no spec knowledge. In `mongo_client.py`, `_retry_internal` gained three span modes: owned, caller-owned, and passthrough. The two decisions most worth scrutinising are the `parent_span` parameter in `_otel.py` (transaction parenting is passed explicitly rather than read from ambient context) and cursor-lifetime span ownership in `cursor_shared.py`, where `_end_operation_telemetry` is idempotent because close, GC, and exception paths can all reach it.

**Already considered:**

- **`bson/json_util.py` is touched.** `_truncate_documents` silently dropped falsy values (`0`, `False`, `""`) from truncated output. It's shared with `pymongo/logger.py`, so this affected structured command logging too. Has a regression test.
- **Change streams opt out of `getMore` nesting**, because a tailing stream's operation span would never end.
- **Background monitors emit no spans**, so `close()` needs no special handling. A monitor's `hello` goes through `Connection.command` without a client, so its tracing options are None, which means untraced; CMAP, heartbeat and SDAM telemetry never touch `_otel` either. Enabling tracing through the environment variable used to be an exception here, tracing every heartbeat, because the old enablement check fell back to reading the environment when there was no client. Fixed, with a regression test.
- **Hand-written tests cover only what the vendored suite can't.** The 24 vendored cases cover per-operation attributes and nesting for ~20 operations with `ignoreExtraSpans: false`; duplicates of those were dropped. What's left is cursor lifetime beyond the single `get_more.json` fixture, exception attributes on an *operation* span (fixtures only assert them on command spans), span status, `killCursors`/`endSessions`, change streams, and config handling.
- **Two vendored fixtures needed fixing.** `update.json` rejected drivers that send `multi`/`upsert` at their defaults, and the fixtures that create collections declared no `initialData`, so they weren't repeatable. Both are fixed in DRIVERS-3597, which has merged, so the fixtures here match `mongodb/specifications` master and carry no local deviation.
- **Forward compatibility with DRIVERS-3454** ([specifications#1966](https://github.com/mongodb/specifications/pull/1966)) was checked. It's purely additive to the spec and propagates the *command* span's context, which is the nesting this PR establishes, parented to the exact retry attempt. The one change it needs is unrelated to this PR: `run_command` encodes the OP_MSG before the command span exists, so span creation has to move ahead of encoding. Tracked in PYTHON-5855.

## Test Plan

- Unit tests for the span primitives, and integration tests covering operation spans, cursor and transaction nesting, retry and failure paths, and sensitive-command redaction.
- The vendored OpenTelemetry spec suite.
- The otel Evergreen variant now covers all three topologies, held at 22 tasks: replica set in full (the only topology where transaction spans run, and the only one with free-threaded Python), sharded on the newest CPython and PyPy, and standalone for its min-deps tasks, which resolve `opentelemetry-api` to its floor.
- Verified on Python 3.10, 3.13, and free-threaded 3.14, since two bugs in this area reproduced only on newer interpreters.

### Checklist for Author
- [x] Is there test coverage?
- [x] Is any followup work tracked in a JIRA ticket? If so, add link(s).

No follow-up work outstanding from this ticket. PYTHON-5978 was filed separately for two pre-existing bugs noticed in passing: the sync driver's `Monitor.join()` and `Topology.cleanup_monitors()` both wrap their `join()` calls in an `asyncio.gather(...)` that is never awaited, so neither actually waits.

### Checklist for Reviewer


bson/json_util.py
```diff
@@ -1109,20 +1109,18 @@ def _truncate_documents(obj: Any, max_length: int) -> tuple[Any, int]:
     if hasattr(obj, "items"):
         truncated: Any = {}
         for k, v in obj.items():
-            truncated_v, remaining = _truncate_documents(v, remaining)
```
blink1073, line 1118: These changes were needed for the new `distinct.json`, which has a falsey value for the field.  These fields should not be dropped.

PR #2902, opened 2026-06-29
PYTHON-5862 - Remove usage of uv lock file

NoahStapp:
[PYTHON-5862](https://jira.mongodb.org/browse/PYTHON-5862)

## Changes in this PR

Removes all uses of `uv.lock` and prevents local or CI usage of `uv` from creating new lockfiles. Excludes dependency releases made in the last 7 days to prevent accidental breaking changes from making it into CI.

## Test Plan

N/A.

### Checklist for Author
- ~[ ] Did you update the changelog (if necessary)?~
- ~[ ] Is there test coverage?~
- ~[ ] Is any followup work tracked in a JIRA ticket? If so, add link(s).~

### Checklist for Reviewer
- [ ] Does the title of the PR reference a JIRA Ticket?
- [ ] Do you fully understand the implementation? (Would you be comfortable explaining how this code works to someone else?)
- [ ] Is all relevant documentation (README or docstring) updated?

PR #2966, opened 2026-07-29
PYTHON-5862 Cleanup lockless uv usage

NoahStapp:
[PYTHON-5862](https://jira.mongodb.org/browse/PYTHON-5862)

## Changes in this PR

Removes the no-op `UV_NO_LOCK` environment variable usage (it doesn't exist anywhere in uv's docs, pure hallucination) in our CI alongside several other confusing and non-functional changes introduced by https://github.com/mongodb/mongo-python-driver/pull/2902. Also fixes the persistent failure of [test-auth-aws-ecs](https://spruce.corp.mongodb.com/task/mongo_python_driver_auth_aws_ecs_macos_test_auth_aws_ecs_patch_48238a1a84dda761030d3d197ed70e1b64ea14a4_6a6a13143706ca0007ae5de0_26_07_29_14_50_10?execution=0) due to the lockfile removal.

## Test Plan

Test infra changes.

### Checklist for Author
- ~[ ] Is there test coverage?~

### Checklist for Reviewer


NoahStapp:
Test failures will be resolved by [PYTHON-5974](https://jira.mongodb.org/browse/PYTHON-5974).

blink1073:
So in general we're still creating lock files, but they're git ignored?  If so we may want to reconsider because our local dev with almost never be the same as CI.

NoahStapp:
> So in general we're still creating lock files, but they're git ignored? If so we may want to reconsider because our local dev with almost never be the same as CI.

There isn't currently a way to prevent invocations of `uv run`, `uv sync`, and the like from creating a lockfile. Multiple invocations of `uv` on one machine will lead to a lockfile being used by the latter runs. That seems fine to me, especially if we decide to have our CI scripts cleanup `uv.lock` files as part of setup/teardown.

blink1073:
What I mean is, we might be better off following astral's [advice](https://docs.astral.sh/uv/guides/projects/#uvlock) and checking in the lock file, and then having a job that periodically updates the lock file.  This would mean going back to using "pip" in the dependabot config to avoid the spurious update PRs.

blink1073 (approved):
LGTM.  I opened https://jira.mongodb.org/browse/PYTHON-5980 as a recommended follow up

PR #2968, opened 2026-07-30
PYTHON-5977 APM/logging hot-path optimizations

NoahStapp:
PYTHON-5977

## Changes in this PR

The changes introduced by [PYTHON-5903](https://jira.mongodb.org/browse/PYTHON-5903), [PYTHON-5846](https://jira.mongodb.org/browse/PYTHON-5846), [PYTHON-5745](https://jira.mongodb.org/browse/PYTHON-5745), and [PYTHON-5676](https://jira.mongodb.org/browse/PYTHON-5676) caused small but measurable regressions for low-latency, high operation count workloads such as the DSI locust_ping benchmark. Much of this is due to constructing telemetry objects even when APM/logging is disabled and other similar changes that lie on the hot path for such workloads.

This PR attempts to optimize all of those regressions away and return performance to its point before their addition.

## Test Plan

Standard test suite, benchmarks.

Results of a simple ping workload benchmark:
| Metric | PR | main | main vs PR |
|---|---|---|---|
| Throughput | 10,634/s | 6,567/s | **−38.8%** |
| Mean latency | 752 µs | 1,219 µs | **+62%** |
| p50 | ~170 µs | ~800 µs | ~4–5× worse |
| p99 | ~5,000 µs | ~7,000 µs | +40% |

### Checklist for Author
- ~[ ] Is there test coverage?~

### Checklist for Reviewer


NoahStapp:
Failures are unrelated.

pymongo/asynchronous/command_runner.py
```diff
@@ -159,11 +163,20 @@ async def _run_command(
...
         orig = cmd
-    if op_id is None:
-        op_id = _op_id.OP_ID.get()
 
-    telemetry = _CommandTelemetry(topology_id, conn, listeners, cmd, dbname, request_id, op_id)
-    telemetry.started(orig, ensure_db)
+    # Fast path: skip telemetry construction when logging and APM are disabled
+    # Inline enabled check here for performance
+    telemetry: Optional[_CommandTelemetry] = None
+    if (topology_id is not None and _COMMAND_LOGGER.isEnabledFor(logging.DEBUG)) or (
```
blink1073: I think this part should be a helper in `_telemetry.py`

pymongo/asynchronous/pool.py
```diff
@@ -1113,7 +1115,11 @@ async def checkin(self, conn: AsyncConnection) -> None:
         self._pinned_sockets.discard(conn)
         async with self.lock:
             self.active_contexts.discard(conn.cancel_context)
-        self._telemetry.checked_in(conn.id)
+        telemetry = self._telemetry
+        if telemetry._should_publish or (
```
blink1073: This logic should also be in `_telemetry.py`, perhaps as a property on the class

pymongo/asynchronous/pool.py
```diff
@@ -1230,12 +1236,20 @@ def __init__(
 
     async def __aenter__(self) -> AsyncConnection:
         pool = self._pool
-        checkout_started_time = pool._telemetry.checkout_started()
+        telemetry = pool._telemetry
+        # Fast path: skip telemetry calls when CMAP events/logging are disabled
+        if not telemetry._should_publish and not (
```
blink1073, line 1238: Also could be factored into `_telemetry.py`

pymongo/asynchronous/topology.py
```diff
@@ -283,10 +285,13 @@ async def _select_servers_loop(
         now = time.monotonic()
         end_time = now + timeout
         logged_waiting = False
-        ss = _ServerSelectionTelemetry(
-            self._topology_id, selector, operation, operation_id, self.description
-        )
-        ss.started()
+        # Server selection does not have APM events, gate only on logging
+        ss: Optional[_ServerSelectionTelemetry] = None
+        if _SERVER_SELECTION_LOGGER.isEnabledFor(logging.DEBUG):
```
blink1073: Also can be factored into `_telemetry.py`

pymongo/asynchronous/topology.py
```diff
@@ -371,15 +377,16 @@ async def select_server(
...
-        log_server_selection_succeeded(
-            self._topology_id,
-            selector,
-            operation,
-            operation_id,
-            self.description,
-            server.description.address[0],
-            server.description.address[1],
-        )
+        if _SERVER_SELECTION_LOGGER.isEnabledFor(logging.DEBUG):
```
blink1073: Same here, we shouldn't need to have this type of logic in multiple places.

pymongo/asynchronous/bulk.py
```diff
@@ -455,7 +456,8 @@ async def execute_command(
             "nRemoved": 0,
             "upserted": [],
         }
-        op_id = _randint()
+        client = self.collection.database.client
+        op_id = _randint() if _should_generate_op_id(client._event_listeners) else None
```
blink1073: Maybe this function should either return a `_randint` or `None` itself

blink1073 (changes requested):

pymongo/asynchronous/command_runner.py
NoahStapp: This is intentionally inlined here for performance since this is called on every single operation, moving to a helper is a small performance hit that compounds across each time the inlined check is abstracted.

pymongo/asynchronous/pool.py
NoahStapp: Same here as above, inlined intentionally for performance.

pymongo/asynchronous/pool.py
NoahStapp: Same here as well.

pymongo/asynchronous/bulk.py
NoahStapp: I like it.

pymongo/asynchronous/topology.py
blink1073: This one needs the new helper function

blink1073 (dismissed):
LGTM!

blink1073 (approved):

PR #2976, opened 2026-08-01
PYTHON-5983 Validate uncompressed size in OP_COMPRESSED messages

dfengliu:
[PYTHON-5983](https://jira.mongodb.org/browse/PYTHON-5983)

## Summary

Validate the `uncompressed_size` field from the OP_COMPRESSED wire protocol compression sub-header against `max_message_size`.

## Details

The `process_compression_header` method in `network_layer.py` previously unpacked the compression sub-header and discarded the `uncompressed_size` field. A malicious or compromised MongoDB server could send a small compressed envelope (passing the envelope size check) that decompresses to a very large payload, causing memory exhaustion.

## Changes

- `process_compression_header` now returns `uncompressed_size` in addition to `op_code` and `compressor_id`
- The caller validates `uncompressed_size` against `self._max_message_size` and raises `ProtocolError` if it exceeds the limit
- Added unit test `test_compression_uncompressed_size_exceeds_max_closes`


blink1073:
Hi @dfengliu, thanks for the PR!  Could you move the bound into `decompress()` in
`pymongo/compression_support.py` instead? That is the right layer for it, and it covers both
the async and sync read paths in one place.

dfengliu:
@blink1073 Thank you for the suggestion. I have moved the size validation into `_decompress()` in `compression_support.py`, which covers both the async and sync read paths. The original `process_compression_header` has been reverted to its previous behavior.

blink1073 (changes requested):
Thanks for moving this forward. One thing still needs to change: the check runs after decompression completes, so it measures the result rather than bounding it. Both zlib and zstd accept a max length on their incremental decompressors, which would apply the limit during decompression instead.

Smaller items:

- The `2**31 - 1` default on `decompress()` is effectively unbounded, and nothing in the
  driver calls it now. Worth dropping it or defaulting to `MAX_MESSAGE_SIZE`.
- Please restore the snappy comment explaining the `bytes(data)` conversion.
- The test payload is small enough that it only exercises the comparison. A large expansion case would be more useful, and `test/test_compression_support.py` is a better home for it.
- `just lint-manual` picks up a few things, and this needs a `doc/changelog.rst` entry.

dfengliu:
@blink1073 Thank you for the detailed review. I have addressed the items you raised:

- Restored `decompress()` as the original standalone function, removing the `2**31 - 1` default.
- Restored the snappy `bytes(data)` comment that was lost during refactoring.
- Moved the size-limit test to `test/test_compression_support.py` with a high-expansion-ratio payload.
- Added a changelog entry.

Regarding the incremental decompressor suggestion: the current approach uses pre-validation (checking `uncompressed_size` from the OP_COMPRESSED sub-header before decompression in both async and sync paths) combined with the post-decompression size check in `_decompress()` as defense-in-depth. The pre-validation blocks oversized messages at zero memory cost, while the post-validation catches servers that misreport the uncompressed size. This provides more complete coverage than relying on the per-call `max_length` parameter in zlib's incremental API alone.

Would this approach be acceptable?

blink1073:
The layered structure is fine to keep, but the max length arguments need to go in as well.
Without them the trailing check can only run once the buffer already exists, so it reports
the size rather than capping it.

```diff
     elif compressor_id == ZlibContext.compressor_id:
         import zlib

-        result = zlib.decompress(data)
+        result = zlib.decompressobj().decompress(data, max_message_size + 1)
     elif compressor_id == ZstdContext.compressor_id:
         if sys.version_info >= (3, 14):
             from compression import zstd
         else:
             from backports import zstd

-        result = zstd.decompress(data)
+        result = zstd.ZstdDecompressor().decompress(data, max_message_size + 1)
```

Snappy has no equivalent, so it keeps relying on the trailing check.

Rest of the commit:

- `decompress()` has no callers in the driver now, and it's a near-copy of `_decompress`.
  Probably worth collapsing to one. The snappy comment is currently only on the unused one.
- The changelog entry landed inside the `operation_id` bullet, between "As a" and
  "result, ...". It needs its own bullet.
- The test's payload is 100 KB against a 1000 byte cap, so it passes either way. Something
  large enough to be obviously wrong unbounded would catch a regression here.

dfengliu:
@blink1073 Thank you for the detailed feedback. I have implemented the changes:

- Added `max_length` bounding during decompression for both zlib (`zlib.decompressobj().decompress(data, max_message_size + 1)`) and zstd (`zstd.ZstdDecompressor().decompress(data, max_message_size + 1)`), so memory is bounded before the size check runs. Snappy has no equivalent API and keeps relying on the trailing check.
- Collapsed `decompress()` and `_decompress()` into a single function with an optional `max_message_size` parameter, and restored the snappy `bytes(data)` comment.
- Fixed the changelog entry to be its own bullet.
- Enlarged the test payload to 10 MB with a high expansion ratio, moved it to `test/test_compression_support.py`.
- Added regression tests for the async/sync pre-validation of `uncompressed_size` and the decompression bomb path.

All 39 unit tests and the end-to-end scenarios pass.

blink1073:
Thank you!  We're getting close.  I have two more requests:

- Make `max_message_size` required rather than defaulting to `None`. Both call sites
  already pass it, and the `None` branch still reaches the unbounded `zlib.decompress`
  and `zstd.decompress` calls, which is the path this PR exists to remove.
- `TestDecompressSizeLimit` still passes if you take the `max_length` arguments back
  out, so it's covering the size comparison rather than the bound itself. It needs to
  fail without the fix.  The property to assert is that peak allocation stays near the
  limit rather than near the payload size, which `tracemalloc` can measure:

  ```python
  payload = zlib.compress(b"\x00" * 100_000_000)
  max_size = 1_000_000
  tracemalloc.start()
  try:
      with self.assertRaises(ProtocolError):
          decompress(payload, ZlibContext.compressor_id, max_message_size=max_size)
      _current, peak = tracemalloc.get_traced_memory()
  finally:
      tracemalloc.stop()
  self.assertLess(peak, 10 * max_size)
  ```

  With the bound that peak lands around 2 MB. Without it, around 215 MB. That is a
  wide enough gap that the threshold won't be flaky across platforms.

dfengliu:
@blink1073 Thank you for raising these points. Both requests are now addressed:

- max_message_size is now a required parameter in decompress() — the None default and the unbounded zlib.decompress/zstd.decompress branches have been removed. Both call sites already pass it.
- Added a tracemalloc-based peak memory test (test_decompression_peak_memory_bounded): it decompresses a 100 MB payload with a 1 MB limit and asserts that peak allocation stays below 10 MB. With the max_length bound it lands around 2 MB; without it, around 215 MB — so the test fails without the fix.
- Also added exact-boundary (== max accepted, max+1 rejected) and snappy over-limit regression tests.

Additionally, I fixed a compatibility issue in the test_compression_commands spy that was causing CI failures after merging main — the spy now accepts and forwards the max_message_size argument.

CI is currently waiting on approval for both GitHub Actions and Evergreen. Could you help approve the runs when you have a moment? Thanks!

dfengliu:
> Thank you! We're getting close. I have two more requests:
> 
> * Make `max_message_size` required rather than defaulting to `None`. Both call sites
>   already pass it, and the `None` branch still reaches the unbounded `zlib.decompress`
>   and `zstd.decompress` calls, which is the path this PR exists to remove.
> * `TestDecompressSizeLimit` still passes if you take the `max_length` arguments back
>   out, so it's covering the size comparison rather than the bound itself. It needs to
>   fail without the fix.  The property to assert is that peak allocation stays near the
>   limit rather than near the payload size, which `tracemalloc` can measure:
>   ```python
>   payload = zlib.compress(b"\x00" * 100_000_000)
>   max_size = 1_000_000
>   tracemalloc.start()
>   try:
>       with self.assertRaises(ProtocolError):
>           decompress(payload, ZlibContext.compressor_id, max_message_size=max_size)
>       _current, peak = tracemalloc.get_traced_memory()
>   finally:
>       tracemalloc.stop()
>   self.assertLess(peak, 10 * max_size)
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   With the bound that peak lands around 2 MB. Without it, around 215 MB. That is a
>   wide enough gap that the threshold won't be flaky across platforms.

@blink1073 Thank you for raising these points. Both requests are now addressed:

- max_message_size is now a required parameter in decompress() — the None default and the unbounded zlib.decompress/zstd.decompress branches have been removed. Both call sites already pass it.
- Added a tracemalloc-based peak memory test (test_decompression_peak_memory_bounded): it decompresses a 100 MB payload with a 1 MB limit and asserts that peak allocation stays below 10 MB. With the max_length bound it lands around 2 MB; without it, around 215 MB — so the test fails without the fix.
- Also added exact-boundary (== max accepted, max+1 rejected) and snappy over-limit regression tests.

Additionally, I fixed a compatibility issue in the test_compression_commands spy that was causing CI failures after merging main — the spy now accepts and forwards the max_message_size argument.

CI is currently waiting on approval for both GitHub Actions and Evergreen. Could you help approve the runs when you have a moment? Thanks!

blink1073:
I agree with the copilot comment about `snappy`.  We should check the length the raw block declares before calling
`uncompress`. For a valid stream that value is the real output length, and `cramjam` rejects streams where it disagrees, so the check is sufficient:

```python
def _snappy_uncompressed_length(data: bytes | memoryview) -> int:
    """Read the varint-encoded uncompressed length from a raw snappy block."""
    result = shift = 0
    for i in range(5):
        if i >= len(data):
            raise ProtocolError("Truncated snappy payload")
        byte = data[i]
        result |= (byte & 0x7F) << shift
        if not byte & 0x80:
            return result
        shift += 7
    raise ProtocolError("Invalid snappy uncompressed length header")
```

Then before `snappy.uncompress(bytes(data))`:

```python
declared = _snappy_uncompressed_length(data)
if declared > max_message_size:
    raise ProtocolError(
        f"Decompressed message size ({declared!r}) is larger than "
        f"server max message size ({max_message_size!r})"
    )
```

`cramjam.snappy.decompress_raw_len(data)` does the same in one line, but `cramjam` is
`python-snappy`'s implementation detail rather than one of ours, and the requirement
has no upper bound, so we should parse it ourselves.

I think it is worth extending `test_decompression_peak_memory_bounded` to cover snappy on the same
pattern once this is in.

dfengliu:
@blink1073 Thank you! I have addressed the snappy gap:

- Added `_snappy_uncompressed_length()` which parses the varint-encoded length of the raw snappy block header, and rejects declarations larger than `max_message_size` before calling `snappy.uncompress()`, so a malicious raw stream can no longer force an unbounded allocation.
- The pre-check runs before importing `snappy`; truncated or overlong varint headers raise `ProtocolError`.
- Extended the `tracemalloc` peak-memory test to snappy and zstd: without the bound, raw `snappy.uncompress` allocation peaks at ~100 MB on a 100 MB payload; with the pre-check it stays near the limit.
- Added unit tests for the varint parser (single/multi-byte, truncated, overlong) and a declared-over-limit rejection test that works even when `python-snappy` is not installed.

pymongo/compression_support.py
```diff
@@ -164,25 +165,53 @@ def compress(data: bytes) -> bytes:
...
+        declared = _snappy_uncompressed_length(data)
+        if declared > max_message_size:
+            raise ProtocolError(
+                f"Decompressed message size ({declared!r}) is larger than "
+                f"server max message size ({max_message_size!r})"
+            )
         import snappy
 
-        return snappy.uncompress(bytes(data))
+        result = snappy.uncompress(bytes(data))
```
blink1073: Small thing: this comment explains the `bytes(data)` conversion, but it now sits above
the varint check, so it reads as if it's describing that instead. Worth moving it down
next to the call it's about.

```suggestion
        declared = _snappy_uncompressed_length(data)
        if declared > max_message_size:
            raise ProtocolError(
                f"Decompressed message size ({declared!r}) is larger than "
                f"server max message size ({max_message_size!r})"
            )
        import snappy

        # python-snappy doesn't support the buffer interface.
        # https://github.com/andrix/python-snappy/issues/65
        # This only matters when data is a memoryview since
        # id(bytes(data)) == id(data) when data is a bytes.
        result = snappy.uncompress(bytes(data))
```

blink1073:
Thanks!

- I think the copilot suggestions make sense.  We should also add tests for the EOF cases for both compressors
- There is a merge conflict in the changelog - hopefully not to bad to address
- Can you please move the "from pymongo.errors import ProtocolError" to the top of the file rather than importing it in the tests?

blink1073:
The new copilot findings look relevant:

For the first two: the two checks need to agree. I think passing `max_message_size - 16` into `decompress()` at both call sites would address the problem.

For the third: it doesn't allow a bigger allocation on its own, since the actual output is still bounded by `decompress()`, but it's a header that should be rejected as malformed and currently isn't. It's worth adding an explicit `uncompressed_size <= 0` check alongside the size comparison.

blink1073 (approved):
Thanks @dfengliu!  I'll wait for one more team member to review since I pushed a few small fixes

blink1073:
@aclark4life can you please review the cleanup commits I made today?

aclark4life (approved):
LGTM

PR #2981, opened 2026-08-12
PYTHON-5909 / PYTHON-5979 Add GA support for Queryable Encryption string queries + Add QE prefix+suffix GA and rename API to string

aclark4life:
EVG: ~https://spruce.corp.mongodb.com/version/6a88c18021f67700074dfead~ https://spruce.corp.mongodb.com/version/6a8f57263d147f0007ebab7a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

---

[PYTHON-5909](https://jira.mongodb.org/browse/PYTHON-5909)
[PYTHON-5979](https://jira.mongodb.org/browse/PYTHON-5979)

## Changes in this PR

Adds general availability support for Queryable Encryption prefix, suffix, and substring string queries against MongoDB 9.0+, and renames the preview API to its GA form. Support landed in libmongocrypt per query type: prefix and suffix require 1.19.0+, substring requires 1.20.0+.

**Public API**

- `pymongo.encryption_options.StringOpts` replaces `TextOpts`. `TextOpts` is kept as a deprecated subclass of `StringOpts` and emits a `DeprecationWarning` on construction. It also stays re-exported from `pymongo.encryption`, so `from pymongo.encryption import TextOpts` continues to work for the deprecation period.
- `Algorithm.STRING` replaces `Algorithm.TEXTPREVIEW`, which is now deprecated.
- New `QueryType.PREFIX`, `QueryType.SUFFIX`, and `QueryType.SUBSTRING`, backing the `$encStrStartsWith`, `$encStrEndsWith`, and `$encStrContains` operators. The existing `QueryType.PREFIXPREVIEW` / `SUFFIXPREVIEW` / `SUBSTRINGPREVIEW` members remain for experimental use against servers older than 9.0.
- `ClientEncryption.encrypt()` and `AsyncClientEncryption.encrypt()` gain a `string_opts` parameter, deprecating `text_opts`. **`text_opts` is not a silent alias:** pymongocrypt renamed this parameter in 1.19 and accepts only one of the two names per release, so passing the name the installed pymongocrypt does not support raises `ConfigurationError` rather than failing deeper in the binding. Passing both names also raises `ConfigurationError`.

**Internal**

- `_string_opts_kwarg()` resolves the name pymongocrypt gives the string index options (`text_opts` through 1.18, `string_opts` from 1.19) by inspecting the installed `ExplicitEncrypter.encrypt` signature rather than comparing versions: the rename landed on master before any release carried it, so a version check would misclassify the master builds the GA query types require. It is `functools.lru_cache`d and does `import inspect` inside the function body, keeping `inspect` off the module import path.
- `_resolve_string_opts()` validates `string_opts` / `text_opts` against that resolved name so the sync and async `encrypt()` paths share one code path.
- `_STRING_QUERY_MIN_LIBMONGOCRYPT` in the test suite declares the minimum libmongocrypt version per query type in one place, so the test gates and the changelog cannot drift apart.

**CI**

- **libmongocrypt now comes from the signed GitHub release assets.** MONGOCRYPT-838 moved the per-variant release builds to a restricted bucket, which left `mciuploads/.../master/latest` frozen at 1.18.0 and skipped the entire prose suite on every variant. `setup_tests.py` now fetches `libmongocrypt-{target}-{version}.tar.gz` from the release page, pinned by `LIBMONGOCRYPT_VERSION = "1.20.2"` (1.20.0 is the floor for substring GA). Release assets are keyed by libc flavor rather than distro, so the old Debian/Ubuntu/RHEL target table collapses to arch plus glibc/musl, which also picks up ppc64le and s390x. A `get_libmongocrypt_base()` helper absorbs the layout difference: release archives put `lib/` at the archive root, while the master builds nested everything under `nocrypto/`. An explicit `LIBMONGOCRYPT_URL` still wins.
- **The `pymongocrypt<1.19` pin now reads `VERSION`, not `MONGODB_VERSION`.** Evergreen never sets `MONGODB_VERSION` — it is only assigned in-process by `run_server.py` — so the 8.x branch of that gate was dead code and those tasks kept installing the master binding.
- **The prose setup only builds the substring fixture where the query type exists.** It previously encrypted it unconditionally, so on libmongocrypt 1.19.x setup raised before the substring cases could skip, failing all 11 cases including the prefix and suffix ones.
- **New `test-string-query-preview` task on server 8.2.** The preview query types need a server at least 8.2 and older than 9.0, but `ALL_VERSIONS` jumps straight from 8.0 to 9.0, so the preview cases skipped on the server version gate no matter which libmongocrypt was installed. `create_string_query_preview_tasks()` emits one replica-set task pinned to 8.2, selected by the six encryption variants. `ALL_VERSIONS` is untouched, so the rest of the matrix does not grow.

## Test Plan

- **New prose tests** — `TestStringExplicitEncryptionProse` implements the spec's "String Explicit Encryption" tests, cases 01-11: find by prefix/suffix/substring, the corresponding no-match cases, `contentionFactor` being required, and the new case-insensitive and diacritic-insensitive prefix/suffix/substring cases.
- **Per-query-type version gating** — the class is gated on `require_version_min(8, 2, -1)` and `require_libmongocrypt_min(1, 18, 1)`, the floor for the preview query types, so both API generations can run. Setup then encrypts with `Algorithm.STRING` where it is available and the deprecated `Algorithm.TEXTPREVIEW` below 1.19.0. Two helpers gate the individual cases, and both look their requirements up in `_STRING_QUERY_MIN_LIBMONGOCRYPT` rather than hardcoding them at the call site: `_params(kind)` returns the `(query_type, collection)` pair for the cases that run against the GA type on 9.0+ and the preview type on earlier servers, and `_require_ga(*query_types)` skips the GA-only cases, naming each query type they exercise. A case is skipped below server 9.0, or when libmongocrypt is too old for any query type it uses.
- **New unit tests** — `TestStringOptsDeprecation` covers the compatibility surface directly: `TextOpts` warns and remains re-exported from `pymongo.encryption`; the resolve helper accepts whichever of `string_opts` / `text_opts` the installed pymongocrypt supports, raises `ConfigurationError` for the other one, and rejects both at once; and the resolved kwarg name is asserted against the installed binding's signature, so a mismatch fails without a server rather than only in the prose suite. These need no server or libmongocrypt, so the compatibility shims are covered on every CI task. Verified locally: 12 tests pass across both suites.

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer


pymongo/asynchronous/encryption.py
```diff
@@ -529,8 +530,15 @@ class Algorithm(str, enum.Enum):
...
     .. versionadded:: 4.4
     """
+    STRING = "String"
+    """String.
+
+    .. versionadded:: 4.18
+    """
     TEXTPREVIEW = "TextPreview"
-    """**BETA** - TextPreview.
+    """**DEPRECATED** - TextPreview.
```
blink1073: Isn't this a hard failure now?

blink1073:
Okay, we need two things to close this out:

- We need to target `pymongcrypt<1.19` when MONGODB_VERSION is 8.0, in `setup_tests.py`, and make sure the `SubstringPreview` tests are not skipped on MongoDB 8.0.
- We need to find out where libmongocrypt is putting the updated release files, because `f"https://s3.amazonaws.com/mciuploads/libmongocrypt/{target}/master/latest/libmongocrypt.tar.gz"` is pointing to 1.18.0-20260313+git9f4f0a1382.  I think https://github.com/mongodb/libmongocrypt/commit/6571317bb8b5d232064dd308d94d9310e9e062c6 is the relevant comment.  Then make sure the `substring` tests without preview run on MongoDB 9.0.

aclark4life:
> We need to find out where libmongocrypt is putting the updated release files,

[MONGOCRYPT-838](https://jira.mongodb.org/browse/MONGOCRYPT-838) switched release-branch builds to a **restricted** bucket, `cdn-origin-libmongocrypt`. That's the root cause: the per-variant release artifacts we were relying on are no longer public.

`master/latest` is still public on both hosts, but frozen:

| Host | Version at `macos/master/latest` |
| --- | --- |
| `s3.amazonaws.com/mciuploads` | `1.18.0-20260317+git9f4f0a1382` |
| `downloads.mongodb.org` | `1.18.0-20260317+git6571317bb8` |

So master builds *did* move to `downloads.mongodb.org` at that commit — but both are stuck in March either way, which is why the prose suite currently skips everywhere.

~Two options:~

~**1. Versioned "all" tarball** — `https://downloads.mongodb.org/libmongocrypt/all/{version}/libmongocrypt-all.tar.gz`~

~Verified resolving for 1.19.0, 1.19.1, 1.20.0 and 1.20.2, and it uses the *same* target names we already map to (`debian10`, `amazon2`, `macos`, …), so `setup_libmongocrypt()` would barely change. The problem is size: **1.76 GB** (`content-length: 1760303227`). Not viable to pull on every encryption task.~

**2. GitHub release assets** — e.g. `libmongocrypt-macos-universal-1.20.2.tar.gz`

Signed, and what the 1.18.0 changelog explicitly tells drivers to migrate to. Every target is nocrypto except `windows-x86_64`, which has no nocrypto variant (it links CNG, so there is no OpenSSL dependency to avoid). Small (macOS is 13 MB). Two differences from what `setup_tests.py` expects:

- Layout is `lib/libmongocrypt.dylib` at the archive root — there's no `nocrypto/` subdir, so `BASE = ROOT / "libmongocrypt/nocrypto"` needs to become conditional.
- Variant names are release names (`linux-x86_64-glibc_2_7-nocrypto`, `linux-arm64-glibc_2_17-nocrypto`, `linux-x86_64-musl_1_2-nocrypto`, `linux-arm64-musl_1_2-nocrypto`, `linux-ppc64le-glibc_2_17-nocrypto`, `linux-s390x-glibc_2_7-nocrypto`, `macos-universal`, `windows-x86_64`) rather than the distro names we map to today (`debian11/12/13`, `rhel-70-64-bit`, `rhel-80-64-bit`, `rhel-82-arm64`, `windows-test`).

blink1073:
@aclark4life, note, the new pymongocrypt only accepts `string_opts`, so we'll have to call it with the appropriate name based on the the pymongocrypt version.

.evergreen/scripts/setup_tests.py
```diff
@@ -353,29 +384,44 @@ def handle_test_env() -> None:
...
-            setup_libmongocrypt()
+        # The "String" algorithm and the GA prefix/suffix/substring query types
+        # need libmongocrypt 1.19.0+, which is only exercised against MongoDB
+        # 9.0+. Servers before 9.0 test the preview query types instead, which
+        # need the "textPreview" algorithm, so pin to the released pymongocrypt
+        # and use the libmongocrypt bundled in its wheel rather than the
+        # unreleased master build.
+        # Evergreen exposes the server version as VERSION, not MONGODB_VERSION
+        # (which is only set inside the separate run_server.py process).
+        use_released_pymongocrypt = os.environ.get("VERSION", "").startswith("8.")
```
blink1073: This name will be confusing once we release 1.20, perhaps `use_pymongocrypt_text_preview`?

pymongo/asynchronous/encryption.py
```diff
@@ -46,9 +48,20 @@
     from pymongocrypt.mongocrypt import MongoCryptOptions  # type:ignore[import]
 
     _HAVE_PYMONGOCRYPT = True
+    # pymongocrypt renamed the text_opts parameter to string_opts in 1.19. The
+    # preview query types still run against pymongocrypt < 1.19, so resolve the
+    # name from the installed signature rather than from a version comparison:
+    # the rename landed on master before any release carried it, so a version
+    # check would misclassify the master builds the GA query types require.
+    _STRING_OPTS_KWARG = (
```
blink1073: Using inspect is expensive, we should defer this calculation and not have it in the import path

pymongo/asynchronous/encryption.py
```diff
@@ -559,25 +582,77 @@ class QueryType(str, enum.Enum):
...
+def _resolve_string_opts(
+    string_opts: Optional[StringOpts], text_opts: Optional[StringOpts]
+) -> Optional[StringOpts]:
+    """Resolve the deprecated text_opts alias for string_opts."""
+    if text_opts is None:
+        return string_opts
+    if string_opts is not None:
+        raise ConfigurationError("Cannot set both string_opts and text_opts")
+    warnings.warn(
+        "The text_opts parameter is deprecated. Use string_opts instead.",
```
blink1073: I don't think this warning makes sense.  I think we should error if string_opts is supported by pymongocrypt but they pass text_opts, and vice versa.

pymongo/encryption_options.py
```diff
@@ -383,12 +383,32 @@ class PrefixOpts(TypedDict):
...
     # strMinQueryLength is the minimum allowed query length. Querying with a shorter string will error.
     strMinQueryLength: int
     # strMaxQueryLength is the maximum allowed query length. Querying with a longer string will error.
     strMaxQueryLength: int
+
+
+class TextOpts(StringOpts):
+    """**DEPRECATED** Options to configure encrypted queries using the text algorithm.
+
+    .. note:: ``TextOpts`` is deprecated. Use :class:`StringOpts` instead.
```
blink1073, line 400: I think we should be more explicit and say which versions apply here.  Like `TextOpts` was support in `pymongocrypt` x.i -> x.j, but removed support in x.k.

blink1073 (changes requested):

blink1073:
All of the scheduled tests are using latest server, can you please schedule some 8.2 variants and make sure the expected tests are run?

aclark4life:
> All of the scheduled tests are using latest server, can you please schedule some 8.2 variants and make sure the expected tests are run?

Like d6865ff770260b37ae76eaf72b78b8ae4d79e966 or something else?

blink1073:
I mean in Evergreen, run the 8.2 encryption tasks specifically

blink1073:
@aclark4life the linked EVG patch build is failing at startup

aclark4life:
> @aclark4life the linked EVG patch build is failing at startup

fixed

.evergreen/scripts/setup_tests.py
```diff
@@ -353,29 +384,44 @@ def handle_test_env() -> None:
         UV_ARGS.append("--extra zstd")
 
     if test_name in ["encryption", "kms"]:
-        # Check for libmongocrypt download.
-        if not (ROOT / "libmongocrypt").exists():
-            setup_libmongocrypt()
+        # The "String" algorithm and the GA prefix/suffix/substring query types
+        # need libmongocrypt 1.19.0+, which is only exercised against MongoDB
```
blink1073, line 389: I think it is too confusing to call out 1.19.0 separately from [1.20.0](https://github.com/mongodb/libmongocrypt/releases/tag/1.20.0).  I think we should just point to 1.20.0 everywhere.

pymongo/encryption_options.py
blink1073: This comment doesn't say when text_opts was added the pymongocrypt.   The changelog also has the same discrepency.

blink1073 (changes requested):

.evergreen/scripts/setup_tests.py
aclark4life: I propose a compromise in which we declare the floor as `1.20` in the changelog but leave a note about `1.19`  in `setup_tests`.

blink1073 (approved):
LGTM

PR #2985, opened 2026-08-13
PYTHON-4008 Support exhaust cursors on mongos 7.1+

carsontung666:
PYTHON-4008

## Changes in this PR

PyMongo refuses an exhaust cursor whenever the client is connected to a mongos:

```python
if self._collection.database.client.is_mongos:
    raise InvalidOperation("Exhaust cursors are not supported by mongos")
```

That check dates to 2.6, when exhaust was an OP_QUERY flag and mongos rejected it outright. PYTHON-1636 carried it across to OP_MSG in 4.0 unchanged. mongos has served exhaust getMore since 7.1, SERVER-57297.

- `MONGOS_EXHAUST_WIRE_VERSION = 22` in `pymongo/common.py`. libmongoc uses the same value.
- `_check_exhaust_supported(conn)` in `pymongo/message.py`, called from `_Query.use_command` and `_GetMore.use_command`. Both already gate exhaust on `conn.max_wire_version`, and both run before anything reaches the wire.
- `Cursor._supports_exhaust` and `Cursor.add_option` keep only their limit check.

The check reads the connection, not a freshly selected server. A `Sharded` topology can hold mongoses of different versions during a rolling upgrade.

Load-balanced deployments are left ungated. The connection reports `SERVER_TYPE.LoadBalancer` there, so the old check never fired either, and `_should_pin_cursor` pins every LB cursor anyway. An older mongos costs nothing extra, and refusing would break clients that work today.

Two timing changes, both in the changelog. Against a pre-7.1 mongos the `InvalidOperation` now comes on first iteration instead of from `find()`, because the check needs a connection. Async `limit` with `EXHAUST` now raises at `find()` instead of on first iteration, matching sync, because that check no longer needs I/O.

`find()` documents the limits. The saving is on the router-to-client leg only. The connection is monopolized for the cursor's lifetime. Nothing steers selection toward a 7.1+ mongos, so mid-upgrade the same call may succeed or raise depending on routing.

A pre-7.1 mongos does not hang a client that asks for exhaust. `exhaustAllowed` is an optional OP_MSG flag, so it is ignored and the cursor falls back to ordinary getMores. The refusal is because it would hold a connection for no benefit.

Motivation: on a workload streaming thousands of documents through mongos, 40-45% of wall time is the client blocked between getMore round trips. Exhaust makes those reads about 23% faster on a single shard, and about 19% on two shards, where the merge leg does not use exhaust.

## Test Plan

- Added unit tests in `test/test_message.py` for the wire version boundary and both `use_command` call sites. Mutation-tested: removing the version comparison, removing the `is_mongos` condition, or deleting either call site each fails the suite.
- The exhaust tests gated with `require_no_mongos` purely because of this restriction now run under `require_exhaust_cursors`. They pass against a two-shard cluster fronted by a 9.0.0-alpha0 mongos, where they previously skipped, and against a standalone of the same build.
- An exhaust cursor raises `InvalidOperation` on `main` and streams the full result set through the same mongos on this branch.
- `tools/synchro.py` reports no drift. ruff, format and mypy pass.

Not run: `test_encryption.py`, for lack of an encryption environment, and the Evergreen matrix. That includes the `load_balancer` suite, which these gate changes newly activate. The value 22 is not pinned by a test. Doing that needs a pre-7.1 mongos.

### Checklist for Author
- [x] Is there test coverage?
- [x] Is any followup work tracked in a JIRA ticket? If so, add link(s). DRIVERS-3231 tracks the cross-driver spec follow-up.

### Checklist for Reviewer

## AI assistance disclosure

I used Claude Opus 5 to help write and review this change. I reviewed and understand every changed line and can explain and maintain the contribution.

### DIFF
--- a/pymongo/asynchronous/cursor.py
+++ b/pymongo/asynchronous/cursor.py
@@ -213,19 +213,15 @@ class AsyncCursor(_AsyncCursorBase[_DocumentType]):
         self._dbname = collection.database.name
         self._collname = collection.name
 
-        # Checking exhaust cursor support requires network IO
-        if _IS_SYNC:
-            self._exhaust_checked = True
-            self._supports_exhaust()  # type: ignore[unused-coroutine]
-        else:
-            self._exhaust = cursor_type == CursorType.EXHAUST
-            self._exhaust_checked = False
+        self._validate_exhaust_handling()
 
-    async def _supports_exhaust(self) -> None:
-        # Exhaust cursor support
+    def _validate_exhaust_handling(self) -> None:
+        """Reject option combinations an exhaust cursor cannot serve.
+
+        Server support is checked against the connection in use, in
+        _check_exhaust_supported.
+        """
         if self._cursor_type == CursorType.EXHAUST:
-            if await self._collection.database.client.is_mongos:
-                raise InvalidOperation("Exhaust cursors are not supported by mongos")
             if self._limit:
                 raise InvalidOperation("Can't use limit and exhaust together.")
             self._exhaust = True
@@ -369,8 +365,6 @@ class AsyncCursor(_AsyncCursorBase[_DocumentType]):
         if mask & _QUERY_OPTIONS["exhaust"]:
             if self._limit:
                 raise InvalidOperation("Can't use limit and exhaust together.")
-            if await self._collection.database.client.is_mongos:
-                raise InvalidOperation("Exhaust cursors are not supported by mongos")
             self._exhaust = True
 
         self._query_flags |= mask
@@ -1125,9 +1119,6 @@ class AsyncCursor(_AsyncCursorBase[_DocumentType]):
 
     async def next(self) -> _DocumentType:
         """Advance the cursor."""
-        if not self._exhaust_checked:
-            self._exhaust_checked = True
-            await self._supports_exhaust()
         if self._empty:
             raise StopAsyncIteration
         if len(self._data) or await self._refresh():
@@ -1137,9 +1128,6 @@ class AsyncCursor(_AsyncCursorBase[_DocumentType]):
 
     async def _next_batch(self, result: list, total: Optional[int] = None) -> bool:  # type: ignore[type-arg]
         """Get all or some documents from the cursor."""
-        if not self._exhaust_checked:
-            self._exhaust_checked = True
-            await self._supports_exhaust()
         if self._empty:
             return False
         if len(self._data) or await self._refresh():
--- a/pymongo/message.py
+++ b/pymongo/message.py
@@ -43,6 +43,7 @@ from bson.raw_bson import (
     RawBSONDocument,
     _inflate_bson,
 )
+from pymongo.common import MONGOS_EXHAUST_WIRE_VERSION
 from pymongo.monitoring import _EventListeners
 
 try:
@@ -52,7 +53,6 @@ try:
 except ImportError:
     _use_c = False
 from pymongo.errors import (
-    ConfigurationError,
     DocumentTooLarge,
     InvalidOperation,
     ProtocolError,
@@ -1220,6 +1220,17 @@ _UNPACK_REPLY: dict[int, Callable[[bytes | memoryview], _OpMsg]] = {
 }
 
 
+def _check_exhaust_supported(conn: _AgnosticConnection) -> None:
+    """Raise if this connection's server will not continue an exhaust stream.
+
+    mongos gained exhaust getMore in 7.1 (SERVER-57297). Load-balanced deployments
+    are left alone: every cursor pins its connection there anyway, so an older mongos
+    costs nothing extra, and refusing would break clients that work today.
+    """
+    if conn.is_mongos and conn.max_wire_version < MONGOS_EXHAUST_WIRE_VERSION:
+        raise InvalidOperation("Exhaust cursors require MongoDB 7.1+ when connected to mongos.")
+
+
 class _Query:
     """A query operation."""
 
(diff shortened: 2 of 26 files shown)


blink1073:
Thanks @carsontung666!  I will review the PR tomorrow or Monday and get back to you 😄.

pymongo/synchronous/collection.py
```diff
@@ -1881,8 +1881,11 @@ def find(self, *args: Any, **kwargs: Any) -> Cursor[_DocumentType]:
 
           - The `limit` option can not be used with an exhaust cursor.
 
-          - Exhaust cursors are not supported by mongos and can not be
-            used with a sharded cluster.
+          - When connected to mongos, exhaust cursors require MongoDB 7.1 or
```
blink1073, line 1884: This isn't true behind a load balancer. `conn.is_mongos` is `False` there, so the check never fires and a pre-7.1 `mongos` gets through. It ignores the `exhaustAllowed` flag, doesn't set `moreToCome` on the reply, and the cursor goes back to ordinary getMores. Results are still correct, you just don't get the streaming. `InvalidOperation` is not raised, which is what this text tells the reader to expect. Could you add a clause for that case?

pymongo/message.py
```diff
@@ -1296,14 +1308,16 @@ def use_command(self, conn: _AgnosticConnection) -> bool:
...
-            # OP_MSG supports exhaust on MongoDB 4.2+
-            use_find_cmd = True
-        elif not self.read_concern.ok_for_legacy:
-            raise ConfigurationError(
-                f"read concern level of {self.read_concern.level} is not valid "
-                f"with a max wire version of {conn.max_wire_version}."
-            )
+        else:
+            _check_exhaust_supported(conn)
+            if conn.max_wire_version >= 8:
```
blink1073: Optional, but since you're already making changes here: MIN_SUPPORTED_WIRE_VERSION is 8, so this is always true and the ConfigurationError underneath can't be reached.

pymongo/message.py
```diff
@@ -1429,9 +1443,11 @@ def use_command(self, conn: _AgnosticConnection) -> bool:
         use_cmd = False
         if not self.exhaust:
             use_cmd = True
-        elif conn.max_wire_version >= 8:
-            # OP_MSG supports exhaust on MongoDB 4.2+
-            use_cmd = True
+        else:
+            _check_exhaust_supported(conn)
+            if conn.max_wire_version >= 8:
```
blink1073: Same here, we can just set `use_cmd` directly

test/test_message.py
```diff
@@ -44,11 +48,91 @@
...
+    def test_exhaust_allowed_on_mongos_71(self):
+        _check_exhaust_supported(self._make_conn(MONGOS_EXHAUST_WIRE_VERSION, is_mongos=True))
+
+    def test_exhaust_refused_on_older_mongos(self):
+        with self.assertRaisesRegex(InvalidOperation, "MongoDB 7.1"):
+            _check_exhaust_supported(
+                self._make_conn(MONGOS_EXHAUST_WIRE_VERSION - 1, is_mongos=True)
+            )
+
+    def test_exhaust_not_gated_behind_a_load_balancer(self):
```
blink1073, line 71: `load_balanced=True` does nothing here: `is_mongos` stays `False` and the exhaust support check never reads `conn.opts.load_balanced`. That makes this a duplicate of `test_exhaust_allowed_on_any_supported_non_mongos` that'd still pass if the check started rejecting load-balanced connections. The real guarantee is `Connection.is_mongo`s being `False` for `SERVER_TYPE.LoadBalancer`, in `pool.py`.

pymongo/asynchronous/cursor.py
```diff
@@ -213,19 +213,11 @@ def __init__(
...
-        # Checking exhaust cursor support requires network IO
-        if _IS_SYNC:
-            self._exhaust_checked = True
-            self._supports_exhaust()  # type: ignore[unused-coroutine]
-        else:
-            self._exhaust = cursor_type == CursorType.EXHAUST
-            self._exhaust_checked = False
+        self._supports_exhaust()
 
-    async def _supports_exhaust(self) -> None:
```
blink1073, line 224: I think this should be called, `_validate_exhaust_handling` or something like that, and make the comment into a docstring that does not mention the function name

pymongo/synchronous/collection.py
carsontung666: I will add a note about the load balancer case to the docstring.

test/test_message.py
carsontung666: Right, that attribute did nothing. Rewrote the test to assert the real mechanism: a load-balanced hello is typed LoadBalancer, not Mongos.

blink1073:
We're getting close!  One thing I released - `test_cursors.py` is not actually run against our load balancer tests.  We use the `load_balancer` pytest marker, which only runs `test_load_balancer.py`.

- `supports_exhaust_cursors()` needs `if self.load_balancer: return True`. Right now it reports False for any LB deployment, because `client_context.is_mongos` is `True` under LB, so 7.0 skips even though `_check_exhaust_supported` would allow it through.
- Add an `exhaust` test to `test_load_balancer.py`.

carsontung666:
Done. I don't have a load balancer locally so this path never ran on my side, and I didn't realize the LB suite only collects test_load_balancer.py. Fixed both: added the early return and an exhaust test in test_load_balancer.py.

blink1073 (approved):
Thank you!   I verified that the load balancer job ran the new `test_exhaust_cursor` test.   Once all of the Evergreen jobs finish I will merge.

PR #2990, opened 2026-08-14
PYTHON-5956 Use shared python/setup action for CI setup

blink1073:
[PYTHON-5956](https://jira.mongodb.org/browse/PYTHON-5956)

Adopts the shared `python/setup` action from mongodb-labs/drivers-github-tools#117, replacing the uv and just install steps that eight jobs in `test-python.yml` each carried.

## Changes in this PR

- Eight jobs get Python, uv, and just from one step instead of two or three.
- Python comes from `setup-python`, with uv pointed at that interpreter.
- The five jobs that need project dependencies keep their own `just install` step.

## Test Plan

CI-only, so the workflow running green is the test: [14 jobs green](https://github.com/blink1073/mongo-python-driver/actions/runs/33670439229).

Interpreters were read from the job logs per matrix entry rather than inferred from a green tick: CPython 3.10.20, free-threaded 3.13.15t, 3.15.0-rc.1, PyPy 7.3.23, and 3.9.25. `static`'s `uv lock --check` passes, and no job sets `UV_EXCLUDE_NEWER`.

### Checklist for Author
- [ ] Did you update the changelog (if necessary)? N/A
- [x] Is there test coverage? Standard GitHub Actions runs
- [ ] Is any followup work tracked in a JIRA ticket? If so, add link(s). N/A

### Checklist for Reviewer


.github/workflows/test-python.yml
```diff
@@ -290,8 +270,8 @@ jobs:
         - uses: actions/checkout@v7.0.1
           with:
             persist-credentials: false
-        - name: Install uv
-          uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+        - name: Install Python tooling
+          uses: mongodb-labs/drivers-github-tools/python/setup@v3
```
aclark4life: This job manages its own `venv` but the action sets `UV_PYTHON` to an absolute interpreter path which takes precedence over the activated `VIRTUAL_ENV` so `uv pip install` now targets the `setup-python` interpreter instead of the `.venv`.

aclark4life (changes requested):

.github/workflows/test-python.yml
blink1073: Good catch!  I'll fix this upstream in github-tools, PR coming soon

blink1073: I think we're good now.  From the logs:

| | Before (`UV_PYTHON` = interpreter path) | After (`UV_PYTHON` = version) |
|---|---|---|
| Exported by the action | `UV_PYTHON: /opt/hostedtoolcache/Python/3.9.25/x64/bin/python` | `UV_PYTHON: 3.9`, `UV_PYTHON_PREFERENCE: only-system` |
| `uv pip install` target | `Using Python 3.9.25 environment at: /opt/hostedtoolcache/Python/3.9.25/x64` | no override line, so the activated `.venv` is used |
| pytest interpreter | `/opt/hostedtoolcache/Python/3.9.25/x64/bin/python` | `/home/runner/work/mongo-python-driver/mongo-python-driver/.venv/bin/python3` |

.github/workflows/test-python.yml
```diff
@@ -82,10 +79,10 @@ jobs:
       - uses: actions/checkout@v7.0.1
         with:
           persist-credentials: false
-      - name: Install uv
-        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+      - name: Install Python tooling
+        uses: mongodb-labs/drivers-github-tools/python/setup@d518d2c7d04fdec10266c4218c36791a4fcf98d8 # v3.0.2
         with:
-          enable-cache: true
+          # The beta Python here relies on the action's prerelease default.
```
aclark4life, line 85: Should we set `allow-prereleases: true` ?

blink1073: That's the default in the action

aclark4life (approved):

PR #2991, opened 2026-08-17
PYTHON-5947 Add OpenTelemetry operation spans

blink1073:
PYTHON-5947

First of five PRs splitting #2964. Base is `otel`; review in order.

| | Branch | Contents |
|---|---|---|
| **1** | **`PYTHON-5947-otel-1-operations`** | **operation spans (this PR)** |
| 2 | `PYTHON-5947-otel-2-transactions` | transaction spans |
| 3 | `PYTHON-5947-otel-3-unified` | unified runner and vendored fixtures |
| 4 | `PYTHON-5947-otel-4-getmore` | getMore spans |
| 5 | `PYTHON-5947-otel-5-error-type` | `error.type` command span attribute |

## Changes in this PR

Adds one span per public API call, containing one command span per command sent to the server, per the OpenTelemetry driver specification. Without it a retried operation's command spans share no parent, so a reader cannot tell which attempts belong to the same call.

- One operation span covers every retry attempt of a call, so retries show up as sibling command spans instead of collapsing into one.
- `killCursors`, `endSessions`, and unacknowledged client bulk writes get operation spans too.
- An operation that fails before anything reaches the wire still produces a conformant span, so a server selection timeout is reported rather than lost.
- Cursor-creating calls (`find`, `aggregate`, `listCollections`, `listIndexes`) are covered, but only for the command that creates the cursor. The `getMore` spec has pending changes, so getMore spans are deferred to PR 4.
- Fixed truncated output dropping falsy values (`0`, `False`, `""`). This affected structured command logging as well as spans.
- Background tasks no longer inherit a caller's span.
- The otel Evergreen variant now covers all three topologies, but remains at 22 tasks.

Opt-in: no behavior change unless the `tracing` client option or `OTEL_PYTHON_INSTRUMENTATION_MONGODB_ENABLED` is set. With tracing off there is no added allocation per command.

Around 1,000 of the added lines are `just synchro` output and another ~745 are tests.

## Test Plan

- 49 new synchronized tests for span attributes and naming, retry and failure paths, sensitive-command redaction, and config handling, plus a regression test that enabling tracing through the environment variable does not trace background heartbeats.
- Regression tests for the truncation bug and for collection names containing dots.
- Locally against a replica set: 98 otel tests pass, along with the cursor, collection, common, json_util, and operation-id suites. `just lint` clean.

### Checklist for Author
- [x] Is there test coverage?
- [x] Is any followup work tracked in a JIRA ticket? If so, add link(s).

### Checklist for Reviewer

### DIFF
--- a/.evergreen/scripts/generate_config.py
+++ b/.evergreen/scripts/generate_config.py
@@ -458,7 +458,23 @@ def create_otel_variants():
     expansions = dict(TEST_NAME="otel", COVERAGE="1")
     return [
         create_variant(
-            [".test-non-standard .standalone-noauth-nossl"],
+            [
+                # All three topologies, subset to keep the task count at 22.
+                #
+                # Replica set in full: the only topology where transaction spans
+                # run at all (they are skipped on standalone and sharded), and
+                # the only one covering free-threaded Python.
+                ".test-non-standard .replica_set-noauth-ssl",
+                # Sharded adds mongos, which rewrites commands and reports a
+                # different server.address, plus auth and ssl, which exercise
+                # sensitive-command redaction. Newest CPython across server
+                # versions, and PyPy for the alternate implementation.
+                ".test-non-standard .sharded_cluster-auth-ssl .python-3.14",
+                ".test-non-standard .sharded_cluster-auth-ssl .python-pypy3.11",
+                # Standalone only for its min-deps tasks, which resolve
+                # opentelemetry-api down to the floor in requirements/.
+                ".test-non-standard .standalone-noauth-nossl .python-3.10",
+            ],
             get_variant_name("OTel", host),
             host=host,
             tags=["pr"],
--- a/.evergreen/scripts/generate_config.py
+++ b/.evergreen/scripts/generate_config.py
@@ -461,7 +461,7 @@ def create_otel_variants():
             [
                 # All three topologies, subset to keep the task count at 22.
                 #
-                # Replica set in full: the only topology where transaction spans
+                # Replica set keeps every task: the only topology where transaction spans
                 # run at all (they are skipped on standalone and sharded), and
                 # the only one covering free-threaded Python.
                 ".test-non-standard .replica_set-noauth-ssl",
--- a/pymongo/_otel.py
+++ b/pymongo/_otel.py
@@ -433,7 +433,8 @@ def start_operation_span(
     backfills the authoritative values once a command exists.
 
     ``parent_span`` becomes an *explicit* parent rather than being read from
-    ambient context, so a concurrent unrelated session cannot be captured.
+    whichever span is current, so a concurrent unrelated session cannot be
+    captured.
 
     ``set_current=False`` leaves the span and the operation-name contextvar
     alone, for a caller that makes it current with ``use_operation_span``.
@@ -496,7 +497,7 @@ def use_operation_span(handle: Optional[_OperationSpanHandle]) -> Iterator[None]
 
 
 def reset_context() -> None:
-    """Clear the OTel ambient span and operation-name contextvar.
+    """Clear the current OTel span and the operation-name contextvar.
 
     ``asyncio.create_task`` freezes the caller's context, so without this a
     long-lived background task parents every span it emits under an unrelated,
--- a/pymongo/_otel.py
+++ b/pymongo/_otel.py
@@ -227,7 +227,10 @@ def _build_query_summary(command_name: str, dbname: str, collection: Optional[st
 # LUT of db.operation.name for operations the spec names differently from our `_Op`
 # values. The nested command span keeps the wire name in db.command.name, so
 # dropping a collection reports "dropCollection" over a "drop" command span.
-# PYTHON-6054 tracks documenting these discrepancies with the spec.
+#
+# PYTHON-6054 tracks the operations the spec's table has yet to name: rename
+# sends renameCollection, countDocuments sends an aggregate but reports "count",
+# and a collection bulkWrite reports "insert" whatever the batch contains.
 _OPERATION_NAME_OVERRIDES = {
     "drop": "dropCollection",
     "create": "createCollection",
--- a/pymongo/_otel.py
+++ b/pymongo/_otel.py
@@ -349,12 +349,14 @@ def end_command_span_success(span: Optional[Span], reply: _DocumentOut) -> None:
     """Set the cursor id (if any open cursor) and end the span."""
     if span is None:
         return
-    cursor = reply.get("cursor")
-    if isinstance(cursor, Mapping) and cursor.get("id"):
-        # Per the spec the attribute is omitted rather than set to 0, so a
-        # cursor-creating command that leaves no cursor open reports nothing.
-        span.set_attribute("db.mongodb.cursor_id", cursor["id"])
-    span.end()
+    try:
+        cursor = reply.get("cursor")
+        if isinstance(cursor, Mapping) and cursor.get("id"):
+            # Per the spec the attribute is omitted rather than set to 0, so a
+            # cursor-creating command that leaves no cursor open reports nothing.
+            span.set_attribute("db.mongodb.cursor_id", cursor["id"])
+    finally:
+        span.end()
 
 
 def _set_exception_attributes(span: Span, exc: BaseException) -> None:
@@ -383,13 +385,17 @@ def end_command_span_failure(
     """Record the exception, set the error status, and end the span."""
     if span is None:
         return
-    span.record_exception(exc)
-    _set_exception_attributes(span, exc)
-    code = failure.get("code")
-    if code is not None:
-        span.set_attribute("db.response.status_code", str(code))
-    span.set_status(Status(StatusCode.ERROR, description=failure.get("errmsg")))
-    span.end()
+    try:
+        span.record_exception(exc)
+        _set_exception_attributes(span, exc)
+        code = failure.get("code")
+        if code is not None:
+            span.set_attribute("db.response.status_code", str(code))
+        span.set_status(Status(StatusCode.ERROR, description=failure.get("errmsg")))
+    finally:
+        # End even if recording raised, so a failure here costs the attributes
+        # rather than leaking an unended span.
+        span.end()
 
 
 class _OperationSpanHandle:
@@ -524,11 +530,15 @@ def end_operation_span_failure(handle: Optional[_OperationSpanHandle], exc: Base
     """Record the exception, set the error status, and end the operation span."""
     if handle is None:
         return
-    handle.span.record_exception(exc)
-    _set_exception_attributes(handle.span, exc)
-    handle.span.set_status(Status(StatusCode.ERROR, description=str(exc)))
-    if handle._cm is None:
-        handle.span.end()
-        return
-    _CURRENT_OPERATION_NAME.reset(handle._name_token)
-    handle._cm.__exit__(None, None, None)
+    try:
+        handle.span.record_exception(exc)
+        _set_exception_attributes(handle.span, exc)
+        handle.span.set_status(Status(StatusCode.ERROR, description=str(exc)))
+    finally:
+        # Unwind even if recording raised, since a span left current would
+        # re-parent every later span in this task.
+        if handle._cm is None:
+            handle.span.end()
+        else:
+            _CURRENT_OPERATION_NAME.reset(handle._name_token)
+            handle._cm.__exit__(None, None, None)
--- a/pymongo/_otel.py
+++ b/pymongo/_otel.py
@@ -132,15 +132,15 @@ def _is_tracing_enabled(tracing_options: Optional[TracingOptions]) -> bool:
 
     ClientOptions folds ``OTEL_PYTHON_INSTRUMENTATION_MONGODB_ENABLED`` into
     ``tracing.enabled`` once at construction, so this is a lookup rather than
-    an os.environ read per command. None means there is no client to read the
-    option from, as for monitor and handshake connections, which are never
+    an os.environ read per command. ``None`` means there is no client to read the
+    option from, such as for monitor and handshake connections, which are never
     traced.
     """
     return _HAS_OPENTELEMETRY and tracing_options is not None and tracing_options["enabled"]
 
 
 def _resolve_tracing_options(tracing_options: _UnresolvedTracingOptions) -> TracingOptions:
-    """Fold both environment variables into a client's validated tracing options.
+    """Fold Otel environment variables into a client's validated tracing options.
 
     Called once when the client is built, so nothing re-reads the environment
     per command. An explicit client value wins, including a
@@ -186,7 +186,7 @@ def _extract_collection_name(
 ) -> Optional[str]:
     """Return the collection name targeted by ``cmd``, or None if it doesn't target one.
 
-    Always None for commands against the admin database: several (e.g. dropUser,
+    Always ``None`` for commands against the admin database: several (e.g. dropUser,
     renameCollection) carry a string command value that names a user, role, or
     namespace rather than a collection.
     """
@@ -224,7 +224,7 @@ def _build_query_summary(command_name: str, dbname: str, collection: Optional[st
     return f"{command_name} {dbname}"
 
 
-# db.operation.name for operations the spec names differently from our `_Op`
+# LUT of db.operation.name for operations the spec names differently from our `_Op`
 # values. The nested command span keeps the wire name in db.command.name, so
 # dropping a collection reports "dropCollection" over a "drop" command span.
 _OPERATION_NAME_OVERRIDES = {
@@ -285,11 +285,11 @@ def start_command_span(
     command_name: str,
     speculative_hello: bool,
 ) -> Optional[Span]:
-    """Start and return a CLIENT-kind span for a server command, or None.
+    """Start and return a CLIENT-kind span for a server command, or ``None``.
 
-    A no-op returning None when tracing is off; a sensitive command also
-    returns None but still backfills the current operation span first. One span
-    per wire-protocol message, parented to the current operation span but never
+    A no-op returning ``None`` when tracing is off; a sensitive command also
+    returns ``None`` but still backfills the current operation span first. There is one
+    span per wire-protocol message, parented to the current operation span but never
     made current itself.
     """
     if not _is_tracing_enabled(tracing_options):
@@ -395,9 +395,9 @@ def end_command_span_failure(
 
 
 class _OperationSpanHandle:
-    """Bundles an operation span with what's needed to end it later.
+    """Bundles an operation span with what is needed to end it later.
 
-    ``_cm`` is the ``start_as_current_span`` context manager, or None in
+    ``_cm`` is the ``start_as_current_span`` context manager, or ``None`` in
     detached mode, where ``use_operation_span`` makes the span current per use.
     """
 
@@ -424,9 +424,9 @@ def start_operation_span(
     collection: Optional[str] = None,
     set_current: bool = True,
 ) -> Optional[_OperationSpanHandle]:
-    """Start a CLIENT-kind span for one logical operation, or None.
+    """Start a CLIENT-kind span for one logical operation, or ``None``.
 
-    Spans all retry attempts of one ``_retry_internal`` call. Namespace
+    Carries across all retry attempts of one ``_retry_internal`` call. Namespace
     attributes are set eagerly from ``dbname``/``collection`` so an operation
     that fails before building any command, such as a server selection
     timeout, still produces a conformant span; ``start_command_span``
@@ -475,7 +475,7 @@ def use_operation_span(handle: Optional[_OperationSpanHandle]) -> Iterator[None]
     """Make a detached operation span current for the duration of the block.
 
     Does not end the span; its owner ends it explicitly. A no-op when
-    ``handle`` is None.
+    ``handle`` is ``None``.
     """
     if handle is None:
         yield
@@ -502,7 +502,7 @@ def reset_context() -> None:
     ``asyncio.create_task`` freezes the caller's context, so without this a
     long-lived background task parents every span it emits under an unrelated,
     long-ended operation. Attaching an empty context makes spans started
-    afterwards trace roots. Deliberately never detached: the task's context is
+    afterwards into trace roots. Deliberately never detached: the task's context is
     wrong for its whole life and dies with it.
     """
     if not _HAS_OPENTELEMETRY:
--- a/pymongo/_otel.py
+++ b/pymongo/_otel.py
@@ -175,7 +175,9 @@ def _build_query_text(cmd: Mapping[str, Any], max_length: int) -> str:
     text = json_util.dumps(truncated_cmd, json_options=_JSON_OPTIONS, default=repr)
     if len(text) > max_length:
         suffix = "..."
-        text = text[: max(0, max_length - len(suffix))] + suffix
+        if max_length < len(suffix):
+            return text[:max_length]
+        text = text[: max_length - len(suffix)] + suffix
     return text
 
 
--- a/pymongo/_otel.py
+++ b/pymongo/_otel.py
@@ -224,8 +224,9 @@ def _build_query_summary(command_name: str, dbname: str, collection: Optional[st
     return f"{command_name} {dbname}"
 
 
-# Spec operation names that differ from our `_Op` values. "dropCollection" and
-# "createCollection" contradict the spec's own command-name rule, unresolved.
+# db.operation.name for operations the spec names differently from our `_Op`
+# values. The nested command span keeps the wire name in db.command.name, so
+# dropping a collection reports "dropCollection" over a "drop" command span.
 _OPERATION_NAME_OVERRIDES = {
     "drop": "dropCollection",
     "create": "createCollection",
@@ -286,18 +287,19 @@ def start_command_span(
 ) -> Optional[Span]:
     """Start and return a CLIENT-kind span for a server command, or None.
 
-    Returns None when tracing is off, and for a sensitive command, which still
-    backfills the current operation span before being suppressed. One span per
-    wire-protocol message, returned rather than made current, so nothing nests
-    inside it.
+    A no-op returning None when tracing is off; a sensitive command also
+    returns None but still backfills the current operation span first. One span
+    per wire-protocol message, parented to the current operation span but never
+    made current itself.
     """
     if not _is_tracing_enabled(tracing_options):
         return None
 
     collection = _extract_collection_name(command_name, dbname, cmd)
-    # Runs per attempt, not per operation: an attempt that fails before building
-    # a command never reaches here, so a retry may be where the span learns its
-    # namespace.
+    # Must stay above the sensitive-command return: the operation span needs
+    # these attributes even when the command itself gets no span. Runs per
+    # attempt, since an attempt that dies before building a command never
+    # reaches here, so a retry may be where the span learns its namespace.
     current_operation = _CURRENT_OPERATION_NAME.get()
     if current_operation is not None:
         current_span = trace.get_current_span()
--- a/pymongo/_otel.py
+++ b/pymongo/_otel.py
@@ -297,8 +297,8 @@ def start_command_span(
     # Must stay above the sensitive-command return: the operation span needs
     # these attributes even when the command itself gets no span. Runs per
     # attempt, since one that dies before building a command never reaches here
-    # and a retry may be where the span first learns its namespace. Repeats
-    # write the same values.
+    # and a retry may be where the span first learns its namespace. A later
+    # attempt sets the same attributes to the same values.
     current_operation = _CURRENT_OPERATION_NAME.get()
     if current_operation is not None:
         current_span = trace.get_current_span()
--- a/pymongo/_otel.py
+++ b/pymongo/_otel.py
@@ -18,11 +18,9 @@ Kept separate from :mod:`pymongo._telemetry` so that module stays free of
 ``opentelemetry`` import guards. Every function here is a no-op when
 ``opentelemetry`` isn't installed or tracing isn't enabled.
 
-This module also owns the specification's naming and attribute rules, such as
-how span names, ``db.operation.name`` and ``db.query.summary`` are built.
-:mod:`pymongo._telemetry` owns span lifecycles. A specification change to a
-name or an attribute value stays here; one that changes when a span starts or
-ends, or what it nests under, affects both.
+This module builds span names and attributes; :mod:`pymongo._telemetry` starts
+and ends spans. So a specification change to a name or an attribute value lands
+here alone, while one to a span's lifetime or nesting touches both.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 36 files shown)


pymongo/asynchronous/cursor.py
```diff
@@ -1099,6 +1115,21 @@ async def _refresh(self) -> int:
 
         return len(self._data)
 
+    async def _send_message_in_operation_span(self, operation: Union[_Query, _GetMore]) -> None:
+        """Send ``operation``, ending the operation span once it completes.
+
+        ``_send_message``'s own error handling already ends the span with the
```
NoahStapp: This docstring is a bit verbose, I don't think we need such detail here.

pymongo/asynchronous/mongo_client.py
```diff
@@ -1988,6 +2016,10 @@ async def _retry_internal(
         :param is_run_command: If this is a runCommand operation, defaults to False
         :param is_aggregate_write: If this is a aggregate operation with a write, defaults to False.
         :param operation_id: Stable operation id shared across retries, defaults to None
+        :param operation_telemetry: A caller-owned operation span outliving this call
```
NoahStapp: This docstring is also too verbose.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2035,6 +2069,8 @@ async def _retryable_read(
         :param is_run_command: If this is a runCommand operation, defaults to False.
         :param is_aggregate_write: If this is a aggregate operation with a write, defaults to False.
         :param operation_id: Stable operation id shared across retries, defaults to None
+        :param operation_telemetry: A caller-owned operation span outliving this call,
```
NoahStapp: How can the operation span outlive this call and this method own a fresh span? Not clear what that means.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2055,7 +2091,61 @@ async def _retryable_read(
...
+        address: Optional[_Address] = None,
+        retryable: bool = True,
+        operation_id: Optional[int] = None,
+        is_run_command: bool = False,
+        is_aggregate_write: bool = False,
+        *,
+        dbname: str,
+        collection: Optional[str] = None,
+    ) -> _CommandCursor:
+        """Run a command-cursor read within its own operation span.
```
NoahStapp:
```suggestion
        """Run a command cursor read within its own operation span.
```
No hyphen between command and cursor.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2055,7 +2091,61 @@ async def _retryable_read(
...
+
+        Takes the same arguments as :meth:`_retryable_read`, plus the namespace
```
NoahStapp: I'm not sure we need to explain the first batch fetching and how it relates to the span here. Or maybe it just needs to be less verbose?

pymongo/asynchronous/mongo_client.py
```diff
@@ -2910,12 +3009,36 @@ def __init__(
         if operation_id is None:
             operation_id = _generate_op_id_or_none(self._client._event_listeners)
         self._operation_id = operation_id
+        # One span covering every attempt. A caller needing it to outlive this
+        # object (a cursor) passes its own and keeps ownership.
```
NoahStapp: `self._owns_telemetry = True` means that we create the span here, otherwise it's passed in because a cursor owns it? Comment doesn't make that clear.

pymongo/_otel.py
```diff
@@ -12,17 +12,27 @@
...
+"""Optional OpenTelemetry span support.
 
 Kept separate from :mod:`pymongo._telemetry` so that module stays free of
 ``opentelemetry`` import guards. Every function here is a no-op when
 ``opentelemetry`` isn't installed or tracing isn't enabled.
+
+This module also owns the specification's naming and attribute rules, such as
+how span names, ``db.operation.name`` and ``db.query.summary`` are built.
+:mod:`pymongo._telemetry` owns span lifecycles. A specification change to a
+name or an attribute value stays here; one that changes when a span starts or
```
NoahStapp: This could be more concise and just as clear.

pymongo/_otel.py
```diff
@@ -31,20 +41,30 @@
...
+    # calls set_tracer_provider(), so this is not bound to a no-op tracer.
     _TRACER: Optional[Tracer] = trace.get_tracer("PyMongo", __version__)
 except ImportError:
     _HAS_OPENTELEMETRY = False
     _TRACER = None
 
+# Name of the active operation span, so start_command_span can backfill that
+# span's name and namespace attributes from the first command built inside it.
+# Call sites that know the namespace up front pass it to start_operation_span;
+# the generic retry path does not, and relies on this backfill.
```
NoahStapp: Could be more concise.

pymongo/_otel.py
```diff
@@ -55,9 +75,9 @@
 class TracingOptions(TypedDict):
     """The shape of the ``MongoClient`` ``tracing`` option.
 
-    ``query_text_max_length`` is None when the client didn't configure it, so
-    the environment variable can be consulted; any explicit value (including
-    0, to force ``db.query.text`` off) overrides the environment variable.
+    ``query_text_max_length`` is None as validated from user input; the options
+    a client holds have been through :func:`_resolve_tracing_options`, so both
+    fields are resolved.
```
NoahStapp: Does this mean that `query_text_max_length` is always None? Not sure what this comment means.

pymongo/_otel.py
```diff
@@ -72,65 +92,72 @@ class TracingOptions(TypedDict):
...
+    """Return True if spans should be created for this client.
 
-    The ``MongoClient`` ``tracing.enabled`` option and the
-    ``OTEL_PYTHON_INSTRUMENTATION_MONGODB_ENABLED`` environment variable both
-    gate enablement; either one being truthy is sufficient.
+    ClientOptions folds ``OTEL_PYTHON_INSTRUMENTATION_MONGODB_ENABLED`` into
+    ``tracing.enabled`` once at construction, so this is a lookup rather than
+    an os.environ read per command. None means there is no client to read the
+    option from, as for monitor and handshake connections, which are never
+    traced.
```
NoahStapp: Could be more concise.

pymongo/_otel.py
```diff
@@ -72,65 +92,72 @@ class TracingOptions(TypedDict):
...
-    Mirrors the truncation approach used for log messages: truncate field
-    values first, which usually keeps the result well-formed JSON (unlike a
-    blind cut of the fully-serialized string), then fall back to a hard
-    string cut as a safety net for whatever the field truncation's size
-    estimate still leaves over ``max_length``. The "..." marker is carved out
-    of the budget (not appended on top of it) so the result never exceeds
-    ``max_length``.
+    Mirrors log-message truncation: shorten field values first, which usually
+    keeps the result valid JSON, then hard-cut as a safety net. The "..."
+    marker comes out of the budget, so the result never exceeds ``max_length``.
```
NoahStapp: Do we need to explain this in full if we already do for log messages?

pymongo/_otel.py
```diff
@@ -172,6 +213,42 @@ def _build_query_summary(command_name: str, dbname: str, collection: Optional[st
     return f"{command_name} {dbname}"
 
 
+# Some `_Op` values are the wire command name ("drop"/"create") rather than the
```
NoahStapp: Why not rename `_Op` for consistency if it's entirely internal?

pymongo/_otel.py
```diff
@@ -200,15 +277,33 @@ def start_command_span(
 ) -> Optional[Span]:
     """Start and return a CLIENT-kind span for a server command, or None.
 
-    Returns None when tracing is disabled/unavailable or the command is
-    sensitive (mirroring the redaction applied to logs).
+    None when tracing is off or the command is sensitive, mirroring the
```
NoahStapp: Should say "returns None" or "No-op" instead of "None", right?

pymongo/_otel.py
```diff
@@ -200,15 +277,33 @@ def start_command_span(
...
+    redaction applied to logs. One span per wire-protocol message, so a retried
+    operation produces one per attempt. Returned rather than made current: it is
+    a leaf under whichever operation span is current, and the caller holds it
```
NoahStapp: Is "leaf" a standard terminology here? Not sure what it means in this context.

pymongo/_otel.py
```diff
@@ -261,8 +376,150 @@ def end_command_span_failure(
...
+    collection: Optional[str] = None,
+    set_current: bool = True,
+) -> Optional[_OperationSpanHandle]:
+    """Start a CLIENT-kind span for one logical operation, or None.
+
+    Spans all retry attempts of one ``_retry_internal`` call. Namespace
+    attributes are set eagerly from ``dbname``/``collection`` so an operation
+    that fails before building any command, such as a server selection
+    timeout, still produces a conformant span; ``start_command_span``
+    backfills the authoritative values once a command exists.
```
NoahStapp: Reads a little too verbose.

pymongo/cursor_shared.py
```diff
@@ -115,6 +116,24 @@ def _prepare_to_die(self, already_killed: bool) -> tuple[int, Optional[_CursorAd
             address = None
         return cursor_id, address
 
+    def _end_operation_telemetry(self, exc: Optional[BaseException] = None) -> None:
+        """End the operation span currently attached to this cursor, exactly once.
+
+        No span is scoped to the cursor's lifetime: the span of the operation
+        that created the cursor ends as soon as that creating command
+        completes, whether it succeeded, failed, or the cursor was abandoned
+        part-way and is being closed by close()/__del__. Idempotent, so every
```
NoahStapp: This could be shortened to something like "completes, regardless of status."

test/asynchronous/test_otel.py
```diff
@@ -48,29 +61,261 @@
...
+        span = self._finished_span("listCollections", dbname="mydb")
+        self.assertEqual(span.name, "listCollections mydb")
+        self.assertEqual(span.attributes["db.operation.summary"], "listCollections mydb")
+        self.assertNotIn("db.collection.name", span.attributes)
+
+    def test_no_eager_attributes_leaves_provisional_name(self):
+        span = self._finished_span()
+        self.assertEqual(span.name, "find")
+        self.assertNotIn("db.namespace", span.attributes)
+        # db.operation.summary is Required (unlike db.namespace, which is only
```
NoahStapp: Intentional uppercase "Required" here and directly below?

pymongo/asynchronous/client_bulk.py
```diff
@@ -630,13 +630,28 @@ async def execute(
...
 
         if not self.write_concern.acknowledged:
-            async with await self.client._conn_for_writes(session, operation) as connection:
-                if connection.max_wire_version < 25:
-                    raise InvalidOperation(
-                        "MongoClient.bulk_write requires MongoDB server version 8.0+."
-                    )
-                await self.execute_no_results(connection)
-                return ClientBulkWriteResult(None, False, False)  # type: ignore[arg-type]
+            # This path never reaches the command-span code that would otherwise
```
NoahStapp, line 634: Do we need a similar fix for collection-level bulk-writes?

test/asynchronous/test_otel.py
```diff
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Test OpenTelemetry command-span support."""
+"""Test OpenTelemetry operation spans, excluding cursor getMores."""
```
NoahStapp, line 15: Are cursor getMores tested elsewhere?

pymongo/_otel.py
```diff
@@ -172,6 +213,42 @@ def _build_query_summary(command_name: str, dbname: str, collection: Optional[st
     return f"{command_name} {dbname}"
 
 
+# Some `_Op` values are the wire command name ("drop"/"create") rather than the
+# spec's db.operation.name ("dropCollection"/"createCollection"). Translate here
+# instead of renaming `_Op`: `_WRITES_WITH_CLUSTER_TIME` in operations.py matches
+# these exact strings to pick which writes get afterClusterTime.
+_OPERATION_NAME_OVERRIDES = {
```
NoahStapp, line 230: I think we need "rename -> renameCollection" here too.

pymongo/_otel.py
```diff
@@ -200,15 +277,33 @@ def start_command_span(
...
     """
     if not _is_tracing_enabled(tracing_options):
         return None
+
+    collection = _extract_collection_name(command_name, dbname, cmd)
+    # Backfill the operation span's name/namespace/summary from the first command
+    # built inside it. Before the sensitive-command return below, since the
+    # operation span needs those attributes even when the command gets no span.
+    current_operation = _CURRENT_OPERATION_NAME.get()
+    if current_operation is not None:
```
NoahStapp, line 305: Does this backfill run on each retry attempt too? If so, does it need to?

pymongo/asynchronous/cursor_base.py
```diff
@@ -175,6 +175,7 @@ async def _die_lock(self) -> None:
             # ___init__ did not run to completion (or at all).
             return
 
+        self._end_operation_telemetry()
```
NoahStapp, line 180: Shouldn't this come after the actual close, so if the close fails this also records the failure?

pymongo/asynchronous/client_bulk.py
```diff
@@ -630,13 +630,28 @@ async def execute(
...
+                            "MongoClient.bulk_write requires MongoDB server version 8.0+."
+                        )
+                    await self.execute_no_results(connection)
+            except BaseException as exc:
+                if operation_telemetry is not None:
+                    operation_telemetry.failed(exc)
+                raise
+            else:
+                if operation_telemetry is not None:
+                    operation_telemetry.succeeded()
```
NoahStapp: Can this not use the `with telemetry or contextlib.nullcontext()` pattern used elsewhere instead? Same question twice in `mongo_client.py` too.

NoahStapp (changes requested):
Can you run a micro benchmark measuring performance impact both when OTel is enabled as well as disabled?

pymongo/asynchronous/mongo_client.py
blink1073: I clarified that `AsyncCursor._refresh` is the intended user.

blink1073:
> Can you run a micro benchmark measuring performance impact both when OTel is enabled as well as disabled?

I ran a test with `find_one()` on a spawn host with PyPy 3.9, Python 3.14, and Python 3.14t.  When tracing is disabled the the op/s is not measurably affected.  Adding tracing costs 32–48% of throughput across all variations of interpreter and sync/async.

pymongo/asynchronous/mongo_client.py
blink1073: Applied

pymongo/asynchronous/mongo_client.py
blink1073: Clarified the comment

pymongo/_otel.py
blink1073: `None` is to allow the env var to override when resolved.  I created `_UnresolvedTracingOptions` to distinguish between the two states.

pymongo/_otel.py
blink1073: Logging wants the wire name, OTel wants the logical name (good times).

pymongo/_otel.py
blink1073: I expanded on the two cases

pymongo/_otel.py
blink1073: I like the term leaf, but point taken.  I clarified the relationships.

pymongo/asynchronous/client_bulk.py
blink1073: Good catch, fixed and added a regression test.

test/asynchronous/test_otel.py
blink1073: Yes, in the last PR in this stack, since it is still pending spec approval.

pymongo/_otel.py
blink1073: Rename is not one of the [supported operations](https://github.com/mongodb/specifications/blob/master/source/open-telemetry/open-telemetry.md#covered-operations).

pymongo/_otel.py
blink1073: Clarified the comment.  Yes, repeating is necessary in case an attempt dies before the command is built, and it is idempotent.

pymongo/_otel.py
NoahStapp: Do you know the motivation behind that choice in the spec? Having two of our telemetry APIs record different names for the same operation seems very confusing for both us and users.

pymongo/_otel.py
NoahStapp: Why is that list prefaced with "including but not limited to the following operations"? Does that imply we should support rename if it's easy to do so?

pymongo/asynchronous/cursor_base.py
blink1073: No, added a comment above this line.

pymongo/asynchronous/client_bulk.py
blink1073: Good call, done

pymongo/_otel.py
blink1073: Turns out I was confused as well, there are two different fields, `db.operation.name` and `db.command.name`.  I made the comment more explicit.

.evergreen/scripts/generate_config.py
```diff
@@ -458,7 +458,23 @@ def create_otel_variants():
     expansions = dict(TEST_NAME="otel", COVERAGE="1")
     return [
         create_variant(
-            [".test-non-standard .standalone-noauth-nossl"],
+            [
+                # All three topologies, subset to keep the task count at 22.
+                #
+                # Replica set in full: the only topology where transaction spans
```
NoahStapp: What does "in full" mean here?

pymongo/asynchronous/cursor.py
```diff
@@ -1099,6 +1115,21 @@ async def _refresh(self) -> int:
...
+        error on every failure path, and an exhausted cursor's close() ends it
+        on the way out; both are idempotent, so this only has to cover the
+        remaining case of a successful send that leaves the cursor open.
+        """
```
NoahStapp, line 1128: Excessively long docstring here for an internal helper method. This also appears to be the only use of this method, can we just inline it instead?

pymongo/asynchronous/mongo_client.py
```diff
@@ -145,6 +150,7 @@
 
 
 T = TypeVar("T")
+_CommandCursor = TypeVar("_CommandCursor", bound=AsyncCommandCursor[Any])
```
NoahStapp: What's the typevar for? Can we just use `AsyncCommandCursor[Any]` in place of it?

pymongo/asynchronous/mongo_client.py
```diff
@@ -633,7 +640,10 @@ def __init__(
         .. seealso:: The MongoDB documentation on `connections <https://dochub.mongodb.org/core/connections>`_.
 
         .. versionchanged:: 4.18
```
NoahStapp: Should all of these `versionchanged` notes for OTel be `4.XX` to avoid missing them later?

pymongo/asynchronous/mongo_client.py
```diff
@@ -2035,6 +2070,7 @@ async def _retryable_read(
         :param is_run_command: If this is a runCommand operation, defaults to False.
         :param is_aggregate_write: If this is a aggregate operation with a write, defaults to False.
         :param operation_id: Stable operation id shared across retries, defaults to None
+        :param operation_telemetry: As for ``_retry_internal``, defaults to None.
```
NoahStapp: What does "as for" mean here?

pymongo/asynchronous/mongo_client.py
```diff
@@ -2055,6 +2091,53 @@ async def _retryable_read(
                 operation_id=operation_id,
                 is_run_command=is_run_command,
                 is_aggregate_write=is_aggregate_write,
+                operation_telemetry=operation_telemetry,
+            )
+
+    async def _retryable_read_cursor(
```
NoahStapp: Should this have a name that more clearly reflects its purpose as an OTel wrapper around `_retryable_read`? The current one doesn't make that clear.

pymongo/_telemetry.py
```diff
@@ -260,6 +260,92 @@ def failed(
...
+    Construct once per call to ``_retry_internal``; call :meth:`succeeded` or
+    :meth:`failed` exactly once when the operation's outcome is known, or use
+    it as a context manager to do so automatically. It is a no-op throughout when
+    tracing is disabled.
+
+    This span is shared by every attempt, while each attempt gets a command
+    span of its own underneath it. Retries are therefore visible as sibling
+    command spans rather than being collapsed into one.
+
+    With ``set_current=False`` the span is not made current at construction.
```
NoahStapp: Should this explain what it means for a span to be current?

pymongo/asynchronous/mongo_client.py
```diff
@@ -1908,9 +1928,17 @@ async def _run_operation(
...
                     operation.conn_mgr.conn,
                 ):
-                    return await run_with_conn(
-                        operation.conn_mgr.conn, operation, operation.read_preference
-                    )
+                    # Exhaust/pinned cursors bypass _retry_internal, so make the
+                    # caller's span current here to keep their command spans
+                    # nested under it.
+                    with (
+                        operation_telemetry.use()
```
NoahStapp, line 1934: I don't think either caller of `_run_operation` ever passes a non-None `operation_telemetry`, so this code never executes.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2055,6 +2091,53 @@ async def _retryable_read(
...
+        """Run a command cursor read within its own operation span.
+
+        Takes the same arguments as :meth:`_retryable_read`, plus the namespace
+        for the span. A command cursor's first batch is fetched inside that
+        call, before the cursor exists, so the span cannot be owned by the
+        cursor the way a find cursor's is; create it here instead.
+
+        The span ends with the command that created the cursor.
+        """
+        operation_telemetry = _operation_telemetry_or_none(
```
NoahStapp, line 2120: We need to pass `is_run_command` here too.

pymongo/asynchronous/cursor.py
```diff
@@ -974,9 +975,13 @@ async def _send_message(self, operation: Union[_Query, _GetMore]) -> None:
...
         try:
             response = await client._run_operation(
-                operation, self._run_with_conn, address=self._address
+                operation,
+                self._run_with_conn,
+                address=self._address,
+                operation_telemetry=self._operation_telemetry,
             )
         except OperationFailure as exc:
+            self._end_operation_telemetry(exc)
```
NoahStapp: We need to wait to call this until the tailable error check below that returns normally. Otherwise we'll emit an error span even if we don't raise an error.

pymongo/_otel.py
```diff
@@ -261,8 +384,151 @@ def end_command_span_failure(
...
+    if handle is None:
+        return
+    handle.span.record_exception(exc)
+    _set_exception_attributes(handle.span, exc)
+    handle.span.set_status(Status(StatusCode.ERROR, description=str(exc)))
+    if handle._cm is None:
+        handle.span.end()
+        return
+    _CURRENT_OPERATION_NAME.reset(handle._name_token)
+    handle._cm.__exit__(None, None, None)
```
NoahStapp: Can any of these calls raise an error and interrupt this cleanup process?

pymongo/_otel.py
```diff
@@ -261,8 +384,151 @@ def end_command_span_failure(
     if span is None:
         return
     span.record_exception(exc)
+    _set_exception_attributes(span, exc)
     code = failure.get("code")
     if code is not None:
         span.set_attribute("db.response.status_code", str(code))
     span.set_status(Status(StatusCode.ERROR, description=failure.get("errmsg")))
     span.end()
```
NoahStapp: Can any of these calls raise an error and interrupt this cleanup process?

pymongo/_otel.py
```diff
@@ -198,17 +283,35 @@ def start_command_span(
...
+    current_operation = _CURRENT_OPERATION_NAME.get()
+    if current_operation is not None:
+        current_span = trace.get_current_span()
+        if current_span.is_recording():
+            summary = _build_query_summary(current_operation, dbname, collection)
+            current_span.update_name(summary)
+            current_span.set_attribute("db.namespace", dbname)
+            current_span.set_attribute("db.operation.summary", summary)
+            if collection:
+                current_span.set_attribute("db.collection.name", collection)
```
NoahStapp, line 313: Can we detect if these are already set correctly and skip this if so to save the cost of these calls?

NoahStapp (changes requested):

.evergreen/scripts/generate_config.py
blink1073: It meant the replica set variant isn't subsetted the way the other two are. Reworded to "Replica set keeps every task".

pymongo/asynchronous/cursor.py
blink1073: Trimmed the docstring. I kept the method rather than inlining it, because PR 4 gives it a second call site and an `own_span` parameter when `getMores` get their own spans, so inlining here would be reverted there.

pymongo/asynchronous/mongo_client.py
blink1073: Good call, removed

pymongo/asynchronous/mongo_client.py
blink1073: Done

pymongo/asynchronous/mongo_client.py
blink1073: Reworded to "Same as _retry_internal's".  Do you think we should repeat the reasoning here?

pymongo/asynchronous/mongo_client.py
blink1073: Renamed to `_retryable_read_cursor_in_span`.

pymongo/_telemetry.py
blink1073: I added some explanation: with `set_current=False` nothing nests under the span until a caller opens a `use` block.

pymongo/asynchronous/mongo_client.py
blink1073: `AsyncCursor._refresh` assigns `self._operation_telemetry` before it sends, so the query path does pass one. I verified by running the code and inspecting:  I got `('_Query', True)`, then `('_GetMore', False)` for each `getMore`.  `AsyncCommandCursor` passes `None`, and so do getMores in this PR, which is why the parameter is optional.

pymongo/asynchronous/mongo_client.py
blink1073: Fixed

pymongo/asynchronous/cursor.py
blink1073: Good call, I added `test_tailable_rollover_is_not_an_error_span` as a regression test

pymongo/_otel.py
blink1073: I wrapped the recording calls in try/finally.

pymongo/_otel.py
blink1073: The OTel API has no getter for a span's attributes, so detecting "already set" means tracking it ourselves in another contextvar.  I don't think it is worth it.

pymongo/_otel.py
blink1073: I wrapped this one in try/finally as well.

pymongo/_otel.py
blink1073: Forgot to reply, updated to point to DRIVERS-3625 for the follow up work.

pymongo/asynchronous/database.py
```diff
@@ -1051,8 +1052,8 @@ async def inner(
                 else:
                     raise InvalidOperation("Command does not return a cursor.")
 
-            return await self.client._retryable_read(
-                inner, read_preference, tmp_session, command_name, None, False
+            return await self.client._retryable_read_cursor_in_span(
```
NoahStapp, line 1055: We need to pass `is_run_command=True` here for consistency with `command()`.

pymongo/asynchronous/cursor.py
```diff
@@ -1017,7 +1028,7 @@ async def _send_message(self, operation: Union[_Query, _GetMore]) -> None:
                 # Update the namespace used for future getMore commands.
                 ns = cursor.get("ns")
                 if ns:
-                    self._dbname, self._collname = ns.split(".", 1)
+                    self._dbname, self._collname = helpers_shared._split_namespace(ns)
```
NoahStapp, line 1031: Do we need this same fix in command cursor?

pymongo/asynchronous/cursor.py
NoahStapp: Should we construct the span before this raise so we record the failure?

pymongo/_otel.py
```diff
@@ -260,9 +388,160 @@ def end_command_span_failure(
...
+        self.span = span
+        self._cm = cm
+        self._name_token = name_token
+        self.operation_name = operation_name
+
+
+def start_operation_span(
+    tracing_options: Optional[TracingOptions],
+    operation: str,
+    parent_span: Optional[Span],
```
NoahStapp, line 429: We always pass this as `None`, is that not the case in a later PR in this stack?

pymongo/_telemetry.py
```diff
@@ -260,6 +260,93 @@ def failed(
...
+    cursor does; that caller makes it current with :meth:`use`.
+    """
+
+    __slots__ = ("handle",)
+
+    def __init__(
+        self,
+        tracing_options: Optional[_otel.TracingOptions],
+        operation: str,
+        session: Optional[Any],
```
NoahStapp, line 288: The `session` parameter is not stored or used.

NoahStapp (changes requested):

pymongo/asynchronous/database.py
blink1073: Added `is_run_command=True` here to match `command()`.

pymongo/asynchronous/cursor.py
blink1073: No. The command cursor is covered for the command that creates it through `_retryable_read_cursor_in_span` (now with `is_run_command=True`). Its getMore spans are deferred to #2994

pymongo/asynchronous/cursor.py
blink1073: Done. The operation span is created before the min/max validation, and the `InvalidOperation` is recorded as a failed span before re-raising.

pymongo/_otel.py
blink1073: #2992 handles it. Inside a transaction it uses `session._transaction.span` as the explicit parent so the operation span nests under the transaction span.

pymongo/_telemetry.py
blink1073: #2992 uses it. When the session is in a transaction it reads `session._transaction.span` as the parent span, same parameter as `start_operation_span` `parent_span`.

pymongo/asynchronous/mongo_client.py
```diff
@@ -2055,6 +2090,54 @@ async def _retryable_read(
                 operation_id=operation_id,
                 is_run_command=is_run_command,
                 is_aggregate_write=is_aggregate_write,
+                operation_telemetry=operation_telemetry,
+            )
+
+    async def _retryable_read_cursor_in_span(
```
NoahStapp: Can we just make `dbname` and `collection` optional parameters to `_retryable_read(...)` and remove this wrapper entirely?

NoahStapp (changes requested):

pymongo/asynchronous/mongo_client.py
blink1073: Done

NoahStapp (approved):
LGTM!

PR #2998, opened 2026-08-18
PYTHON-5981 Reject aggregate/pipeline as aggregation options

aclark4life:
[PYTHON-5981](https://jira.mongodb.org/browse/PYTHON-5981)

## Changes in this PR

`_AggregationCommand.__init__` now raises `ConfigurationError` if `aggregate` or `pipeline` is passed as a keyword option.

Both names are fields of the `aggregate` command itself, so previously passing them as options silently replaced the generated command's target namespace or its pipeline rather than adding an option. They now join `explain` as rejected option names, with the message `The <name> option cannot be specified as a keyword argument.`

This affects the helpers that build an aggregate command:

- `AsyncCollection.aggregate` / `Collection.aggregate`
- `AsyncCollection.aggregate_raw_batches` / `Collection.aggregate_raw_batches`
- `AsyncDatabase.aggregate` / `Database.aggregate`
- `AsyncCollection.list_search_indexes` / `Collection.list_search_indexes`

## Test Plan

Added `test_aggregate_reserved_options` to `test/asynchronous/test_collection.py` covering `aggregate=` on all four helpers plus `pipeline=` on `list_search_indexes`, asserting `ConfigurationError` in each case.

### Checklist for Author
- [x] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/test/asynchronous/test_collection.py
+++ b/test/asynchronous/test_collection.py
@@ -1569,7 +1569,9 @@ class AsyncTestCollection(AsyncIntegrationTest):
             await coll.aggregate([{"$out": "output-collection"}])
 
     async def test_aggregate_reserved_options(self):
-        # Reserved command fields must not be settable as options.
+        # "aggregate" and "pipeline" are fields of the aggregate command itself,
+        # so they must not be settable as keyword options: doing so would
+        # replace the command's target namespace or its pipeline.
         db = self.db
         with self.assertRaises(ConfigurationError):
             await db.test.aggregate([], aggregate="other")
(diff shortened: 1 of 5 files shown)


test/asynchronous/test_collection.py
```diff
@@ -1568,6 +1568,22 @@ async def test_aggregate(self):
         with self.write_concern_collection() as coll:
             await coll.aggregate([{"$out": "output-collection"}])
 
+    async def test_aggregate_reserved_options(self):
+        # "aggregate" and "pipeline" are fields of the aggregate command itself,
+        # so they must not be settable as keyword options: doing so would
+        # replace the command's target namespace or its pipeline.
+        db = self.db
+        with self.assertRaises(ConfigurationError):
```
blink1073: Do we need to test all combinations of method and kwargs?
Maybe something like kwarg_options = [{"pipeline": ..>}, {"aggregate":..}, {"pipeline": ..>, "aggregate":..}] and then feed that to each of the test functions?

blink1073 (approved):
LGTM

PR #3003, opened 2026-08-21
PYTHON-3419 Use memoryview to avoid byte copies when decoding larger RawBSONDocuments

NoahStapp:
[PYTHON-3419](https://jira.mongodb.org/browse/PYTHON-3419)

## Changes in this PR

RawBSONDocuments of size >= 4KB are decoded to memoryview slices of their buffer rather than bytes copies. This is a very significant performance improvement at no cost to documents smaller than the 4KB threshold:

```
inflate 100 10KB docs: ~17% faster
inflate 10 1MB docs: ~95% faster
decode 50 100KB docs: ~82% faster
additional memory used for 20 5MB docs: 86MB baseline vs ~0MB
inflate 1000 <= 4KB docs: flat
```

## Test Plan

Added new unit tests in `test/test_raw_bson_shared.py` and new integration tests to `test/(a)synchronous/test_raw_bson.py`.

### Checklist for Author
- [X] Did you update the changelog (if necessary)?
- [X] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/bson/__init__.py
+++ b/bson/__init__.py
@@ -243,28 +243,18 @@ _UNPACK_TIMESTAMP_FROM = struct.Struct("<II").unpack_from
 def get_data_and_view(data: Any) -> tuple[Any, memoryview]:
     if isinstance(data, (bytes, bytearray)):
         return data, memoryview(data)
-    # Copy buffer-protocol inputs so the decode (and any raw document views
-    # taken of it) can't observe later mutations of the caller's buffer.
+    # Copy other inputs so decoding is immutable
     data = memoryview(data).tobytes()
     return data, memoryview(data)
 
 
 def _raw_as_bytes(raw: Union[bytes, bytearray, memoryview]) -> bytes:
-    """Coerce a raw BSON buffer (bytes, bytearray, or memoryview) to bytes."""
+    """Convert a raw BSON buffer (bytes, bytearray, or memoryview) to bytes."""
     return raw if isinstance(raw, bytes) else bytes(raw)
 
 
 def _raw_slice(data: Any, view: memoryview, position: int, end: int, obj_size: int) -> Any:
-    """Return the raw BSON document spanning data[position:end + 1] for use
-    as a RawBSONDocument's buffer.
-
-    A document spanning an entire immutable buffer is passed through as-is.
-    Other documents at least _RAW_BSON_VIEW_THRESHOLD bytes are exposed as
-    read-only memoryview slices of the parent buffer instead of bytes copies.
-    Views are only taken of immutable buffers: documents in a mutable buffer
-    (e.g. a bytearray) are copied so the caller can't mutate the document
-    out from under us.
-    """
+    """Return the raw BSON document spanning ``position` to ``end`` for use as a buffer."""
     whole_span = position == 0 and obj_size == len(data)
     if view.readonly:  # data is immutable (bytes).
         if whole_span:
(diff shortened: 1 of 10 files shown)


blink1073:
`decode_iter()` does not have the improvements this PR adds to `decode_all()`, so this test fails:

```python
def test_decode_iter_matches_decode_all_view_backing(self):
    opts = CodecOptions(document_class=RawBSONDocument)
    big = {"payload": b"x" * (_RAW_BSON_VIEW_THRESHOLD + 100)}
    stream = encode(big) * 3

    all_types = [type(d.raw) for d in decode_all(stream, opts)]
    iter_types = [type(d.raw) for d in decode_iter(stream, opts)]

    self.assertEqual(all_types, iter_types)
```

Also, `elements_to_dict` computes `offset = string - PyBytes_AS_STRING(buffer_owner)` and slices on it without checking that `offset` and `offset + max` land inside `buffer_owner`. Nothing enforces that today; it works only because all three call sites happen to derive `string` and `buffer_owner` from the same object. We should an assertion so a future mismatch fails loudly instead of silently returning truncated data:

```c
Py_ssize_t offset = string - PyBytes_AS_STRING(buffer_owner);
assert(offset >= 0 && offset + max <= PyBytes_GET_SIZE(buffer_owner));
PyObject* top_view = PyMemoryView_FromObject(buffer_owner);
```

NoahStapp:
4.2 failures are expected until https://github.com/mongodb/mongo-python-driver/pull/2892 is merged.

NoahStapp:
> `decode_iter()` does not have the improvements this PR adds to `decode_all()`, so this test fails:
> 
> ```python
> def test_decode_iter_matches_decode_all_view_backing(self):
>     opts = CodecOptions(document_class=RawBSONDocument)
>     big = {"payload": b"x" * (_RAW_BSON_VIEW_THRESHOLD + 100)}
>     stream = encode(big) * 3
> 
>     all_types = [type(d.raw) for d in decode_all(stream, opts)]
>     iter_types = [type(d.raw) for d in decode_iter(stream, opts)]
> 
>     self.assertEqual(all_types, iter_types)
> ```
> 
> Also, `elements_to_dict` computes `offset = string - PyBytes_AS_STRING(buffer_owner)` and slices on it without checking that `offset` and `offset + max` land inside `buffer_owner`. Nothing enforces that today; it works only because all three call sites happen to derive `string` and `buffer_owner` from the same object. We should an assertion so a future mismatch fails loudly instead of silently returning truncated data:
> 
> ```c
> Py_ssize_t offset = string - PyBytes_AS_STRING(buffer_owner);
> assert(offset >= 0 && offset + max <= PyBytes_GET_SIZE(buffer_owner));
> PyObject* top_view = PyMemoryView_FromObject(buffer_owner);
> ```

I don't think we want to make the same changes to `decode_iter` for the following reasons:

1. We don't currently implement a C-extension version of `decode_iter`. This implies that `decode_iter` is not intended for maximum performance and that users should use `decode_all` instead for such workloads, which is what I would expect. To get a speedup on the same scale as this PR does for `decode_all`, we'd need to add a C-extension implementation.
2. The C `_bson_to_dict` changes in this PR already improve `decode_iter`'s performance since both APIs call into it. The remaining difference is that `decode_iter` still copies each document as part of decoding. For a generator method, I think that copying is what we want to prevent the entire input buffer from being held in memory for the lifetime of the generator. 
3. We use `decode_all` within the driver, so a performance improvement there speeds up every operation. We don't use `decode_iter` at all, so I see it as less of a priority.
4. I'm fine with `decode_iter` and `decode_all` having different behavior on large documents. If necessary, we can document that and specify more explicitly the intended use cases for each to reduce confusion.

Agreed with the safety assertion change!

blink1073 (approved):
LGTM

PR #3009, opened 2026-08-27
PYTHON-5290 - Align client bulkwrite prose test with spec

NoahStapp:
[PYTHON-5290](https://jira.mongodb.org/browse/PYTHON-5290)

## Changes in this PR

The `test_timeout_in_multi_batch_bulk_write` must only be run against standalone topologies according to [the spec
](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/tests/README.md#11-multi-batch-bulkwrites), but we currently apply no such limitation. This PR aligns us with the spec.

## Test Plan

N/A

### Checklist for Author
- [ ] Is there test coverage?

### Checklist for Reviewer

### DIFF
--- a/test/asynchronous/test_client_bulk_write.py
+++ b/test/asynchronous/test_client_bulk_write.py
@@ -634,7 +634,7 @@ class TestClientBulkWriteCSOT(AsyncIntegrationTest):
 
     @async_client_context.require_version_min(8, 0, 0, -24)
     @async_client_context.require_failCommand_fail_point
-    @flaky(reason="PYTHON-5290", max_runs=3, affects_cpython_linux=True)
+    @async_client_context.require_standalone
     async def test_timeout_in_multi_batch_bulk_write(self):
         if sys.platform != "linux" and "CI" in os.environ:
             self.skipTest("PYTHON-3522 CSOT test runs too slow on Windows and MacOS")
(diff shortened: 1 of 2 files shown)


test/asynchronous/test_client_bulk_write.py
```diff
@@ -29,7 +29,6 @@
 from pymongo.errors import (
     ClientBulkWriteException,
     DocumentTooLarge,
-    InvalidOperation,
```
aclark4life, line 32: Maybe we should leave this explicit import because without it it's only covered by `from pymongo.operations import *` accidentally. `InvalidOperation` still appears to be in use on lines 52 and 535.

aclark4life (changes requested):

test/asynchronous/test_client_bulk_write.py
NoahStapp: If it's being imported somehow and the file doesn't fail to run then I don't see the need for a second explicit import.

aclark4life: The reason is  (leaving aside the "code smell" of `*` imports …) `from pymongo.operations import *` is (typically) for things in `pymongo.operations` and so if `operations` one day decides to _not_ import `InvalidOperation` then _this_ test unexpectedly breaks.

NoahStapp: Resolved by removing the `*` import entirely and explicitly importing instead.

aclark4life (approved):

PR #3037, opened 2026-09-03
PYTHON-6071 Fix ML-KEM OCSP tests and run min-deps OCSP task on PRs

blink1073:
PYTHON-6071

## Changes in this PR

- Replaced real ML-KEM key generation in two OCSP tests with `MagicMock(spec=...)`, so they no longer depend on `cryptography`'s backend supporting ML-KEM.
- Added a min-deps stapled OCSP task to the PR-triggered Evergreen variant, alongside the existing latest-CPython one.
- Scoped both PR-gated OCSP tasks to one algorithm (`ecdsa`) to keep PR CI at two tasks instead of four.

## Test Plan

- `pytest test/test_ocsp_support.py::TestVerifySignature -m ocsp`: 9 passed, including under a min-deps (`cryptography==47.0.0`) environment where they previously failed with `UnsupportedAlgorithm`.
- `just lint` and `just typing`: clean.

### Checklist for Author
- [x] Did you update the changelog (if necessary)? Not needed (test/CI-only fix).
- [x] Is there test coverage? Yes, and it now runs on every PR.
- [x] Is any followup work tracked in a JIRA ticket? If so, add link(s). None; PYTHON-6071 is caused-by PYTHON-6032.

### Checklist for Reviewer

### DIFF
--- a/.evergreen/generated_configs/tasks.yml
+++ b/.evergreen/generated_configs/tasks.yml
@@ -898,6 +898,7 @@ tasks:
       - ocsp-ecdsa
       - "4.4"
       - ocsp-staple
+      - pr
   - name: test-ocsp-ecdsa-valid-cert-server-staples-v5.0-python3.10-min-deps
     commands:
       - func: run tests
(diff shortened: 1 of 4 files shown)


caseyclements (approved):
LGTM

PR #3039, opened 2026-09-03
PYTHON-6076 Request workflows permission in create-release-branch

blink1073:
PYTHON-6076

## Changes in this PR
The Create Release Branch workflow fails at the push step because the `Prep branch` commit edits `.github/workflows/release-python.yml`, and the App token minted by `secure-checkout@v3` does not request `workflows` permission. GitHub rejects the push with GH013.

- Replaced `secure-checkout@v3` with `actions/create-github-app-token@v3`, requesting `workflows: write` in addition to `contents` and `pull-requests`.
- Added `actions/checkout@v7` with `persist-credentials: false`.
- Added a step that configures git push auth from `GH_TOKEN`. `setup@v3` only sets `user.name`/`user.email`, and the third-party `bump-version` push step depends on checkout-configured credentials.

## Test Plan
Rerun the Create Release Branch workflow with `branch_name=v4.18`, `version=4.18.1.dev0`, `base_ref=main` and confirm the `v4.18` branch is created and pushed. Validated locally by pre-commit's "Validate GitHub Workflows" hook.

### Checklist for Author
- [x] Did you update the changelog (if necessary)?  N/A: internal release workflow, no user-facing change.
- [x] Is there test coverage?  N/A: workflow change; verified by the workflow validator and a live run.
- [x] Is any followup work tracked in a JIRA ticket? If so, add link(s).  PYTHON-6076.

### Checklist for Reviewer

### DIFF
--- a/.github/workflows/create-release-branch.yml
+++ b/.github/workflows/create-release-branch.yml
@@ -33,10 +33,30 @@ jobs:
     outputs:
       version: ${{ steps.pre-publish.outputs.version }}
     steps:
-      - uses: mongodb-labs/drivers-github-tools/secure-checkout@v3
+      - name: Create app token
+        id: app-token
+        uses: actions/create-github-app-token@v3
         with:
-          app_id: ${{ vars.APP_ID }}
-          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          permission-contents: write
+          permission-pull-requests: write
+          permission-workflows: write
+      - name: Store token
+        shell: bash -eu {0}
+        # zizmor: ignore[template-injection] GH_TOKEN is a GitHub App installation token, not attacker-controllable
+        run: echo "GH_TOKEN=${{ steps.app-token.outputs.token }}" >> "$GITHUB_ENV"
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release-branch.yml
```diff
@@ -33,10 +33,30 @@ jobs:
...
+      - name: Store token
+        shell: bash -eu {0}
+        # zizmor: ignore[template-injection] GH_TOKEN is a GitHub App installation token, not attacker-controllable
+        run: echo "GH_TOKEN=${{ steps.app-token.outputs.token }}" >> "$GITHUB_ENV"
+      - uses: actions/checkout@v7.0.1
+        with:
+          ref: ${{ github.ref }}
+          token: ${{ env.GH_TOKEN }}
+          fetch-depth: 1
+          persist-credentials: false
```
aclark4life, line 54: I think we can drop `persist-credentials: false` and `Configure git auth for push` because `actions/checkout` already writes the auth header so we don't need to not-save it and then re-create it.

aclark4life (changes requested):

.github/workflows/create-release-branch.yml
blink1073: We don't want the auth stored on disk

aclark4life (approved):

PR #3047, opened 2026-09-10
PYTHON-6085 Prep for 4.18.1 release

aclark4life:
PYTHON-6085

## Changes in this PR

Bumps the version on `v4.18` to 4.18.1 and adds the 4.18.1 changelog section.

## Test Plan

No functional change. Version and changelog only.

### Checklist for Author
- [x] Is there test coverage? N/A
- [x] Is any followup work tracked in a JIRA ticket? If so, add link(s). N/A


doc/changelog.rst
```diff
@@ -1,6 +1,23 @@
...
+
+Version 4.18.1 is a bug fix release.
+
+- Use an exact match for the file ID in GridFS delete methods
+  (`PYTHON-5994`_).
+
+Issues Resolved
+...............
+
+See the `PyMongo 4.18.1 release notes in JIRA`_ for the list of resolved issues
```
sleepyStick: nit: i don't think this line is usually in the changelog? (only in the announcements?) but i might be misremembering

aclark4life: Good catch

doc/changelog.rst
```diff
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+Changes in Version 4.18.1 (2026/09/10)
+--------------------------------------
+
+Version 4.18.1 is a bug fix release.
+
+- Use an exact match for the file ID in GridFS delete methods
```
sleepyStick, line 9: okay i don't think i noticed this until now (sorry for not realizing sooner) but can we not embed the link like `text <link>` as opposed to doing it after? the reference below syntax is (maybe) new to me but it feels a bit silly to put the link two lines down from the original text when there's only one link lol

Jibola (approved):

doc/changelog.rst
sleepyStick: (i see that jib approved lol this doesn't matter that much if you wanna go ahead and merge without this)

aclark4life: k

PR #3054, opened 2026-09-15
PYTHON-6040 Preserve version-to-name mapping in Client Metadata

blink1073:
Implements the DRIVERS-3251 handshake metadata update: PyMongo's `driver.name` and `driver.version` are now pipe-delimited lists with a 1:1 index correspondence. Each name segment (`|c`, `|async`, or a framework) has a matching version entry, so the number of `|` in `name` always equals the number of `|` in `version`.

## Changes in this PR
- Reject the `|` delimiter in `DriverInfo` fields at construction time.
- Append empty version entries for the `|c` and `|async` name suffixes so name/version stay index-aligned.
- Rewrite `_update_metadata` to always append the delimiter for name and version, deduplicate appended drivers by whole `DriverInfo` object, and track appended drivers.
- Fix `_truncate_metadata` to keep name/version index-aligned when metadata is truncated to the 512-byte limit.
- Update the metadata, dedup, and handshake tests.

## Test Plan
- Standard evergreen tasks

### Checklist for Author
- [ ] Did you update the changelog (if necessary)? N/A
- [x] Is there test coverage?
- [ ] Is any followup work tracked in a JIRA ticket? If so, add link(s). N/A

### Checklist for Reviewer


pymongo/pool_options.py
```diff
@@ -226,34 +246,44 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
-        else:
-            metadata["driver"]["name"] = _METADATA["driver"]["name"]
+        # Trim wrapper version and name content first, dropping paired segments
+        # only as a last resort, so name and version stay 1:1 aligned.
+        while True:
+            encoded_size = len(bson.encode(metadata))
+            if encoded_size <= _MAX_METADATA_SIZE:
+                break
+            overflow = encoded_size - _MAX_METADATA_SIZE
+            previous = (driver.get("name"), driver.get("version"))
```
NoahStapp:
```suggestion
            previous = (driver.get("name", ""), driver.get("version", ""))
```

test/asynchronous/test_client.py
```diff
@@ -134,6 +134,19 @@
 _IS_SYNC = False
 
 
+def _driver_version(base_version: str, name: str, last_version: str | None = None) -> str:
```
NoahStapp: This doesn't have to be synchronized, there's no async or sync-specific code. We can put this in `test/utils_shared.py` instead.

test/asynchronous/test_client.py
```diff
@@ -438,19 +463,98 @@ async def test_metadata(self):
...
+        # Empty strings are treated as unset, so a duplicate differing only in
+        # None vs "" is a no-op.
+        client = self.simple_client(connect=False)
+        client.append_metadata(DriverInfo("library", None, "Library Platform"))
+        names = client.options.pool_options.metadata["driver"]["name"]
+        vers = client.options.pool_options.metadata["driver"]["version"]
+        client.append_metadata(DriverInfo("library", "", "Library Platform"))
+        metadata = client.options.pool_options.metadata
+        self.assertEqual(metadata["driver"]["name"], names)
+        self.assertEqual(metadata["driver"]["version"], vers)
```
NoahStapp, line 567: Can we split this up into several smaller testcases each testing a subset of the current body?

test/asynchronous/test_client_metadata.py
```diff
@@ -232,6 +232,124 @@ async def test_handshake_documents_include_backpressure(self):
...
+                assert version0 is not None
+                driver_name = name0.split("|")[0]
+                driver_version = version0.split("|")[0]
+
+                def resolve(value: Optional[str]) -> Optional[str]:
+                    if value is None:
+                        return None
+                    return value.format(driver_name=driver_name, driver_version=driver_version)
+
+                # Append each DriverInfoOptions in order.
```
NoahStapp:
```suggestion
                # Append each DriverInfo in order.
```

pymongo/pool_options.py
```diff
@@ -368,28 +410,36 @@ def __init__(
...
+
+            _truncate_metadata(metadata)
+
+            self.__metadata = metadata
 
-        self.__metadata = metadata
+            # Only track drivers whose appended name/version pair survived
+            # truncation (i.e. the name gained a segment), so __appended_drivers
+            # stays bounded and the dedup membership check stays fast.
+            if metadata["driver"]["name"].count("|") > name_delims:
```
NoahStapp: This only checks if the name survived, not the version.

pymongo/pool_options.py
```diff
@@ -368,28 +410,36 @@ def __init__(
...
     def _update_metadata(self, driver: DriverInfo) -> None:
-        """Updates the client's metadata"""
-        if driver.name and driver.name.lower() in self.__metadata["driver"]["name"].lower().split(
-            "|"
-        ):
-            return
+        """Updates the client's metadata."""
+        with self.__metadata_lock:
+            driver = _normalize_driver(driver)
+            if driver in self.__appended_drivers:
```
NoahStapp, line 448: We should document in the changelog that `append_metadata` now checks exact `DriverInfo` match and not just name.

pymongo/pool_options.py
```diff
@@ -336,6 +368,8 @@ def __init__(
         self.__load_balanced = load_balanced
         self.__credentials = credentials
         self.__metadata = copy.deepcopy(_METADATA)
+        self.__appended_drivers: list[DriverInfo] = []
+        self.__metadata_lock = _create_lock()
```
NoahStapp: This uses a sync lock even on the async API, which will block the loop.

test/asynchronous/test_client_metadata.py
```diff
@@ -232,6 +232,124 @@ async def test_handshake_documents_include_backpressure(self):
...
+            with self.subTest(description=description):
+                client = await self.async_rs_or_single_client(
+                    "mongodb://" + self.server.address_string,
+                    maxIdleTimeMS=1,
+                )
+                # Capture the driver's own name and version from the first handshake.
+                name0, version0, _, _ = await self.send_ping_and_get_metadata(client, True)
+                await asyncio.sleep(0.005)
+
+                assert name0 is not None
```
NoahStapp: These should be unittest-style asserts for consistency with our test suite.

test/asynchronous/test_client_metadata.py
```diff
@@ -232,6 +232,124 @@ async def test_handshake_documents_include_backpressure(self):
...
+                    ),
+                )
+                self.assertEqual(
+                    version1,
+                    version0
+                    + expected_version_suffix.format(
+                        driver_name=driver_name, driver_version=driver_version
+                    ),
+                )
+                await client.close()
```
NoahStapp: These clients will get leaked if a subtest fails, we need a try-catch-finally block here.

test/asynchronous/test_client_metadata.py
```diff
@@ -232,6 +232,124 @@ async def test_handshake_documents_include_backpressure(self):
...
+                    driver=DriverInfo("library", "1.2", "Library Platform"),
+                )
+                # Send initial handshake.
+                name0, version0, platform0, _metadata = await self.send_ping_and_get_metadata(
+                    client, True
+                )
+                await asyncio.sleep(0.005)
+                # Appending metadata containing the delimiter raises.
+                with self.assertRaises(ValueError):
+                    DriverInfo(name, version, platform)
```
NoahStapp, line 347: This is inconsistent with the comment above: it doesn't append metadata, but constructs a `DriverInfo`.

pymongo/pool_options.py
```diff
@@ -368,28 +410,36 @@ def __init__(
...
 
-        metadata = copy.deepcopy(self.__metadata)
+            name_delims = self.__metadata["driver"]["name"].count("|")
+            metadata = copy.deepcopy(self.__metadata)
 
-        if driver.name:
             metadata["driver"]["name"] = "{}|{}".format(
-                metadata["driver"]["name"],
-                driver.name,
+                metadata["driver"]["name"], driver.name or ""
```
NoahStapp: Since the `_normalize_driver(driver)` call above already normalized the name, version, and platform fields, we don't need to do `or ""` afterwards.

NoahStapp (changes requested):

pymongo/pool_options.py
blink1073: Good call, I think we don't need a lock in async at all, since the only condition that would cause a race is an async client being used in two threads, which we don't support.

pymongo/pool_options.py
blink1073: Done

test/asynchronous/test_client.py
blink1073: Done

test/asynchronous/test_client.py
blink1073: Done

test/asynchronous/test_client_metadata.py
blink1073: Fixed

test/asynchronous/test_client_metadata.py
blink1073: Done

test/asynchronous/test_client_metadata.py
blink1073: Fixed

pymongo/pool_options.py
blink1073: Fixed

test/asynchronous/test_client_metadata.py
blink1073: I went with adding cleanup instead to avoid the indent

pymongo/pool_options.py
blink1073: Fixed

pymongo/pool_options.py
```diff
@@ -226,34 +247,44 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
+                driver["name"] = "|".join(n_parts)
+            elif len(n_parts) > 1:
+                n_parts.pop()
+                v_parts.pop()
+                driver["name"] = "|".join(n_parts)
+                driver["version"] = "|".join(v_parts)
+            else:
+                break
+
+            if previous == (driver.get("name"), driver.get("version")):
```
NoahStapp: Do we need this? The only way to reach this condition is if we have shortened the metadata, which will eventually cause the `encoded_size` to drop below `_MAX_METADATA_SIZE`.

pymongo/pool_options.py
```diff
@@ -226,34 +247,44 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
     if plat:
-        plat = plat[:-overflow]
+        plat = _truncate_utf8(plat, overflow)
     if plat:
         metadata["platform"] = plat
     else:
         metadata.pop("platform", None)
     encoded_size = len(bson.encode(metadata))
     if encoded_size <= _MAX_METADATA_SIZE:
         return
```
NoahStapp: Do we need this when we do it again just below inside the truncation loop?

pymongo/pool_options.py
```diff
@@ -226,34 +247,44 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
-        else:
-            metadata["driver"]["name"] = _METADATA["driver"]["name"]
+        # Trim wrapper version and name content first, dropping paired segments
+        # only as a last resort, so name and version stay 1:1 aligned.
+        while True:
+            encoded_size = len(bson.encode(metadata))
+            if encoded_size <= _MAX_METADATA_SIZE:
+                break
+            overflow = encoded_size - _MAX_METADATA_SIZE
+            previous = (driver.get("name", ""), driver.get("version", ""))
```
NoahStapp: This can go away as well if we remove the check at the bottom of the loop.

pymongo/pool_options.py
```diff
@@ -226,34 +247,44 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
     if plat:
         metadata["platform"] = plat
     else:
         metadata.pop("platform", None)
     encoded_size = len(bson.encode(metadata))
     if encoded_size <= _MAX_METADATA_SIZE:
         return
-    # 5. Truncate driver info.
-    overflow = encoded_size - _MAX_METADATA_SIZE
+    # 5. Truncate driver info, keeping name and version 1:1 index-aligned.
```
NoahStapp, line 280: What if we store pairs of `(name, version)` directly in a list and then do a single join on "|" when we actually need to serialize the metadata? Then we could remove a lot of the string splitting and make our pair requirement easier to follow and maintain.

NoahStapp (changes requested):

pymongo/pool_options.py
blink1073: Done

pymongo/pool_options.py
blink1073: Removed

pymongo/pool_options.py
blink1073: Removed

pymongo/pool_options.py
blink1073: yep, removed

pymongo/pool_options.py
```diff
@@ -226,34 +247,31 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
+        # trimming wrapper content and dropping paired segments only as a
+        # last resort.
+        pairs = list(zip(driver.get("name", "").split("|"), driver.get("version", "").split("|")))
+        while True:
+            driver["name"] = "|".join(name for name, _ in pairs)
+            driver["version"] = "|".join(version for _, version in pairs)
+            overflow = len(bson.encode(metadata)) - _MAX_METADATA_SIZE
+            if overflow <= 0 or len(pairs) <= 1:
+                break
+            last_name, last_version = pairs[-1]
```
NoahStapp, line 302: Should we first pop any empty pairs? For example, if for some reason we had a platform-only `DriverInfo` get appended, it would also add a blank pair of name and version that we could truncate for free instead of potentially losing real data.

pymongo/pool_options.py
```diff
@@ -226,34 +247,31 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
-            metadata["driver"]["name"] = _METADATA["driver"]["name"]
+        # Keep the name and version segments paired so they stay 1:1 aligned,
+        # trimming wrapper content and dropping paired segments only as a
+        # last resort.
+        pairs = list(zip(driver.get("name", "").split("|"), driver.get("version", "").split("|")))
+        while True:
+            driver["name"] = "|".join(name for name, _ in pairs)
+            driver["version"] = "|".join(version for _, version in pairs)
+            overflow = len(bson.encode(metadata)) - _MAX_METADATA_SIZE
+            if overflow <= 0 or len(pairs) <= 1:
```
NoahStapp: Can we add a comment here stating that a single pair can never exceed `_MAX_METADATA_SIZE`?

test/asynchronous/test_client.py
```diff
@@ -380,12 +381,31 @@ async def test_read_preference(self):
         )
         self.assertEqual(c.read_preference, ReadPreference.NEAREST)
 
+    def _metadata_with_appended_driver(
```
NoahStapp: We can move this into `test/utils_shared.py` by adding a `is_sync` parameter that controls the addition of `async`.

test/test_client_metadata.py
```diff
@@ -232,6 +232,125 @@ def test_handshake_documents_include_backpressure(self):
...
+                )
+                self.assertEqual(
+                    version1,
+                    version0
+                    + expected_version_suffix.format(
+                        driver_name=driver_name, driver_version=driver_version
+                    ),
+                )
+
+    def test_11_appending_metadata_containing_the_delimiter_raises_an_error(self):
```
NoahStapp, line 326: This test effectively duplicates `test_metadata_rejects_delimiter` in much more code, can we remove it?

pymongo/pool_options.py
```diff
@@ -226,34 +247,31 @@ def _truncate_metadata(metadata: MutableMapping[str, Any]) -> None:
...
-        else:
-            metadata["driver"]["name"] = _METADATA["driver"]["name"]
+        # Keep the name and version segments paired so they stay 1:1 aligned,
+        # trimming wrapper content and dropping paired segments only as a
+        # last resort.
+        pairs = list(zip(driver.get("name", "").split("|"), driver.get("version", "").split("|")))
+        while True:
+            driver["name"] = "|".join(name for name, _ in pairs)
+            driver["version"] = "|".join(version for _, version in pairs)
+            overflow = len(bson.encode(metadata)) - _MAX_METADATA_SIZE
```
NoahStapp: This might be excessive performance chasing (even though this code is going to run a lot), but we don't actually need to `bson.encode(metadata)` each loop since we know exactly what we're truncating. Instead we could calculate overflow once and then subtract the size of each truncated string.

That also lets us avoid joining the pairs each loop and just operate purely on the list itself, joining only once truncation is done.

pymongo/pool_options.py
```diff
@@ -368,28 +401,38 @@ def __init__(
...
-        if driver.version:
+        """Updates the client's metadata."""
+        with self.__metadata_lock:
+            driver = _normalize_driver(driver)
+            if driver in self.__appended_drivers:
+                return
+
+            name_delims = self.__metadata["driver"]["name"].count("|")
+            version_delims = self.__metadata["driver"]["version"].count("|")
+            metadata = copy.deepcopy(self.__metadata)
```
NoahStapp: We only mutate `metadata["driver"]`, can we avoid the cost of the deepcopy here?

pymongo/pool_options.py
```diff
@@ -336,6 +356,11 @@ def __init__(
         self.__load_balanced = load_balanced
         self.__credentials = credentials
         self.__metadata = copy.deepcopy(_METADATA)
+        self.__appended_drivers: list[DriverInfo] = []
```
NoahStapp: We can make this a set for faster dedup checking.

NoahStapp (changes requested):

pymongo/pool_options.py
blink1073: Done

pymongo/pool_options.py
blink1073: Done

pymongo/pool_options.py
blink1073: Done

pymongo/pool_options.py
blink1073: Done

pymongo/pool_options.py
blink1073: Done

test/asynchronous/test_client.py
blink1073: Done

test/test_client_metadata.py
blink1073: test_11 is a prose test, so I removed `test_metadata_rejects_delimiter`

pymongo/pool_options.py
NoahStapp: Can we do a similar optimization here for potential overflow counting? Keep only the first `bson.encode(metadata)` call and then reduce that known length with each truncation instead of calling `bson.encode(metadata)` after each step's reduction?

NoahStapp (changes requested):

pymongo/pool_options.py
blink1073: Done

pymongo/pool_options.py
```diff
@@ -368,28 +442,40 @@ def __init__(
...
+
+            self.__metadata = metadata
+
+            # Only track drivers whose appended name/version pair survived
+            # truncation (i.e. both gained a segment), so __appended_drivers
+            # stays bounded.
+            if (
+                metadata["driver"]["name"].count("|") > name_delims
+                and metadata["driver"]["version"].count("|") > version_delims
+            ):
```
NoahStapp: Can we just check if the actual `name, version` pair of `driver` survived truncation instead? That would be easier to understand than counting delimiters. I'm worried that this hides an edge case with truncation somewhere.

NoahStapp (changes requested):

pymongo/pool_options.py
blink1073: Done

NoahStapp (approved):

